Issue #4, opened 2016-11-06
Build tools


jsotuyod:
Right now we have a large and complex Ant build, and a smaller but incomplete Maven build.

Keeping the two requires effort and time with dubious benefits if we are not using both ourselves, and given it's status I believe none of us actually use Maven right now.

Ant needs too much configuration, and if you want up-to-date checking, you have to build it yourself, task by task. This alone makes me want to run away from it. I understand Maven has it's issues too, and specially for those that got to use it before Maven 3, raises mistrust on how reliable builds are (I've been there too, throwing the towel and just performing a `mvn clean`).

I'm particularly fond on Gradle after using it and contributing to it for the last couple years. It's not XML, it's flexible (the DSL is powerful, but you can write Groovy code if you feel like it), creating custom plugins and tasks is easy (I've done it several times), and has amazing support for up-to-date and incremental builds. Gradle even can use [directly ant tasks thanks to Groovy's support](http://docs.groovy-lang.org/latest/html/documentation/ant-builder.html).

Also, [FindBugs is officially supported by Gradle (no external plugins)](https://docs.gradle.org/current/userguide/findbugs_plugin.html). They are cool guys :P

I'm eager to know your point of view.

mebigfatguy:
I agree the current build process is a disaster.

I personally prefer ant, to maven or gradle, however, if others want to use another build process, i'm ok with that, so long as those 'others' take responsibility for maintaining the pom, or whatever the gradle thing is, as I dont' want to have to burn brain cells trying to figure out that voodoo.

But before we do that, i think we should decide on what the shape of the project should be.

The codebase currently has many tenacles, a bunch of which, i doubt any of us are interested in.

Things like the BlueJ integration, and the cloud thing, do we want to maintain those? I don't. There's a mess of folders that i dont' even know what they are... NullPointerBenchmark? sandbox? etc etc.

If any of you do want to keep those, i'd still hope we could move them to a separate repository (within spotbugs project sure). Even the GUI, and eclipse plugin, i'd be happy to move to separate repositories.

Anyhow, I think we should decide on all that first.

jsotuyod:
I agree on killing cloud.
Update checks can go too.
Dubious projects such as `NullPointerBenchmark` and `sandbox` are good to go without asking, if it's not being part of a published binary it's no use.

BlueJ we can probably drop.

[...]

Issue #1, opened 2016-11-06
missing libs?


mebigfatguy:
looks like asm and bcel jars didn't come over, or am i missing something?

mebigfatguy:
hmm is see you added dynamic pulling but i get

/home/dave/dev/spotbugs/build.xml:18: The following error occurred while executing this line:
/home/dave/dev/spotbugs/findbugs/build.xml:1295: org.mozilla.javascript.EcmaError: ReferenceError: "Java" is not defined.

jsotuyod:
Those changes were introduced back in FindBugs, however, it seems you can access Java [classes directly](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Scripting_Java), I'm submitting a PR for you to check.

However, we should start discussing build process unification. Right now we have Ant & Maven, where Maven is incomplete.

I'll open a new issue to discuss it.

PR #11, opened 2016-11-06
use simple <get> task to get bcel, and asm

mebigfatguy:


iloveeclipse:
What is the reason for the change? Are there any troubles with the current approach? For the proposed change: does it mean that files are downloaded on every execution?

mebigfatguy:
i had multiple problems with the current approach, one that juan seemed to fix, as seen here

https://github.com/spotbugs/spotbugs/issues/1

the next one, which remains, as

/home/dave/dev/spotbugs/findbugs/build.xml:1295: org.mozilla.javascript.EcmaError: ReferenceError: "print" is not defined.
	at org.mozilla.javascript.ScriptRuntime.constructError(ScriptRuntime.java:3951)



We can remove the tasks from the build target, there's no reason to run them all the time, if you like

There's also a 

usetimestamp="true"

attribute which only downloads if a version is newer., which we can add.

i'm fine with leaving the current as is, if someone can fix it, it just seemed to be broken.

jsotuyod:
Bare in mind `usetimestamp` relies on local modification time of the file. If you `touch` an empty file with that name, it will still consider it up to date. Nothing too serious, but still to consider.

The current download code however, is giving me issues when trying to get Travis to run using Java9.

```
BUILD FAILED
/home/travis/build/spotbugs/spotbugs/build.xml:37: The following error occurred while executing this line:
/home/travis/build/spotbugs/spotbugs/findbugs/build.xml:1295: Unable to load a script engine manager (org.apache.bsf.BSFManager or javax.script.ScriptEngineManager)
```

https://travis-ci.org/spotbugs/spotbugs/jobs/173798833#L2064-L2066

On Java 8 it runs 'though... https://travis-ci.org/spotbugs/spotbugs/jobs/173798832

findbugs/build.xml
```diff
@@ -9,7 +9,7 @@
     <!--
             The local.properties properties file contains the location of eclipsePlugin.dir
 
-            This value is likely to be different for each checkout of the plugin,
+            This valueg is likely to be different for each checkout of the plugin,
```
henrik242: What's this `valueg` thing you are talking about? :)

jsotuyod:
I tracked down my build issue. ant `<script>` task is broken with modular jdk9 due to a known ant issue: https://bz.apache.org/bugzilla/show_bug.cgi?id=58271 

This issue is due to change in behaviour of `ClassLoader.getResource` for resources in named modules, and is fixed on ant 1.9.6.

I am manually installing ant 1.9.7 on Travis, but I approve of this change once the typo marked by @henrik242.

Ant sucks at dependency management anyway, but we will get to discussing that when we attack #4

jsotuyod:
I'll revert the download of Ant 1.9.7 on Travis based on this to speed up build / make sure build works on all Java versions for everyone.

PR #12, opened 2016-11-06
Add initial Travis support

jsotuyod:
- Based on @KengoTODA's work at https://github.com/findbugsproject/findbugs/pull/48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: java
+sudo: false
+jdk:
+  - oraclejdk8
+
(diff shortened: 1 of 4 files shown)


iloveeclipse:
Is there a way to setup a project mail so that everyone will be informed if build will fail? Something like gerrit does?

jsotuyod:
Actually, I'm removing notifications, and enabling the GitHub hook with Travis, so that merge is forbidden if the build for the PR doesn't pass. This, along with the current prohibition of pushing to master without a PR should be enough without noisy email notifications every time someone sends a broken PR.

Issue #17, opened 2016-11-07
Deal with Locale provider changes in JRE9


jsotuyod:
As stated [on Java's issue database](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8152154):

> DK 8 has JRE as default locale provider, whereas in JDK 9 onwards CLDR is the default  locale provider. (http://openjdk.java.net/jeps/252 )
To enable behavior compatible with JDK 8, set the system property java.locale.providers to a value with JRE ahead of CLDR.
> 
> In JDK 9 ,if we run the test case as " java  -Djava.locale.providers=JRE,CLDR JI9032375" it is successful.
Closing as not an issue.

This change means date parsing is failing on the latest Java 9 early access since `DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT, Locale.ENGLISH)` will require the format "M/d/yy, h:mm a", instead of the traditional "M/d/yy h:mm a" of JRE8 (notice there is a comma).

This break is evidenced by #16

SpotBugs should work as is, without extra JVM arguments, for all users on all environments. So we need to work around this. Possibly creating a `SimpleDateFormat` with an explicit pattern and Locale rather than using the default should work.

jsotuyod:
Closed by #18

PR #18, opened 2016-11-07
Fix issue #17

jsotuyod:
- Since we have a hardwired Locale, and are already assuming the format,
    just use that format directly. This avoids issues with the LocaleProvider
    changes in JRE9.

Confirmed, the failing test passes with this change on all tested JRE versions.

### DIFF
--- a/findbugs/src/java/edu/umd/cs/findbugs/BugInstance.java
+++ b/findbugs/src/java/edu/umd/cs/findbugs/BugInstance.java
@@ -27,6 +27,7 @@ import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Instead of hard-coding the format I would rather change the test and make sure we don't persist dates with the unsupported format elsewhere.

jsotuyod:
I'm not sure what you mean with "unsupported format" here, nor how you propose to change the test.

Non the less, maybe some explanation on my reasoning can help.

The format is part of SpotBugs public API. It's written to XML reports, and it's sent over /read from cloud plugins. That means, different machines / VMs should be able to share that info seamlessly otherwise a whole organization would have to switch to using JRE9 (not JDK) at the same time. The same way, CI servers may find issues, since the build can be performed on a different VM version than the one on which the CI server runs and which it uses to parse results (think of Hudson / Jenkins for instance).

As part of the public API, and in an effort to keep backwards compatible, making sure all JRE versions produce the same output as previously expected seemed as the most reasonable way forward.

Issue #10, opened 2016-11-06
Unit testing


jsotuyod:
As per #4, particularly https://github.com/spotbugs/spotbugs/issues/4#issuecomment-258701234 and https://github.com/spotbugs/spotbugs/issues/4#issuecomment-258702861 we need to improve upon current test architecture.

This issue keeps open for discussion as to the approach to testing to move forward.

jsotuyod:
With #12 and #16 we have completed CI integration. Unit testing revamping still pending.

jsotuyod:
Given the current test architecture has proven incapable of actually testing violations on lambdas, which are the corner stone of all tasks we need to perform to have full Java 8 support, I'm raising this task's priority.

As previously stated I like @h3xstream's approach in find-sec-bugs. The hamcrest matcher is easy to use and flexible, and can easily be integrated with the current use of the engine within tests. Some helpers could be added to convert `@ExpectWarning` and other annotations to assertions, meaning we could get rid of the `CheckExpectedWarnings` detector.

I'll try to get a working draft on this for review adding the find-sec-bugs testutils lib directly.

h3xstream:
In case it is needed, you can reuse find-sec-bugs test utility (move the code partially or as-is). It is LGPL anw.

jsotuyod:
@h3xstream thanks! I'll make sure to provide proper acknowledgements

PR #21, opened 2016-11-08
Resolve #20

### LINKED ISSUE #20: @CheckForNull and @Nonnull not working within lambdas
Original: https://github.com/findbugsproject/findbugs/issues/95 - reported by @lotsabackscatter

Null pointer de-reference checks do not appear to be working within Java 8 lambdas.

Anonymous classes work fine but converting these to lambdas effectively suppresses any warnings.

See below for an example test case:

```
package sfBugsNew;

import javax.annotation.CheckForNull;
import edu.umd.cs.findbugs.annotations.ExpectWarning;

/**
 */
public class Bug {

  /**
   * @return may be null
   */
  @CheckForNull
  public static String methodThatMightReturnNull() {
    if (Boolean.getBoolean("test")) {
      return null;
    }

    return "test";
  }

  /**
   */
  @ExpectWarning("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE")
  public void testNPLambda() {
    Runnable lambda = () -> {
      String maybeNull = methodThatMightReturnNull();

      // would expect findbugs warning about possible null dereferencing here
      boolean empty = maybeNull.isEmpty();
      System.out.println(empty);
    };
    lambda.run();
  }

}
```

jsotuyod:
- Add the failing test scenario
 - Add a helper in BCELUtil to detect lambdas.
 - Avoid BCELUtil from returning true to `isSynthetic` for lambdas,
    not technically correct, but semantically, when detectors ask for
    `isSynthetic` they just want to know if the developer wrote the code
    and for lambdas, that is true.
 - Workaround the testing framework to deal with the lack of expectation
    warnings for lambdas. This needs a refactor.

Fixes #20


jsotuyod:
Ok, this is probably not ready to be merged, but having the code available will certainly make discussion much easier.

I realize the change to `BCELUtility.isSynthetic` is bad. I am returning false if it's a lambda, even if it's a synthetic method. This is strictly broken, but semantically when asking for synthetic methods, what we want to ask when checking it is not "was this method created by the compiler?" but rather "was this method **not** written by the developer?". The difference is small, yet important. The body of a lambda method *was* written by the developer, and is therefore as interesting as any non-synthetic method.

What I wanted to discuss with you is the semantics we would like to offer for  proper fix to this.

We can't prevent plugin developers from accessing `isSynthetic` on any of the multiple classes and helpers that offer that method, but we may offer some higher-level abstractions such as `isAutogeneratedMethod` or `isLambda`; and refactor SpotBugs source to start using those instead of directly checking for synthetic fields / methods.

Not sure what to name such helper / where to put it, nor the semantics for those methods (is really `isAutogeneratedMethod` an appropriate name to tell apart accessors / autogenerated enum methods from lambdas and user methods?)

Also, this scenario shows once again the need of addressing #10, since the current testing integration is insufficient to properly / easily test these scenarios.

jsotuyod:
Has anyone got the chance to think about this one?

jsotuyod:
@henrik242 good call not merging this. It's marked as WIP, we still need to decide what are proper abstractions to present to detectors / plugin developers to make it easier to work with lambdas.

henrik242:
@jsotuyod Yup. You should maybe change the title of this PR, though, since #20 is resolved :)

jsotuyod:
Ok, a bunch of changes:

- I replaced the test to use the new test infrastructure (#10 & #35)
- I deprecated `BCELUtil.isSynthetic` (all versions), in favor of the corresponding higher level abstraction on the new `MemberUtils.isUserGenerated`. This change has already been impacted on `BugInstance`. Other detectors may be need to be updated, but should be analyzed separately.

jsotuyod:
This is ready for review as far as I'm concerned. I feel the current abstraction is good enough.

findbugs/src/main/java/edu/umd/cs/findbugs/bytecode/MemberUtils.java
```diff
@@ -0,0 +1,129 @@
...
+ * Utility to analyze class members.
+ */
+public final class MemberUtils {
+
+	private MemberUtils() {
+		throw new AssertionError("Utility classes can't be instantiated");
+	}
+
+	private static boolean internalIsSynthetic(final FieldOrMethod m) {
+        if (m.isSynthetic()) {
```
KengoTODA: Checkstyle also reported problems:
```
[ant:checkstyle] [ERROR] /findbugs/src/main/java/edu/umd/cs/findbugs/SAXBugCollectionHandler.java:573:9: File contains tab characters (this is the first instance). [FileTabCharacter]
[ant:checkstyle] [ERROR] /findbugs/findbugs/src/main/java/edu/umd/cs/findbugs/Version.java:49:13: File contains tab characters (this is the first instance). [FileTabCharacter]
[ant:checkstyle] [ERROR] /findbugs/src/main/java/edu/umd/cs/findbugs/ba/generic/GenericUtilities.java:264:13: File contains tab characters (this is the first instance). [FileTabCharacter]
```

jsotuyod: You are absolutely right. I'l fix this and have Checkstyle run on Travis too.

jsotuyod:
There you go. I'll enable Checkstyle on a separate PR since there are a few other issues to solve and don't want to mix things on this PR.

PR #22, opened 2016-11-10
Load findbugs.xml from core plugin JAR file.

KengoTODA:
This PR is same with https://github.com/findbugsproject/findbugs/pull/102
This change lets implementation follows documentation, but actually it's breaking change. It's better to merge after build and test become stable.

----

Even though javadoc of #getResource(String) says as follows, current
implementation of #getCoreResource(String) loads resource from
filesystem first, even though we have it in JAR file.

> We try findResource because (based on experiment) we can trust it
> to prefer resources in the jarfile to resources on the filesystem

This is why we have problem discussed in FB-Discuss mailing list:
https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2012-September/003632.html

### DIFF
--- a/findbugs/src/java/edu/umd/cs/findbugs/PluginLoader.java
+++ b/findbugs/src/java/edu/umd/cs/findbugs/PluginLoader.java
@@ -54,6 +54,7 @@ import java.util.zip.ZipFile;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.WillClose;
 
 import org.dom4j.Document;
@@ -423,6 +424,7 @@ public class PluginLoader {
         plugin = null;
     }
 
+    @Nonnull
     private static URL computeCoreUrl() {
         URL from;
         String findBugsClassFile = ClassName.toSlashedClassName(FindBugs.class) + ".class";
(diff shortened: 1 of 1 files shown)


jsotuyod:
@KengoTODA thanks for the PR. Most interesting.

Please, check Travis, this changeset is breaking when running `ant test`

KengoTODA:
Thanks, I'll check. It might be better to fix javadoc if it's hard to fix test.

KengoTODA:
I've rebased to the latest master, and found a compilation error:

```
    [javac] /home/kengo/GitHub/spotbugs/eclipsePlugin/src/de/tobject/findbugs/FindbugsPlugin.java:252: error: incompatible types: String cannot be converted to Plugin
    [javac]         ResourcesPlugin.getWorkspace().addSaveParticipant(PLUGIN_ID, saveParticipant);
```

Not sure why, I will check it later.

KengoTODA:
I found that my Eclipse was too fresh, I should use old version to build.

And in local I can compile this branch successfully... I'll push again.

KengoTODA:
It seems that we cannot `wget` Eclipse zip file now...

KengoTODA:
I confirmed [3.6 (Helios)](http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.6.2-201102101200/eclipse-SDK-3.6.2-linux-gtk-x86_64.tar.gz) is 404 but [3.7 (Indigo)](http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.7.2-201202080800/eclipse-SDK-3.7.2-linux-gtk-x86_64.tar.gz) is downloadable. I'll try to use Indigo in `.travis.yml`.

jsotuyod:
Yes, I had just found that same problem, and submitted #26 for review. You can check out that changes from there.

KengoTODA:
@jsotuyod I also made a similar diff in local. It should work, thanks! I'll rebase to `master` after #26 has been merged.

KengoTODA:
Thank you for merging!

PR #23, opened 2016-11-10
fix two problems in special kind handling

KengoTODA:
This MR is same with https://github.com/findbugsproject/findbugs/pull/86

### DIFF
--- a/findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
+++ b/findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
@@ -31,6 +31,9 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


jsotuyod:
This looks great. I don't see much of an issue with changing types for `specialKindNames`. Since `ConcurrentMap` extends `Map` we are safe for well-behaved code.

The only breaking cases would be:

* Storing it to a local `HashMap`
* Passing it to a method that expects a `HashMap`

Good practice calls for people to access it by interface as a `Map` (which SpotBugs should have done either way), and even if not, it's a trivial error to spot (it will throw a compile error or runtime exception) and fix so I'm not particularly against breaking this API.

On the contrary, I'm hesitant on changing `defineNewSpecialKind`. Even though it's true it's return value is off-by-one, [it's been broken for almost 4 years](https://github.com/spotbugs/spotbugs/commit/83f8c89b0b0992292cad40f3fda4672bea868ace#diff-ea91b1571671d3af2c38cc26772a6aae).

This could either mean, no one is using it and we are safe, or they just workaround it on their own, meaning this fix would break that code. If it's the second scenario, the fix may go unnoticed and cause more harm than good (plugins may not break, though they would misbehave).

I've so far failed to find a plugin that uses `defineNewSpecialKind` so I tend to think we are mostly on the first scenario, but I wonder what @iloveeclipse and @mebigfatguy think, since both of them were around at the time of the breaking change.

jsotuyod:
Unless @mebigfatguy has some insight, I guess the best we can do without breaking third parties is:

1. `@Deprecate` the `defineNewSpecialKind` method, leaving it's implementation as-is.
1. create a new method named `defineSpecialKind` or whatever you see fit, with the correct behavior.

People using the old code will work as expected, new users can use the good method, and in a future SpotBugs 4 release, we can safely kill the deprecated method.

Comments anyone?

mebigfatguy:
@jsotuyod 's approach seems sane. i really have no clue about that code

KengoTODA:
@mebigfatguy @jsotuyod Thanks for your review! I made another pull-request #27 to add methods, please check. I will close this pull-request.

PR #24, opened 2016-11-13
more branding changes

mebigfatguy:


jsotuyod:
Awesome! Check that the Eclipse plugin build is failing though, since it references libs you renamed.

findbugs/src/antTask/edu/umd/cs/findbugs/anttask/tasks.properties
```diff
@@ -1,5 +1,4 @@
-findbugs=edu.umd.cs.findbugs.anttask.FindBugsTask
-findBugs=edu.umd.cs.findbugs.anttask.FindBugsTask
```
jsotuyod: I think we should keep these 2 lines for the time being, in spite of adding the `spotbugs` definition. That way we can remain compatible with all other tools (maven plugin, gradle, etc.) that would only require to change the dependency from findbugs to spotbugs to keep on working. We will eventually drop them, but retaining them for the time being is probably for the best.

findbugs/src/gui/edu/umd/cs/findbugs/gui2/help/About.html
```diff
@@ -8,27 +8,27 @@
 
 
 <p align="center">
-						<a href="http://findbugs.sourceforge.net/"><img src="umdFindbugs.png"
-								alt="FindBugs logo" border="0" /> </a>
+						<a href="https://github.com/spotbugs/"><img src="umdFindbugs.png"
```
jsotuyod: we probably should replace `umdFindbugs.png` with the spotbugs logo (there are a bunch in the repository under different directories), with apropriate copies of [our-logo](https://github.com/spotbugs/spotbugs.github.io/blob/master/spotbugs-icon.png) in a proper size. Will add a task to do so during the week.

jsotuyod:
Now it's green! Other from that comment on keeping the old ant task names around, this looks great

PR #26, opened 2016-11-14
Update Eclipse version to be used

jsotuyod:
- All mirrors are throwing a 404 on the old 3.6.2 version, so we upgrade to 3.7.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ matrix:
            - oracle-java9-installer
      
 install:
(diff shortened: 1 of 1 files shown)


KengoTODA:
Now I can download Helios from [here](http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.6.2-201102101200/eclipse-SDK-3.6.2-linux-gtk-x86_64.tar.gz), this PR might be needless.

jsotuyod:
Confirmed, mirror are back again. Closing this on.

PR #27, opened 2016-11-15
create new methods to handle special kind in proper way

KengoTODA:
To fix problem in current way, mark them as deprecated and create new methods.

refs #23

### DIFF
--- a/findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
+++ b/findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
@@ -273,7 +273,7 @@ public class OpcodeStack implements Constants2 {
         int TYPE_ONLY = 24;
 
         /**
-         * @deprecated will be deleted at v4 release. use @{link {@link #getSpecialKindName(int)}} instead.
+         * @deprecated will be deleted at v4 release. use {@link #getSpecialKindName(int)} instead.
          */
         @Deprecated
         @edu.umd.cs.findbugs.internalAnnotations.StaticConstant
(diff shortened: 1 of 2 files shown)


findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
```diff
@@ -268,11 +272,25 @@
         public static final @SpecialKind
         int TYPE_ONLY = 24;
 
+        /**
+         * @deprecated will be deleted at v4 release. use @{link {@link #getSpecialKindName(int)}} instead.
```
jsotuyod: double `@link`?

findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
```diff
@@ -268,11 +272,25 @@
...
+         */
+        @Deprecated
         @edu.umd.cs.findbugs.internalAnnotations.StaticConstant
         public static final HashMap<Integer, String> specialKindNames = new HashMap<Integer, String>();
```
jsotuyod: we should probably leave a `TODO` for ourselves to convert this into a `private ConcurrentMap`

KengoTODA: Which we prefer: TODO comment or Issue at GitHub? Personally I prefer latter one.

findbugs/src/java/edu/umd/cs/findbugs/OpcodeStack.java
KengoTODA: Oops... I'll fix.

jsotuyod:
Ok, Github issue it is. I just created a new 4.0.0 milestone so we can tag it.

KengoTODA:
@jsotuyod I've issued #28, please tag it with new milestone.

jsotuyod:
@KengoTODA done! thanks!

PR #32, opened 2016-11-20
exchange findbugs.ico for spotbugs.ico

mebigfatguy:


KengoTODA:
Does Spotbugs break backward compatibility around environment variable? I think it's better to support legacy variables like `FINDBUGS_HOME`, to support existing users.

e.g. Jenkins still supports `HUDSON_HOME`. https://github.com/jenkinsci/jenkins/blob/08def67a18eee51de9f3f99bc2a792fee1c160e0/core/src/main/java/hudson/Main.java#L77-L81

jsotuyod:
@KengoTODA this changeset just swaps the variables used / exported by the scripts. Your observation, however is correct, but should be taken into account directly [into the code](https://github.com/mebigfatguy/spotbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/FindBugs.java#L140) were we may be getting called by more... ad-hoc means.

The correct behavior should be to privilege `spotbugs.home`, fallback to `findbugs.home` if not found. Will submit a PR right now.

PR #37, opened 2016-11-22
fix NPE

KengoTODA:
`File#getParentFile()` may return null, so we should check its nullness before we refer it.
https://docs.oracle.com/javase/8/docs/api/java/io/File.html#getParentFile--

In my case, `findbugsJar` was `/home/kengo/GitHub/spotbugs/findbugs/lib/spotbugs.jar` and `libDir` was `null`. Tested with `1.8.0_112-b15` on Ubuntu 14.04.

### How to reproduce

1. run `ant build` to make package
2. `cd findbugs/lib`
3. execute `java -jar spotbugs.jar`, then you'll find NPE


jsotuyod:
Thanks @KengoTODA! We definitely need to kill that method at some point in the future... at least rewrite it using `Paths` and for sure not as part of that class

Issue #30, opened 2016-11-17
Travis: deploy SNAPSHOT artifact to Maven repository


KengoTODA:
I'm working on my FindBugs plugin to build with latest SpotBugs, but currently no Maven repository provides `com.github.spotbugs:spotbugs:3.1.0-SNAPSHOT` so I need to install it by own like below:

```sh
$ cd findbugs && ant build srcJar && mvn install:install-file \
  -Dfile=lib/spotbugs.jar \
  -DpomFile=pom.xml \
  -Dsources=spotbugs-src.jar
```

It's not convenient, so I wish Travis CI deploys SNAPSHOT artifact to [Sonatype Maven repository](https://oss.sonatype.org/content/repositories/snapshots/) automatically. Maybe it's not so hard:

1. Create a script file which executes [`deploy:deploy-file`](http://maven.apache.org/plugins/maven-install-plugin/install-file-mojo.html)
2. Kick it at [after_success](https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle) only when Travis build SNAPSHOT version in `master` branch

Maybe we can refer https://github.com/JakeWharton/butterknife/commit/bac2342500971a3c6664629c3a63ca74d6e4819c as example.

How do you think? If it's OK, I will propose a PR.

jsotuyod:
I think this is great, but first we would have to apply to the central repository, which we haven't done already.

I'll try to get this underway ASAP.

henrik242:
@jsotuyod Did you get a reply from the central repository yet?

jsotuyod:
Haven't applied yet, need to collect everyone's users.

PR #38, opened 2016-11-24
Gradle build

jsotuyod:
As per #4 I rewrote most tasks (the meaningful ones) in Gradle. It's ~380 LoCs, against the ~1800 of ant; and lots of it could be removed as be tidy up the project structure.

Right now I focus on findbugs and finbugsTestCases, the rest is mostly trash to be deleted, or could be easily added as needed.

I also added some useful tasks (IntelliJ / Eclipse project setup), so we can add those to the list of files to be deleted.

Finally, I also added a sources jar generation (needed to publish to Maven Central), not sure why ant wasn't doing it.

I believe none of the non-translated tasks are of any interest. We can probably kill all ant / maven config once this gets merged (and after changing Travis config accordingly), and if ever needed we can get it back from the git history.

With that in place, getting jar signing and publication to Maven Central should be really easy, to the point of getting #30 in place.


jsotuyod:
That was quick!

I'll convert the `eclipsePlugin` build script first, then proceed to killing ant and maven.

jsotuyod:
Initial stub of grale building the eclipse plugin is [here](https://github.com/spotbugs/spotbugs/tree/gradle-eclipseplugin). The jar and zip are generated, the update site still missing.

Expect a PR tomorrow.

henrik242:
Great job!

PR #40, opened 2016-11-26
Kill ant & maven

### LINKED ISSUE #4: Build tools

jsotuyod:
- Resolves #4 
 - Travis is now building with Gradle - took the chance to make the code compilation compatible with JDK9.

All idea / eclipse files are removed form git and ignored. One you fetch this changeset, you will have to run `./gradlew eclipse` or `./gradlew idea` to regenerate them locally.


h3xstream:
Another thing to remember .. **Please test** that it is actually **usable** in Eclipse/IntelliJ for development.
In IntelliJ, the Gradle support is pretty good.

.travis.yml
```diff
@@ -22,7 +22,6 @@ install:
   - echo eclipsePlugin.dir=$(pwd)/eclipse/plugins > eclipsePlugin/local.properties
 
 script:
-  - ./gradlew assemble
-  # Manually switch to JDK9 for tests if needed
+  # Manually switch to JDK9 if needed
   - if [[ $JDK_FOR_TEST == "oraclejdk9" ]]; then remove_dir_from_path $JAVA_HOME/bin; export JAVA_HOME=/usr/lib/jvm/java-9-oracle; export JAVA_HOME=/usr/lib/jvm/java-9-oracle; fi
-  - ./gradlew test smoketest
+  - ./gradlew assemble test smoketest -S
```
h3xstream: If anybody doing the review .. `-S` is just printing the stacktrace in case of failure. https://docs.gradle.org/current/userguide/gradle_command_line.html

eclipsePlugin/build.gradle
```diff
@@ -18,28 +18,36 @@ sourceSets {
...
   compile.extendsFrom embeddedLibs
+  compile.extendsFrom includedLibs
 }
 
 def localProps = new Properties()
 localProps.load(new FileInputStream("$projectDir/local.properties"))
 
-def requiredLibs = fileTree(dir:'lib', include:['jsr305.jar',
+// TODO : convert these to external dependencies
+def requiredLibs = fileTree(dir:'lib', include:[
```
h3xstream: @jsotuyod Could the require libraries use the basic compile directive instead of this additional custom directives?

jsotuyod: I probably could if I changed the eclipse dependencies from `compile` to `compileOnly`, although I would have to make sure to filter dependencies from extended configurations (`compileOnly`, `embeddedLibs`) when creating the distribution jars / zips...

Probably better to address that as we change those dependencies to actual maven dependencies instead of local libs...

jsotuyod:
Totally agree on the changes to move to a more conventional project structure. I strongly believe that getting this done however, absolutely needed us having Travis check on our every step of the way. So this PR as is.

As for testing on IDEs. I personally tested on Eclipse, but don't have a IntelliJ license, feedback from users would be welcome. I did however check the diff on the generated files and saw no differences on project structure and setup. The code style embedded on the project files is gone, but was too unreliable anyway (other IDEs wouldn't be using it and the code is messy as it stands).

Tidying it up and setting up Checkstyle on Travis to enforce a coding style is still down the road.

henrik242:
👍

jglick:
Any instructions for how to run this build?

```
spotbugs$ ./gradlew clean build

FAILURE: Build failed with an exception.

* Where:
Build file '…/spotbugs/eclipsePlugin/build.gradle' line: 27

* What went wrong:
A problem occurred evaluating project ':eclipsePlugin'.
> …/spotbugs/eclipsePlugin/local.properties (No such file or directory)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 0.667 secs
```

jsotuyod:
@jglick that's been around for ages (the ant build used it). The eclipse plugin needs to have access to an eclipse install. It will be used both as dependencies, and to generate the update site metadata. `local.properties` holds a reference to the location of the eclipse install to use. You can check how we [set it up on Travis](https://github.com/spotbugs/spotbugs/blob/master/.travis.yml#L22). In the future we can tune the gradle build to either skip the eclipse plugin build or fail with a better error if the property is missing.

Other than that, `./gradlew build` should work.

henrik242:
Skipping the eclipsePlugin build by default sounds like a good idea.  I don't have eclipse installed at all, so I won't be able to build spotbugs..

jsotuyod:
In the meantime, you can simply do `./gradlew :findbugs:build` to just build the library.

It will eventually change to `./gradlew :spotbugs:build` as discussed on previous comments.

henrik242:
Just curious: What is missing for the review to get through?  Does anyone else but @mebigfatguy  have approval rights?

KengoTODA:
I will check generated .jar later, to ensure that gradle can generate .jar as expected (no big difference with Ant).

henrik242:
Should pom/ and future-pom.xml also have been a part of this PR?

jsotuyod:
@henrik242 absolutely right. I missed those. Would you mind submitting a PR killing those? I don't think I'll be able to do so before tomorrow afternoon.

KengoTODA:
@jsotuyod I found several files which isn't packaged by gradle:

* version.properties
* edu/umd/cs/findbugs/annotations/package-info.class
* edu/umd/cs/findbugs/version.properties

And properties file is necessary for `findbugs/src/java/edu/umd/cs/findbugs/Version.java`
Could you check?

jsotuyod:
That package info is discarded by the compiler for being empty. There are no annotations there https://github.com/spotbugs/spotbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/annotations/package-info.java

To retain it would require compiler flags and ia useless. 


As for the version properties, I'll check tomorrow. I was sure it wasn't being used, but I seem to have been wrong. 

Thanks for checking it!

jsotuyod:
That file seems mostly useless or related to update checks. I'll have a deeper look tomorrow, but killing it may be an option over generating the properties file.

KengoTODA:
@jsotuyod Agree, thanks for your quick response!

henrik242:
@jsotuyod PR submitted in #41

PR #41, opened 2016-11-30
Get rid of remaining Maven and Eclipse files

henrik242:
Ref #40 

There's still some remaining `build.xml`/`pom.xml`/`.project` in `plugins/*/*`.  Should those go as well, or are we waiting for them to be gradle-i-fied?


jsotuyod:
This is great! Thanks! 

We can leave those until we go over those projects. Most probably we will just delete them completely.

PR #43, opened 2016-11-30
Load version from Manifest

jsotuyod:
- Simplify the code, get rid of a bunch of version flags that had no real value
 - Resolve issues raised at https://github.com/spotbugs/spotbugs/pull/40#issuecomment-263771827


henrik242:
👍 Nice cleanup

mebigfatguy:
whats the checking for file protocol for? Should work otherwise, right? what would the other be, jnlp?

jsotuyod:
@mebigfatguy actually, I'm not sure (it was already like that in the code, I just load the manifest instead of  `version.properties`). I think it's intending to check if we are loading from a jar (`"jar:file://..."`) or not, but I may be wrong.

If we are not running from the jar (running from IDE), the getResourceAsStream would return null.

PR #45, opened 2016-11-30
Fix Travis flakyness

jsotuyod:
- It's randomly failing under JRE9 due to timeouts
 - Setup gradle properties for better results


jsotuyod:
Travis is still killing the executor... will have to test more in depth...

PR #55, opened 2016-12-02
Workaround for JDT illegal signatures

jsotuyod:
- See https://bugs.eclipse.org/bugs/show_bug.cgi?id=494198
 - The issue was raised on FindBugs https://github.com/findbugsproject/findbugs/issues/140


jsotuyod:
@mebigfatguy there I added test cases for all, `!+`, `!-` and `!*` at a higher level and was able to detect and fix an issue. I know believe this will work as expected.

PR #61, opened 2016-12-05
Fix path of class files to analyse in integration test

KengoTODA:
Currently gradle generates class files under `findbugsTestCase/build/classes/main` directory, but out test still uses `findbugsTestCase/build/classes`.

To find similar problems asap, I replace `Assume` in `AbstractIntegrationTest` with `Assert`.

### DIFF
--- a/findbugs/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
+++ b/findbugs/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
@@ -28,7 +28,6 @@ import java.io.IOException;
 import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
-import org.junit.Assume;
 
 import edu.umd.cs.findbugs.config.UserPreferences;
 import edu.umd.cs.findbugs.internalAnnotations.SlashedClassName;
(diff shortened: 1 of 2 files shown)


KengoTODA:
I cannot reproduce this build failure in my local, I will run Travis build again.

jsotuyod:
I've no idea why this PR fails. I checked out this PR, added some Travis config to make tests show failure causes, and it passes. See [Travis Build](https://travis-ci.org/spotbugs/spotbugs/builds/181448551) and the [branch](https://github.com/spotbugs/spotbugs/compare/pr-61?expand=1)

jsotuyod:
I'm working on figuring this one out. So far, I've been able to determine the issue is detected, but with priority Medium and not High on Travis for the merge commit.

I'm going deep now on why this priority is reached and fixing the root cause. Will update hopefully soon.

jsotuyod:
I just figured it out. The problem is that [this static `XMethod`](https://github.com/spotbugs/spotbugs/blob/master/findbugs/src/main/java/edu/umd/cs/findbugs/detect/ExplicitSerialization.java#L44) is populated on the first test to load `ExplicitSerialization`. However, at that point, the `ClassDescriptor` for `java.io.ObjectOutputStream` may have not been already referenced, and therefore, that field will be initialized as an `UnresolvedXMethod`.

When running again, even if now the analyzed class references `java.io.ObjectOutputStream`, and makes those methods interned, the fields can't be reinitialized and will remain as unresolved, failing to detect calls to those methods, and therefore, reducing the confidence (and the priority) of the reports.

We either make those fields non-static, or we ask junit to fork after every test class (notice different test cases on the same test file can still screw each other up).

@iloveeclipse I'm curious as to how the Eclipse plugin solves this. As far as I know, FindBugs is run on the same JVM as Eclipse, so this issue would affect it too...

All this makes me favor the non-static approach. I'll submit a separate PR in  couple minutes.

PR #63, opened 2016-12-05
Do not merge - Pr 61

jsotuyod:
This is the same as #61, but with extra gradle config to make failing tests show the failure cause.

This is just a test to see why that PR fails.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ subprojects {
   apply plugin: 'java'
   apply from: "$rootDir/gradle/eclipse.gradle"
   apply from: "$rootDir/gradle/idea.gradle"
+  apply from: "$rootDir/gradle/test.gradle"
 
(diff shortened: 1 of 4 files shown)


jsotuyod:
Ok, this is weird. The HEAD of the branch built just fine. The PR fails consistently on Travis (never locallly) with

```
edu.umd.cs.findbugs.DetectorsTest > testAllRegressionFilesJavac FAILED
    java.lang.AssertionError: Unexpected bugs (2):
    missing SE_BAD_FIELD In RFE3062724.java
    missing SE_BAD_FIELD In RFE3062724.java
        at org.junit.Assert.fail(Assert.java:91)
        at edu.umd.cs.findbugs.DetectorsTest.checkForUnexpectedBugs(DetectorsTest.java:142)
```

I'm going to test converting that test to an integration test.

PR #65, opened 2016-12-05
DO NOT MERGE Fix integration tests without cache

KengoTODA:
Same with #61 but it does not cache build result in `.gradle` directory


KengoTODA:
This change cannot make PR stable, I will close.

PR #66, opened 2016-12-06
Analysis dependant variables should not be static

jsotuyod:
- For more details, see https://github.com/spotbugs/spotbugs/pull/61#issuecomment-265057575


KengoTODA:
I will test this branch with rebasing to `pr-61`, to confirm that this fix can stabilize Travis build.

PR #71, opened 2016-12-06
Add downloaded eclipse archive to cache

henrik242:
The eclipse binary might take a long time to download from eclipse.org.  The travis cache is probably a lot faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,9 @@ matrix:
            - oracle-java9-installer
      
 install:
-  - wget http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.6.2-201102101200/eclipse-SDK-3.6.2-linux-gtk-x86_64.tar.gz -O eclipse-3.6.2-linux64.tar.gz
-  - tar xzvf eclipse-3.6.2-linux64.tar.gz eclipse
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,8 +17,9 @@ matrix:
            - oracle-java9-installer
      
 install:
-  - wget http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.6.2-201102101200/eclipse-SDK-3.6.2-linux-gtk-x86_64.tar.gz -O eclipse-3.6.2-linux64.tar.gz
-  - tar xzvf eclipse-3.6.2-linux64.tar.gz eclipse
+  - mkdir -p deps
+  - wget -c http://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops/R-3.6.2-201102101200/eclipse-SDK-3.6.2-linux-gtk-x86_64.tar.gz -O deps/eclipse.tar.gz
```
KengoTODA: Great, I didn't know this `-c` option. Useful!

jsotuyod: this will still fail if eclipse returns a 404... I think we still need the `if [[ ! -e` from @KengoTODA version

Issue #75, opened 2016-12-07
Build fails on Eclipse plugin - missing local.properties


ThrawnCA:
I can't yet build the project; the ``eclipsePlugin`` module complains that I'm missing local.properties. Anyone have any suggestions? Am I meant to have copied it from another file?

jsotuyod:
It needs to point to an Eclipse plugin folder. Check [how we do it on Travis](https://github.com/spotbugs/spotbugs/blob/master/.travis.yml#L23)

ThrawnCA:
Umm...what if I (or whoever is building it) don't care about building an Eclipse plugin?

jsotuyod:
This was already discussed in #40. In the future we can tune the gradle build to either skip the eclipse plugin build or fail with a better error if the property is missing.

In the meantime, you can simply do `./gradlew :findbugs:build` to just build the library.

ThrawnCA:
OK, that works. Thanks.

Do you want me to work on said fine-tuning?

jsotuyod:
Feel free to do so if you are comfortable with Gradle! If you aren't don't worry, we will eventually get to that.

PR #77, opened 2016-12-07
skip Eclipse plugin build if config isn't present #75

ThrawnCA:


README.md
```diff
@@ -6,3 +6,11 @@
 
 
 More information at the [official website](https://spotbugs.github.io/). A lot of things can still be found at the [old FindBugs website](http://findbugs.sourceforge.net).
+
+# Build
+
+SpotBugs is built using Gradle. If you don't have Gradle installed, you can simply run the `gradlew` wrapper, which will automatically download Gradle if needed.
```
jsotuyod: actually, even with gradle installed, this is the recommended way. Otherwise, they would need a matching version of gradle (I think 3.1+ for the current config).

Issue #81, opened 2016-12-07
Dogfooding - should classes be Serializable?


ThrawnCA:
Running FindBugs over the SpotBugs codebase, there are 8 hits for SE_BAD_FIELD, ie a non-serializable field in a Serializable class. Do we really need these classes to be Serializable? If so, we should make it work properly; if not, we shouldn't declare it.

`edu.umd.cs.findbugs.BugInstance`
`edu.umd.cs.findbugs.ba.type.ExceptionSetFactory`
`edu.umd.cs.findbugs.gui2.MainFrameMenu`
`edu.umd.cs.findbugs.gui2.MainFrameTree`
`edu.umd.cs.findbugs.log.Profiler$ClassNameComparator`

jsotuyod:
Most likely than not, none of those needs to be `Serializable`. I think we can remove those `implements` and see what breaks. Either you will get a compile error, or a brand new `SE_BAD_FIELD` report somewhere else on the chain.

The only of those that seems to have any reason to be `Serializable` is `BugInstance` due to the cloud plugins, but that is gone now, so it's most likely useless.

mebigfatguy:
maybe BugInstance was for cloud? no idea

jsotuyod:
I've been looking into this, it seems to stem down from the fact that `MainFrame`, as an awt component, is `Seralizable`.

AWT serialization is not used. It's mainly intended for Applets (not our case) and may be used for a sort-of prevalence, but it's not our intended use case either.

We should probably just mark those fields as `transient` on `MainFrame`, implement `readObject` with something along the lines:

```java
    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
        in.defaultReadObject();
        // re-init non-serializable fields
        logger = new ConsoleLogger(this);
        backgroundExecutor = Executors.newCachedThreadPool();
        mainFrameInitialized = new CountDownLatch(1);
        ....
    }
```

mebigfatguy:
i think you can just rip it all out with impunity

jsotuyod:
Yeah, that could also be, but we would have to suppress the `SE_BAD_FIELD` warnings on the awt compontents.

@ThrawnCA will you take this over now or shall someone else do it?

ThrawnCA:
OK, I've removed the `serialVersionUID` from the class we changed. There are a bunch of other classes that define it but aren't `Serializable`; will get to them later.

jsotuyod:
Resolved on #85

PR #85, opened 2016-12-09
drop half-baked efforts to make things serializable #81

ThrawnCA:


findbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -110,7 +109,7 @@
  * @author David Hovemeyer
  * @see BugAnnotation
  */
-public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Serializable, Cloneable {
+public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Cloneable {
 
     private static final long serialVersionUID = 1L;
```
jsotuyod: we should drop the `serialVersionUID`s too

ThrawnCA:
OK, I've removed the `serialVersionUID` from the class we changed. There are a bunch of other classes that define it but aren't `Serializable`; will get to them later.

jsotuyod:
Awesome! Yeah, we should kill all those `serialVersionUID` and suppress `SE_BAD_FIELD` from gui components on a separate PR. Lots of dogfooding.

PR #88, opened 2016-12-10
Remove dead code

jsotuyod:
- The DockLayout was never compiled (not even on the original code), so
    we get rid of it and all related code.


h3xstream:
@jsotuyod I would like to be removed from the list of assignees in the future.
I know little about SpotBugs core. I will otherwise miss the reviews that I am really needed.

It would be nice to create [teams in the organisation](https://github.com/orgs/spotbugs/teams). So that it is possible to **[at]** mentions developers or admins.

jsotuyod:
@h3xstream np, sorry for the spam. I just added an admins, everyone and code-devs teams. Let me know if you think of any other teams that make sense to you.

PR #94, opened 2016-12-23
remove some deprecations around BCELs change from Constants to Const

mebigfatguy:
BCELs Constants was an interface, Const is now a class.

SpotBugs in places has 'implements Constants', which i didn't change as that could theoretically change the contract for plugins. Probably we should just go ahead and remove those implements Constants, and switch the rest to Const, but figured I'd break that up in case folks had a problem with that.


findbugs/src/main/java/edu/umd/cs/findbugs/ba/constant/Constant.java
```diff
@@ -94,7 +94,7 @@ public int getConstantInt() {
     }
 
     /**
-     * Merge two Constants.
+     * Merge two Const.
```
jsotuyod: Seearch & Replace hit a false positive here :fearful:

findbugs/src/main/java/edu/umd/cs/findbugs/ba/constant/ConstantFrame.java
```diff
@@ -21,7 +21,7 @@
 import edu.umd.cs.findbugs.ba.Frame;
 
 /**
- * Dataflow Frame for Constants.
+ * Dataflow Frame for Const.
```
jsotuyod: and here

findbugs/src/main/java/edu/umd/cs/findbugs/ba/constant/ConstantFrameModelingVisitor.java
```diff
@@ -33,7 +33,7 @@
  *
  * <p>
  * For now, only String constants are modeled. In the future we can add other
- * kinds of constants.
+ * kinds of Const.
```
jsotuyod: and here

findbugs/src/main/java/edu/umd/cs/findbugs/detect/MutableStaticFields.java
```diff
@@ -124,7 +124,7 @@ static boolean mutableSignature(String sig) {
      * Eclipse uses reflection to initialize NLS message bundles. Classes which
      * using this mechanism are usually extending org.eclipse.osgi.util.NLS class
      * and contains lots of public static String fields which are used as
-     * message constants. Unfortunately these fields cannot be final, so FB
+     * message Const. Unfortunately these fields cannot be final, so FB
```
jsotuyod: and here

findbugs/src/main/java/edu/umd/cs/findbugs/detect/Naming.java
```diff
@@ -380,7 +380,7 @@ private static boolean mightInheritFromException(ClassDescriptor d) {
      * Eclipse uses reflection to initialize NLS message bundles. Classes which
      * using this mechanism are usually extending org.eclipse.osgi.util.NLS class
      * and contains lots of public static String fields which are used as
-     * message constants. Unfortunately these fields often has bad names which
+     * message Const. Unfortunately these fields often has bad names which
```
jsotuyod: and here

PR #101, opened 2017-01-08
add `--add-opens` option to solve InaccessibleObjectException

KengoTODA:
New Java9 package `oracle-java9-installer=9b149+9b149arm-1~webupd8~0` was released at Jan/4th, and it introduced [a build error](https://github.com/spotbugs/spotbugs/pull/69#issuecomment-271070661) to spotbugs.

This pull-request solves it by adding `--add-opens` option to `GRADLE_OPTS` envvar.

PR #103, opened 2017-01-15
temporally drop build with java9, it will be back in #101

KengoTODA:
It seems that supporting Java9 has taken too long time, so let's disable it temporally to keep making progress.
I will continue working on #101 to enable the build with Java9.


KengoTODA:
This PR solves build errors in all other PR, please merge this first.

PR #104, opened 2017-01-21
WIP: Delete jar files from repository

KengoTODA:
Use normal dependency management to depend on other libraries.
Need to modify `distributions` in `findbugs/build.gradle` to keep packaging dependencies into dist zip file.

PR #107, opened 2017-01-24
Create a new subproject for annotations

KengoTODA:
Similar with #106, but it also contains:

* Stop making jar file which contains jcip annotations and jsr305 annotations. We've switched to transitive dependency.
* Delete `jsr305.jar` from `findbugs/lib`, and use `jsr305-3.0.1.jar` on Maven central instead.

Note that currently the artifactId for annotation jar is `spotbugs-annotations`.


findbugs/build.gradle
```diff
@@ -213,11 +130,12 @@ distributions {
 
         // TODO : These exclusions can be removed when we move dependencies from lib to dependency management
         exclude 'AppleJavaExtensions.jar'
+        exclude 'hamcrest-all-1.3.jar'
         exclude 'junit.jar'
 
         include '**/*.jar'
       }
-      from([jar, spotbugsAnnotationsJar]) {
+      from([jar]) {
```
jsotuyod: no need to use a list anymore

findbugs/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java
```diff
@@ -223,24 +228,16 @@ private void setUpEngine(String... analyzeMe) throws IOException {
...
-                throw new RuntimeException(e);
-            }
-            project.addAuxClasspathEntry(rootDirectory);
-        } else {
-            String[] cpParts = classpath.split(File.pathSeparator);
-            for (String cpStr : cpParts) {
-                File file = new File(cpStr);
-                if (file.isDirectory()) {
-                    project.addAuxClasspathEntry(file.getCanonicalPath());
+        File annotationsLib = new File("../spotbugs-annotations/build/libs");
```
jsotuyod: I'm not comfortable hardwiring this into the code. Specially when it's a build dir, and therefore a run from IDE may fail if the annotations project was cleaned from CLI.

What was wrong with the old code?

eclipsePlugin/build.gradle
```diff
@@ -41,9 +41,9 @@ dependencies {
   compile fileTree(dir:localProps.getProperty('eclipsePlugin.dir'), include:'**/*.jar',
     exclude:['**/datanucleus-enhancer*.jar', 'edu.umd.cs.findbugs.**/*.jar'])
 
-  includedLibs(project(path:':findbugs', configuration:'annotations')) {
-    transitive = false
-  }
+  includedLibs project(':spotbugs-annotations')
+  includedLibs 'com.google.code.findbugs:jsr305:3.0.1'
```
jsotuyod: if you do this, you should probably remove `jsr305.jar` from the libs folder, remove it from line 31 of this build script, and change the manifest accordingly.

findbugs/build.gradle
```diff
@@ -213,11 +130,12 @@ distributions {
 
         // TODO : These exclusions can be removed when we move dependencies from lib to dependency management
         exclude 'AppleJavaExtensions.jar'
+        exclude 'hamcrest-all-1.3.jar'
```
jsotuyod: who is adding hamcrest as part of the compile configuration? AFAIK it's only used as a test dependency.

findbugs/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java
KengoTODA: It couldn't find the jar file of spotbugs-annotation. So nullness check couldn't run as expected.

findbugs/build.gradle
KengoTODA: It comes from Lin directory.
https://github.com/spotbugs/spotbugs/tree/47d9a90d697542d1cef48a7a0d4104d6b128817d/findbugs/lib

Refs: f038a08505ed8dcdf6d1b85cccc371f6b69132a6

KengoTODA: I'll delete it in #104.

spotbugs-annotations/build.gradle
```diff
@@ -1,7 +1,5 @@
 apply from: "$rootDir/gradle/checkstyle.gradle"
 apply from: "$rootDir/gradle/javadoc.gradle"
-apply plugin: 'maven'
-apply plugin: 'signing'
```
jsotuyod: don't you need these plugins to be able to sign and publish to maven central the annotations artifacts?

findbugs/build.gradle
jsotuyod: Ok, great!

findbugs/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java
jsotuyod: That's weird... having the project dependency, it should be in the classpath for both Gradle and Eclipse / IntelliJ runs.... we definitely need some way around this

KengoTODA: I guess we can simply delete `if (file.isDirectory())`, because gradle added jar file path to CLASSPATH.

```java
System.out.println("CP:" + System.getProperty("java.class.path"));
->
    ~/spotbugs/findbugs/build/classes/test:
    ...
    ~/spotbugs/spotbugs-annotations/build/libs/spotbugs-annotations-3.1.0-SNAPSHOT.jar
    ...
```

findbugs/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java
```diff
@@ -237,9 +236,7 @@ private void setUpEngine(String... analyzeMe) throws IOException {
             String[] cpParts = classpath.split(File.pathSeparator);
             for (String cpStr : cpParts) {
                 File file = new File(cpStr);
-                if (file.isDirectory()) {
-                    project.addAuxClasspathEntry(file.getCanonicalPath());
-                }
+                project.addAuxClasspathEntry(file.getCanonicalPath());
```
jsotuyod: have you tested this running from the IDE? Just curious, I clearly runs from Gradle... I don't remember if it was getting this way, or `classpath` was null on that scenario

findbugs/build.gradle
```diff
@@ -286,19 +204,6 @@ publishing {
         classifier 'javadoc'
       }
     }
-
-    annotations(MavenPublication) {
```
jsotuyod: we do need publishing config / maven + signing plugins in the new project to publish annotations artifacts.

jsotuyod:
This is looking really good

findbugs/build.gradle
KengoTODA: Yes we need. I'm working on another branch `issue-30`.

https://github.com/KengoTODA/spotbugs/commits/issue-30

findbugs/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java
KengoTODA: It works in my Eclipse.
But I have another problem: `test-harness` project introduced a cyclic dependency.

Gradle has no problem because it computes dependency based on tasks. However Eclipse computes it based on projects...

We should specify fixed version instead of `project(':test-harness')`. I'll handle this after we release 3.1.0.

KengoTODA:
I've commented to all review comments, please check and merge.

PR #109, opened 2017-01-26
Delete jar files in findbugs/lib directory

KengoTODA:
Almost same with #104, but this pull request does not delete jar files in `findbugsTestCases` because these files are necessary for `AbstractIntegrationTest` class in `:test-harness` project.

This change is necessary to generate `pom.xml` which contains proper `<dependency>` elements.


henrik242:
@KengoTODA Just a hint: You could have just force pushed onto the same branch as in the #104, and the PR would have updated accordingly.

KengoTODA:
@henrik242 OK I will reuse PR from next.

PR #115, opened 2017-02-02
Make test harness portable

KengoTODA:
Currently `AbstractIntegrationTest` in `test-harness` has several hard-coded value which depends on project structure of this repository, such as `"../findbugsTestCases"`, `"lib"` and `"/build/classes/main/"`. To make this library portable, it's better to remove them.

And one more problem, is that, we provide utilities as abstract class so it might be hard to use with another solution which uses the same way. Basically `Rule` is better choice for JUnit.

This pull request proposes:

* several refactorings
    * extract `CountMatcher`
    * extract `AnalysisRunner`
* create new `ExternalResource` named `SpotBugsRule`
* move `AbstractIntegrationTest` to `findbugs` subproject
    * this class shouldn't be a part of test-harness, because it contains hard-coded values
* use `SpotBugsRule` in `FindUnsatisfiedObligationTest`


test-harness/src/main/java/edu/umd/cs/findbugs/AnalysisRunner.java
```diff
@@ -0,0 +1,81 @@
+package edu.umd.cs.findbugs;
```
jsotuyod: we should probably move this to a more appropriate package... maybe `edu.umd.cs.findbugs.test`?

test-harness/src/main/java/edu/umd/cs/findbugs/CountMatcher.java
```diff
@@ -0,0 +1,40 @@
+package edu.umd.cs.findbugs;
```
jsotuyod: idem

test-harness/src/main/java/edu/umd/cs/findbugs/SpotBugsRule.java
```diff
@@ -0,0 +1,118 @@
+package edu.umd.cs.findbugs;
```
jsotuyod: idem

KengoTODA:
@jsotuyod Renamed at c32aabc. This change needed to change visibility of several legacy classes, please check.

PR #116, opened 2017-02-04
WIP: Close #91 (UselessCode FP)

### LINKED ISSUE #91: UselessCode FP
the useless code detector fires when a condition is known only from an assert. It should ignore asserts as pre-conditions.

ie

           assert cmp <= 0;

            if (cmp < 0)
            {
                break; 
            }
            else if (cmp == 0)   // <-- UC Fires here
            {
                 ....
            }

KengoTODA:
This PR contains two changes:

* reproduced problem by JUnit
* solve problem by changing `ASSUME_ASSERTIONS_ENABLED` to false

The 2nd change would affect other detectors, I need to check carefully.

### DIFF
--- a/findbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/bcel/CFGFactory.java
+++ b/findbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/bcel/CFGFactory.java
@@ -118,7 +118,7 @@ public class CFGFactory extends AnalysisFactory<CFG> {
 
         // Remove CFG edges corresponding to failed assertions.
         boolean changed = false;
-        boolean ASSUME_ASSERTIONS_ENABLED = true;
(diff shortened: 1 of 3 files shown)


KengoTODA:
As 362c9fb4 (not in this branch) I created a sub-class of `CFG` to let specific detector ignores `assert`, but it failed.
It seems that the problem exists in product architecture: analysis cache doesn't suppose the possibility to have multiple kinds of `CFG`. I think it's really hard to introduce this way which can keep backward compatibility of other detectors.

So we should choose other methodology, for instance:
* Use `SystemProperties.ASSERTIONS_ENABLED` instead of `ASSUME_ASSERTIONS_ENABLED`
* Create an option (system properties?) to switch the value of `ASSUME_ASSERTIONS_ENABLED`

Note that both of above affects all detectors which depends on `CFG`.
Any other ideas are welcome :)

PR #125, opened 2017-02-13
Fixes #123 - restore Version.RELEASE

### LINKED ISSUE #123: Errors in findbugs-maven-plugin
@jsotuyod can I ask you to handle this? Maybe this is caused by 7fb6ff52bb20363a2684be64e845753f909e1b54.
I tested with `findbugs-maven-plugin:3.0.4`. You can run `mvn findbugs:findbugs` and `mvn site` in https://github.com/KengoTODA/findbugs-slf4j/tree/spot-bugs-3.1 to test.

----

Current `master` cannot work with `findbugs-maven-plugin`. `findbugs:findbugs` works but at `site` goal, it reports following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.4:site (default-site) on project test-case: Execution default-site of goal org.apache.maven.plugins:maven-site-plugin:3.4:site failed: No such property: RELEASE for class: edu.umd.cs.findbugs.Version
```

It seems that we've accidentally removed a public field in `Version` class. Refs: https://github.com/findbugsproject/findbugs/blob/94ec4ad/findbugs/src/java/edu/umd/cs/findbugs/Version.java#L133

----

And generated `findbugsXml.xml` has `version="(unknown)"`, this is caused by NPE:

```
     [java] java.lang.NullPointerException
     [java] 	at java.util.jar.Manifest$FastInputStream.fill(Manifest.java:441)
     [java] 	at java.util.jar.Manifest$FastInputStream.readLine(Manifest.java:375)
     [java] 	at java.util.jar.Manifest$FastInputStream.readLine(Manifest.java:409)
     [java] 	at java.util.jar.Attributes.read(Attributes.java:376)
     [java] 	at java.util.jar.Manifest.read(Manifest.java:199)
     [java] 	at java.util.jar.Manifest.<init>(Manifest.java:69)
     [java] 	at edu.umd.cs.findbugs.Version.<clinit>(Version.java:49)
```

It seems that `Version.class.getResourceAsStream("META-INF/MANIFEST.MF")` in Version.java returns `null`. I confirmed that `MANIFEST.MF` is packaged to `spotbugs.jar`.

I tried to change `"META-INF/MANIFEST.MF"` to `"/META-INF/MANIFEST.MF"`, but it cannot be solution because it doesn't load file from `spotbugs.jar`. Here is entries in loaded manifest:

```
     [java] Ant-Version: Apache Ant 1.8.4
     [java] Created-By: 1.6.0_65-b14-466-11M4802 (Apple Inc.)
     [java] Manifest-Version: 1.0
```

I guess it loaded data from Maven plugin itself.

jsotuyod:
- Mark it as deprecated
 - On next version it should be removed for good.
 - Fixes #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v0.3](http://
 ### Deprecated
 
 * `OpcodeStack.Item.defineNewSpecialKind(String)` ([#27](https://github.com/spotbugs/spotbugs/pull/27))
+* `Version.RELEASE` ([#125](https://github.com/spotbugs/spotbugs/pull/125))
 
 ### Removed
 
--- a/findbugs/src/main/java/edu/umd/cs/findbugs/Version.java
+++ b/findbugs/src/main/java/edu/umd/cs/findbugs/Version.java
@@ -35,6 +35,12 @@ public class Version {
     public static final String WEBSITE = "https://spotbugs.github.io/";
     
     public final static String VERSION_STRING;
+
+    /**
+     * @deprecated Use {@link #VERSION_STRING} instead.
+    */
+    @Deprecated
+    public final static String RELEASE;
     
     private static String applicationName = "";
     private static String applicationVersion = "";
@@ -55,7 +61,7 @@ public class Version {
             version = "Development";
         }
         
-        VERSION_STRING = version;
+        RELEASE = VERSION_STRING = version;
     }
 
     public static void registerApplication(String name, String version) {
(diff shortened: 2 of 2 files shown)


KengoTODA:
May we need to apply some solution like http://stackoverflow.com/a/1273196

jsotuyod:
@KengoTODA sorry! I hadn't noticed the second issue.

Yes, it looks like doing that is going to be needed. Possibly adding checks to make sure we are on the correct manifest... I'll try and think about something.

KengoTODA:
Yes, thank you!

One more note: CHANGELOG.md has a place to list deplications, please add Versions.RELEASE to there.

jsotuyod:
@KengoTODA there you go! I believe it should work now...

KengoTODA:
Great work @jsotuyod! I confirmed that:

* `version` attribute in `findbugsXml.xml` has been filled as expected
* `site` goal can generate HTML report as expected

PR #128, opened 2017-02-18
[performance improvement] Delete outdated hack for legacy JVM

KengoTODA:
According to my benchmark, SpotBugs is 6% slower than FindBugs 3.0.1. And some of bottlenecks can be solved easily without adding new dependency nor upgrading library. This pull-request proposes these easy fixes.

## Benchmark script

I use Google Guava as target, because it has no big dependencies but it depends on JSR305 so we can use almost full-set feature of SpotBugs.

```
#!/bin/bash -e
for i in {1..20}; do
  /usr/bin/time java -Xmx2g -Xms2g \
  -jar findbugs/build/distributions/spotbugs-3.1.0-SNAPSHOT/lib/spotbugs.jar \
  -textui -auxclasspath ~/.m2/repository/com/google/code/findbugs/jsr305/3.0.1/jsr305-3.0.1.jar \
  -output /dev/null ~/.m2/repository/com/google/guava/guava/19.0/guava-19.0.jar 2>&1 | \
  grep real | sed 's/ \{1,\}/ /g' | cut -d ' ' -f 2,4,6
done;
```

## Note

* I use this script on OS X, your `time` might not be the same one if you're using different OS.
* I run benchmark 20 times, and use its MEDIAN as performance indicator.
* After applying these optimizations, bottleneck still remains especially in collection handling. I will list the big one as follows, and will consider to introduce Guava, FastUtil or other libraries.
    * edu.umd.cs.findbugs.classfile.DescriptorFactory.getMethodDescriptor(String, String, String, boolean)
        * We creates instance first, and check the existing instance in collection.
        * `HashMap#get(key)` costs much time even though `MethodDescriptor` implements `hashCode()` and `equals()` in acceptable quality.
        * Deleting this cache doesn't improve performance (26.2 sec).
    * edu.umd.cs.findbugs.classfile.impl.AnalysisCache.getClassAnalysis(Class, ClassDescriptor)	
        * `HashMap#get(key)` costs much time even though `ClassDescriptor` implements `hashCode()` and `equals()` in acceptable quality.
    * edu.umd.cs.findbugs.classfile.DescriptorFactory.getClassDescriptor(String)
        * `HashMap#get(key)` costs much time even though we use `String` as key.
    * edu.umd.cs.findbugs.classfile.analysis.ClassInfo.findMethod(String, String, boolean)
        * We use `O(n)` algorithm to search. Better to introduce [Table](https://github.com/google/guava/wiki/NewCollectionTypesExplained#table) or some other collection to search by `O(log(n))` algorithm.
* About detailed report, please check [SpotBugsPerformance_3adfcbf.ods.zip](https://github.com/spotbugs/spotbugs/files/784678/SpotBugsPerformance_3adfcbf.ods.zip)

### DIFF
--- a/findbugs/src/main/java/edu/umd/cs/findbugs/ba/Frame.java
+++ b/findbugs/src/main/java/edu/umd/cs/findbugs/ba/Frame.java
@@ -77,7 +77,7 @@ public abstract class Frame<ValueType> {
     /**
      * Array storing the values of local variables and operand stack slots.
      */
-    private final ArrayList<ValueType> slotList;
+    private ArrayList<ValueType> slotList;
 
     /**
      * Flag marking this frame as a special "TOP" value. Such Frames serve as
@@ -605,23 +605,8 @@ public abstract class Frame<ValueType> {
      */
     public void copyFrom(Frame<ValueType> other) {
         lastUpdateTimestamp = other.lastUpdateTimestamp;
-        if (true) {
-            int size = slotList.size();
-            if (size == other.slotList.size()) {
-                for (int i = 0; i < size; i++) {
-                    slotList.set(i, other.slotList.get(i));
-                }
-            } else {
-                slotList.clear();
-                for (ValueType v : other.slotList) {
-                    slotList.add(v);
-                }
-            }
-        } else {
-            slotList.clear();
-            slotList.addAll(other.slotList);
+        slotList = new ArrayList<>(other.slotList);
 
-        }
         /*
          * Andrei, 27.02.2008: "optimized" code below takes ~18% overall FB
          * execution time, code above only 5% int size = slotList.size(); if
--- a/findbugs/src/main/java/edu/umd/cs/findbugs/ba/Frame.java
+++ b/findbugs/src/main/java/edu/umd/cs/findbugs/ba/Frame.java
@@ -606,14 +606,6 @@ public abstract class Frame<ValueType> {
     public void copyFrom(Frame<ValueType> other) {
         lastUpdateTimestamp = other.lastUpdateTimestamp;
         slotList = new ArrayList<>(other.slotList);
-
-        /*
-         * Andrei, 27.02.2008: "optimized" code below takes ~18% overall FB
-         * execution time, code above only 5% int size = slotList.size(); if
-         * (size == other.slotList.size()) { for (int i = 0; i < size; i++)
-         * slotList.set(i, other.slotList.get(i)); } else { slotList.clear();
-         * for (ValueType v : other.slotList) slotList.add(v); }
-         */
         isTop = other.isTop;
         isBottom = other.isBottom;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v0.3](http://
 
 * `OpcodeStack.Item.defineNewSpecialKind(String)` ([#27](https://github.com/spotbugs/spotbugs/pull/27))
 * `Version.RELEASE` ([#125](https://github.com/spotbugs/spotbugs/pull/125))
+* `DescriptorFactory.canonicalizeString(String)` ([#128](https://github.com/spotbugs/spotbugs/pull/128))
 
 ### Removed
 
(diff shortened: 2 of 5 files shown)


jsotuyod:
I wonder how we are 6% slower if we didn't add code, but on the contrary killed quite some...

This PR covers the contents of https://github.com/findbugsproject/findbugs/pull/132 regarding string canonicalization (original implementation), I strongly support it.

The rest is an alternative of part (not all) the changes suggested on https://github.com/findbugsproject/findbugs/pull/135, you may want to look at the other changes proposed in that PR.

https://github.com/findbugsproject/findbugs/pull/134 also introduces performance improvements (avoid an array allocation and iniialization).

findbugs/src/main/java/edu/umd/cs/findbugs/ba/Frame.java
```diff
@@ -605,23 +605,8 @@ public boolean sameAs(Frame<ValueType> other) {
...
-                }
-            }
-        } else {
-            slotList.clear();
-            slotList.addAll(other.slotList);
+        slotList = new ArrayList<>(other.slotList);
 
-        }
         /*
          * Andrei, 27.02.2008: "optimized" code below takes ~18% overall FB
```
jsotuyod: this comment should be removed if the implementation is dropped.

KengoTODA:
I run benchmark with Apache Cassandra 3.1.0, and confirmed that this change makes SpotBugs slow (284 sec -&gt; 338 sec).  It seems that [@iloveeclipse 's suggestion](https://github.com/findbugsproject/findbugs/pull/132#issuecomment-254489753) is nice, I will use system property to switch strategy.

KengoTODA:
> This PR covers the contents of findbugsproject/findbugs#132 regarding string canonicalization (original implementation)

I've run the same benchmark with Apache Cassandra 3.1.0, which I believe that really big project (6.3 MB, about 3,600 classes). This optimisation could improve performance from 284s to 265s. I think we can merge this fix if we have no another OSS which can prove potential problem.

> https://github.com/findbugsproject/findbugs/pull/134/files
I will try to introduce this change to this branch.

KengoTODA:
findbugsproject/findbugs/pull/134 doesn't affect performance, but it's more intuitive implementation so I will add it to this pull-request.

KengoTODA:
> * edu.umd.cs.findbugs.classfile.analysis.ClassInfo.findMethod(String, String, boolean)
>     * We use O(n) algorithm to search. Better to introduce Table or some other collection to search by O(log(n)) algorithm.

I tried introducing Table in a9ac1838, but it cannot improve performance (25.83s -&gt; 26.46s) in microbench with Guava 19.0.

Refs: [SpotBugsPerformance_a9ac1838.o.zip](https://github.com/spotbugs/spotbugs/files/1016269/SpotBugsPerformance_a9ac1838.ods.zip)

jsotuyod:
@KengoTODA I'm theorizing here, but the main key for the table is the method name, under high overloading this means several conflicts... Also, 2 nested maps + an array means lots of jumps in memory to get to the value...

Possibly simplifying the table to:

```java
class MethodDescriptor {
  private final String name;
  private final String signature;
  private final boolean static;

  // hashCode() and equals()
}

Map<MethodDescrptor, XMethod> table = new HashMap<>();
```

may produce better results. But there is only one way to know for sure...

KengoTODA:
note: I run a microbenchmark on cloud (could be unstable), and it says that `SpotBugs 3.1.12 & 4.0.6` is faster than `FindBugs 3.0.1` (4% slower) and `SpotBugs 4.1.4` (1% slower).

https://github.com/KengoTODA/spotbugs-benchmark/runs/1362695598?check_suite_focus=true#step:7:36

```
Benchmark #1: SpotBugs 4.1.4
  Time (mean ± σ):     69.282 s ±  1.517 s    [User: 134.499 s, System: 1.448 s]
  Range (min … max):   67.957 s … 70.950 s    5 runs
 
Benchmark #2: SpotBugs 4.0.6
  Time (mean ± σ):     68.837 s ±  2.301 s    [User: 134.689 s, System: 1.511 s]
  Range (min … max):   66.999 s … 72.762 s    5 runs
 
Benchmark #3: SpotBugs 3.1.12
  Time (mean ± σ):     68.759 s ±  1.073 s    [User: 134.635 s, System: 1.490 s]
  Range (min … max):   67.019 s … 69.968 s    5 runs
 
Benchmark #4: FindBugs 3.0.1
  Time (mean ± σ):     71.121 s ±  1.054 s    [User: 139.361 s, System: 1.468 s]
  Range (min … max):   69.945 s … 72.370 s    5 runs
 
Summary
  'SpotBugs 3.1.12' ran
    1.00 ± 0.04 times faster than 'SpotBugs 4.0.6'
    1.01 ± 0.03 times faster than 'SpotBugs 4.1.4'
    1.03 ± 0.02 times faster than 'FindBugs 3.0.1'
```

Issue #60, opened 2016-12-05
ClassFormatException: Constant pool at index 0 is null


barney2k7:
The problem described in findbugsproject/findbugs#141 also affects spotbugs, I just verified it with a current snapshot. The analysis of the following code fails:
```
import java.io.IOException;
import java.io.InputStream;
import java.net.URL;
import java.util.Properties;
import java.util.stream.Stream;

public class MyClass {
  
  public static void create(URL url) {
    try (InputStream in = url.openStream()) {
      Properties p1 = new Properties();
      p1.load(in);
    } catch (IOException e) {
    }
  }
  
  public Stream<String> keys() {
    return Stream.<Properties>of().flatMap(p -> p.stringPropertyNames().stream());
  }

}
```

Stacktrace:
```
The following errors occurred during analysis:
  Exception analyzing MyClass using detector edu.umd.cs.findbugs.detect.FindUnsatisfiedObligation
    org.apache.bcel.classfile.ClassFormatException: Constant pool at index 0 is null.
      At org.apache.bcel.classfile.ConstantPool.getConstant(ConstantPool.java:258)
      At org.apache.bcel.classfile.ConstantPool.getConstantString(ConstantPool.java:293)
      At org.apache.bcel.generic.FieldOrMethod.getReferenceType(FieldOrMethod.java:116)
      At edu.umd.cs.findbugs.ba.obl.InstructionActionCache.getActions(InstructionActionCache.java:107)
      At edu.umd.cs.findbugs.ba.obl.ObligationAnalysis.transferInstruction(ObligationAnalysis.java:139)
      At edu.umd.cs.findbugs.ba.obl.ObligationAnalysis.transferInstruction(ObligationAnalysis.java:68)
      At edu.umd.cs.findbugs.ba.AbstractDataflowAnalysis.transfer(AbstractDataflowAnalysis.java:135)
      At edu.umd.cs.findbugs.ba.obl.ObligationAnalysis.transfer(ObligationAnalysis.java:160)
      At edu.umd.cs.findbugs.ba.obl.ObligationAnalysis.transfer(ObligationAnalysis.java:68)
      At edu.umd.cs.findbugs.ba.Dataflow.execute(Dataflow.java:376)
      At edu.umd.cs.findbugs.classfile.engine.bcel.ObligationDataflowFactory.analyze(ObligationDataflowFactory.java:75)
[...]

PR #129, opened 2017-02-18
refs #6: support INVOKEDYNAMIC at DismantleBytecode

KengoTODA:
Note that I cannot find any other place which needs to add `INVOKEDYNAMIC`, but I lack the view of whole SpotBugs architecture so we may have more classes to change.


KengoTODA:
It seems that this change breaks backward compatibility, then I will close this PR.

jsotuyod:
@KengoTODA interestingly, the 2 failing tests are for #20 and #60, both fixes introduced by us on SpotBugs... if we fixed them some other way, allowing this change to stand, I think it would be cool regarding backwards compatibility.

PR #132, opened 2017-02-20
Update pom.xml to add name, description and developer information

KengoTODA:
This pull-requests adds developer information, name and description into pom.xml. Please check developer information in `gradle/maven.gradle`, and let me know if you want to change/fix it.

One more change exists in `test-harness/build.gradle`. This module should be licensed under LGPL 3.0 in my understanding.

### DIFF
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -36,3 +36,14 @@ task sourcesJar(type: Jar) {
 artifacts {
   archives jar, javadocJar, sourcesJar
 }
+
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        name 'SpotBugs Annotations'
+        description 'Annotation the SpotBugs tool supports'
+      }
+    }
+  }
+}
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -42,7 +42,7 @@ uploadArchives {
     mavenDeployer {
       pom.project {
         name 'SpotBugs Annotations'
-        description 'Annotation the SpotBugs tool supports'
+        description 'Annotations the SpotBugs tool supports'
       }
     }
   }
--- a/spotbugs-ant/build.gradle
+++ b/spotbugs-ant/build.gradle
@@ -33,7 +33,7 @@ uploadArchives {
     mavenDeployer {
       pom.project {
         name 'SpotBugs Ant Task'
-        description 'AntTask to run SpotBugs'
+        description 'Ant Task to run SpotBugs'
       }
     }
   }
--- a/spotbugs-ant/build.gradle
+++ b/spotbugs-ant/build.gradle
@@ -27,3 +27,14 @@ task sourcesJar(type: Jar) {
 artifacts {
   archives jar, javadocJar, sourcesJar
 }
+
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        name 'SpotBugs Ant Task'
+        description 'AntTask to run SpotBugs'
+      }
+    }
+  }
+}
--- a/test-harness/build.gradle
+++ b/test-harness/build.gradle
@@ -26,3 +26,20 @@ task sourcesJar(type: Jar) {
 artifacts {
   archives jar, javadocJar, sourcesJar
 }
+
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        licenses {
+          // This project contains copied source files from SonarSource and find-sec-bugs,
+          // which is licensed under LGPL 3.0, so we should use v3.0 instead of 2.1.
+          license {
+            name 'GNU LESSER GENERAL PUBLIC LICENSE, Version 3.0'
+            url 'https://www.gnu.org/licenses/lgpl-3.0.en.html'
+          }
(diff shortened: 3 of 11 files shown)


test-harness/build.gradle
```diff
@@ -26,3 +26,23 @@ task sourcesJar(type: Jar) {
...
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        name 'Test Harness for SpotBugs Plugin'
+        description 'A test harness library for SpotBugs plugin developers'
+
+        licenses {
+          // This project contains copied source files from SonarSource and find-sec-bugs,
+          // which is licensed under LGPL 3.0, so we should use v3.0 instead of 2.1.
```
KengoTODA: @h3xstream please check this. Is my understanding correct?

KengoTODA:
note: https://github.com/KengoTODA/spotbugs/commit/d677b56c6af9a6f77ed5452f77623b0c10788ac7 will be tagged as `3.1.0-rc1`, after this branch is merged to `master` and we deployed artifacts to central.

test-harness/build.gradle
h3xstream: find-sec-bugs is not just me. But the code use was the small test utility (hamcrest matchers mostly) which I was only the only contributor. I gave authorization to @jsotuyod to transfer the code to spotbugs.

2017-02-20 11_47_33-improve infrastructure for better integration tests by jsotuyod pull request

-----

The code use from SonarSource is actually not under LGPL3.0. It is a SmapParser that was reused. Its main license is just to keep a header mentioning the original author.
https://github.com/spotbugs/spotbugs/blob/1aa089627fb3639634d7cfa4dbfeaf063e9917cd/findbugs/src/junit/org/sonar/plugins/findbugs/resource/SmapParser.java#L39

Original source: http://stackoverflow.com/a/11299757

This segment could be removed : https://github.com/spotbugs/spotbugs/blob/1aa089627fb3639634d7cfa4dbfeaf063e9917cd/findbugs/src/junit/org/sonar/plugins/findbugs/resource/SmapParser.java#L2

The few changes were made allow the support additional JSP compilers. I do not recommend using the original Stackoverflow source.

In conclusion, the dependency to Sonar-FindBugs license should not exist. Only the reference to Michael Schierl in the source.

----

It may not be 100% legally accurate. But since I produce the test utility and the small fixes to the SMAPParser, I can assure I won't sue the project. ;)

jsotuyod: :+1: I guess that counts as written authorization :P. I guess this is the only change then.

KengoTODA: Thanks for reply!

> This segment could be removed : https://github.com/spotbugs/spotbugs/blob/1aa089627fb3639634d7cfa4dbfeaf063e9917cd/findbugs/src/junit/org/sonar/plugins/findbugs/resource/SmapParser.java#L2

Oh I didn't find that this java file has two comments about license... I will remove the 1st one (LGPL 3.0 for SonarQube Findbugs Plugin), and keep the 2nd one (Copyright (c) 2012 Michael Schierl).

KengoTODA: @h3xstream So, in my understanding, we can distribute this module (test utility) under LGPL 2.1 which is used for other modules. Then I will remove this `license { ... }` from `test-harness/build.gradle`

spotbugs-annotations/build.gradle
```diff
@@ -36,3 +36,14 @@ task sourcesJar(type: Jar) {
 artifacts {
   archives jar, javadocJar, sourcesJar
 }
+
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        name 'SpotBugs Annotations'
+        description 'Annotation the SpotBugs tool supports'
```
ThrawnCA: Should say 'Annotations', plural.

spotbugs-ant/build.gradle
```diff
@@ -27,3 +27,14 @@ task sourcesJar(type: Jar) {
 artifacts {
   archives jar, javadocJar, sourcesJar
 }
+
+uploadArchives {
+  repositories {
+    mavenDeployer {
+      pom.project {
+        name 'SpotBugs Ant Task'
+        description 'AntTask to run SpotBugs'
```
ThrawnCA: 'Ant task' should be two words.

test-harness/src/main/java/org/sonar/plugins/findbugs/resource/SmapParser.java
```diff
@@ -1,22 +1,3 @@
-/*
```
ThrawnCA: Is it appropriate for us to drop this licence info?

spotbugs-annotations/build.gradle
KengoTODA: I copied from FindBugs pom.xml, but yes you're right. I will fix!

test-harness/src/main/java/org/sonar/plugins/findbugs/resource/SmapParser.java
KengoTODA: I think so, please refer https://github.com/spotbugs/spotbugs/pull/132#discussion_r102060390

In my understanding this class is licensed by Michael Schierl, not SonarSource.

PR #138, opened 2017-02-22
add Javadoc badges

KengoTODA:
refs: https://github.com/spotbugs/spotbugs.github.io/pull/4#issuecomment-275974136


jsotuyod:
I recommend caution to not add too many badges. Badges should carry important info, if everything is important, nothing is.

I'm approving this one, but if we decide to add more, we should probably review carefully which ones add the most value to users hitting this repository.

PR #140, opened 2017-02-23
Support plugin development by test harness

KengoTODA:
Current implementation supports SpotBugs core plugin only, to support other plugin developers, we need to load plugin jar file in advance.

Note that SpotBugs manages loaded plugin in static fields, so we need this hack in static initializer

And currently we have no API to create Plugin instance without loading it,
so this code catches `DuplicatePluginIdException` to judge we load core plugin or not.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/Plugin.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/Plugin.java
@@ -637,6 +637,11 @@ public class Plugin {
 
     /**
      * Loads the given plugin and enables it for the given project.
+     * 
+     * @param f
+     *      A non-null jar file of custom plugin.
+     * @param project
+     *      A nullable target project
      */
     public static Plugin loadCustomPlugin(File f, @CheckForNull Project project)
             throws PluginException {
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/SortedBugCollection.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/SortedBugCollection.java
@@ -47,6 +47,7 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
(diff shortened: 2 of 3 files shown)


jsotuyod:
> Current implementation supports SpotBugs core plugin only, to support other plugin developers, we need to load plugin jar file in advance.

@KengoTODA could you elaborate on this? I'm not sure I follow you through.

KengoTODA:
@jsotuyod Current implementation in `develop` branch does not load plugin, so it cannot test any plugin implementation in `src/main/java`.

To let SpotBugs loads plugin, we need to prepare a `.jar` file which contains necessary resources such as `findbugs.xml` and `message.xml`, then we can invoke `Plugin. loadCustomPlugin(File, Project)`. For instance, test-util in find-sec-bug loads plugin at [here](https://github.com/find-sec-bugs/find-sec-bugs/blob/17da3c9/findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/FindBugsLauncher.java#L96-L108).

jsotuyod:
Ok, now I get it. Looks good.

Any reason for not doing this as part of a JUnit `@Rule`? That way we could automatically have it remove the temp file once the test suite finishes, and would only apply to tests using it (plugin developers), but not to core development.

KengoTODA:
> Any reason for not doing this as part of a JUnit @Rule?

It is really good, however, currently SpotBugs loads plugin information to static field so it's little hard to handle. 

If we want to unload plugin after each test method execution, we need to call [Plugin.removeCustomPlugin(Plugin)](https://github.com/spotbugs/spotbugs/blob/d1e60f8dbeda0a454f2d497ef8dcb878fa8e3852/findbugs/src/java/edu/umd/cs/findbugs/Plugin.java#L702), or create a new classloader for each test execution.

I will check how calling `removeCustomPlugin(Plugin)` affects performance.

KengoTODA:
Oops, I found that we need to consider parallel JUnit execution. And to support it, we cannot remove custom plugin after each test execution. If we support both of parallel execution and plugin unloading, we need to create a class loader for each test thread but it should affects test method behaviour and performance.

Ideally we should not use static fields to manage state of Plugin, I will issue a ticket to enhance.
And for this PR, I will keep current implementation.

KengoTODA:
@jsotuyod Could you recheck and merge this?

KengoTODA:
@jsotuyod Thanks!

PR #145, opened 2017-02-28
Summarise how we release the fixed version of SpotBugs

KengoTODA:
We have several kinds of product to release, let me summarise how-to.

About `TBU` part, I will try to research and explain it. If you already know, please help me.


RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,29 @@
+# Release procedure
+
+When you release fixed version of SpotBugs, please follow these procedures.
+
+## Release to Maven Central
+
+Add necessary properties to `gradle.properties` and run `gradle clean build smoketest uploadArchives`.
```
jsotuyod: clean shouldn't be necessary, Gradle is smart enough

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,29 @@
...
+Make sure that you have proper `eclipsePlugin/local.properties` file to release Eclipse plugin at the same time.
```
jsotuyod: Maybe this should be before running gradle

RELEASE_PROCEDURE.md
jsotuyod: Prefer `./gradlew` over system version of `gradle`

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Check [SonaType official page](http://central.sonatype.org/pages/gradle.html) for detail.
+
+## Release to Eclipse Marketplace
+
+TBU
```
jsotuyod: @iloveeclipse is the most (only one?) experienced on this

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,29 @@
...
+## Release to Eclipse Update Site
+
+Send pull-request to spotbugs/spotbugs.github.io, to update contents.
+As files to upload, you can use the zip file which is uploaded to Maven Central when you release to there.
+
+See [this pull-request](https://github.com/spotbugs/spotbugs.github.io/pull/12) as example.
+
+## Release to Gradle Plugin Portal
+
+TBU
```
jsotuyod: this one should be `./gradlew publishPlugins`, but it needs to have the api key on a gradle properties (not in git obviously), usually that is `~/.gradle/gradle.properties`. AFAIK the plugins on the portal are owned by a single account (that is, no teams), so we should create a "spotbugs" account, and share the credentials out-of-band (ie: private email)

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Release to ReadTheDocs
+
+TBU
```
jsotuyod: I was hoping you knew how to do this....

RELEASE_PROCEDURE.md
iloveeclipse: That is now changed. So all what we need are:

1. Update site URL with actual plugin update site (not zip, not maven site)
2. Unique feature id (distinct from the old one)
3. For the maintainers: account at eclipse.org
4. Once the plugin entry at marketplace is created, the account owner can invite others to maintain it. They only need an account name at eclipse.org. So *all* members in the team with eclipse.org account *can* maintain the plugin entry.

KengoTODA: > we should create a "spotbugs" account, and share the credentials out-of-band (ie: private email)

Who can handle this? Can I ask you @jsotuyod?

KengoTODA: > Update site URL with actual plugin update site (not zip, not maven site)

Will be https://spotbugs.github.io/eclipse/

> Unique feature id (distinct from the old one)

For now we haven't change this yet. IMO it's better to change.

> For the maintainers: account at eclipse.org

I have, its mail address is `skypencil@gmail.com`

> Once the plugin entry at marketplace is created, the account owner can invite others to maintain it. They only need an account name at eclipse.org. So all members in the team with eclipse.org account can maintain the plugin entry.

I'll try to handle this later.

jsotuyod: Agreed, we should change the eclipse plugin id, this would also resolve #147.

Yes, I'll create a spotbugs account on Gradle plugin portal and email it to you.

iloveeclipse: For the marketplace, **feature** id need to be changed only. In theory, we can let the plugin id as is and change it in the next release.

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,34 @@
...
+
+Send pull-request to spotbugs/spotbugs.github.io, to update contents.
+As files to upload, you can use the zip file which is uploaded to Maven Central when you release to there.
+
+See [this pull-request](https://github.com/spotbugs/spotbugs.github.io/pull/12) as example.
+
+## Release to Eclipse Marketplace
+
+No action necessary. Just push latest plugin to Eclipse Update Site then it's enough.
+If you need to update [entry at Eclipse Marketplace](https://marketplace.eclipse.org/content/spotbugs-eclipse-plugin), please contact with @KengoTODA or @iloveeclipse.
```
KengoTODA: @iloveeclipse Could you review this part?

I also want to ask you to confirm you can access https://marketplace.eclipse.org/content/spotbugs-eclipse-plugin with your account.

iloveeclipse: I'm fine with the description, there shouldn't be much more to say.

Regarding the spotbugs-eclipse-plugin entry: I can't access https://marketplace.eclipse.org/content/spotbugs-eclipse-plugin, it says 403 access denied. Have you got any confirmation or anything else after creation of the entry? In worst case we should just open a ticket at https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Community. I can do it, if you like.

KengoTODA: Thanks @iloveeclipse for your review.

About marketplace entry, let's release stable plugin first, because currently [our Eclipse Update page](https://spotbugs.github.io/eclipse) has no plugin for now.

KengoTODA:
In my understanding, this PR has no remaining tasks. Please review and merge.

iloveeclipse:
Question: where is the description how to build manual (I mean something like this: http://findbugs.sourceforge.net/manual/index.html) and bug descriptions list (I mean something like this: http://findbugs.sourceforge.net/bugDescriptions.html)?
*Can* we build them at all today?

KengoTODA:
Bug description page will be added by #163. About manual, we have no build script for now so I'm going to moveing its contents to `docs` directory.

iloveeclipse:
Sorry, why can't we build manual and why do you plan to move docs? For the missing manual, do we have a task / plan to restore that functionality? This is an essential part of the distribution, if we don't have it, how we can release SpotBugs?

KengoTODA:
Simply speaking, we don't know how to build that manual.
And I think it cannot be blocker, old site still exists at http://findbugs.sourceforge.net/ and interface is not changed, so users can refer that. About updated part, [Migration guide](http://spotbugs.readthedocs.io/en/latest/migration.html) can be help.

KengoTODA:
Please refer https://github.com/spotbugs/spotbugs.github.io/pull/4 for process to introduce new site

iloveeclipse:
The old ant build file did this:

`
    <target name="docs" depends="classes,version,bugdesc,checkdocs" unless="docs.uptodate" description="Make documentation">
        <loadproperties srcFile="${doc.props.file}"/>

        
        <antcall target="generatedocs">
            <param name="docoutput.dir" value="${doc.dir}"/>
            <param name="googleanalytics.fragment" value=""/>
        </antcall>
        <property name="webdocs.uptodate" value="true"/>

        
        <antcall target="generatemanual">
            <param name="docoutput.dir" value="${doc.dir}"/>
            <param name="googleanalytics.fragment" value=""/>
        </antcall>
        <property name="manual.uptodate" value="true"/>

        
        <antcall target="generatepdfmanual">
            <param name="docoutput.dir" value="${doc.dir}"/>
        </antcall>
        <property name="pdfmanual.uptodate" value="true"/>
    </target>`

Now that we have much better gradle build, I guess this shouldn't be a problem to do same in gradle? Or if gradle can't do that, can't it just call ant task?

KengoTODA:
It should be possible even for Gradle.
So is it better to do double-maintenance for DocBook and Sphinx?

iloveeclipse:
> It should be possible even for Gradle.

Then it would be good to have it. One main reason to move project away from FindBugs is to have full control over the project and to be able to release any parts. Manual *is* a part of the distribution, because this explains how to use it, and we can't simply point to some unrelated and out-of-date site for that.

> So is it better to do double-maintenance for DocBook and Sphinx? 

Yes, as long as we don't migrated fully. This is better as not to have manual at all.

> In my opinion, it's better to migrate from DocBook to Sphinx eventually, motivation is listed in spotbugs/spotbugs.github.io#4

Sure, but this probably requires much more time then "just fix" the Gradle build.

KengoTODA:
I confirmed that `ant generatedocs` in `findbugs` directory can generate documents... I will propose PR to fix Gradle build. Thanks!

KengoTODA:
Updated, now we can build old manual site even by Gradle.

I tried to use [saxon-gradle](https://github.com/eerohele/saxon-gradle/) but it needs many tasks to generate source file (we need to replace many tokens), and it generates XML only, so I just trigger Ant from Gradle.

jsotuyod:
@KengoTODA careful here. The build.xml is the complete old one, including all tasks, not only site-related ones. Most of them will fail (missing directories / jars), and using the findbugs names instead of spotbugs. We should probably clean it up until we finish moving to readthedocs and can drop the old manual.

Issue #152, opened 2017-04-18
spotbugsTestCases has compile errors in Eclipse


iloveeclipse:
The new cool Gradle build generates too restrictive .classpath file for spotbugsTestCases plugin: the examples code uses some com.sun.* classes which are forbidden now, and so the Eclipse project has compile errors.

This is a no-go. Projects should compile out of the box. Please fix the Gradle file or the test code.

Issue #151, opened 2017-04-18
Eclipse setup is wrong, cycle project dependencies


iloveeclipse:
I've tried to import spotbugs into Eclipse for the first time, this failed badly because a new "test-harness" project is added which is used by spotbugs but which also depends on it. Eclipse denies to compile both projects because of the cycle. Of course I can change compiler settings in Eclipse but why did we added this cycle at all???

This is a no-go.

The solution is to extract spotbugs tests code (src/test/java) from spotbugs project and place it under extra spotbugs-tests project. This was a mess before in findbugs but we should not keep this mess and increase the pain by adding cycles.

Just found out that even new gradlePlugin project continues with the bad habit to include everything in one place, resulting in another cycle. The test code under src/test/java belongs to the extra gradlePlugin-tests project.

KengoTODA:
refs #108 for cyclic reference.

iloveeclipse:
@KengoTODA : please check the commit above. Today I've tried to import spotbugs to understand Eclipse plugin issues but found out that current setup is broken on Eclipse. Also I'm not sure but it seems that the findbugs unit tests I've moved were not executed at all? Anyway, 3 of them seem to fail now because the new AnalysisRunner seem to be confused by an unexpected findbugs.xml location:
`
java.lang.ExceptionInInitializerError
	at edu.umd.cs.findbugs.test.SpotBugsRule.before(SpotBugsRule.java:52)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:46)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
[...]

PR #154, opened 2017-04-18
Fixes for #151 and #152

iloveeclipse:
Moved test code out of SpotBugs project, enabled JUnit tests, muted 3 failing tests, fixed compile errors in test examples plugins.


iloveeclipse:
@jsotuyod : I see the 3 commits were squashed into one - is this per purpose or default project setting? I don't like this because not the commit with 3 disabled tests is merged together with other commits and can't be simply reverted.

If this is a github project setting, can we please change the defaults to use commits 1:1. If this was my fault: please explain how to crate pull requests without squashing everything.

jsotuyod:
@iloveeclipse github settings, although it can be overriden by merger (I didn't, sorry). Usually one PR = one fix / feature.

Issue #153, opened 2017-04-18
Eclipse plugin debug is broken


iloveeclipse:
Findbugs project was used as a valid Eclipse bundle (it had a valid META-INF/MANIFEST.MF) so that one could debug Eclipse plugin and use hot code replace because all required classes were there in the FindBugs bundle. Now someone removed that manifest and so one can't debug Eclipse plugin anymore. OMG.
Also I see that almost all libraries are now inside the hidden local gradle folder and so aren't on the spotbugs bundle classpath. Is there a way to put them somewhere where the bundle could reference it via MANIFEST.MF?

iloveeclipse:
I can understand that Gradle is cool and modern, but what I see now is a complete, absolute, pointless mess.

Not only the manifest from the findbugs project was removed, the manifest of eclipse plugin is pointing now to non existing libraries in the lib folder, which will be copied there by gradle at the very end of the build. 

So debugging a plugin is effectively not possible anymore. WTF?!?

I can't debug Eclipse plugin anymore and so this project is dead for me.

jsotuyod:
@iloveeclipse I understand your frustration, please relax a bit. It's not no one cares about the Eclipse plugin, but mostly that none of us other than you has experience with them.

The issues you mention have been noted, and we will work on getting them straight.

iloveeclipse:
For whatever reason I've expected that the gradle replaces ant without breaking existing functionality, and now I'm even unable to debug things.

What is needed is: 3 projects (spotbugs, eclipsePlugin and eclipsePluginTests) need valid MANIFEST.MF files pointing to existing libraries. This could be achieved for example if the gradle build would copy those jars or create links from lib/ folders to the local versions of those hidden jar files now located somewhere on every user home directory. But since I have no idea how to make it with gradle, and ant build files are now removed, I can't help here, and so my frustration.

jsotuyod:
@iloveeclipse would you mind giving me some minimum input on how to run / debug the eclipse plugin from eclipse? Thanks in advanced

iloveeclipse:
Import spotbugs and eclipsePlugin projects into Eclipse.
Run -> Debug Configurations -> Eclipse Application -> New -> Debug.

This will fail right now because the spotbugs plugin can't find any of spotbugs main classes and dependencies.

This would not fail if spotbugs project would became the manifest back, the spotbugs plugin would fix the dependency to spotbugs (right now it is findbugs) and all the libraries required by the plugin would be really located at the lib directory.

[...]

PR #159, opened 2017-04-23
Enhance performance of Eclipse Plugin

KengoTODA:
Current Eclipse Plugin costs much time to update (redraw) dialog with progress bar.
By this change, we can reduce the UI change and it can shorten time to analyse.

For instance, in my machine, previous version needs 150 seconds to analyse `spotbugsTestCases` project, and this change shortens it to 20 seconds.


eclipsePlugin/test/de/tobject/findbugs/reporter/ThrottledProgressMonitorTest.java
```diff
@@ -0,0 +1,52 @@
+package de.tobject.findbugs.reporter;
```
iloveeclipse: Please do not add test code to the plugin. We have a dedicated tests plugin for tests.

eclipsePlugin/build.gradle
```diff
@@ -19,6 +19,11 @@ sourceSets {
       include '**/*.xml'
     }
   }
+  test {
```
iloveeclipse: please undo this change and enable tests plugin code which is in a dedicated tests plugin.

eclipsePlugin/build.gradle
```diff
@@ -56,6 +61,9 @@ dependencies {
   embeddedLibs(project(':spotbugs')) {
     transitive = false
   }
+
```
iloveeclipse: please undo this change and add this to the dedicated tests plugin

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+	private static final long NOT_TRIGGERED = Integer.MIN_VALUE;
+
+	private final IProgressMonitor delegate;
+	private final LongSupplier currentTimeMillis;
+	private long lastRenamed = NOT_TRIGGERED;
+	private long lastWorked = NOT_TRIGGERED;
+	private int accumulatedWork = 0;
+
+	ThrottledProgressMonitor(@Nonnull IProgressMonitor delegate, @Nonnull LongSupplier currentTimeMillis) {
+		assert delegate != null;
```
iloveeclipse: Please in Eclipse code, do not use assert keyword but use 
`org.eclipse.core.runtime.Assert.isNotNull(delegate);`

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+	 * Span of throttling, in milliseconds.
+	 */
+	static final int THROTTLE_SPAN = 300;
+	private static final long NOT_TRIGGERED = Integer.MIN_VALUE;
+
+	private final IProgressMonitor delegate;
+	private final LongSupplier currentTimeMillis;
+	private long lastRenamed = NOT_TRIGGERED;
+	private long lastWorked = NOT_TRIGGERED;
+	private int accumulatedWork = 0;
```
iloveeclipse: One do not need to init field values to the defaults.

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+		assert currentTimeMillis != null;
+
+		this.delegate = delegate;
+		this.currentTimeMillis = currentTimeMillis;
+	}
+
+	@Override
+	public void setTaskName(String name) {
+		long currentTime = currentTimeMillis.getAsLong();
+		if (currentTime - lastRenamed >= THROTTLE_SPAN) {
```
iloveeclipse: I would make sure we have set this at least once at the beginning. Otherwise it will not set the task name at all.

So something like if(lastRenamed == Integer.MIN_VALUE || ...)

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+			lastRenamed = currentTime;
+			delegate.setTaskName(name);
+		}
+	}
+
+	@Override
+	public void worked(int work) {
+		long currentTime = currentTimeMillis.getAsLong();
+		accumulatedWork += work;
+		if (currentTime - lastWorked >= THROTTLE_SPAN) {
```
iloveeclipse: same here. We should at least increment the value once.

iloveeclipse:
I've just found out that the eclipse tests plugin isn't functional anymore, thanks to #40 . I've entered issue #160 .

eclipsePlugin/test/de/tobject/findbugs/reporter/ThrottledProgressMonitorTest.java
KengoTODA: I see, but I still want to put this test at here because:

* According to https://github.com/spotbugs/spotbugs/issues/153#issuecomment-296539032, currently we have no CLI entry to run JUnit test for Eclipse plugin
* It is good habit to run test continuously. It would be nice to add a CLI entry (`./gradlew test` in this case) to test plugin.
* Then we can use Travis CI (or any other CI solution) to keep testing.

I think we can use `eclipsePlugin-test` for scenario test, and this `eclipsePlugin/test` for unit test.
How do you feel?

iloveeclipse: Test code and test dependencies do not belong to the production plugin project. There is nothing to discuss, this is how Eclipse plugins are structured and deployed. We have had many Eclipse tests running before, now thanks to all gradle changes we don't have it, not even manual one. Not even test code is compiled anymore - because if it would compile, one would see that we managed to break the test code too.

If you wish to add test code (which is great), please do this into the test plugin and please help to fix the gradle build to make the test plugin working again.

Until then I would just don't touch Eclipse plugin code, because no one knows which regressions one can introduce.

KengoTODA: OK Got it. Then I will fix #160 first, and add a new project to add CLI entry for JUnit tests.

> this is how Eclipse plugins are structured and deployed

I see.

Even for me, a Maven hacker, it's natural to put test cases into the same project of production code (`src/main/java` and `src/test/java`). I didn't know the culture in Eclipse plugin community. I'll learn from you and that community.

Thanks again for your review!

iloveeclipse: That's the problem (one from many) with maven. There was a life before it, but people tend to forget it :-) 

One of (from many) really stupid things maven people did was to break the habit to have clean modular code, now everything need to be in one place, this is awful.

Production code should be always separated from tests, and thanks to the fact Eclipse was developed before maven and without maven, they also had a very good choice to have clean separated test bundle projects and clean separated production bundle projects.

eclipsePlugin-junit/build.gradle
```diff
@@ -0,0 +1,7 @@
+apply from: "$rootDir/gradle/checkstyle.gradle"
+
+dependencies {
+  compile project(':eclipsePlugin')
+  testCompile 'junit:junit:4.12'
+  testCompile 'org.mockito:mockito-core:2.7.22'
+}
```
KengoTODA: added new project, because `eclipsePlugin-test` is not built by Travis CI &amp; that is for plugin (scenario) test.

KengoTODA:
@iloveeclipse I've fixed all pointed problems, please check.

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final long NOT_TRIGGERED = Integer.MIN_VALUE;
+
+    private final IProgressMonitor delegate;
+    private final LongSupplier currentTimeMillis;
+    private long lastRenamed = NOT_TRIGGERED;
+    private long lastWorked = NOT_TRIGGERED;
+    private int accumulatedWork;
+
+    ThrottledProgressMonitor(@Nonnull IProgressMonitor delegate, @Nonnull LongSupplier currentTimeMillis) {
+        isNotNull(delegate);
```
iloveeclipse: I personally prefer to read the code like "assert that this is not null" but due the static import it reads now "is not null, so *what*?"
I would prefer to write this like Assert.isNotNull(delegate), which leaves no question what is "isNotNull(delegate)" does.

eclipsePlugin-junit/build.gradle
KengoTODA: My motivation is run my JUnit test by build tool (Gradle for now).
Currenly our build tool cannot run tests in `eclipsePlugin-test` so I added new project.

If we can run tests in `eclipsePlugin-test` just by build tool, then it's enough &amp; better solution, I think.

KengoTODA: If you reviewer can agree my opinion, I will add `README.md` to both projects to explain their responsibility.

eclipsePlugin/src/de/tobject/findbugs/reporter/ThrottledProgressMonitor.java
KengoTODA: Solved by d8fcc8f5c, thanks for your review!

eclipsePlugin-junit/build.gradle
iloveeclipse: Yep, with the TODO that the new project should be merged to the old one as soon as we have working build/test cycle again.

KengoTODA: Wmm, sorry, I couldn't get your point. Do you mean that it's better to keep current design? It is OK for me, then I will trigger unit test by hands, not in CI build.

iloveeclipse: My point is: ideally
1. we can run Eclipse tests automatically
2. we have one project for Eclipse tests (otherwise questions arise which test to put where and this is not obvious).

Unfortunately point 1. is not true and so you have started a new project which allows some subset of tests to run, and this is against point 2. . 

My previous comment was: if you are going to add a new project for Eclipse tests because we can't run the tests in the old project, please add readme files in each of them explaining the situation, and a TODO to merge the new project with the old one as soon as we can run all tests automatically.

KengoTODA: I got your point, please check 3cd28c2.

PR #162, opened 2017-04-28
Improve eclipsePlugin config to allow PDE debugging

jsotuyod:
I detected and fixed a bunch of minor issues on the generated site, but most importantly , improved on the ability to import the project directly into eclipse to be used with PDE.

Currently, however, running the plugin from PDE throws:

> Caused by: java.lang.ClassNotFoundException: edu.umd.cs.findbugs.PluginException cannot be found by com.github.spotbugs.plugin.eclipse_3.1.0.20170427-1493341245436
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:461)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:372)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:364)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:161)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 80 more

Notice with the generated `MANIFEST.MF` / `build.properties` all other classes seem to be found just fine.


iloveeclipse:
It is simple, and I've explained this already. spotbugs (not only the plugin project!) must have META-INF/MANIFEST.MF file which will define spotbugs as a valid bundle and point to the right classpath. Generate this manifest, similar to one we had before it was deleted, and eclipse plugin will be able to see spotbugs classes if running in the debugger.

iloveeclipse:
P.S.
Small thing: /spotbugsTestCases/src/java/lambdas/Issue60.java has now a compile issue because it has no proper package definition.

iloveeclipse:
Bigger issue: eclipsePlugin-test still has old eclipse plugin id as dependency in the manifest file, it must be spotbugs plugin id

jsotuyod:
I'm back after several weeks out of town. I'm picking this up and finishing with the extra input. If no new issues appear I expect it to have it fixed tomorrow.

Today I'll just get it rebased to master.

jsotuyod:
I just added the generation of a Spotbugs MANIFEST.MF file based on [findbugs' one](https://github.com/findbugsproject/findbugs/blob/master/findbugs/META-INF/MANIFEST.MF), but I fear I keep finding the same `ClassNotFoundException` error.

iloveeclipse:
Please check #216 which is based on this one.

PR #163, opened 2017-05-02
Generate bug description page

KengoTODA:
This chnage will generate bug description page like [this page in findbugs.sourceforge.net](http://findbugs.sourceforge.net/bugDescriptions.html). It will help us to keep the document updated.

Note that we still need another pull request for localized pages (currently `<Detail>` part is not localized).

## Sample

* http://spotbugs-in-kengo-toda.readthedocs.io/en/latest/bugDescriptions.html
* http://spotbugs-in-kengo-toda.readthedocs.io/ja/latest/bugDescriptions.html

## Sample image

2017-05-02_18h30_14


KengoTODA:
Hello, who can review and merge this PR? Thanks in advance.

KengoTODA:
> The docs/bugDescriptionList.rst seem to be generated. Why is this file added to the repo?

Because readthedocs.org has no feature to trigger build tool like Gradle. So it's necessary to commit generated files too.

* http://docs.readthedocs.io/en/latest/builds.html

PR #166, opened 2017-05-08
Fix HTML formatting

orihalcon128:
Fix HTML formatting even in messages.xml and messages_fr.xml.

- Modify source code block(&lt;blockquote&gt; -&gt; &lt;code&gt;).
- Change to the bulleted list(&lt;ul&gt;&lt;/ul&gt;).
- Change Datails of the BugCategory into the HTML format.

### DIFF
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -93,146 +93,71 @@ This plugin contains all of the standard SpotBugs detectors.
   <BugCategory category="CORRECTNESS">
     <Description>Correctness</Description>
     <Abbreviation>C</Abbreviation>
-    <Details>
-<![CDATA[
-<p>
-    Probable bug - an apparent coding mistake
+    <Details>Probable bug - an apparent coding mistake
             resulting in code that was probably not what the
-            developer intended.
-</p>
-<p>
-             We strive for a low false positive rate.
-</p>
-]]>
-    </Details>
+            developer intended. We strive for a low false positive rate.</Details>
   </BugCategory>
-  
   <BugCategory category="NOISE">
     <Description>Bogus random noise</Description>
     <Abbreviation>N</Abbreviation>
-    <Details>
-<![CDATA[
-<p>
-    Bogus random noise: intended to be useful
+    <Details>Bogus random noise: intended to be useful
     as a control in data mining experiments, not in finding actual bugs in software
-</p>
-]]>
             </Details>
   </BugCategory>
-  
   <BugCategory category="SECURITY">
     <Description>Security</Description>
     <Abbreviation>S</Abbreviation>
-    <Details>
-<![CDATA[
-<p>
-    A use of untrusted input in a way that could create a remotely exploitable security vulnerability.
-</p>
-]]>
+    <Details>A use of untrusted input in a way that could create a remotely exploitable security vulnerability.
     </Details>
   </BugCategory>
-  
   <BugCategory category="BAD_PRACTICE">
     <Description>Bad practice</Description>
     <Abbreviation>B</Abbreviation>
-    <Details>
-<![CDATA[
-<p>
-    Violations of recommended and essential
+    <Details>Violations of recommended and essential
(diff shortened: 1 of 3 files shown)


spotbugs/etc/messages.xml
```diff
@@ -93,73 +93,149 @@ This plugin contains all of the standard SpotBugs detectors.
...
+<p>
+    Probable bug - an apparent coding mistake
             resulting in code that was probably not what the
-            developer intended. We strive for a low false positive rate.</Details>
+            developer intended.
+</p>
+<p>
+             We strive for a low false positive rate.
+</p>
+]]>
```
KengoTODA: We need to confirm that we can use `<![CDATA[` in `<Details>` or not.

KengoTODA: For now we can NOT use `<![CDATA[` in `<Details>` in `<BugCategory>`, according to Javadoc in `BugCategory.java`:

```java
    /**
     * ...
     * @param detailText
     *            full description of the bug category (no HTML markup, may be
     *            null)
     */
    public BugCategory(String category, String shortDescription, String abbrev, String detailText) {
```

So please revert changes for BugCategory, it should keep using plain text instead of HTML markup in `<![CDATA[`.

orihalcon128: OK, revert changes for BugCategory.

KengoTODA:
Currently our JDK9 build is blocked by [maintenance at www.oracle.com](http://www.oracle.com/splash/java.net/maintenance/index.html). I'll retry building later.

spotbugs/etc/messages.xml
```diff
@@ -3712,17 +3725,16 @@ blocking and deadlock behavior. See <a href="http://www.javalobby.org/java/forum
     <Details>
       <![CDATA[
   <p> The code synchronizes on a boxed primitive constant, such as a Boolean.</p>
-<pre>
+<pre><code>
 private static Boolean inited = Boolean.FALSE;
-...
```
KengoTODA: I think this `...` is necessary, or code becomes non intuitive (why we can code fields and logic at the same place?)

spotbugs/etc/messages.xml
```diff
@@ -3736,18 +3748,17 @@ and possible deadlock.</p>
       <![CDATA[
   <p> The code synchronizes on an apparently unshared boxed primitive,
 such as an Integer.</p>
-<pre>
+<pre><code>
 private static final Integer fileLock = new Integer(1);
-...
```
KengoTODA: This `...` is also necessary

spotbugs/etc/messages.xml
```diff
@@ -3763,14 +3774,13 @@ throughout the JVM, leading to very confusing behavior and potential deadlock.
     <Details>
       <![CDATA[
   <p> The code synchronizes on a boxed primitive constant, such as an Integer.</p>
-<pre>
+<pre><code>
 private static Integer count = 0;
-...
```
KengoTODA: Let's keep this `...`

spotbugs/etc/messages_ja.xml
```diff
@@ -99,135 +99,70 @@
...
     <Description>セキュリティ</Description>
     <Abbreviation>S</Abbreviation>
-    <Details>
-<![CDATA[
-<p>
-リモートから悪用可能なセキュリティ脆弱性を引き起こすかもしれない信頼できない入力を使用しています。
-</p>
-]]>
-    </Details>
+    <Details>リモートから悪用可能なセキュリティ脆弱性を引き起こすかもしれない信頼できない入力を使用しています。 </Details>
```
KengoTODA: Thanks for your fix! Sorry I didn't check this problem in previous PR.

KengoTODA:
No other problem I found. Could you consider to revert changes about `...`?

orihalcon128:
Yes, revert changes about `...` ,
The lost reason is because it copied from messages_ja.xml. 
Also fix messages_ja.xml.

Issue #161, opened 2017-04-26
TypeMatcher is missing in filter documentation


Gubbel:
`TypeMatcher` is not documented in the [filter documentation](http://findbugs.sourceforge.net/manual/filter.html) (also missing in spotbugs/spotbugs/src/doc/manual.xml).

Example Filter:

	<Match>
		<Bug pattern="BC_UNCONFIRMED_CAST" />
		<Type
			descriptor="Ltest/IgnoredCastTarget;"
			role="TYPE_EXPECTED" />
	</Match>

(also reported here: https://github.com/findbugsproject/findbugs/issues/159)

PR #168, opened 2017-05-14
Migrate filter file description to new site.

KengoTODA:
@iloveeclipse Please review the description for `<Type>`, it's missing even in legacy site so I want to add description by this PR. By merging this PR, we can solve #161.

To ease review, I will share zipped generated page: [Filter file.zip](https://github.com/spotbugs/spotbugs/files/999143/Filter.file.zip)


spotbugs/src/doc/FilterFile.txt
```diff
@@ -1,125 +1 @@
...
-     <Match class="com.foobar.AnotherClass">
-       <Or>
-         <Method name="nonOverloadedMethod" />
-         <Method name="frob" params="int,java.lang.String" returns="void" />
-         <Method name="blat" params="" returns="boolean" />
-       </Or>
-       <BugCode name="DC" />
-     </Match>
-</FindBugsFilter>
+Refer http://spotbugs.readthedocs.io/en/latest/filter.html instead of this page.
```
iloveeclipse: Why should we keep this file?
If we want to keep it, may be add that "the filter.html is generated from docs/filter.rst"?

KengoTODA: Reasonable & intuitive, I will add path of `docs/filter.rst` to this file.

KengoTODA:
Currently our JDK9 build is blocked by [maintenance at www.oracle.com](http://www.oracle.com/splash/java.net/maintenance/index.html). I'll retry building later.

PR #169, opened 2017-05-14
stop JDK9 build temporally to stabilize build result

KengoTODA:
Currently JDK9 build in `master` branch (and others) are broken, because we cannot download JDK from oracle.com. It says that it's under maintenance.

I'm not sure when the maintenance for oracle.com will be finished, so I'll propose this as workaround.


KengoTODA:
The oracle.com is still under maintenance, let's merge this asap.

PR #180, opened 2017-05-25
No more JSR305, use SpotBugs internal annotation instead

### LINKED ISSUE #130: Annotations should not be deprecated.
The [findbugs annotations](https://github.com/spotbugs/spotbugs/tree/master/findbugs/src/java/edu/umd/cs/findbugs/annotations) e.g. [`CheckForNull`](https://github.com/spotbugs/spotbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/annotations/CheckForNull.java) [`NonNull`](https://github.com/findbugsproject/findbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/annotations/NonNull.java) etc. should not be marked as deprecated and users should not be directed to use the [JSR-305](https://jcp.org/en/jsr/detail?id=305) equivalents.

JSR-305 is dormant, has been for a long while and shows no hope of ever producing an agreed set of annotations in our lifetime. Further more these annotations use `javax.` packages which it is not possible to use according to the [Oracle Java binary licence](http://www.oracle.com/technetwork/java/javase/terms/license/index.html), so applications can not use and ship these dependencies along with a JRE without violating the Oracle licence agreement.

> F. JAVA TECHNOLOGY RESTRICTIONS. You may not create, modify, or change the behavior of, or authorize your licensees to create, modify, or change the behavior of, classes, interfaces, or subpackages that are in any way identified as "java", "javax", "sun", “oracle” or similar convention as specified by Oracle in any naming convention designation.

The JSR-305 group has not defined any official releases according to its [jsr page](https://jcp.org/en/jsr/detail?id=305) so the only implementations is a seemingly random implementation provided by the FindBugs team. Even if the team where experts on the JSR (which some where) they are not official as there has been no vote and are not available from the JSR hompage - so the  `javax` package name restriction still applies.

So please remove the `@Deprecated` annotation from the spotbugs specific annotations or update the links to point to the new preferred non-jsr305 implementation to use instead.

KengoTODA:
This PR will close #130 and cancel deprecation on SpotBugs annotations.
Let's have discussion about we should cancel this deprecation or not.

I think we don't need to recommend jsr305 in current situation, but this change will affect public interface so I'm marking this PR for SpotBugs 4.0.0 release. Please do not merge this PR for now.

### DIFF
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -7,7 +7,7 @@ apply from: "$rootDir/gradle/javadoc.gradle"
 apply from: "$rootDir/gradle/maven.gradle"
 
 dependencies {
-  compile 'com.google.code.findbugs:jsr305:3.0.1'
+  compile 'com.google.code.findbugs:jsr305:3.0.2'
 }
 
 def manifestSpec = manifest {
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckForNull.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckForNull.java
@@ -33,15 +33,12 @@ import javax.annotation.meta.When;
  *
  * When this annotation is applied to a method it applies to the method return
  * value.
- *
- * @deprecated - use {@link javax.annotation.CheckForNull} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.MAYBE)
 @TypeQualifierNickname
-@Deprecated
 public @interface CheckForNull {
 
 }
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/NonNull.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/NonNull.java
@@ -32,15 +32,12 @@ import javax.annotation.meta.When;
  *
  * Annotated Fields must only not be null after construction has completed.
  * Annotated methods must have non-null return values.
- *
- * @deprecated - use {@link javax.annotation.Nonnull} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.ALWAYS)
 @TypeQualifierNickname
-@Deprecated
 public @interface NonNull {
 
 }
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/Nullable.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/Nullable.java
@@ -36,15 +36,12 @@ import javax.annotation.meta.When;
  *
  * When this annotation is applied to a method it applies to the method return
  * value.
- *
- * @deprecated - use {@link javax.annotation.Nullable} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.UNKNOWN)
 @TypeQualifierNickname
-@Deprecated
 public @interface Nullable {
 
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased
 
+### Changed
+
+* SpotBugs annotation is recommended instead of JSR305 annotation ([#130](https://github.com/spotbugs/spotbugs/pull/130))
+
 ### Fixed
 
 * Wrong Class-Path in MANIFEST.MF ([#407](https://github.com/spotbugs/spotbugs/pull/407))
--- a/spotbugs-annotations/build-template.properties
+++ b/spotbugs-annotations/build-template.properties
@@ -1,6 +1,0 @@
-src.excludes=META-INF/MANIFEST-TEMPLATE.MF
-bin.includes=META-INF/
-jars.compile.order=spotbugs-annotations.jar
-src.includes=META-INF/
-output.spotbugs-annotations.jar=bin_eclipse/
-bin.excludes=META-INF/MANIFEST-TEMPLATE.MF
--- a/spotbugs-annotations/build-template.properties
+++ b/spotbugs-annotations/build-template.properties
@@ -1,0 +1,6 @@
+src.excludes=META-INF/MANIFEST-TEMPLATE.MF
+bin.includes=META-INF/
+jars.compile.order=spotbugs-annotations.jar
+src.includes=META-INF/
+output.spotbugs-annotations.jar=bin_eclipse/
+bin.excludes=META-INF/MANIFEST-TEMPLATE.MF
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckReturnValue.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckReturnValue.java
@@ -29,13 +29,10 @@ import java.lang.annotation.Target;
  * be checked when invoking the method.
  *
  * The checker treats this annotation as inherited by overriding methods.
- *
- * @deprecated - use {@link javax.annotation.CheckReturnValue} instead.
  */
 @Documented
 @Target({ ElementType.METHOD, ElementType.CONSTRUCTOR })
 @Retention(RetentionPolicy.CLASS)
-@Deprecated
 public @interface CheckReturnValue {
 
     @Deprecated
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/Confidence.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/Confidence.java
@@ -19,8 +19,6 @@
 
 package edu.umd.cs.findbugs.annotations;
 
-import javax.annotation.Nonnull;
-
 /**
  * Describes the confidence with which FindBugs reports a bug instance.
  */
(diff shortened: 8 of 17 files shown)


jsotuyod:
I'm ok with ceasing to publish jsr305, and removing those from our codebase.

As for reintroducing the old annotations, I'm ok with it, but why not simply telling users to adopt the annotations from JSR-308 (checker framework), which **are** an approved and published JSR that includes both `@Nullable` and `@NonNull` annotations? Using SpotBugs is still easier than using the checker framework for compile-time validations.

iloveeclipse:
JSR 305 is dead, yes, what about adopting JDT null annotations (available on maven too)? 
- https://wiki.eclipse.org/JDT_Core/Null_Analysis
- https://search.maven.org/#artifactdetails%7Corg.eclipse.jdt%7Corg.eclipse.jdt.annotation%7C2.1.0%7Cjar

stephenc:
> why not simply telling users to adopt the annotations from JSR-308 (checker framework), which are an approved and published JSR that includes both @Nullable and @NonNull annotations?

One thing is that the checker framework actually includes 10 different `@NonNull` annotations (https://checkerframework.org/api/) so that is not going to help user confusion.

Another thing is that presumably spotbugs will want to pick up some enhanced checks that leverage JSR-308 in which case - presumably - spotbugs would catch any of the bugs in checker and more besides.

The final thing is that, at least for these type of annotations, the fewer dependencies in the annotation JAR the better. Zero dependencies is ideal, so pushing people to checker for the Nullable / NonNull / CheckForNull annotations but keeping them with spotbugs for the OverrideMustInvoke, etc is not a good plan IMHO

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckForNull.java
```diff
@@ -33,15 +33,12 @@
  *
  * When this annotation is applied to a method it applies to the method return
  * value.
- *
- * @deprecated - use {@link javax.annotation.CheckForNull} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.MAYBE)
```
stephenc: I think these should probably be removed if removing "JSR-305" stuff

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/NonNull.java
```diff
@@ -32,15 +32,12 @@
  *
  * Annotated Fields must only not be null after construction has completed.
  * Annotated methods must have non-null return values.
- *
- * @deprecated - use {@link javax.annotation.Nonnull} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.ALWAYS)
```
stephenc: more "JSR-305" stuff that should probably be removed

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/Nullable.java
```diff
@@ -36,15 +36,12 @@
  *
  * When this annotation is applied to a method it applies to the method return
  * value.
- *
- * @deprecated - use {@link javax.annotation.Nullable} instead.
  **/
 @Documented
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
 @Retention(RetentionPolicy.CLASS)
 @javax.annotation.Nonnull(when = When.UNKNOWN)
```
stephenc: more "JSR-305" stuff that should probably be removed

KengoTODA:
@iloveeclipse Does JDT annotations provide build tool integration or static analysis integration? 
If not, I think [Checkerframework Qualifier](https://mvnrepository.com/artifact/org.checkerframework/checker-qual) is better to choose.

iloveeclipse:
> Does JDT annotations provide build tool integration or static analysis integration?

What do you mean by that? It is just plain java annotations in a library. One can use it as one likes, but they do not do anything except being annotations.

> If not, I think Checkerframework Qualifier is better to choose.

Sorry, I've found no obvious links on their homepage (https://checkerframework.org/) pointing to the *annotations library* they use. Can you please provide a direct link?

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckForNull.java
KengoTODA: It seems really hard to delete dependency on jsr-305, or we cannot support feature depending on `@TypeQualifierDefault` such as `ReturnValuesAreNonnullByDefault`.

spotbugs-annotations/build.gradle
```diff
@@ -7,7 +7,7 @@ apply from: "$rootDir/gradle/javadoc.gradle"
 apply from: "$rootDir/gradle/maven.gradle"
 
 dependencies {
-  compile 'com.google.code.findbugs:jsr305:3.0.1'
+  compile 'com.google.code.findbugs:jsr305:3.0.2'
```
KengoTODA: refer https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2017-March/004354.html

stephenc: Any jar that has classes in `java` or `javax` packages violates the Oracle JVM redistribution license. By adding back such a dependency you restore this limitation

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckForNull.java
stephenc: If you need defaults, you need your own defaults annotation. There is no JSR 305 and this there are no annotations from it *until* the JSR page publishes at least a draft javadocs.

spotbugs-annotations/build.gradle
KengoTODA: OK then delete `ReturnValuesAreNonnullByDefault` and ask users to migirate to `DefaultAnnotation` or others :+1:

KengoTODA: Not only `ReturnValuesAreNonnullByDefault` but also `CheckForNull` needs migration, because current spotbugs-annotation does not support `When`. and `TypeQualifierNickname`.

If the distribution is the problem, we can change this dependency to `compileOnly`, then our distribution package will not include `jsr-305-3.0.1.jar`.

KengoTODA: Oops, `TypeQualifierAnnotation` depends on `When` in jsr-305...

KengoTODA:
@iloveeclipse

> Sorry, I've found no obvious links on their homepage (https://checkerframework.org/) pointing to the annotations library they use. Can you please provide a direct link?

https://mvnrepository.com/artifact/org.checkerframework/checker-qual

Especially for nullness handling: https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/package-summary.html

spotbugs-annotations/build.gradle
KengoTODA: Note: here is license http://www.oracle.com/technetwork/java/javase/terms/license/index.html

> F. JAVA TECHNOLOGY RESTRICTIONS. You may not create, modify, or change the behavior of, or authorize your licensees to create, modify, or change the behavior of, classes, interfaces, or subpackages that are in any way identified as "java", "javax", "sun", “oracle” or similar convention as specified by Oracle in any naming convention designation.

IANAL but who violate this is not us but jsr305.jar, is it also bad to depend on it? What do you mean by "you restore this limitation"?

stephenc: When a class is loaded, if the annotation is not present... no foul...

Now what happens if there is a reboot JSR that superseded JSR 305, and it adds `javax.annotation.Nonnull`, but decided to use `value` for the when clause so that it is `@Nonnull(___)` rather than forcing everyone to type in `@Nonnull(when=___)`... save 5 characters many times... that's a win...

In that case when a class is loaded... and that includes an annotation... you will have the *entire* class fail to load as the annotation is present but the signature is invalid.

You have to be exceedingly careful in how you evolve the fields in an annotation.

Oracle are aware of this issue, which is why they have that restriction, because if they do not keep the guard of the "must be published as part of a JSR process" then it would be *their* fault for releasing a JSR-xyz that had the alternative `javax.annotation.Nonnull`.

Right now, however, there is no publication from JSR-305, so it will be your fault if you have those class references in your class files.

Additionally, a really nit-picking lawyer told me that the classloading failure described above could be interpreted as "change the behavior of"... so if you are distributing an oracle JVM, then - in his overly pedantic view - you should not even have those annotations referenced in your class files... 

now that last bit is not an argument that I personally put much strength in... but I do think that at some point in time there may be a JSR-305 reboot, and we do not know what they would produce from that, hence we should not be preparing to shoot ourselves in the foot, especially given that the IDE that redistribute a JVM in their installers have gone to great lengths not to include those annotations in their classfiles... indicating to me that they share somewhat similar concerns

seamlik:
Forgive me if I'm not familiar with the history of FindBugs, but I guess JSR-305 was (or still considered is) a sub-project (or partner project) of FindBugs. I don't know why JSR-305 was abandoned in the first place, but since SpotBugs is going to supersede FindBugs, why not also adopt JSR-305 and keep it maintained?

KengoTODA:
[JSR305 is not sub-project nor partner project of FindBugs, but its spec lead is William Pugh the author of FindBugs.

spotbugs-annotations/build.gradle
KengoTODA: In my personal opinion, we can keep current dependency for 3.1.0 release, to keep SpotBugs as much as same with FindBugs. Of course, we all know this dependency is negative legacy, we should have issue for 4.0.0 to delete them.

buzz3791:
> [JSR305](https://jcp.org/en/jsr/detail?id=305) is not sub-project nor partner project of FindBugs, but its spec lead is William Pugh the author of FindBugs. I'm personally not sure the historical background about this.
> 
> This spec is already marked as [Dormant](https://jcp.org/en/introduction/glossary#status), so we have no motivation to keep it maintained.

A bit more info about JSR 305... on Jul 25 at 20:55, Mark Reinhold Chief Architect, Java Platform Group at Oracle wrote, "JSR 305, which would’ve standardized javax.annotation.NonNull, never completed because its spec lead [William Pugh] went AWOL. It had nothing to do with any decision by Oracle.", source: Mark commented on Ludwig Weinzierl's Stackoverflow answer to jaxzin's question "Which @NotNull Java annotation should I use?
", https://stackoverflow.com/a/42695253/110126.

PR #181, opened 2017-05-25
update gradle toolVersion and maven coordinates

brharrington:
Currently it is a bit of a pain to try out the spotbugs
RCs using the spotbugs gradle plugin because it pulls in
the older findbugs release.


jsotuyod:
Thanks for the PR! really cool form you.

PR #191, opened 2017-06-01
Specify project dependencies to enable ./gradlew clean test

### LINKED ISSUE #192: ./gradlew clean test does not run
There is no gradle-style dependency specified between projects.
Projects depend on each other in a quite unusual way by specifying another project's class directories
We should make dependencies explicit and gradle style

kzaikin:
fix #192


KengoTODA:
This change makes coverage check by coveralls broken, I will check.

PR #195, opened 2017-06-03
Follow latest version automatically

KengoTODA:
This change will reduce manual task like #181


KengoTODA:
I referred #162 to introduce Gradle 3.5

PR #198, opened 2017-06-04
Report coverage not only from spotbugs subproject but also other subprojects

KengoTODA:
Currently we analyze spotbugs subproject only. After we apply this change, we can check coverage even in other subprojects such as gradlePlugin and eclipsePlugin-junit.


KengoTODA:
Coverage should be decreased by this change: please merge even though coveralls makes pre-merge check failed.

PR #199, opened 2017-06-04
Change dependencies from local to maven, remove unused dependencies

kzaikin:
I was able to replace almost all of local dependencies with their modern versions from maven. I was even able to remove some unused local jars without adding remote references.
Unfortunately j2ee dependency was not able to be resolved from maven so it had to be left as is


kzaikin:
I have to investigate coverage decrease

Issue #203, opened 2017-06-09
Gradle build fails to build eclipsePlugin under macOS


sewe:
Under macOS, the directory layout of Eclipse installations is different (IIRC, since Eclipse Mars). The `eclipse` binary can be found at `Eclipse.app/Contents/MacOS/eclipse` whereas the `plugins` directory can be found at `/Users/sewe/Desktop/spotbugs/Eclipse.app/Contents/Eclipse/plugins`.

If one specifies the latter path as `eclipsePlugin.dir` in `eclipsePlugin/local.properties`, then the build fails with a message like

    :eclipsePlugin:generateCandidateP2Metadata FAILED
    
    FAILURE: Build failed with an exception.
    
    * What went wrong:
    Execution failed for task ':eclipsePlugin:generateCandidateP2Metadata'.
    > A problem occurred starting process 'command '/Users/sewe/Documents/spotbugs/Eclipse.app/Contents/Eclipse/plugins/../eclipse''

Ideally, Gradle would detect the operating system and chose its relative paths accordingly. And while we are at it, it may be worth switching from  `eclipsePlugin.dir` to  `eclipse.dir`. Then, Gradle would need to manage only OS-dependent path *suffixes*, which I believe is clearer.

PR #205, opened 2017-06-10
Account for OS-dependent Eclipse directory layout

sewe:
This replaces the `eclipsePlugin.dir` property in `eclipsePlugin/local.properties` with `eclipseRoot.dir` and automatically computes the OS-dependent `eclipseExecutable` and `eclipsePluginsDir` properties from it.

This change fixes spotbugs/spotbugs#203.

### DIFF
--- a/eclipsePlugin/build.gradle
+++ b/eclipsePlugin/build.gradle
@@ -34,6 +34,16 @@ def eclipsePluginId = 'com.github.spotbugs.plugin.eclipse'
 def localProps = new Properties()
 localProps.load(new FileInputStream("$projectDir/local.properties"))
 
+def eclipseRootDir = new File(localProps.getProperty('eclipseRoot.dir'))
+// Eclipse 4.5+ uses a different directory layout under macOS. Try to detect this first.
+def eclipseExecutable = new File(eclipseRootDir, "Contents/MacOS/eclipse")
+def eclipsePluginsDir = new File(eclipseRootDir, "Contents/Eclipse/plugins")
+if (!eclipseExecutable.exists()) {
+   // Fall back to non-macOS directory layout.
+   eclipseExecutable = new File(eclipseRootDir, "eclipse")
+   eclipsePluginsDir = new File(eclipseRootDir, "plugins")
+}
+
 // TODO : convert these to external dependencies
 def requiredLibs = fileTree(dir:'lib', include:[
   'jsr305.jar',
@@ -46,7 +56,7 @@ def requiredLibs = fileTree(dir:'lib', include:[
 dependencies {
   includedLibs requiredLibs
 
-  compile fileTree(dir:localProps.getProperty('eclipsePlugin.dir'), include:'**/*.jar',
+  compile fileTree(dir:eclipsePluginsDir, include:'**/*.jar',
     exclude:['**/datanucleus-enhancer*.jar', 'edu.umd.cs.findbugs.**/*.jar', 'com.github.spotbugs.**/*.jar'])
 
   includedLibs project(':spotbugs-annotations')
@@ -198,7 +208,7 @@ task siteDailyXml(type:Copy) {
 task generateP2Metadata(type:Exec) {
   inputs.file 'local.properties'
   dependsOn pluginJar, featureJar, siteXml
-  commandLine "${localProps.getProperty('eclipsePlugin.dir')}/../eclipse", '-nosplash',
+  commandLine "${eclipseExecutable}", '-nosplash',
     '-application', 'org.eclipse.equinox.p2.publisher.UpdateSitePublisher',
     '-metadataRepository', "file:${buildDir}/site/eclipse", // TODO : May not work on Windows
     '-artifactRepository', "file:${buildDir}/site/eclipse", // TODO : May not work on Windows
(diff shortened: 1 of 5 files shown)


sewe:
Would be great if a reviewer could test this under Windows to see if there are any slash vs. backslash issues.

eclipsePlugin/build.gradle
```diff
@@ -34,6 +36,12 @@ def eclipsePluginId = 'com.github.spotbugs.plugin.eclipse'
 def localProps = new Properties()
 localProps.load(new FileInputStream("$projectDir/local.properties"))
 
+ext {
+  eclipseRootDir = localProps.getProperty('eclipseRoot.dir')
+  eclipseExecutable = eclipseRootDir + (Os.isFamily(Os.FAMILY_MAC) ? "/Contents/MacOS/eclipse" : "/eclipse")
+  eclipsePluginsDir = eclipseRootDir + (Os.isFamily(Os.FAMILY_MAC) ? "/Contents/Eclipse/plugins" : "/plugins")
+}
```
KengoTODA: SpotBugs Eclipse Plugin supports legacy Eclipse such as 3.6.2, so the condition to switch directory should be more complex. For instance, `eclipsePluginsDir ` for Eclipse 3.6.2 for OS X is not `${eclipseRootDir}/Contents/Eclipse/plugins` but `${eclipseRootDir}/plugins`.

I guess it's better to check not OS but existence of files like below:

```groovy
eclipseExecutable = eclipseRootDir + '/eclipse'
if (!new File(eclipseExecutable).isFile()) {
  eclipseExecutable = eclipseRootDir + '/Contents/MacOS/eclipse'
}
```

iloveeclipse: I think we schould throw away 3.6, may be even 3.8 support. There is really no reason  to support ancient Eclipse releases if this makes the code more complex.

KengoTODA: Sounds nice, but according to #203, 4.3 and 4.4 follows the same directory layout with 3.6. Then we need to drop support not only for 3.6 & 3.8 but also 4.3 and 4.4.

- [x] Oxygen (4.7)
- [x] Neon (4.6)
- [x] Mars (4.5) which starts new directory layout
- [x] Luna (4.4)
- [x] Kepler (4.3)
- [ ] Juno (4.2, 3.8)
- [ ] Previous to Juno (<=4.1)

KengoTODA: > At least, the release which doesn't support Java8 should be dropped.

Kepler is the first version which supports Java8. https://wiki.eclipse.org/JDT_Core/Java8

KengoTODA: Current stable release is Neon, and Mars isn't supported for now according to [Wikipedia](https://en.wikipedia.org/wiki/Eclipse_(software)#Releases) (sorry I cannot find official page).

Then why we do not drop support for Kepler, Luna and Mars?
It is also necessary to update our `.travis.yml` to use Neon to build the plugin.

sewe: > I guess it's better to check not OS but existence of files like below:

Makes sense. While I don’t think it will make the Gradle script more portable (the `MacOS` in `/Contents/MacOS/eclipse` is pretty telling), it would support building against a pre-Mars Eclipse. Also, you could at least cross-compile against an Eclipse installation for a different OS (`./gradlew :eclipsePlugin:check`), even if you cannot build the update site (part of `./gradlew :eclipsePlugin:build`), as that would require executing the `${eclipseExecutable}`.

I will modify my pull request accordingly.

> Then why we do not drop support for Kepler, Luna and Mars?

Because a lot of them are still used in the wild. Here are some numbers from the Eclipse Code Recommenders to back this up:

In the last 30 days, there have these many downloads of the recommendation model for the JRE itself (the one library every Java project has to use) using the different Eclipse versions:

Eclipse Version | Downloads
-|-
Neon | 493,589
Mars | 222,210
Luna | 48,596
Kepler | 14,557

As the model is downloaded only once per workspace, these numbers are **lower bounds** on the number of **new workspaces** created using the various Eclipse versions; there may in fact be many more users who happily use Kepler, but haven’t felt the need to create a new workspace in months (and thus haven’t downloaded new models).

I hope you find these numbers useful.

KengoTODA:
> would be great if someone could double-check this under Windows.

I will.

sewe:
> I will.

Thank you, @KengoTODA. Great to see progress on SpotBugs/FindBugs again.

KengoTODA:
I confirmed that it can work even on Windows. One point is that a backslash `\` should be written as `\\` in properties file.

sewe:
> I confirmed that it can work even on Windows. One point is that a backslash \ should be written as \\ in properties file.

Is this something that needs to be documented? Or should be assume that developers know that a plain backslash is the escaping character in a `.properties` file? Personally, I tend towards the latter point of view.

KengoTODA:
> Is this something that needs to be documented?

It is expected behaviour according to [javadoc](http://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-). I think developer can handle this trouble by own, for instance, it is easy to reach [SO page](https://stackoverflow.com/q/5784895) by just google 'properties file backslash'.

eclipsePlugin/build.gradle
iloveeclipse: Please note that SpotBugs *requires* Java 8. So if someone would try to install it on Eclipse before Luna (4.4) it will most likely fail. I doubt there are many using Kepler with additional JDT patch for Java 8, so first really useful Eclipse verson for Java 8 is 4.4 (Luna).

Also there is still FindBugs plugin which supports older Eclipse installations.

So for me we can target Eclipse >= 4.5. Everyone who want to use SpotBugs plugin with older Eclipse versions should clone the code, create a patch and use it.

Issue #210, opened 2017-06-12
Allow additional keywords to search for SpotBugs preference page


sewe:
The preference page of the SpotBugs Eclipse plugin can now be found under _Java_ > _SpotBugs_. Consequently, users who have upgraded from FindBugs to SpotBugs may not find it that easily.

I suggest that the SpotBugs preference page should also be found using the following keywords:

- bugs
- findbugs

Any prefix like “bug”, “spot”, or “find” would work as well, of course.

FWIW, I’m willing to prepare a pull request.

sewe:
Here’s a screenshot that illustrates the new behavior:

![preferences-searching-for-findbugs](https://user-images.githubusercontent.com/491981/27045822-7eafb76c-4fa2-11e7-8523-4fbcf83a8365.png)

KengoTODA:
#211 closed this issue.

Issue #160, opened 2017-04-23
EclipsePlugin-tests plugin classpath is broken


iloveeclipse:
It was broken via #40 .
eclipsePlugin-tests .classpath is wrong, developing tests in Eclipse is not possible anymore.

Also after fixing the main problem with broken .classpath, MANIFEST.MF must be updated to change dependency from old findbugs plugin to the new id.

KengoTODA:
Current status: I found that we should do following to fix.

* `spotbugs` and `spotbugs-annotation` should follow OSGI and export their classes to users
    * `spotbugs-annotation` depends on JSR305, and need to export packages in it
    * Both of them need to copy dependencies into `lib` directory in project (we can ignore them by `.gitignore`)

## MANIFEST.MF in jsr-305

```
Manifest-Version: 1.0
Bundle-Description: JSR305 Annotations for Findbugs
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: org.jsr-305
Archiver-Version: Plexus Archiver
Built-By: lan
Bnd-LastModified: 1490936130302
Bundle-ManifestVersion: 2
Tool: Bnd-2.1.0.20130426-122213
Export-Package: javax.annotation;uses:="javax.annotation.meta";version
 ="3.0.2",javax.annotation.concurrent;version="3.0.2",javax.annotation
 .meta;uses:="javax.annotation";version="3.0.2"
Bundle-Name: FindBugs-jsr305
Bundle-Version: 3.0.2
Created-By: Apache Maven Bundle Plugin
Build-Jdk: 1.8.0_101
```

iloveeclipse:
> Both of them need to copy dependencies into lib directory in project
By dependencies you mean probably all the libraries they have specified on the classpath. Yes. This (manifests and dependencies in lib/) would restore one part of previous state.

Another issue is that .classpath files for plugin projects aren't properly generated - they seem to be generated for plain Java projects, but they should be generated for OSGI bundles (aka Eclipse plugins).

KengoTODA:
@jsotuyod It seems that we need to use OSGI plugin to generate proper manifest file.

This https://github.com/KengoTODA/spotbugs/commit/d490eb0cca07cdd5de09e4fd5162939b79e8f8e2 is my commit to update `spotbugs-annotation` project, however generated Eclipse project seems still invalid.

PR #211, opened 2017-06-12
Additional keywords to search for SpotBugs preference page (#210)

sewe:
This adds the keywords “bugs“ and “findbugs” to “spotbugs” (the default).

### DIFF
--- a/eclipsePlugin/plugin.xml
+++ b/eclipsePlugin/plugin.xml
@@ -1319,7 +1319,13 @@
              class="de.tobject.findbugs.properties.FindbugsPropertyPage"
              id="FindBugsPreferencePage"
              name="SpotBugs">
+          <keywordReference id="com.github.spotbugs.plugin.eclipse.keywords"/>
        </page>
+    </extension>
+    <extension
+          point="org.eclipse.ui.keywords">
+       <keyword id="com.github.spotbugs.plugin.eclipse.keywords"
+             label="bugs findbugs"/>
     </extension>
 
    <extension


sewe:
> Please use full qualified name for id attribute, just in case someone woulduse same "keywords" id.

You are right. Mistakenly thought that IDs without dot in them would automatically be prefixed by the plugin ID. But apparently that’s not the case.

iloveeclipse:
Thanks. 
Questions:
1. You seem to manage Eclipse plugin debugging, despite all gradle related issues. Can you please share a short how to (there was old one inside the plugin, you could just update it). 
2. If you plan to provide more patches and intetested, would you like tobe a committer on the project? I have no much free time and it would be nice to have another committer with Eclipse experience on board.

iloveeclipse:
Oops, looks like my approval was not enough to merge. Stupid interface.

sewe:
> 1. You seem to manage Eclipse plugin debugging, despite all gradle related issues. Can you please share a short how to (there was old one inside the plugin, you could just update it).

I’ve got Buildship set up to compile stuff in Eclipse, but I test by building the update site (`./gradlew :eclipsePlugin:build`) and installing into Eclipse with remote debugging enabled. So yeah, I *manage* Eclipse plugin debugging, but have far more experience with Maven/Tycho than with Gradle – and so far, my Gradle experience wasn’t that great.

> 2. If you plan to provide more patches and intetested, would you like tobe a committer on the project? I have no much free time and it would be nice to have another committer with Eclipse experience on board.

I can imagine becoming a committer, yes. I have a few questions in return, however:

1. Does there exist a communication channel to gather opinions among committers? An issue tracker seems less well suited than, say, a mailing list for this.

2. Who is the Gradle expert in residence? Maybe the old FindBugs Ant build was worse, but the build is an area I find very frustrating at the moment – and am not in a position to help much with, as a Gradle newbie.

iloveeclipse:
Currently the communication channel is bug tracker. 
Remote debugging is not an option for me, this is as if using vi instead of IDE. 

Gradle expert should be @jsotuyod, and I'm also not happy about transition from ant - see #153 and #160. Since I have no time to fix this, I have no working setup and so I effectively can't do much anymore on the project, and this is pity. My hope is that @jsotuyod would fix gradle setup sooner or later.

PR #213, opened 2017-06-19
Use gradle-wrapper v3.5

henrik242:
Somehow the gradle-wrapper jar was still on version 3.2.1


KengoTODA:
The result of `md5sum` for this jar is `4651e6c1060850294c8b720e07b66b36`, which is not same with `8f39e17d09d3b68fa08b8fe95a7be8d7` (https://mvnrepository.com/artifact/org.gradle/gradle-wrapper/3.5) nor `6c6df07ecfa2004b601193824e10ef42` (result of `./gradlew wrapper` in my local). How you generated this jar file?

henrik242:
This was the result of running `./gradlew wrapper` on my machine.  The existing wrapper jar has `3.2.1` in its `MANIFEST.MF` file.

KengoTODA:
I checked the result of `./gradlew wrapper` in another environment (macOS) and found that its result `8841de9a436fcfbdbd0d7d78cff2c56a` is also different. It seems that `wrapper` task does not download jar file but generate it locally.

I will compare .class files instead of checksum.

PR #216, opened 2017-06-25
An attempt to fix gradle eclipse build

iloveeclipse:
Should fix issue #153 and issue #160

I've tried to rebase and use #162 but I didn't manage it, so this PR is "inspired" by #162.

I'm not a gradle expert, so probably I've did something not gradle-like, but at least this one allows me finally to run Eclipse in debugger and to run Eclipse plugin tests from Eclipse.

I would like to integrate this before the first SpotBugs release.


eclipsePlugin/build.gradle
```diff
@@ -102,18 +183,15 @@ def distSpec = copySpec {
...
+  from('../spotbugs/optionalPlugin/noUpdateChecks.jar') { // FIXME : ugly, but update checks are gonna die too
     into 'plugin/'
   }
-  from(configurations.includedLibs) {
-    into 'lib/'
-    rename 'annotations-.*', 'annotations.jar'
-  }
-  from('META-INF/MANIFEST.MF') {
+  from('META-INF/MANIFEST-DIST.MF') {
     filter { line ->
```
jsotuyod: we probably don't need this anymore, since the manifest is being updated before creating the jar, and the version is set there

gradle/eclipse.gradle
```diff
@@ -2,11 +2,26 @@ import org.gradle.plugins.ide.eclipse.model.AccessRule
 
 apply plugin: 'eclipse'
 
+tasks.eclipse.dependsOn(cleanEclipse)
```
jsotuyod: I'm not sure why this would be required

iloveeclipse: To have a consistent build, so that whenever it is executed it does everything from scratch.

jsotuyod: The thing is I don't see how it would ever be inconsistent short of going to the eclipse config and manually screwing it up (which is not a use case we should seriously care about)... This configuration doesn't depend from other settings in non-conventional ways... adding / removing dependencies would force an update on their own...

Issue #215, opened 2017-06-23
Earlier/LaterSubtypes ordering constraint supports only supertypes from SpotBugs itself


sewe:
The following ordering constraints

    <SplitPass>
        <EarlierSubtypes super="org.example.MyDetector" />
        <Later class="org.example.PostMyDecectors" />
    </SplitPass>

will fail at runtime with an `edu.umd.cs.findbugs.PluginException` (Unknown class org.example.MyDetector in constraint selector node). The problem apparently is that [`PluginLoader` uses `Class.forName(String)`](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java#L1308) rather than the three-argument version which could use the class loader from `PluginLoader.classLoader` rather than the class loader `PluginLoader` itself has been loaded with.

sewe:
Considering doing a pull request for this at the moment.

This raises the question: Do we have a test harness in place to test that SpotBugs’s plugin mechanism works as expected (as opposed to a test harness for individual detectors)?

sewe:
> Considering doing a pull request for this at the moment.
> 
> This raises the question: Do we have a test harness in place to test that SpotBugs’s plugin mechanism works as expected (as opposed to a test harness for individual detectors)?

Pull request is ready. Testing has been done manually, however, in a scenario akin to the `EarlierSubtypes/@super = "org.example.MyDetector"` case above.

sewe:
Thanks for merging the pull request.

PR #223, opened 2017-06-26
Earlier/LaterSubtypes supports only built-in supertypes (#215)

sewe:
This lets the `PluginLoader` load the supertype with the plugin's class loader, not the class loader that loaded the SpotBugs core.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
@@ -1268,7 +1268,7 @@ public class PluginLoader {
         }
     }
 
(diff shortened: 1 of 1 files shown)


KengoTODA:
BTW what is the purpose of this change? Bugfix?

KengoTODA:
If so, better to add an entry to CHANGELOG.md.

sewe:
> BTW what is the purpose of this change? Bugfix?

Yes, `EarlierSubtypes`/`LaterSubtypes` didn’t allow you to use your own supertype; you were stuck with the types defined in the SpotBugs JAR. Now you are able to define your own marker interfaces to easily order the detectors in the plugin you contribute; you are no longer stuck with `NonReportingDetector` and friends.

> If so, better to add an entry to CHANGELOG.md.

Will do.

KengoTODA:
Thank you :)

Issue #209, opened 2017-06-12
Compatibility with (and update for) existing FindBugs Eclipse plugin


sewe:
I just stumbled over the fact that the `edu.umd.cs.findbugs.plugin.eclipse.findbugsPlugins` extension point got renamed to `com.github.spotbugs.plugin.eclipse` (interestingly, while leaving the `<findbugsPlugin>` element name unchanged).

I hence wonder what compatibility guarantees towards existing FindBugs plugins are given by SpotBugs? Maybe we can use this issue to discuss/document this.

kzaikin:
IMO it should be possible to replace findbugs with spotbugs reusing all existing goodnes like find-sec-bugs, fb-contrib etc. I don't see a reason to change public API for plugins

sewe:
> IMO it should be possible to replace findbugs with spotbugs reusing all existing goodnes like find-sec-bugs, fb-contrib etc. I don't see a reason to change public API for plugins

Agreed. While the codebase could certainly use some clean-up, version 3.1.0 is probably not the time to do it.

In light of this, I think the `edu.umd.cs.findbugs.plugin.eclipse.findbugsPlugins` shouldn't have been renamed. If you agree, I will open a bug.

iloveeclipse:
If we are talking aboud SpotBugs (not Eclipse) plugins, we should stay compatible. For SpotBugs Eclipse plugin, however, we can't stay compatible simply because we can run in Eclipse along with FindBugs Eclipse plugin and then our contributions would be partly overridden or broken by that. 
So in Eclipse plugin we can't keep compatibility because of the fork.
This is sad but this is unavoidable.

sewe:
>  For SpotBugs Eclipse plugin, however, we can't stay compatible simply because we can run in Eclipse along with FindBugs Eclipse plugin and then our contributions would be partly overridden or broken by that.
So in Eclipse plugin we can't keep compatibility because of the fork.
This is sad but this is unavoidable.

Are you sure? It may be possible to [use `p2.inf`](https://dev.eclipse.org/mhonarc/lists/p2-dev/msg05503.html) to declare the `com.github.spotbugs.plugin.eclipse` as an update of both older versions of itself *and* of the `edu.umd.cs.findbugs.plugin.eclipse` feature. WDYT?

iloveeclipse:
[...]

PR #225, opened 2017-06-30
Keep extension point IDs compatible with FindBugs (#209)

sewe:


KengoTODA:
@sewe Could you update README.md for this change?

sewe:
> Could you consider to update README.md for this change?

@KengoTODA Are you sure that’s necessary? After all, the whole point of this change was to **not** necessitate a change on the client side. And the previous, incompatible change didn’t end up in the changelog, so we don’t need to *undo* it now.

Issue #232, opened 2017-07-02
Eclipse plugin installation data still uses some FindBugs references


iloveeclipse:
License, copyright etc should be changed from University of Maryland to SpotBugs Project, also icon and some texts must be fixed.

iloveeclipse:
The result looks like this:

![installed_spotbugs](https://user-images.githubusercontent.com/964108/27770251-405209b8-5f3b-11e7-8a31-c6c0a021bced.png)

KengoTODA:
I'm not sure but can we change copyright? I think what we can do is just forking based on OSS licence, not changing copyright.

e.g. jenkins's copyright is still Sun, not Oracle nor Cloudbees.

iloveeclipse:
@KengoTODA , you have right, I had to leave existing notice and only add the new one, will push in a minute.

AndreasAumayr:
Not sure if it belongs here, but this is also true for the context menu.
Right click in the package explorer (on a project or file): Spot Bugs > Spot Bugs 
Right click in an editor: Find Bugs > Spot Bugs

iloveeclipse:
I will push the fix in a moment. I think the top most level should be the project label (to *whom* we are talking?), and the action name should impress *what* we do, therefore: "SpotBugs -> Find Bugs".

PR #233, opened 2017-07-02
Update installation texts and icon for Eclipse plugin, see #232

iloveeclipse:
Also removed p2.inf file since we don't use it in the right way.


iloveeclipse:
@KengoTODA : do you have time for a review? If not, I will ask someone else.

PR #238, opened 2017-07-04
Allow upgrades of both FindBugs and SpotBugs features (#209)

sewe:
As noted in #209, this works for both updates from FindBugs and older versions of SpotBugs. The only thing we *may* want to discuss is which maximum version of FindBugs SpotBugs should be considered an update for.

In the pull request, I choose 3.0.2, which is as of this writing the latest version of FindBugs. If FindBugs evolves in a different direction, then we won’t overwrite it, but the present state of affairs is IMHO completely superseded by SpotBugs. Makes sense?

### DIFF
--- a/eclipsePlugin/feature_p2.inf
+++ b/eclipsePlugin/feature_p2.inf
@@ -1,0 +1,7 @@
+# Consider this feature to be an update of both earlier versions of SpotBugs
+# and all versions of FindBugs up to 3.0.2, the last release prior to the fork.
+update.matchExp = providedCapabilities.exists(iu | iu.namespace == 'org.eclipse.equinox.p2.iu' && \
+ ((iu.name == '@FEATURE_ID@.feature.group' && iu.version ~= range('[0, $version$)')) || \
+  (iu.name == 'edu.umd.cs.findbugs.plugin.eclipse.feature.group' && iu.version ~= range('[0, 3.0.3)'))))
+update.severity = 0
+update.description = Update FindBugs to SpotBugs.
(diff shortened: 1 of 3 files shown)


iloveeclipse:
I will test tomorrow. Code looks OK. I don't think FindBugs will evolve further (how this can happen if there is not a single committer left?) except we all move back one day.

eclipsePlugin/feature_p2.inf
```diff
@@ -0,0 +1,7 @@
+# Consider this feature to be an update of both earlier versions of SpotBugs
+# and all versions of FindBugs up to 3.0.2, the last release prior to the fork.
+update.matchExp = providedCapabilities.exists(iu | iu.namespace == 'org.eclipse.equinox.p2.iu' && \
+ ((iu.name == '@FEATURE_ID@.feature.group' && iu.version ~= range('[0, $version$)')) || \
+  (iu.name == 'edu.umd.cs.findbugs.plugin.eclipse.feature.group' && iu.version ~= range('[0, 3.0.3)'))))
+update.severity = 0
+update.description = Update FindBugs to SpotBugs.
```
iloveeclipse: Updating SpotBugs to SpotBugs also shows this message, which is a bit awkward.
I would propose to use here same as in our regular feature description:
"SpotBugs is a program which uses static analysis to look for bugs in Java code, a spiritual successor of FindBugs."

sewe: > Let merge and fix cosmetic issue later.

Sounds good. FWIW, I think `update.description` is optional, so we may just leave it out and use the default.

Issue #242, opened 2017-07-04
Don’t distribute test-only JDepend dependency


sewe:
At the moment, SpotBugs has `jdepend:jdepend:2.9.1` as a *compile*-scoped dependency, even though it is only needed by (a single) unit test. The dependency’s scope should IMHO be reduced to *test* and JDepend should no longer be included in the distribution ZIP (the tests aren’t either).

Note that this potentially affects existing detectors, who may assume that JDepend are on the classpath, but I think this is only a hypothetical issue (I’ve checked `fb-contri` and `findsecbugs`; they do not use JDepend).

iloveeclipse:
I think we are done here now.

PR #243, opened 2017-07-04
Don’t distribute test-only JDepend dependency (#242)

sewe:
@iloveeclipse @jsotuyod Can you have a look at the Eclipse plugin. `grep` turned up a few references to `lib/jdepend.jar` in `MANIFEST-TEMPLATE.MF` and similar files, but I honestly can’t figure out how the Eclipse plugin build is suppossed to work (being used to only Maven+Tycho). The non-Eclipse portions of the build look good to me, though.


iloveeclipse:
I will have a look tomorrow. I don't think we have any dependency to jdepend anywhere except tests, so most likely the reference from manifest can be simply removed.

PR #246, opened 2017-07-05
Cleanup

iloveeclipse:
Fixed various smaller issues with build and Eclipse plugin

### DIFF
--- a/eclipsePlugin/about.properties
+++ b/eclipsePlugin/about.properties
@@ -1,4 +1,6 @@
 blurb=SpotBugs \n\
 \n\
-(c) SpotBugs Project. Licensed under the LGPL.\n\
-https://spotbugs.github.io\n\
+Copyright (c) 2017 SpotBugs Project. All rights reserved.\n\
+Licensed under the LGPL.\n\
+Copyright 2003-2008 by the University of Maryland. \n\
+Visit https://spotbugs.github.io\n\
--- a/eclipsePlugin/about.properties
+++ b/eclipsePlugin/about.properties
@@ -1,6 +1,6 @@
 blurb=SpotBugs \n\
 \n\
 Copyright (c) 2017 SpotBugs Project. All rights reserved.\n\
-Licensed under the LGPL.\n\
 Copyright 2003-2008 by the University of Maryland. \n\
-Visit https://spotbugs.github.io\n\
+Licensed under the LGPL.\n\
+Visit https://spotbugs.github.io/\n\
--- a/eclipsePlugin/build.gradle
+++ b/eclipsePlugin/build.gradle
@@ -68,6 +68,10 @@ dependencies {
   }
 }
 
+clean {
+	delete "lib"
+}
+
 task copyLibsForEclipse (type: Copy) {
     dependsOn configurations.embeddedLibs
     from configurations.embeddedLibs.files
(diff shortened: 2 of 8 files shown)


eclipsePlugin/about.properties
```diff
@@ -1,4 +1,6 @@
 blurb=SpotBugs \n\
 \n\
-(c) SpotBugs Project. Licensed under the LGPL.\n\
-https://spotbugs.github.io\n\
+Copyright (c) 2017 SpotBugs Project. All rights reserved.\n\
+Licensed under the LGPL.\n\
+Copyright 2003-2008 by the University of Maryland. \n\
```
sewe: I would move the two copyright lines together. Also, why is the date range (2003-2008) different from the ranges used below (2006-2011)?

eclipsePlugin/about.properties
```diff
@@ -1,4 +1,6 @@
...
+Visit https://spotbugs.github.io\n\
```
sewe: The URI should end with a trailing slash: `https://spotbugs.github.io/` (this is the canonical form)

eclipsePlugin/build.gradle
```diff
@@ -74,6 +74,7 @@ task copyLibsForEclipse (type: Copy) {
     into "lib"
     include "*.jar"
     exclude "*xml-apis*.jar"
+    exclude "jdepend*.jar"
```
sewe: Why is this still necessary? Can’t we switch `configurations.embeddedLibs` to the *compile* rather than *test* scope of `project(':spotbugs')`?

@jsotuyod This may need a Gradle expert. (As a Maven guy myself I am continuously confused by the script-ness of Gradle…)

eclipsePlugin/about.properties
iloveeclipse: Regarding the date: I just kept what was in the file before. I'm not a layer, so just trying to be polite.
Regarding the lines: OK.

eclipsePlugin/about.properties
iloveeclipse: OK

eclipsePlugin/build.gradle
iloveeclipse: If I only know how :-)

iloveeclipse:
Ping..

sewe:
> Ping..

@iloveeclipse Are you pinging me or @jsotuyod?

From my side, the change looks good – with the exception of the `eclipsePlugin/build.gradle`, which could use some Gradle expertise.

iloveeclipse:
Thanks Andreas for fast feedback.
I just want merge that stuff into and deploy latest greatest snapshot on our update site *without* jdepend.
Given the fact that the exclude list in copyLibsToEclipse was not invented in this PR, I think we can merge & it would be great if you would approve that.

iloveeclipse:
Thanks Andreas!

eclipsePlugin/build.gradle
jsotuyod: Sorry, I missed this notification. This exclude should not be necessary. `jdepend` is a `testCompile` dependency under `:spotbugs`, and therefore not part of the default configuration being used here as dependency.

As to moving from `embeddedLibs` to compile, we would need to change the eclipse jars from `compile` to `compileOnly`, and make sure those are not transitively included in the jar, but I believe it should work just fine.

jsotuyod: FYI, I'll be tackling this and some other minor gradle improvements. I'll submit a PR after I test it.

PR #247, opened 2017-07-07
Convert unit tests of spotbugs project from JUnit 3 to 4

sewe:
This conversion was done mostly automatically, thanks to a neat [Eclipse plugin](http://blog.vogella.com/2014/11/03/migration-to-junit-4/). I limited myself to the `spotbugs` project, however, as `spotbugs-tests` contains some abstract base classes like `FindBugsTestCase` I am not 100% sure the Eclipse plugin will convert truthfully. These  are left for a later clean-up pass.

At any rate, this change already gets rid of 232 warnings in Eclipse about use of deprecated methods (`junit.framework.Assert.*` and friends). I also made two ignored test cases more explicit (`@Ignore` rather commented out).


sewe:
OK, these three changes are good to go.

@iloveeclipse Would you care to review them?

PR #248, opened 2017-07-07
Cleanup Eclispe plugin build.gradle

jsotuyod:
- Little by little we are going towards a more standard build and less
ad-hoc setup

### DIFF
--- a/eclipsePlugin-junit/build.gradle
+++ b/eclipsePlugin-junit/build.gradle
@@ -2,6 +2,7 @@ apply from: "$rootDir/gradle/checkstyle.gradle"
 
 dependencies {
   compile project(':eclipsePlugin')
+  testCompile project(path: ':eclipsePlugin', configuration: 'compileOnly')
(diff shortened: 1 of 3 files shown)


sewe:
@jsotuyod Nice of you to ask me for a review, but I honestly don’t have enough Gradle experience to review changes to the build, which at the moment still seems to suffer from much the script-ness of the previous Ant build (not your fault!).

Let’s hope we can move towards a more declarative structure soon. If not, I hereby volunteer to port the build to Maven + Tycho. 😉

jsotuyod:
@sewe just checking I didn't miss anything and I'm actually putting back in something we wanted out should be enough.

I believe the build complexity can be greatly reduced on all modules except maybe this one. We can certainly improve it a lot still, but I don't think we will be able to get rid of all of it (ie: stuff like MacOs vs rest of the world eclipse structure). We may though at least organize it better (move that to a separate file under `gradle/`).

PR #252, opened 2017-07-10
Update building_spotbugs_plugin.txt

Bananeweizen:
* Windows uses double backslashes in properties files
* Since the new smart import wizard it is important to not try to import the root level project, since there would be a name clash between the root spotbugs project and the spotbugs\spotbugs project.

I managed to setup my workspace with these changed instructions.

### DIFF
--- a/eclipsePlugin/doc/building_spotbugs_plugin.txt
+++ b/eclipsePlugin/doc/building_spotbugs_plugin.txt
@@ -9,7 +9,7 @@ Install Eclipse
 ================
 1) Download and install Eclipse 4.6.+ (minimum supported version).
 2) Download and install EGit plugin for Git.
-    Chose [Help | Software Updates... | Available Software | Add Site...] and use
+    Choose [Help | Software Updates... | Available Software | Add Site...] and use
     http://download.eclipse.org/egit/updates as update site url. Select both checkboxes
     [Eclipse Git Team Provider] and [JGit]
 3) Uncheck "Task focused interface" child element of "Eclipse Git Team Provider".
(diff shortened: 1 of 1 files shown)


eclipsePlugin/doc/building_spotbugs_plugin.txt
```diff
@@ -41,9 +41,10 @@ Import the projects
 8) Choose [File | Import General | Existing Projects into Workspace | Next]. The Import project import wizard opens.
 9) In the wizard, proceed as follows:
    a) Choose [Select root directory].
-   b) Enter ~/git/spotbugs. Click [Browse]
-   c) Click [Finish].
-   g) Now "spotbugs" project will be compiled by Eclipse and should not have any errors,
+   b) Enter ~/git/spotbugs. Click [Browse]. Make sure the checkbox "Search for nested projects" is selected.
```
sewe: @henrik242 Shouldn’t we clone to a different directory (say `spotbugs-project`) instead? AFAICT, the above does not make **all** projects available in Eclipse, which is not a good basis to start contributing on SpotBugs; you miss out on important files like the root `build.gradle`.

(FWIW, I’ve opened a bug with the Eclipse Buildship project about this problematic import behavior a while ago: [Bug 518089](https://bugs.eclipse.org/bugs/show_bug.cgi?id=518089))

henrik242: This is @Bananeweizen's commit, not mine.  I thought it looked reasonable, but we can of course change it to something better :)

Bananeweizen: @sewe It depends on your point of view. The document is titled "Building spotbugs plugin" (and lives in the Eclipse plugin directory), which makes me think the original author only had the Eclipse plugin in mind. I guess you contribute to both the plugin and the main project. I myself see the main project just as a dependency for the plugin, and basically don't care about it. :)
However, of course feel free to make it work for both cases. My change just assures it works _at least_ for someone who wants to build the eclipse plugin (since it did not work at all with the old description).

PR #254, opened 2017-07-12
Only create SourceFinder once to improve performance. Resolve #253

### LINKED ISSUE #253: Performance issue with large number of source dirs
The method Project.addSourceDir() creates a new sourceFinder each time it is called. The constructor of the SourceFinder will check all files in the SourceFinder for existence. In all places Project.addSourceDir is called this happens in a loop. This leads to behaviour of quadratic complexity in checking file existence.

When using the Gradle Findbugs plugin it will call .allJava() on the source directories leading it to list all java files separately as source dirs. One of our projects has 15,000 java files and thus it takes 7 minutes to scan about 100 million files.

I have submitted pull request #254 for this.

piderman314:
Applying this fix to both Find/Spotbugs itself and the Jenkins plugin reduced our build time from 30 minutes to about 15.


spotbugs/src/main/java/edu/umd/cs/findbugs/SAXBugCollectionHandler.java
```diff
@@ -662,7 +665,9 @@ public void endElement(String uri, String name, String qName) throws SAXExceptio
         } else if (nestingOfIgnoredElements > 0) {
             // ignore it
         } else if ("Project".equals(qName)) {
-            // noop
+            Project project = this.project;
+            assert project != null;
```
iloveeclipse: Please remove assert or use Objects.requireNonNull.

spotbugs/src/main/java/edu/umd/cs/findbugs/Project.java
```diff
@@ -237,18 +237,26 @@ public boolean addFile(String fileName) {
...
-     *            the directory to add
-     * @return true if the source directory was added, or false if the source
-     *         directory was already present
+     * @param dirNames
+     *            the directories to add
+     * @return true if a source directory was added, or false if all source
+     *         directory were already present
      */
-    public boolean addSourceDir(String dirName) {
+    public boolean addSourceDirs(Collection<String> dirNames) {
```
iloveeclipse: Please leave the old method there but make it deprecated. We don't want to break API yet.

spotbugs/src/main/java/edu/umd/cs/findbugs/Project.java
```diff
@@ -237,18 +237,26 @@ public boolean addFile(String fileName) {
...
     /**
-     * Add a source directory to the project.
+     * Add source directories to the project.
      *
-     * @param dirName
-     *            the directory to add
-     * @return true if the source directory was added, or false if the source
-     *         directory was already present
+     * @param dirNames
+     *            the directories to add
```
iloveeclipse: I hate such API.

The old javadoc was fuzzy - what is a "directoryName"? Is this just a name? a relative path? an absolute path? 
Looking at the FindBugsWorker we see that this is full absolute path to the directory, NOT a name. Looking into the makeAbsoluteCwdCandidates() we see it can be almost anything.

So obviously the javadoc is lying here. It means: either full absolute paths or paths relative to any of the "working" directories added via addWorkingDir().

Please fix this.

eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
```diff
@@ -233,9 +234,7 @@ public void work(List<WorkItem> resources) throws CoreException {
 
     private void configureSourceDirectories(Project findBugsProject, Map<IPath, IPath> outLocations) {
         Set<IPath> srcDirs = outLocations.keySet();
-        for (IPath iPath : srcDirs) {
-            findBugsProject.addSourceDir(iPath.toOSString());
-        }
+        findBugsProject.addSourceDirs(srcDirs.stream().map(IPath::toOSString).collect(Collectors.toList()));
```
henrik242: A static import of toList() is slightly prettier.

henrik242:
@piderman314  Do you have updates for this PR?

piderman314:
Sorry I've been a little bit busy of late and I was struggling with the javadoc a bit. I'll see if I can make time for this in the weekend.

spotbugs/src/main/java/edu/umd/cs/findbugs/SAXBugCollectionHandler.java
piderman314: Done.

spotbugs/src/main/java/edu/umd/cs/findbugs/Project.java
piderman314: Done.

spotbugs/src/main/java/edu/umd/cs/findbugs/Project.java
piderman314: Ok I have updated the javadoc. I hope it's better like this.

PR #257, opened 2017-07-13
Remove calls to no-op canonicalizeString methods

sewe:
These methods have been deprecated and turned into no-ops with #128. This change removes the calls (but not the methods themselves, so not to break API) to reduce the number of distracting warnings (12 warnings less).

@KengoTODA As you implemented the performance improvement in #128 that turned these methods to no-ops, would you care to review this change please?


sewe:
Could someone please review this pull request? Change is pretty trivial, but we would get rid of a few more deprecation warnings this way.

PR #263, opened 2017-07-15
Use rules by gitignore.io

koppor:
[gitignore.io](https://www.gitignore.io/) generates useful `.gitignore` file. They build on [GitHub's .gitignore colleciton](https://github.com/github/gitignore/) and extend them with [their own patches](https://github.com/dvcs/gitignore).

This PR makes use of gitignore.io's patterns with the aim to reduce maintenance issues of `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,6 @@
-.gradle
-build
 .idea
-.project
-.settings
-.classpath
 *.iml
-*.swp
 classesEclipse
-.DS_Store
 deps/
 eclipse/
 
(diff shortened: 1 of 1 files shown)


henrik242:
Hmm, `.gitignore` really doesn't see many changes in spotbugs, and we know that the current version works well for us (I think?).  I'm not necessarily against the PR, I just don't see why this is needed.

sewe:
A `.gitignore` IMHO has two types of entries:

1. Build outputs
2. Editor/IDE-specific configuration files

I’d really like the former to be a short, hand-crafted list, as we really ought to know what the outputs of our build are and where they are placed.

For the latter I am not against copying long lists of externally-provided entries, although things like `.cproject` (from Eclipse CDT) should never occur and I see little value in cluttering out `.gitignore` with them. But if someone else has already done the legwork of compiling IntelliJ-specific entries (which I can imagine being used), then fine – let’s use them.

iloveeclipse:
Same here.

Issue #262, opened 2017-07-14
Avoid using Constants interface deprecated in BCEL 6.x


sewe:
At the moment, the `spotbugs` project shows more than 1,800 warnings, *1,506* of which are caused by our use of the `org.apache.bcel.Constants` interface that has been [deprecated with BCEL 6.0](http://commons.apache.org/proper/commons-bcel/xref/org/apache/bcel/Constants.html).

IMHO, we should switch to the BCEL 6.0 replacement, `org.apache.bcel.Const`, to get our list of warnings down to a manageable level.

This change would then come in two parts:

1. Use the constants in `Const` rather than `Constants` ourselves.
2. No longer implement `Constants` in our classes.

The first part is a purely internal change, which doesn’t affect our clients. But no longer implementing the `Constants` interface *may* affect them.

Now, the question is whether we strive to be binary compatible (old plugins should run on the new SpotBugs version) or also source compatible (old plug-ins should still compile against the new SpotBugs version) for SpotBugs 3.1.0.

Binary compatibility is easy: Most constants are primitive and inlined during compilation and the few constants which are not (`OPCODE_NAMES`, `NO_OF_OPERANDS`, …) are accessed using `invokestatic`. Hence, existing clients which extend a SpotBugs class that previously implemented `Constants` would just continue to use `invokestatic` to access `Constants.*`; they wouldn’t notice that they no longer (transitively) implement the interface. This leaves only the oddball case of someone doing an `instanceof Constants`, which I am confident nobody did (because: why?).

Source compatibility is probably broken, as I suspect not everyone who extends, e.g., `BytecodeScanningDetector` adds an `implements Constants` clause to their detector when accessing one of the constants themselves.

Now the **question**: Should we do both parts of this change for SpotBugs 3.1.0, even though part 2 breaks source compatibility, or should we leave part 2 to SpotBugs 4.0.0 and only do part 1 for 3.1.0? (FYI, part 1 would already get rid of most of the deprecation warnings.)

Thoughts?

iloveeclipse:
Without looking into the code: 3.1.0 release should be compatible as much as we can (to simplify transition from FB to SB), 4.0.0 is by definition "open" for new features and API can be broken (not that this is nice, but where we can't avoid this, we can't keep compatibility).

sewe:
[...]

PR #265, opened 2017-07-16
This implements part 1 of the changes described in #262

sewe:
This brings the number of warnings on the `spotbugs` project down from 1,739 to 233.

Unfortunately, we cannot yet use `import static org.apache.bcel.Const.*`, as we still implement the deprecated `org.apache.bcel.Constants` for compatibility reasons (part 2 of the changes described in #262) and the constants therein get picked over the ones from the static import.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v0.3](http://
 
 * `jdepend:jdepend:2.9.1` is no longer a compile-scoped dependency but only test-scoped. ([#242](https://github.com/spotbugs/spotbugs/issues/242))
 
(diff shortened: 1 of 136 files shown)


sewe:
As @iloveeclipse mentioned on issue #268, he’s to busy for reviews. Could someone else (@henrik242, @KengoTODA , @h3xstream) please have a look at my pull requests from the last few days. Thank you.

sewe:
> phew, that was a long list of changes. lgtm.

Thanks for reviewing them anyway, @henrik242. Once we can use `import static` (part 2 of #262), the delta to the before-`Const` state will be limited to the `import` section and a removed `implements Constant` clause. It’s just the intermediate state that is ugly due to all the explicit `Const.` qualifications.

PR #269, opened 2017-07-18
Clean up CountMatcher Hamcrest Matcher

sewe:
Don’t duplicate `Matcher` factory methods, but keep them in `Matcher` class following Hamcrest's convention. Also, make `CountMatcher` a `TypeSafeMatcher`; it will no longer consider `null` or non-`Iterable`s a match for count 0.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
@@ -89,10 +89,6 @@ public class AbstractIntegrationTest {
         return bugReporter.getBugCollection();
     }
 
-    protected static <T> Matcher<Iterable<T>> containsExactly(final Matcher<T> matcher, final int count) {
-        return new CountMatcher<T>(count, matcher);
-    }
-
     /**
      * Sets up a FB engine to run on the 'spotbugsTestCases' project. It enables
      * all the available detectors and reports all the bug categories. Uses a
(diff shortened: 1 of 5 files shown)


KengoTODA:
@sewe you've deleted public method `SpotBugsRule#containsExactly`, then please update update `CHANGELOG.md` as usual. And please update following contents which depends on it:

1. [javadoc](https://github.com/spotbugs/spotbugs/blob/6a3ff9d19f73ad716629609f33e5848c4432d55f/test-harness/src/main/java/edu/umd/cs/findbugs/test/SpotBugsRule.java#L29)
2. [maven-archetype](https://github.com/spotbugs/spotbugs-archetype/blob/master/src/main/resources/archetype-resources/src/test/java/MyDetectorTest.java)

Thanks in advance! 🙏

henrik242:
@KengoTODA Should we keep that method (as deprecated) in 3.1.0 to avoid breaking compatibility?

sewe:
> Should we keep that method (as deprecated) in 3.1.0 to avoid breaking compatibility?

@henrik242 I don’t think this method ever was API of FindBugs 3.0.x, as neither the test cases nor the test harness was [distributed on Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.google.code.findbugs%22) before. Adding it to the changelog as *removed* should be enoygh for those users who following the RCs alongs.

@KengoTODA Will make the changes you pointed out later tonight.

KengoTODA:
Thanks @sewe, I agree that `@Deprecated` is needless for this case.

Issue #214, opened 2017-06-19
unable to infer FindBugs version using gradle plugin 1.1


brharrington:
I get an exception when trying the latest version of the gradle plugin:

```
Caused by: org.gradle.api.GradleException: Unable to infer the version of FindBugs from currently specified FindBugs classpath: [spotbugs-3.1.0-RC3.jar, asm-debug-all-6.0_ALPHA.jar, jcip-annotations-1.0.jar, dom4j-1.6.1.jar, commons-lang-2.6.jar, jaxen-1.1.6.jar, jdepend-2.9.1.jar, jFormatString-3.0.0.jar, spotbugs-annotations-3.1.0-RC3.jar, xml-apis-1.0.b2.jar, jsr305-3.0.1.jar]
        at com.github.spotbugs.internal.spotbugs.FindBugsClasspathValidator.getFindbugsVersion(FindBugsClasspathValidator.java:43)
        at com.github.spotbugs.internal.spotbugs.FindBugsClasspathValidator.validateClasspath(FindBugsClasspathValidator.java:19)
        at com.github.spotbugs.SpotBugsTask.run(SpotBugsTask.java:208)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
        at org.gradle.api.internal.project.taskfactory.DefaultTaskClassInfoStore$StandardTaskAction.doExecute(DefaultTaskClassInfoStore.java:141)
        at org.gradle.api.internal.project.taskfactory.DefaultTaskClassInfoStore$StandardTaskAction.execute(DefaultTaskClassInfoStore.java:134)
        at org.gradle.api.internal.project.taskfactory.DefaultTaskClassInfoStore$StandardTaskAction.execute(DefaultTaskClassInfoStore.java:123)
        at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:692)
        at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:675)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.execute(ExecuteActionsTaskExecuter.java:115)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.execute(ExecuteActionsTaskExecuter.java:109)
        at org.gradle.internal.Transformers$4.transform(Transformers.java:169)
[...]

PR #273, opened 2017-07-21
Release 3.1.0 RC4

KengoTODA:
This MR will update versions in `build.gradle` and `CHANGELOG.md`. I've released followings:

* SpotBugs 3.1.0 RC4 (will be available at Maven Central several hours later)
* Gradle Plugin v1.2 (already available at https://plugins.gradle.org/plugin/com.github.spotbugs)

[Eclipse update site](https://spotbugs.github.io/eclipse-candidate/) will be updated by another PR in [spotbugs/spotbugs.github.io](https://github.com/spotbugs/spotbugs.github.io).


brharrington:
For what it's worth, I tried the 1.2 gradle plugin and it seems to work fine. I had problems with both 1.0 (#181) and 1.1 (#214).

boris-petrov:
For me there are still issues even with Gradle plugin 1.2 and RC4. Here's a couple:

1) I'm getting errors:

```
Scanning archives (765 / 765)
2 analysis passes to perform
Pass 1: Analyzing classes (1308 / 1308) - 100% complete
Pass 2: Analyzing classes (572 / 572) - 100% complete
Done with analysis
  accept
  apply
  iterator
  test
  run
  compare
  isLoggable
  get
  subscribe
  handleEvent
  call
  createExecuteAsyncRunnable
  onEvent
  applyAsInt
  uncaughtException
```

And the build fails.

2) I'm using the following configuration:

```gradle
tasks.withType(FindBugs) {
	reports {
		xml.enabled = false
		html.enabled = true
	}
}
```

(First of all - are you going to change the task name from `FindBugs` to `SpotBugs`?)
This worked fine with FindBugs, but now an XML report is generated instead of an HTML one.

3) The tasks are called `findbugs*SourceSetName*`. Is this going to change?

KengoTODA:
@boris-petrov I think it's renamed. Try to follow [official migration guide](http://spotbugs.readthedocs.io/en/latest/migration.html#findbugs-gradle-plugin) first

brharrington:
@boris-petrov 

1. Checking the logs more closely I also see that, but I'm not sure what harm it causes. It doesn't seem to fail the builds for me. I did check that I can change the source to make spotbugs fail the build due to a rule violation.

2. The following works for me:

    ```gradle
      tasks.withType(com.github.spotbugs.SpotBugsTask) {
        reports {
          xml.enabled = false
          html.enabled = true
        } 
      }
    ```

    I think the previous one referencing FindBugs still works (in the sense that it doesn't fail the build) because the findbugs plugin is included as one of the core gradle code quality plugins.

3. The task names are still `findbugs${SourceSet}`. Since the configuration blocks are called spotbugs, I agree it can be confusing as to where to use "findbugs" and where to use "spotbugs". It hasn't come up much for me because I typically run the reports via `check` or `build` rather than directly. Not sure how the project owners feel about changing it to be consistently spotbugs at least for the gradle usage, but I created a PR (#275) to see what others think.

boris-petrov:
@KengoTODA - I did follow it a long time ago. Nowhere does it mention the issues that I discussed here (or maybe that has changed, I'm sorry if that's the case).

@brharrington - thanks for the answers!

1) You're right, that wasn't why the build was failing - I saw what the issue was once I fixed 2). You'd better mention that somewhere in the migration guide - or maybe I'm just blind and haven't seen it, sorry if that's the case.

3) The question is that we have a bunch of sourcesets and we don't want all of them to be checked - so we have some custom code that runs the tasks - and that's why I said that the task is still called `findbugs`. It really doesn't matter to us, it's just confusing. :)

Now everything seems to be working fine, thank you very much! Good luck with the project!

brharrington:
> You'd better mention that somewhere in the migration guide - or maybe I'm just blind and haven't seen it, sorry if that's the case.

#276 adds it to the migration guide.

KengoTODA:
@henrik242 Your commit 7a6eb4163a4db12142c077e64664bcc602aa5474 seems LGTM. Can you merge this PR?

henrik242:
@KengoTODA It seems you need to resolve the conflicts this time.  GitHub doesn't show the resolve interface.

PR #286, opened 2017-07-25
Add sonar badges

kzaikin:


iloveeclipse:
@kzaikin : I see you and @KengoTODA as members of spotbugs in https://sonarcloud.io/organizations/spotbugs/members. I would like to be member to configure the stuff - what should I do?

kzaikin:
@iloveeclipse here you go

kzaikin:
My experience with Sonar is that new values are more important.
Let me leave New Debt and new Coverage

Issue #289, opened 2017-07-27
Remove the YourKit profiling callback


sewe:
As discussed in spotbugs/discuss#3, none of the current maintainers is using the `findbugs.yourkit.enabled` option or holds an YourKit license, but we include a very specific version of the `yjp-controller-api-redist.jar` with every SpotBugs download (which may not even work with newer versions of YourKit).

We hence should remove the YourKit profiling callback.

sewe:
Remvoed with 9e668cd. Closing.

PR #290, opened 2017-07-27
Remove the YourKit profiling callback (#289)

sewe:
This follows a discussion in spotbugs/discuss#3.


sewe:
@henrik242 Care to have a look as well? If you agree with this change, please merge and close.

PR #298, opened 2017-08-02
GH#297 add the full xml headers

xenoterracide:


xenoterracide:
btw, you might want to put the xsd reference somewhere other than raw git, but that's all I have right now.

KengoTODA:
I think it is ok to use raw git, but it is better to use tag instead of master branch.

henrik242:
> I think it is ok to use raw git, but it is better to use tag instead of master branch.

@xenoterracide Are you looking at this?

xenoterracide:
I don't know what tag I should use, I mean it should basically be the current or next release. I mean I could do one for the existing release... 3.0.x but that has a different path from the 3.1.x so... kind of hard to know what to change it to.

henrik242:
@xenoterracide I'd go for https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0_RC5/spotbugs/etc/findbugsfilter.xsd

PR #294, opened 2017-07-31
Fixes issue #287

iloveeclipse:
Delete site directories before generating them again to avoid multiple
or obsoleted artifacts on the update sites. Also delete build data on
clean build.

Signed-off-by: Andrey Loskutov <loskutov@gmx.de>

PR #302, opened 2017-08-06
Deploy latest Eclipse Plugin automatically to gh-pages

KengoTODA:
This is related with https://github.com/spotbugs/spotbugs.github.io/pull/29
I found that we don't need `git submodule`, just make new repository https://github.com/spotbugs/eclipse-latest/ then we can host webpage at https://spotbugs.github.io/eclipse-latest/.

When changes have merged to `master` branch, Travis CI will force-push newly built update site to GitHub pages. So this is not daily-build, but more frequent so I've renamed it to `eclipse-latest`.


iloveeclipse:
@KengoTODA : good idea! I will check this tomorrow. 
Can you please also review #294?

iloveeclipse:
See https://travis-ci.org/spotbugs/spotbugs/builds/261718490?utm_source=email&utm_medium=notification

This seem to fail the build if there were no build before, also I see that travis still uses Eclipse 3.6 from cache. I think we should fix .travis.yml, I will push PR shortly.

PR #307, opened 2017-08-07
GH#296 GH#297 fix XSD so that it requires nested elements

xenoterracide:


spotbugs/etc/findbugsfilter.xsd
```diff
@@ -264,137 +317,151 @@
...
+                <element ref="fb:BugPattern"/>
+            </choice>
         </sequence>
     </complexType>
 
     <complexType name="NotType">
         <annotation>
-            <documentation></documentation>
-        </annotation>
-        <sequence maxOccurs="unbounded" minOccurs="0">
```
sewe: Why not simply switch from `minOccurs="0"` to `minOccurs="1"` here? Why making the bigger change to the nested `choice` element?

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -264,137 +317,151 @@
...
         <restriction base="int">
-            <minInclusive value="1"></minInclusive>
-            <maxInclusive value="20"></maxInclusive>
+            <minInclusive value="1"/>
+            <maxInclusive value="20"/>
         </restriction>
     </simpleType>
 
     <complexType name="MatchType">
-        <sequence maxOccurs="unbounded" minOccurs="0">
```
sewe: Why not simply switch from `minOccurs="0"` to `minOccurs="1"` here? Why making the bigger change to the nested `choice` element?

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -264,137 +317,151 @@
...
-            <element ref="fb:Method" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Not" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Or" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Package" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Priority" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Rank" maxOccurs="unbounded" minOccurs="0"></element>
-            <element ref="fb:Source" maxOccurs="unbounded" minOccurs="0"></element>
+            <documentation/>
+        </annotation>
+        <sequence maxOccurs="unbounded" minOccurs="2">
```
sewe: Is `minOccurs="2"` really needed? Granted, `minOccurs="1"` is not very useful, but I can imagine a code generator emitting this as a special case.

spotbugs/etc/findbugsfilter.xsd
xenoterracide: same as above

spotbugs/etc/findbugsfilter.xsd
xenoterracide: I did it on these because with an AND and an OR I would consider it an XML issue to only have 1 sub element

xenoterracide:
ok, fixed the changes such that they are as minimal as possible, sorry my IDE did some of that, I shouldn't have sent the PR without cleaning it up.

 unfortunately a sequence of choice is mandatory to get the truly desired behavior, I tried like 6 different things before finding one that marked things invalid and suggested things appropriately. 

If you really want, I can amend the commit further so that `<And>` and `Or` don't require 2 sub-elements, but I would consider not having 2 to be a bug in your configuration, what does that mean. Especially given when if you think it does nothing, my original problem was that an empty `<And>` which would intuitively match nothing, did not.

iloveeclipse:
> guess I'm talking to myself?

I hope not always ...

PR #311, opened 2017-08-09
deploy update sites before SNAPSHOT artifacts

KengoTODA:
This is just a trial: we've generated site contents in `after_success` phase so when we enter `deploy` phase we should have generated site in `build` directory.

I'm still not sure why site dirs are deleted by uploading SNAPSHOT artifacts, but this try will help us to identify the root cause (maybe related with #294, because deploy had no problem before we merge that...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ script:
   - ./gradlew build smoketest sonarqube -S
 
 after_success:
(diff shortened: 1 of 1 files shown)


iloveeclipse:
OMG

iloveeclipse:
But finally we have it deployed !!! Great!!!
See https://spotbugs.github.io/eclipse-latest/site.xml

PR #313, opened 2017-08-11
Deploy released Eclipse Plugin to Update Site automaticallt

KengoTODA:
By these two `deploy` tasks, we can publish Eclipse Plugin to Update Site automatically.

For `eclipse` site, we deploy changes which is tagged by non-RC version. So the commit with tag `3.1.0_RCx` will not be deployed, but the commit with tag `3.1.0` will be deployed to https://spotbugs.github.io/eclipse/.

For `eclipse-candidate` site, we deploy all tagged changes. So both of `3.1.0_RCx` and `3.1.0` will be hosted at https://spotbugs.github.io/eclipse-candidate/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,14 @@ deploy:
     on:
       tags: true
       condition: "$JDK_FOR_TEST = oraclejdk8"
+  - provider: pages
+    skip_cleanup: true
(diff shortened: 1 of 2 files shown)


KengoTODA:
I forgot to update document, I will reopen this later.

Issue #323, opened 2017-08-17
All command line scripts are broken


iloveeclipse:
After commit 8b518ec0 all command line scripts do not forward arguments, so all command line tasks work now without any arguments. This is a blocker.

They all should append `$@` at the end of the new command line to forward arguments to `fbwrap` script.

sewe:
Sorry about that. I’ll fix this ASAP.

iloveeclipse:
Closed via #324.

PR #324, opened 2017-08-17
Fix broken command-line scripts (#323)

sewe:
In addition to now correctly passing on argument lists, this change also further improves upon the scripts; they now `exec` the script they warp, thus saving an extra process. Also, a few scripts which are effectively synonyms of another script (`unionBugs`/`unionResults` and `updateBugs`/`computeBugHistory`) now `exec` canonical scripts.

### DIFF
--- a/spotbugs/src/scripts/deprecated/bugHistory
+++ b/spotbugs/src/scripts/deprecated/bugHistory
@@ -1,11 +1,5 @@
 #! /bin/sh
 
-@GET_FBHOME@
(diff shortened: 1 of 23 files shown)


sewe:
@iloveeclipse I hope you don’t mind me doing some minor improvements beyond fixing the (admittedly stupid) parameter bug: The scripts now `exec` their replacement and a few synonym scripts are made more explicit (and should probably be deleted post 3.1.0).

PR #325, opened 2017-08-18
use isolated account for GitHub automation

KengoTODA:
Previously we use GitHub token generated by @kengotoda so post from SonarQube uses his icon.
It makes our communication complicated.

Now we've created @spotbugs-bot and this change will use new token generated by her.


KengoTODA:
I've confirmed that new GitHub token works as expected; It updated eclipse-latest successfully. https://github.com/spotbugs/eclipse-latest/commit/4478b714ec87fdcdd247683b2c583c2a5b73d7be

PR #337, opened 2017-08-20
Add more tests to gradle plugin, and close #320

### LINKED ISSUE #320: Cannot convert the provided notation to a File or URI: classesDirs
The full error is:

```
Cannot convert the provided notation to a File or URI: classesDirs.            
The following types/formats are supported:                                     
  - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.                                                                             
  - A String or CharSequence URI, for example 'file:/usr/include'.             
  - A File instance.                   
  - A Path instance.                   
  - A URI or URL instance.
```

This appears when using Gradle plugin version 1.3. Probably this commit broke it:

https://github.com/spotbugs/spotbugs/commit/850dc13c1d7d0727024f13814b6e26d1b2ff6371

KengoTODA:
I've created a simple test for gradlePlugin, but it is not so nice because it cannot find packaged artifacts (spotbugs, spotbugs-annotations etc.) automatically so we need to install them into Maven local repository before we run test.

If you have idea to delete `./gradlew spotbugs:install -x test` from travis configuration, please let me know.


KengoTODA:
Hi @johnscancella, can I ask you to review this change?

johnscancella:
Yes I will take a look

johnscancella:
Gradle support for testing the functionality of a plugin in unit tests has always been sorely lacking. So I am thinking instead of hacking around it and installing the plugin locally, does it make more sense to create a small project that uses the plugin instead to test actual functionality?

If you would rather keep it as a unit test here are my comments:
It seems like testSpotBugsTaskCanRun was copied from TestSpotBugsTasksExist. Because there is much overlap I think it would be better to pull out the duplicated logic of setting up the test project into a `@Before` annotated method.
for example
```
@Before
  public void setup() throws Exception {
    File spotBugsLibraryFolder = new File("../spotbugs/build/libs/");
    System.err.println(spotBugsLibraryFolder.getAbsolutePath());
    String buildScript = "plugins {\n" +
        "  id 'java'\n" +
        "  id 'com.github.spotbugs'\n" +
        "}\n" +
        "version = 1.0\n" +
        "repositories {\n" +
        "  mavenCentral()\n" +
        "  mavenLocal()\n" +
        "}";

    File sourceDir = folder.newFolder("src", "main", "java");
    File to = new File(sourceDir, "Foo.java");
    File from = new File("src/test/java/com/github/spotbugs/Foo.java");
    Files.copy(from, to);

    File buildFile = folder.newFile("build.gradle");
    Files.write(buildScript.getBytes(), buildFile);
  }
```

KengoTODA:
Thanks @johnscancella for your review!

I felt that making project is overkill for now, so I extracted a method to generate a project for unit test.

johnscancella:
Also, you might want to take a look at https://docs.gradle.org/4.1/userguide/test_kit.html
it seems like if you create a task to write the classpath, you can then import that during a test and not have to deal with installing the plugin locally. It does make harder to then run the tests in an IDE, but not impossible.

gradlePlugin/src/test/java/com/github/spotbugs/SpotBugsPluginTest.java
```diff
@@ -21,65 +23,81 @@
...
-        "version = 1.0\n" +
-        "repositories {\n" + 
-        "  mavenCentral()\n" + 
-        "  mavenLocal()\n" + 
-        "}";
-    File root = folder.newFolder();
+  /**
+   * The root directory of generated project by {@link #createProject()}
+   */
+  private File root;
```
johnscancella: I would use http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html here instead, since it will automatically clean the folder after each test, thus ensuring no contamination between tests.

KengoTODA: Nice review! I misunderstood that `TemporaryFolder#getRoot()` returns temporary directory like `/tmp`, but it actually returns `/tmp/junitfoobar`. I will follow your suggestion.

KengoTODA:
@johnscancella Thanks for your nice idea!

It seems that we still need additional hack to use this method, because [SpotBugs gradle plugin depends on SpotBugs](https://github.com/spotbugs/spotbugs/blob/b559146/gradlePlugin/src/main/java/com/github/spotbugs/SpotBugsPlugin.java#L99) so what we need is not only class files but also artifact metadata (maybe pom.xml).

Then let's merge current version with `./gradlew spotbugs:install -x test` for now, we can enhance it later.

KengoTODA:
TO SpotBugs members,
I've resolved conflict, please review and merge this PR. Thanks in advance!

KengoTODA:
Thanks @iloveeclipse @johnscancella !

PR #343, opened 2017-08-23
[docs] Generate bug description page not by gradle but by sphinx extension

KengoTODA:
Previously we needed to commit generated files to Git, because readthedocs does not support executing Gradle script before we build HTML pages. By this change, we can stop commiting them to Git.

It also makes possible to translate not only short/long descriptions but also detailed description.

For reviewer: Here is [staging site](http://spotbugs-in-kengo-toda.readthedocs.io/ja/translate-bug-description-page/) to check these changes.

cc @orihalcon128

PR #345, opened 2017-08-25
stop listing deprecated bugs in docs

KengoTODA:
This change will remove deprecated bugs such as `TLW_TWO_LOCK_NOTIFY` from bug descriptor list.

staging site: http://spotbugs-in-kengo-toda.readthedocs.io/en/remove-deprecated-bug-descriptions-from-docs/bugDescriptions.html

refs #343, cc: @orihalcon128

### DIFF
--- a/docs/extensions/generate_bug_description.py
+++ b/docs/extensions/generate_bug_description.py
@@ -34,6 +34,8 @@ def generate_bug_description(language):
             bug_description_page.write('\n\n')
 
             for bug_pattern in findbugs.findall(".//BugPattern[@category='%s']" % category):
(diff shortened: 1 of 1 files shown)


iloveeclipse:
I think I don't like this. All bug warnings we ship should have descriptions in the manual. Otherwise users will ask about them.

May be better to move deprecated stuff to the end and extra section.

orihalcon128:
Are bug patterns which is deprecated detected as bugs?

orihalcon128:
@KengoTODA I have thought various ways to do it.
How about marking the deprecation on the left or right side of the bug pattern name?
I think that it can be realized more easily than moving to extra section.

PR #350, opened 2017-08-26
Improve wording

Bananeweizen:
"are changed" seems grammatically wrong. Also the artificial line break
should not exist in the middle of the message.


iloveeclipse:
Thanks.

PR #353, opened 2017-08-26
Use new style filter tree

Bananeweizen:
Since Eclipse 3.5 filter text controls can embed the "clear" button
directly into the text control instead of showing a separate button.
This also removes a deprecation warning.


iloveeclipse:
I will try to review this and two other Ecliose PRs once I will have access to my PC (right now I'm a bit disabled). Thanks for contributing!

Bananeweizen:
Take your time. I know you are spending lots of time in many Eclipse related projects. :)

iloveeclipse:
Thanks!

PR #359, opened 2017-09-03
Minimum Eclipse SDK version for SpotBugs plugin is 4.6

### LINKED ISSUE #354: Minimum Eclipse version confusion
The Eclipse plugin is a bit confusing regarding which versions of Eclipse are intended to be supported. The build instructions say to build against an Eclipse 4.6 target platform. The manifest.template however declares org.eclipse.ui 3.6 as the only dependency with a version number.

If people really build against a 4.6 target platform, they will start using API that is not available in older versions. That in turn will lead to the Eclipse plugin binary being runtime incompatible with older Eclipse versions, although its manifest allows installation with Eclipse 3.6 or better.

Therefore both the instructions as well as the manifest.template should refer to the same version. And that version number should be declared by some committer, probably @iloveeclipse.

iloveeclipse:
Set restriction on lower version bound for org.eclipse.core.runtime to 3.12.0 which corresponds to the bundle version from Eclipse 4.6.0 SDK.

This fixes #354.

### DIFF
--- a/eclipsePlugin/META-INF/MANIFEST-TEMPLATE.MF
+++ b/eclipsePlugin/META-INF/MANIFEST-TEMPLATE.MF
@@ -4,9 +4,9 @@ Bundle-Name: SpotBugs
 Bundle-Activator: de.tobject.findbugs.FindbugsPlugin
 Bundle-Vendor: SpotBugs Project
 Require-Bundle: org.eclipse.ui.workbench.texteditor,
- org.eclipse.ui;bundle-version="3.6.0",
+ org.eclipse.ui,
  org.eclipse.ant.core,
- org.eclipse.core.runtime,
+ org.eclipse.core.runtime;bundle-version="3.12.0",
  org.eclipse.core.resources,
  org.eclipse.ui.ide,
  org.eclipse.ui.editors,
--- a/eclipsePlugin/plugin_feature-candidate.xml
+++ b/eclipsePlugin/plugin_feature-candidate.xml
@@ -26,7 +26,7 @@ http://www.gnu.org/licenses/lgpl.html
 
    <requires>
       <import plugin="org.eclipse.core.resources"/>
-      <import plugin="org.eclipse.core.runtime" version="3.6.0" match="greaterOrEqual"/>
+      <import plugin="org.eclipse.core.runtime" version="3.12.0" match="greaterOrEqual"/>
       <import plugin="org.eclipse.jdt"/>
       <import plugin="org.eclipse.jdt.core"/>
       <import plugin="org.eclipse.jdt.launching"/>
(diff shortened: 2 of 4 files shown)


Bananeweizen:
👍

sewe:
I honestly don’t understand why we want to **artificially** restrict the number of Eclipse versions  with which SpotBugs is compatible. Our are we really using an `org.eclipse.core.runtime` feature that’s only available since 3.12.0?

I understand that manually checking compatibility with older releases is cumbersome, but that’s what build servers are for. For Eclipse Code Recommenders, for example, we regularly [build against different Eclipse/JDK combinations](https://hudson.eclipse.org/recommenders/job/org.eclipse.recommenders-master/). So even if you locally build only against the latest Eclipse, your build server will complain if you accidentally use an API that isn’t available in all target versions.

Taking a step further back, I think it is a mistake to simply use the running Eclipse as the “target platform“ to build and test against. That’s what PDE’s `.target` files are for. That way, you can use the latest Eclipse Oxygen to develop the plugin with, but still test and debug against, say, Eclipse Mars.

iloveeclipse:
The point is: there is no reason to support older platform for us. We are very limited in resources, so better to say "we don't support Eclipse 4.5" and disallow installation on it as to allow to install the plugin and get a bug report about some half working stuff. And I personally don't see any value to support older Eclipse platforms as 4.6, given the fact that the current one is 4.7 and almost all 4.x releases before 4.6 were unusable anyway (IMHO).

Regarding the target etc: I don't know what you want to discuss here - sure, no one should just use "some current Eclipse" for testing or validation. Current difficulty we have is that we obviously chosen Gradle but don't have people/knowledge how to setup Eclipse platform and tests with Gradle / Travis. Feel free to provide a patch :-).

Bananeweizen:
@sewe I would also prefer to have one or more .target files based on the http URL of the Eclipse update site repositories, similar to what many eclipse.org projects are using, e.g. https://github.com/eclipse/egit/tree/master/org.eclipse.egit.target. However, I'm not sure if that contradicts with the gradle based generation of some of the settings files. (I guess not)

If this is wanted, I can try to create an actual 4.6 target file, then we can merge that together with this change, and afterwards we can try to create additional 4.5, 4.4. etc. targets and then lower the above mentioned dependency. But right at the moment, that high dependency version reflects what get's tested regularly (even though still just manually).

Since I'm not a core team member, it is generally your choice which versions of Eclipse to support. But I somewhat with @iloveeclipse here. I have the impression you want to bring Spotbugs to speed again (compared to Findbugs). Reducing your technical debt by restricting your compatibility range seems a reasonable choice (at least at this still early stage of the Spotbugs project).

KengoTODA:
I have motivation to merge this, how about you @sewe ?

In my understanding, this change is similar with dropping Java7 support.

iloveeclipse:
Ping. Can we please merge this?

iloveeclipse:
@sewe : any feedback?

iloveeclipse:
Anyone: please, can we merge this? No one will ever fix any bugs on Eclipse < **4.7**, and the bounds we are talking here are for **4.6**!

Issue #329, opened 2017-08-18
Duplicate resources in *-sources.jar


sewe:
Found *another* problem with the Gradle build: The `*-sources.jar` generates contains duplicate ZIP entries for `.png`, `.db`, and other resource files. These makes the files larger than necessary.

    > jar tf spotbugs/build/libs/spotbugs-3.1.0-SNAPSHOT-sources.jar | sort | uniq -c | awk '$1 > 1'
    2 edu/umd/cs/findbugs/FindBugsAnnotationDescriptions.properties
    2 edu/umd/cs/findbugs/FindBugsAnnotationDescriptions_ja.properties
    2 edu/umd/cs/findbugs/UserDesignations.properties
    2 edu/umd/cs/findbugs/UserDesignations_ja.properties
    2 edu/umd/cs/findbugs/ba/npe/jdkBaseNonnullReturn.db
    2 edu/umd/cs/findbugs/ba/npe/jdkBaseUnconditionalDeref.db
    2 edu/umd/cs/findbugs/detect/longInstant.db
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_de.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_en.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_es.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_et.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_fi.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_fr.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_ja.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_pl.properties
    2 edu/umd/cs/findbugs/gui/bundle/findbugs_zh_CN.properties
    2 edu/umd/cs/findbugs/gui/help/About.html
    2 edu/umd/cs/findbugs/gui/help/Acknowledgements.html
    2 edu/umd/cs/findbugs/gui/help/License.html
    2 edu/umd/cs/findbugs/gui/help/spotbugs.png
    2 edu/umd/cs/findbugs/gui/help/surelogic.png
    2 edu/umd/cs/findbugs/gui2/SplashBug1.png
    2 edu/umd/cs/findbugs/gui2/SplashBug1reverse.png
    2 edu/umd/cs/findbugs/gui2/SplashBug2B.png
[...]

PR #362, opened 2017-09-03
Stop including resources to main source

KengoTODA:
This change eliminates duplicated resources in *-sources.jar, which is reported at #329.
I confirmed that this change reduces size of sources .jar file from 2,746,646 bytes to 2,545,922 bytes.

```bash
$ jar tf spotbugs/build/libs/spotbugs-3.1.0-SNAPSHOT-sources.jar | sort | uniq -c | awk '$1 > 1'
// no output
```

### DIFF
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -5,6 +5,10 @@ sourceSets {
   main {
     java {
       srcDirs = ['src/main/java', 'src/gui']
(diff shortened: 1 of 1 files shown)


KengoTODA:
@sewe could you review and merge this? RC6 will be shipped by #370 soon.

PR #363, opened 2017-09-04
GitHub 82

ThrawnCA:
Cleaning up issues that would cause Spotbugs to fail Findbugs checks.


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AppendingToAnObjectOutputStream.java
```diff
@@ -81,16 +81,14 @@ public void sawOpcode(int seen) {
             OpcodeStack.Item item = stack.getStackItem(0);
             Object value = item.getConstant();
             sawOpenInAppendMode = value instanceof Integer && ((Integer) value).intValue() == 1;
-        } else if (!sawOpenInAppendMode) {
```
iloveeclipse: Why this?

iloveeclipse: I see it now, the if check at line 70 and return at 77.

PR #370, opened 2017-09-11
Release 3.1.0 rc6

KengoTODA:
It seems that we need more time to use BCEL 6.1, so let's release RC6 without it.

After we merge this PR, we tag 2565477 or rebased commit then Travis CI will deploy artifacts to Sonatype Nexus, Gradle plugin repository and Eclipse Update Sites. This is the first version to use this deploy automation, I will recover if it fails.


KengoTODA:
We have many proposed PR, I will close this until they are successfully merged.

PR #384, opened 2017-09-15
Update manifest before generate Eclipse projects

KengoTODA:
We define this dependency in eclipsePlugin/build.gradle,
but it is possible to ignore this if developer has no local.properties.

So we need to have this dependency even in spotbugs/build.gradle
and spotbugs-annotation/build.gradle.

### DIFF
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -19,6 +19,7 @@ task updateManifest {
     // write manifests
     manifestSpec.writeTo("$projectDir/META-INF/MANIFEST.MF")
 }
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Sorry, "we define this dependency" - *which* dependency? I honestly don't get the point.

KengoTODA:
@iloveeclipse I meant the task dependency from `eclipse` task to `updateManfest` task. We have it at [eclipsePlugin/build.gradle](https://github.com/spotbugs/spotbugs/blob/3.1.0_RC5/eclipsePlugin/build.gradle#L106).

Issue #389, opened 2017-09-21
Errors analyzing I2F and I2D


iloveeclipse:
While testing new BCEL version I've stumbled on errors we get while analyzing code with  I2F and I2D instructions. The errors are not new and aren't caused by new BCEL version, I've just never run SpotBugs over the related code yet.

```
eclipse.buildId=4.8.0.I20170911-2000
java.version=1.8.0_131
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data file:/data/4x_platform_workspace/ -os linux -ws gtk -arch x86_64

com.github.spotbugs.plugin.eclipse
Error
Thu Sep 21 09:17:34 CEST 2017
Error processing opcode i2f @ 96 in org.eclipse.jdt.debug.tests.eval.CharAssignmentOperatorsTests.testCharPlusAssignmentFloat : ()V

java.lang.ClassCastException: java.lang.Character cannot be cast to java.lang.Number
	at edu.umd.cs.findbugs.OpcodeStack.constantToFloat(OpcodeStack.java:2222)
	at edu.umd.cs.findbugs.OpcodeStack.sawOpcode(OpcodeStack.java:2044)
	at edu.umd.cs.findbugs.bcel.OpcodeStackDetector.afterOpcode(OpcodeStackDetector.java:83)
	at edu.umd.cs.findbugs.visitclass.DismantleBytecode.visit(DismantleBytecode.java:884)
	at edu.umd.cs.findbugs.detect.FieldItemSummary.visit(FieldItemSummary.java:130)
	at edu.umd.cs.findbugs.visitclass.BetterVisitor.visitCode(BetterVisitor.java:218)
	at edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitCode(PreorderVisitor.java:243)
	at edu.umd.cs.findbugs.bcel.OpcodeStackDetector.visitCode(OpcodeStackDetector.java:65)
	at org.apache.bcel.classfile.Code.accept(Code.java:132)
	at edu.umd.cs.findbugs.visitclass.PreorderVisitor.doVisitMethod(PreorderVisitor.java:315)
	at edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitJavaClass(PreorderVisitor.java:403)
	at org.apache.bcel.classfile.JavaClass.accept(JavaClass.java:213)
[...]

PR #392, opened 2017-09-21
Fix for i2f, i2d and i2l instructions if the lhs is a character

iloveeclipse:
Fix for i2f, i2d and i2l instructions if the lhs is a character. Fixes issue #389.


iloveeclipse:
Thanks Dave!

PR #395, opened 2017-09-23
Support JUnit5

KengoTODA:
JUnit5 is released recently. It does not support `@Rule` natively, so it is nice to provide [JUnit5 extension](http://junit.org/junit5/docs/current/user-guide/#extensions) instead.

To provide isolated module for each of JUnit 4 and JUnit 5, I create new module `test-harness-core` to share core logic.

Here is [example of usage](https://github.com/KengoTODA/findbugs-slf4j/blob/f84a7992a7fa4726e36047f933db2fa0331a5f52/bug-pattern/src/test/java/jp/skypencil/findbugs/slf4j/GettingClassFromArrayTest.java). Currently this module provide no matcher support, because JUnit5 itself does not depend on hamcrest.


henrik242:
Please rebase against the latest master so we can merge

Issue #373, opened 2017-09-12
Consume final ASM 6


iloveeclipse:
Bad news: final ASM 6 will break API compared with ASM 6 BETA.
Good news: the change is small and they plan to release by the end of September.
So we should consider to switch to the final ASM 6 for the 3.1 release.

See https://mail.ow2.org/wws/arc/asm/2017-09/msg00000.html

pdillinger:
The SpotBugs 3.1.0 description says "First SpotBugs release. Same features as FindBugs, plus proper support for Java 8."

Doesn't ASM 5.x have proper support for Java 8? Doesn't ASM 6 support Java 9? Do you mean Java 9? Is there a plan for Java 9 support?

KengoTODA:
> Doesn't ASM 5.x have proper support for Java 8?

ASM 5 supports Java 8. ASM 6 was [introduced at FindBugs 3.1.0 preview version](https://github.com/spotbugs/spotbugs/commit/16657d9ad740ce3ef054c164d5471e644ebfd655), to support Java 9.

> Doesn't ASM 6 support Java 9?

ASM 6 beta support Java 9. You can refer [official release note](https://mail.ow2.org/wws/arc/asm/2017-08/msg00000.html).

> Do you mean Java 9? Is there a plan for Java 9 support?

We have plan to support Java 9, but it has no progress for now. Java 9 is not supported by ASM and other bytecode manupilation tools.

So Java 9 will not be supported by SpotBugs 3.1.0, I guess.

iloveeclipse:
> Java 9 is not supported by ASM and other bytecode manupilation tools.

??? ASM 6 **has** support for Java 9.

> So Java 9 will not be supported by SpotBugs 3.1.0, I guess.

I guess we will support it (we had it in FindBugs already, see https://github.com/findbugsproject/findbugs/issues?q=java+9+label%3Ajava9), we just need to validate if the new Java 9 builds broke something again.

KengoTODA:
> ??? ASM 6 has support for Java 9.

Then it's OK. [Their release note](https://mail.ow2.org/wws/arc/asm/2017-08/msg00000.html) says that `does not work with javac 9 (yet !)`, so I wasn't so sure.

KengoTODA:
> we had it in FindBugs already

But Jigsaw broke FindBugs/SpotBugs so current `master` branch cannot support Java 9 for now :sob:

iloveeclipse:
[...]

PR #397, opened 2017-09-24
Upgrade ASM to 6.0

KengoTODA:
This PR tries to upgrade ASM from 6.0 beta to to 6.0.
Currently Maven Central Repository has no 6.0 binaries, so we need to add entry to `repository` in `build.gradle`. Before we merge this, it is better to wait release on Maven Central Repository and revert the commit 6670219.

refs #373

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,9 @@ subprojects {
 
   repositories {
     jcenter()
+    maven {
+      url 'https://repository.ow2.org/nexus/content/repositories/releases/'
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


KengoTODA:
It should have no problem for CLI/Ant/Eclipse users.
But for Maven/Gradle users, they need some hack to use repository.ow2.org in their build script. I think it's not so good.

KengoTODA:
I changed my mind, it is OK to ask users to hack their build script, it will be solved eventually.

PR #398, opened 2017-09-24
Reserve Jigsaw module name for future release

KengoTODA:
According to [Gradle guide](https://guides.gradle.org/building-java-9-modules/#optional_add_code_automatic_module_name_code_manifest_entries_for_all_other_projects) and [Stephen Colebourne's blog](http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html), it is recommended to put module name as `Automatic-Module-Name` into `MANIFEST.MF`.

### DIFF
--- a/test-harness/build.gradle
+++ b/test-harness/build.gradle
@@ -37,3 +37,8 @@ uploadArchives {
     }
   }
 }
+
+// Module name should be reverse-DNS (com.github.spotbugs) just like package name
+// http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html
+ext.moduleName = 'com.github.spotbugs.test'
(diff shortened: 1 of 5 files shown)


test-harness/build.gradle
```diff
@@ -37,3 +37,8 @@ uploadArchives {
     }
   }
 }
+
+// Module name should be reverse-DNS (com.github.spotbugs) just like package name
+// http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html
+ext.moduleName = 'com.github.spotbugs.test'
```
iloveeclipse: Schouldn't this be .test.harness?

KengoTODA: I think .test is OK, we use `edu.umd.cs.findbugs.test` as package name.

PR #414, opened 2017-09-28
Avoid ArithmeticExceptions while interpreting ldiv/lrem values

### LINKED ISSUE #413: Errors interpreting ldiv values
Revealed by logging added via #392. The code below produces analysis error `Exception processing 'pushByLongMath' with opcode 109, lhs < J, 1 > and rhs < J, 0, cbz >` :
```
package ghIssues;
public class Issue413 {
	public long error_in_ldiv() {
		return 1L / (1L / 2);
	}
}
```

iloveeclipse:
This fixes #413.

Note: this is not a new issue, the commit bf2470ea8 just added more error logging to find out the reason for a try/catch(RuntimeException) block. This PR fixes two cases where the ArithmeticExceptions were thrown and catched immediately by the old code. We don't need to throw ArithmeticException if we know that 1/0 or 1%0 will never fly with long values.


iloveeclipse:
Yea. Thanks Dave.

KengoTODA:
Please update changelog later @iloveeclipse

PR #416, opened 2017-09-28
stop replacing filename on-demand

KengoTODA:
Currently we're replacing filename on-demand, and it will introduce trouble when our version becomes the same with some of dependencies. e.g. when we release spotbugs version 6.0, renaming method will rename not only spotbugs relatives but also ASM 6.0.

To stop replacing filename, we can use `archiveName` attribute to remove version from jar file name.


KengoTODA:
pushed unrelated commits by accident, I will recover.

Issue #423, opened 2017-09-30
gradle failed with java.lang.IllegalArgumentException: path may not be null or empty string. path='null'


LintaoLin:
### stacktrace

````
java.lang.IllegalArgumentException: path may not be null or empty string. path='null'
        at org.gradle.api.internal.file.BaseDirFileResolver.doResolve(BaseDirFileResolver.java:72)
        at org.gradle.api.internal.file.AbstractFileResolver.resolve(AbstractFileResolver.java:79)
        at org.gradle.api.internal.file.AbstractFileResolver.resolve(AbstractFileResolver.java:61)
        at org.gradle.api.internal.tasks.CompositeTaskOutputPropertySpec$1.computeNext(CompositeTaskOutputPropertySpec.java:66)
        at org.gradle.api.internal.tasks.CompositeTaskOutputPropertySpec$1.computeNext(CompositeTaskOutputPropertySpec.java:56)
        at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:143)
        at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:138)
        at com.google.common.collect.Iterators$5.hasNext(Iterators.java:542)
        at org.gradle.api.internal.tasks.TaskPropertyUtils.collectFileProperties(TaskPropertyUtils.java:33)
        at org.gradle.api.internal.tasks.DefaultTaskOutputs.getFileProperties(DefaultTaskOutputs.java:151)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository.getOutputPropertyNamesForCacheKey(CacheBackedTaskHistoryRepository.java:148)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository.getHistory(CacheBackedTaskHistoryRepository.java:69)
        at org.gradle.api.internal.changedetection.changes.DefaultTaskArtifactStateRepository.getStateFor(DefaultTaskArtifactStateRepository.java:63)
        at org.gradle.api.internal.changedetection.changes.ShortCircuitTaskArtifactStateRepository.getStateFor(ShortCircuitTaskArtifactStateRepository.java:47)
        at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:43)
[...]

PR #428, opened 2017-10-02
[doc] Add document to help Android developers

KengoTODA:
This document helps Android developer to avoid problem like #423.

@johnscancella @jsotuyod Do you know in which timing we can validate Gradle Plugin configuration? If we can throw error with reasonable message (e.g. please set `destination` for SpotBugs report), it should be better solution.


johnscancella:
I would recommend putting it in the SpotBugsSpecBuilder, somewhere around here:
https://github.com/spotbugs/spotbugs/blob/master/gradlePlugin/src/main/java/com/github/spotbugs/internal/spotbugs/SpotBugsSpecBuilder.java#L166

KengoTODA:
Wmm, it seems that Gradle throws exception before it reaches to `SpotBugsSpecBuilder`.

It seems that Gradle plugin verify config in `project.afterEvaluate`, I will have a try.

https://android.googlesource.com/platform/tools/base/+/android-7.0.0_r34/build-system/gradle/src/main/groovy/com/android/build/gradle/BasePlugin.java#574

johnscancella:
Looks good to me. I would only suggest editing the phrasing a little bit. So instead of 
> Task '%s' has no destination for TEXT report. Set reports.text.destination to this task.

I would change it to something like:
> Task '%s' has no destination for TEXT report. Please set the reports.text.destination for this task or in the spotbugs configuration closure.

KengoTODA:
@johnscancella 

> spotbugs configuration closure

Do you mean `spotbugs { ... }`? In my understanding it does not support report configuration.

http://gradle.monochromeroad.com/docs/dsl/org.gradle.api.plugins.quality.FindBugsExtension.html

johnscancella:
Sorry I can't look at the link right now, for some reason my employer is blocking it.

Since it a direct copy from Gradle's findbugs implementation it should be the same as their example https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.FindBugsExtension.html#org.gradle.api.plugins.quality.FindBugsExtension:reportsDir

so you should be able to do 
```
spotbugs{
  reportsDir = file("$rootProject.projectDir/build/reports/MY_REPORT_DIR")
}
```

KengoTODA:
@johnscancella Yes extension (closure) supports setting `reportsDir`, not name of report file. To configure report file name, user needs to set property of Task.

Note that basically user does not need to configure report name, because SpotBugs Gradle Plugin uses name of sourceSet by default (e.g. `main.xml` for `main` sourceSet). In the reported case, user @LintaoLin had to set manually, because his project had no sourceSet. This is reproducible case for Android hackers so I wanna suggest us to enhance error message.

johnscancella:
@KengoTODA ok, that makes sense. But looking at the code that is in the pull request it all refers to the report destination, not report name.  is `reports.text.destination` used for the report name as well?

KengoTODA:
@johnscancella Yes for our plugin. Refer [SpotBugsSpecBuilder](https://github.com/spotbugs/spotbugs/blob/3.1.0_RC6/gradlePlugin/src/main/java/com/github/spotbugs/internal/spotbugs/SpotBugsSpecBuilder.java#L165-L166) and [Gradle API Javadoc](http://gradle.monochromeroad.com/docs/javadoc/org/gradle/api/reporting/Report.html#getDestination--) for detail.

KengoTODA:
@members please review and approve this change. Thanks in advance!

PR #431, opened 2017-10-04
Add anchor with name of bug patterns

### LINKED ISSUE #430: Bug descriptions
The URL for bug descriptions does not support anchor with bug id.

https://github.com/spotbugs/spotbugs/blob/bfeb57de30ab97746b7195f1e693883d261df5a4/spotbugs/etc/messages.xml#L18-L21

For example :
https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#FI_EMPTY

KengoTODA:
staging site: http://spotbugs-in-kengo-toda.readthedocs.io/en/issue-430/bugDescriptions.html#FI_EMPTY

This PR closes #430. This change will add new anchor, so existing anchor will not deleted.

### DIFF
--- a/docs/extensions/generate_bug_description.py
+++ b/docs/extensions/generate_bug_description.py
@@ -37,7 +37,7 @@ def generate_bug_description(language):
                 if (bug_pattern.get('deprecated') == 'true'):
                     continue
                 type = bug_pattern.get('type')
-                bug_description_page.write(".. _%s:\n\n" % type)
+                bug_description_page.write(".. raw:: html\n\n  <a id='%s'></a>\n\n" % type)
                 message = messages.find(".//BugPattern[@type='%s']" % type)
                 pattern_title = generate_pattern_title(bug_pattern, message)
                 bug_description_page.write("%s\n%s\n\n" % (pattern_title, '^' * len(pattern_title)))
--- a/docs/extensions/generate_bug_description.py
+++ b/docs/extensions/generate_bug_description.py
@@ -20,7 +20,7 @@ def generate_bug_description(language):
     else:
         messages = parse('../spotbugs/etc/messages.xml')
 
-    with codecs.open('generated/bugDescriptionList.rst', 'w', encoding='UTF-8') as bug_description_page:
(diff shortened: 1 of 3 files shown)


docs/extensions/generate_bug_description.py
```diff
@@ -37,6 +37,7 @@ def generate_bug_description(language):
                 if (bug_pattern.get('deprecated') == 'true'):
                     continue
                 type = bug_pattern.get('type')
+                bug_description_page.write(".. raw:: html\n\n  <a id='%s'></a>\n\n" % type)
```
iloveeclipse: The anchor is added to the **previous** bug pattern:
```
<div class="section" id="np-equals-method-does-not-check-for-null-argument-np-equals-should-handle-null-argument">
<h3>NP: equals() method does not check for null argument (NP_EQUALS_SHOULD_HANDLE_NULL_ARGUMENT)<a class="headerlink" href="#np-equals-method-does-not-check-for-null-argument-np-equals-should-handle-null-argument" title="Permalink to this headline">¶</a></h3>
<p>
This implementation of equals(Object) violates the contract defined
by java.lang.Object.equals() because it does not check for null
being passed as the argument.  All equals() methods should return
false if passed a null value.
</p><a id='FI_EMPTY'></a></div>
<div class="section" id="fi-empty-finalizer-should-be-deleted-fi-empty">
<h3>FI: Empty finalizer should be deleted (FI_EMPTY)<a class="headerlink" href="#fi-empty-finalizer-should-be-deleted-fi-empty" title="Permalink to this headline">¶</a></h3>
<p> Empty <code>finalize()</code> methods are useless, so they should
be deleted.</p><a id='FI_NULLIFY_SUPER'></a></div>
```

KengoTODA: This is necessary to display bug descriptor name (`<h3>`) in the first view. If we put anchor like `<a id='FI_EMPTY'>` after `<div class="section">` or `<h3>`, first view will not display descriptor name.

Ideally we should put this anchor between `<div class="section">` and `<h3>`, but it is not supported by Sphinx.

iloveeclipse: OK, thanks.

PR #434, opened 2017-10-06
Parse package-info from aux classpath

### LINKED ISSUE #429: CheckReturnValue annotation applied to package is not enforced
The `RV_RETURN_VALUE_IGNORED_BAD_PRACTICE` rule detects whether methods that are marked `@CheckReturnValue` have their return value ignored, and warns if so.

This should catch a class of bugs in fluent assertion libraries like "[AssertJ](http://joel-costigliola.github.io/assertj/)" and "[Google Truth](https://github.com/google/truth)", i.e.

```
assertThat(a == b); // BUG: this is a no-op

assertThat(a).isEqualto(b); // correct assertion
```

See the discussion in the AssertJ library about this issue: http://joel-costigliola.github.io/assertj/assertj-core.html#faq-incorrect-api-usage and https://github.com/joel-costigliola/assertj-core/pull/695

The AssertJ library has applied `@CheckReturnValue` to each of their methods, but the "Google Truth" library has applied `@CheckReturnValue` to their entire package: https://github.com/google/truth/blob/9cfceb957c0e42b76ca598c07900414081a81144/core/src/main/java/com/google/common/truth/package-info.java#L28

The former works with Findbugs/Spotbugs, but the latter is not detected. I think it should be, as the `javax.annotation.CheckReturnValue` annotation is allowed at the package level, so the designers clearly intended this kind of usage.

## Repro case:

Code like the following should warn with `RV_RETURN_VALUE_IGNORED_BAD_PRACTICE`, but do not:

```
import org.junit.*;
import static com.google.common.truth.Truth.assertThat;

public class MyTest {
  @Test
  public void testExample() {
    assertThat(System.currentTimeMillis() > 0); // BUG: this is a no-op
  }
}
```

KengoTODA:
This change is enough to close #429. AFAIK there are still known problems (e.g. annotation on class in aux classpath does not work), I will make separated issue for them. Maybe we need refactoring around `Database` classes.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/CheckReturnAnnotationDatabase.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/CheckReturnAnnotationDatabase.java
@@ -33,6 +33,7 @@ import org.apache.bcel.Const;
 import org.apache.bcel.Repository;
 import org.apache.bcel.classfile.JavaClass;
 
+import edu.umd.cs.findbugs.FindBugs2;
 import edu.umd.cs.findbugs.annotations.CheckReturnValue;
 import edu.umd.cs.findbugs.annotations.Confidence;
 import edu.umd.cs.findbugs.classfile.CheckedAnalysisException;
@@ -310,7 +311,9 @@ public class CheckReturnAnnotationDatabase extends AnnotationDatabase<CheckRetur
             return null;
         } catch (CheckedAnalysisException e) {
             // ignore unexpected error to keep backward compatibility
-            e.printStackTrace();
+            if (FindBugs2.DEBUG) {
+                e.printStackTrace();
+            }
             return null;
         }
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/CheckReturnAnnotationDatabase.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/CheckReturnAnnotationDatabase.java
@@ -19,14 +19,29 @@
 
 package edu.umd.cs.findbugs.ba;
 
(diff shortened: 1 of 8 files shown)


RichardBradley:
Thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/CheckReturnAnnotationDatabase.java
```diff
@@ -258,7 +273,68 @@ public CheckReturnValueAnnotation getResolvedAnnotation(Object o, boolean getMin
...
+        ClassDescriptor descriptor = DescriptorFactory.createClassDescriptor(className);
+        XClass clazz;
+        try {
+            clazz = descriptor.getXClass();
+        } catch (ResourceNotFoundException e) {
+            // no annotation on package
+            return null;
+        } catch (CheckedAnalysisException e) {
+            // ignore unexpected error to keep backward compatibility
+            e.printStackTrace();
```
iloveeclipse: Can you please either don't print to std err at all or report analysis error? I think we should just join both catch blocks together.

KengoTODA: I don't agree, it will make debugging difficult and user cannot understand behavior in corner cases. 

How about printing stderr only when `FindBugs2.DEBUG` is set?

iloveeclipse: OK.

KengoTODA:
@iloveeclipse updated, thanks!

PR #435, opened 2017-10-06
Add windows builds through appveyor

stephengroat:
example: https://ci.appveyor.com/project/stephengroat/spotbugs

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,13 @@
+environment:
+  matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.7.0
+    - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+install:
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Sorry for asking stupid questions, what is the purpose of this file? Would it add another automated build to every pull request, this time on windows? Why do we need this? To confirm of we are windows compatible?

stephengroat:
yup, you can build on windows. it currently fails on `jdk1.7`, that's why i was including it

iloveeclipse:
Yea, does it mean, *every* pull request will also have to wait for the confirmation from appveyor? BTW Java 7 is not supported for a long time anyway.

stephengroat:
ok, i removed java7.

you can make appveyor blocking or not, at a minimum it can report status back on windows builds

KengoTODA:
currently we have no automatic test for Windows specific solutions (e.g. .bat files), then appveyor isn't so valuable I think.

PR #444, opened 2017-10-13
Clean up source codes

KengoTODA:
Eclipse cleaned up source codes automatically:

* add missing Override annotations
* organize imports
* delete trailing spaces


jolkdarr:
Hi
Why not add the following option (if available)?
"Use lambda where possible"

iloveeclipse:
> Why not add the following option (if available)?
> "Use lambda where possible"

Why not to push a commit to this branch :-) ?

Please note that test cases could be affected by this, so please do not apply it on bug examples.

jolkdarr:
OK => new separate pull request: #445

PR #446, opened 2017-10-13
Release 3.1.0-RC7

iloveeclipse:
See #111 for details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
   id 'org.sonarqube' version '2.5'
 }
 
-version = '3.1.0-SNAPSHOT'
+version = '3.1.0-RC7'
 
 apply from: "$rootDir/gradle/java.gradle"
 apply from: "$rootDir/gradle/jacoco.gradle"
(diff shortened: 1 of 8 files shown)


iloveeclipse:
@KengoTODA : I've pushed the changes, please check.

build.gradle
```diff
@@ -2,7 +2,7 @@ plugins {
   id 'org.sonarqube' version '2.5'
 }
 
-version = '3.1.0-RC7'
+version = '4.0.0-SNAPSHOT'
```
KengoTODA: use '3.1.0-SNAPSHOT' instead. we should use '4.0.0-SNAPSHOT' after '3.1.0' release.

iloveeclipse: OK, but why this causes the build failure?

iloveeclipse:
OK, the build passes now. But what a surprise that 4.0.0 is not allowed, end the gradle error was not really helpful for fixing the problem!

iloveeclipse:
@KengoTODA : is it OK now?

PR #448, opened 2017-10-14
Enhance test and javadoc

KengoTODA:
Add more unit test to enhance test coverage.
This PR also added `@deprecated` javadoc tag to util methods which should be replaced with methods in `Objects` class.


henrik242:
Please rebase against the latest master

PR #451, opened 2017-10-15
[docs] partially stop filling versions in .rst file directly

KengoTODA:
Introduce [substitutions](http://www.sphinx-doc.org/en/stable/markup/inline.html#default-substitutions) and [sphinx.ext.extlinks](http://www.sphinx-doc.org/en/stable/ext/extlinks.html) to remove hard-coded versions in .rst files.

Here is staging site: http://spotbugs-in-kengo-toda.readthedocs.io/en/read-the-docs/installing.html

About remaining parts, I'm still investigating preferred methodology. It seems that [we need to implement extension by own](https://groups.google.com/forum/#!topic/sphinx-users/b3yPzQhpqo0).


KengoTODA:
this PR is ready to merge.

PR #452, opened 2017-10-16
upgrade SpotBugs Gradle Plugin to the latest

KengoTODA:
This PR uses the latest SpotBugs Gradle Plugin which depends on SpotBugs 3.1.0-RC7.
I'm not sure when we can merge this PR, though.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [ ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code
- [ ] Added an entry into `gradlePlugin/CHANGELOG.md` if you have changed Gradle plugin code

PR #455, opened 2017-10-16
Update missing RC6 -> RC7 entries

henrik242:
After the RC7 update, some entries seems to be missing.

Ref. https://github.com/spotbugs/discuss/issues/15

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ import os
 html_context = {
   'version' : '3.1',
   'full_version' : '3.1.0-RC7',
-  'maven_plugin_version' : '3.1.0-RC6',
+  'maven_plugin_version' : '3.1.0-RC7',
   'gradle_plugin_version' : '1.5'
 }
 
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -311,7 +311,7 @@ Complete Example
     <FindBugsFilter
 		xmlns="https://github.com/spotbugs/filter/3.0.0"
 		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0_RC6/spotbugs/etc/findbugsfilter.xsd">
+		xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0_RC7/spotbugs/etc/findbugsfilter.xsd">
     <Match>
       <Class name="com.foobar.ClassNotToBeAnalyzed" />
     </Match>
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -86,5 +86,5 @@ apply from: "$rootDir/gradle/jigsaw.gradle"
 spotbugs {
   effort = "max"
   reportLevel = "high"
-  toolVersion '3.1.0-RC6'
+  toolVersion '3.1.0-RC7'
 }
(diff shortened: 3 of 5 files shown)


spotbugs-annotations/build.gradle
```diff
@@ -86,5 +86,5 @@ apply from: "$rootDir/gradle/jigsaw.gradle"
 spotbugs {
   effort = "max"
   reportLevel = "high"
-  toolVersion '3.1.0-RC6'
+  toolVersion '3.1.0-RC7'
```
iloveeclipse: Why do we need those versions here at all? I'm not the gradle expert, but can't we somehow say "use the latest"?

henrik242: We should at least be able to use a global variable (e.g. `ext.spotbugsVersion`) in the root build.gradle

KengoTODA: or delete `toolVersion` and update version of gradle version like https://github.com/spotbugs/spotbugs/pull/452/files

then it will use RC7.

iloveeclipse: I'm +1 on the solution which will remove duplicated version mentions all around, but my Gradle knowledge is too bad to propose or to prefer any one.

wrprice: This is a "proper" subproject, right?  If I understand what you're going for, I think you can do this little bit of foo to always depend on the current build instead of using version numbers:

```
dependencies {
  spotbugs project(':spotbugs')
}

spotbugs {
  toolVersion '3.1.0' // overridden by explicit spotbugs configuration dependency
}
```

henrik242: But regardless, RC6 is obviously wrong, so this patch should go into master, no?

iloveeclipse: I silently hoped that someone would have a silver bullet to get rid of this versions once and forever :-)

iloveeclipse: Looks like @KengoTODA just found the right bullet, see #451.
Kendo: *this* PR is not needed then, right?

henrik242: Cool. Does it take filter.rst as well?

henrik242: There were still five RC6 entries after #451 got merged, I've updated my commit to match that.

docs/conf.py
```diff
@@ -18,7 +18,7 @@
 html_context = {
   'version' : '3.1',
   'full_version' : '3.1.0-RC7',
-  'maven_plugin_version' : '3.1.0-RC6',
+  'maven_plugin_version' : '3.1.0-RC7',
```
KengoTODA: Maven plugin has no RC7 for now.

http://repo.maven.apache.org/maven2/com/github/spotbugs/spotbugs-maven-plugin/

docs/filter.rst
```diff
@@ -311,7 +311,7 @@ Complete Example
     <FindBugsFilter
 		xmlns="https://github.com/spotbugs/filter/3.0.0"
 		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0_RC6/spotbugs/etc/findbugsfilter.xsd">
+		xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0_RC7/spotbugs/etc/findbugsfilter.xsd">
```
iloveeclipse: I think we should fix it to "3.1.0" during the 3.1.0 version release, and let it be fixed in the future.

iloveeclipse:
See PR #461. I think we can close this one.

PR #457, opened 2017-10-18
fix broken travis build

KengoTODA:
Oracle changed download URL for Oracle Java 9, so our build in `master` branch has been broken.
But recently Travis CI supported `oraclejdk9` so we can stop using `apt` to install Java9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,11 @@ language: java
 sudo: false
 dist: trusty
 
+jdk:
+  - oraclejdk8
+  - oraclejdk9
+
 env:
-  matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,7 +32,6 @@ env:
 addons:
   apt:
     packages:
-      - oracle-java9-installer
       - oracle-java8-installer
```
iloveeclipse: do we still need java 8 **installer**?

KengoTODA: YES we need, we compile by Java8 to test on Java9.

PR #447, opened 2017-10-13
[docs] List detectors and don't render hidden elements

lqc:
This is a draft patch for the code that generates the bug list. Changes:

* Bug patterns are sorted inside Categories
* All sections have predictable anchors (not translation sensitive) to use with `:ref:`
* Hidden categories (like Noise) are omitted - not sure if that's ok
* Detectors are listed as an extra category with list of links to bugs it reports
* Detectors which are disabled by default have "(disabled by default)" under their title

I have trouble translating that last one. Is it ok to add some fake entries to "messages.xml" ?

If you think this is usefull, I would also like to add information about Rank for each Bug and speed for each detector.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [X] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code
- [X] Added an entry into `gradlePlugin/CHANGELOG.md` if you have changed Gradle plugin code

PR #465, opened 2017-10-22
Release 3.1.0

iloveeclipse:
Final 3.1.0 release, see #111.

----

- [x] Added an entry into `gradlePlugin/CHANGELOG.md` if you have changed Gradle plugin code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased - 2017-??-??
+
 ## 3.1.0 - 2017-10-25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased - 2017-??-??
+## 3.1.0 - 2017-10-23
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
-## 3.1.0 - 2017-10-23
+## 3.1.0 - 2017-10-25
 
 ### Fixed
 
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
   id 'org.sonarqube' version '2.5'
 }
 
-version = '3.1.0-SNAPSHOT'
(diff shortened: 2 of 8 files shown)


iloveeclipse:
@KengoTODA : similar to last time, the build fails due some version issues. Could you please check this?

KengoTODA:
I hope rebuilding fixes this problem. Not sure why Gradle failed to download SpotBugs from Maven repository.

Please solve conflict, and use version `1.6.0` for gradle-plugin [to follow semver2](https://github.com/spotbugs/discuss/issues/12), then we can merge this. It is also good to wait #447, but it should not be mandatory.

KengoTODA:
No idea for now. What I know is that, SpotBugs Gradle Plugin adds dependency dynamically like below,

https://github.com/spotbugs/spotbugs/blob/3.1.0_RC7/gradlePlugin/src/main/java/com/github/spotbugs/SpotBugsPlugin.java#L149-L156

Not sure this is really related with our build failure in this branch.

iloveeclipse:
To be able to build I had to temporarily disable spotbugs task execution on spotbugs and spotbugs-annotations projects. I have no idea which logic is there but it seems that any NON SNAPSHOT version in the main build file causes gradle dependency resolution on gradlePlugin project going amok and require something which I can't understand. Please see the commit error message what I mean.

I think we can release with this temporary fix, but it would be great if someone with gradle knowledge would look at our build scripts to understand the root cause and may be even to provide a proper fix.

iloveeclipse:
Arrrgh, the build for eclipse plugin must generate NOT com.github.spotbugs.plugin.eclipse_3.1.0.jar but com.github.spotbugs.plugin.eclipse_3.1.0.timestamp.jar, otherwise Eclipse will not update existing nightly builds with the release build!!!

iloveeclipse:
> Arrrgh, the build for eclipse plugin must generate NOT com.github.spotbugs.plugin.eclipse_3.1.0.jar

Fixed with #473 , which must be merged before this one.

iloveeclipse:
The build issue with spotbugs task and weird gradle error message is caused most likely by https://github.com/gradle/gradle/issues/2723

KengoTODA:
If so, we can try to downgrade Gradle to 4.0.

And I think current change has conflict with #473, could you merge (or rebase to) latest `master` branch?

KengoTODA:
I've confirmed that Gradle 4.0 cannot be solution: https://travis-ci.org/spotbugs/spotbugs/builds/292073473

iloveeclipse:
I've rebased it now.

KengoTODA:
I've added several changes to prepare for coming development.

We will tag rebased 94f4e9f3316c6082210856f8536789b1ae5814d9 as `4.0.0`, and create `release-3.1` branch from it. See https://github.com/spotbugs/discuss/issues/11 for detail.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2017-??-??
 
+## 3.1.0 - 2017-10-23
```
iloveeclipse: This date was too optimistic ... Should we change?

KengoTODA: OK I will.

iloveeclipse:
> We will tag rebased 94f4e9f as 4.0.0

You mean probably 3.1.0?

KengoTODA:
> You mean probably 3.1.0?

Oh, yes. Sorry for my mistake.

iloveeclipse:
And probably now it will be 82b2bab

PR #473, opened 2017-10-23
Release versions of Eclipse plugins must be higher as nightly versions

### LINKED ISSUE #472: Release builds of Eclipse plugin should have version higher as nightly builds
Current build generates 3.1.0 version for Eclipse plugin, but this version is semantically smaller as 3.1.0.20171023-1508734123102 we generate for nightly. We should generate either same style or add something like 3.1.0.r20171023-1508734123102

iloveeclipse:
Fixes #472, additionally uses commit hash to identify the build instead
of the timestamp.

This PR is the precondition for #465, otherwise no user with already installed SpotBugs Eclipse plugin will be able update to the released 3.1.0 version.

----

### DIFF
--- a/eclipsePlugin/build.gradle
+++ b/eclipsePlugin/build.gradle
@@ -1,12 +1,26 @@
 // TODO: deploy not only jar but also sources and javadoc, to pass validation by Sonatype nexus
 // apply from: "$rootDir/gradle/maven.gradle"
+plugins {
+  id 'org.ajoberstar.grgit' version '1.7.2'
+}
+
+def readLastCommitHash() {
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Arrgh... looks like CLI git is not on path of travis build? OMG.

iloveeclipse:
I've not noticed yet that our build id missed also hours and minutes, so I've added them too. The reason is to allow multiple builds be produced from the same version with the guarantee that the last one can always be installed over the previous one. Without hours and minutes this was not possible if two builds were generated on same day.

PR #475, opened 2017-10-25
delete all gradle plugin related files

KengoTODA:
Gradle Plugin has been moved to https://github.com/spotbugs/spotbugs-gradle-plugin
So we do not host their files in this repository.

refs https://github.com/spotbugs/discuss/issues/12


KengoTODA:
I think it is better to merge `release-3.1` to `master` after we merge #476, then we can merge other fixes in  `release-3.1`. I'll close this.

PR #505, opened 2017-11-29
Release 3.1.1

KengoTODA:
Recently we have no progress on bugfix, then release current version to get more feedback.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,8 +17,8 @@ import os
 
 html_context = {
   'version' : '3.1',
-  'full_version' : '3.1.0',
-  'maven_plugin_version' : '3.1.0-RC6',
+  'full_version' : '3.1.1',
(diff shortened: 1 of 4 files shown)


KengoTODA:
@spotbugs/core-devs could you review this? thank you in advance!

docs/conf.py
```diff
@@ -17,8 +17,8 @@
 
 html_context = {
   'version' : '3.1',
-  'full_version' : '3.1.0',
-  'maven_plugin_version' : '3.1.0-RC6',
+  'full_version' : '3.1.1',
+  'maven_plugin_version' : '3.1.0-RC7',
```
henrik242: Would be awesome to have a release version on the maven plugin, but no biggie :)

henrik242: This is actually in RC8 now, by the way...

PR #512, opened 2017-12-27
Replaced explicit null-checks by automatically generated

denis-zhdanov:
This *PR* applies [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin to the project. The idea is to achieve the following:  
* enforce *null*-checks [implied by package-level ParametersAreNonnullByDefault annotations](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/package-info.java#L5)
* remove explicit *null*-checks like [this](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/AnalysisContext.java#L176)  

Let's consider [AnalysisContext](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/AnalysisContext.java) as an example:  

*Source code*  

```java
public AnalysisContext(@Nonnull Project project) {
    this.project = project;
    ...
}
```  

Resulting bytecode looks like if it's compiled from the source below:  

```java
public AnalysisContext(@Nonnull Project project) {
    if (project == null) {
        throw new NullPointerException("Argument 'project' of type Project (#0 out of 1, zero-based) is marked by @javax.annotation.Nonnull but got null for it");
    }
    this.project = project;
    ...
}
```  

*Bytecode*  

```
javap -c ./spotbugs/build/classes/java/main/edu/umd/cs/findbugs/ba/AnalysisContext.class
...
public edu.umd.cs.findbugs.ba.AnalysisContext(edu.umd.cs.findbugs.Project);
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: aload_1
       5: ifnonnull     18
       8: new           #2                  // class java/lang/NullPointerException
      11: dup
      12: ldc           #3                  // String Argument 'project' of type Project (#0 out of 1, zero-based) is marked by @javax.annotation.Nonnull but got null for it
      14: invokespecial #4                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      17: athrow
```  

**Note:** added a number of *CheckForNull* annotations in the source code because tests show that *null* is given as method parameters and generated checks fail otherwise. So, the change already made the code better documented.
----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2017-??-??
 
-## 3.1.0 - 2017-10-25
-
 ### Changed
 
 * Replaced explicit *null*-checks by automatically generated ([#512](https://github.com/spotbugs/spotbugs/pull/512))
+
+## 3.1.0 - 2017-10-25
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## 3.1.0 - 2017-10-25
 
+### Changed
+
+* Replaced explicit *null*-checks by automatically generated ([#512](https://github.com/spotbugs/spotbugs/pull/512))
+
 ### Fixed
 
 * Do not try to parse module-info.class ([#408](https://github.com/spotbugs/spotbugs/issues/408))
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/AnalysisCache.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/AnalysisCache.java
@@ -19,8 +19,6 @@
 
 package edu.umd.cs.findbugs.classfile.impl;
 
-import static java.util.Objects.requireNonNull;
-
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -139,7 +137,7 @@ public class AnalysisCache implements IAnalysisCache {
     static final AbnormalAnalysisResult NULL_ANALYSIS_RESULT = new AbnormalAnalysisResult();
 
     @SuppressWarnings("unchecked")
-    static <E> E checkedCast(Class<E> analysisClass, Object o) {
+    static <E> E checkedCast(Class<E> analysisClass, @CheckForNull Object o) {
         if (SystemProperties.ASSERTIONS_ENABLED) {
             return analysisClass.cast(o);
         }
@@ -238,7 +236,6 @@ public class AnalysisCache implements IAnalysisCache {
     @Override
     @SuppressWarnings("unchecked")
     public <E> E getClassAnalysis(Class<E> analysisClass, @Nonnull ClassDescriptor classDescriptor) throws CheckedAnalysisException {
-        requireNonNull(classDescriptor, "classDescriptor is null");
         // Get the descriptor->result map for this analysis class,
         // creating if necessary
         Map<ClassDescriptor, Object> descriptorMap = findOrCreateDescriptorMap(classAnalysisMap,
@@ -313,7 +310,6 @@ public class AnalysisCache implements IAnalysisCache {
 
     @Override
     public <E> E getMethodAnalysis(Class<E> analysisClass, @Nonnull MethodDescriptor methodDescriptor) throws CheckedAnalysisException {
-        requireNonNull(methodDescriptor, "methodDescriptor is null");
         ClassContext classContext = getClassAnalysis(ClassContext.class, methodDescriptor.getClassDescriptor());
         Object object = classContext.getMethodAnalysis(analysisClass, methodDescriptor);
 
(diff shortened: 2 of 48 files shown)


denis-zhdanov:
A CI build ended with exception *Could not determine java version from '9.0.1'*. Looks like something is wrong with the CI setup.

KengoTODA:
We need to upgrade grade to 4.2.1 or later.
https://discuss.gradle.org/t/could-not-determine-java-version-from-9-0-1/24457

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## 3.1.0 - 2017-10-25
 
+### Changed
+
+* Replaced explicit *null*-checks by automatically generated ([#512](https://github.com/spotbugs/spotbugs/pull/512))
+
```
KengoTODA: 3.1.0 has been released, please put your line under 'Unreleased'.

denis-zhdanov: Done

spotbugsTestCases/src/java/bugIdeas/Ideas_2011_08_05.java
```diff
@@ -49,8 +49,6 @@ public int test4OK() {
 
     @NoWarning("RCN")
     public int test5(@Nonnull String x) {
-        if (x == null)
```
iloveeclipse: This is a test idea example, why is this changed? It makes no sense.

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/AnalysisCache.java
```diff
@@ -238,7 +236,6 @@ public void dispose(){
     @Override
     @SuppressWarnings("unchecked")
     public <E> E getClassAnalysis(Class<E> analysisClass, @Nonnull ClassDescriptor classDescriptor) throws CheckedAnalysisException {
-        requireNonNull(classDescriptor, "classDescriptor is null");
```
iloveeclipse: this will create NPE at line 249, where it will be not clear WHAT was null.

Maaartinus: @iloveeclipse AFAIK the NPE gets thrown immediately on method enter by the [Traute plugin](http://traute.oss.harmonysoft.tech/), assuming it's used and it works. Which may be a problem.

denis-zhdanov: Yes, an exception is thrown from an automatically generated check which is inserted into the method's beginning. The error message explicitly mentions target problem parameter's name.

spotbugsTestCases/src/java/bugIdeas/Ideas_2011_08_05.java
denis-zhdanov: As the parameter is marked by *Nonnull*, *Traute* inserts a *null*-check there and the test fails like below:  

```
edu.umd.cs.findbugs.DetectorsTest > testAllRegressionFilesJavac FAILED
    java.lang.AssertionError: Unexpected bugs (1):
    unexpected RCN Redundant null check at Ideas_2011_08_05.java:[line 52]
        at org.junit.Assert.fail(Assert.java:88)
        at edu.umd.cs.findbugs.DetectorsTest.checkForUnexpectedBugs(DetectorsTest.java:142)
```  

Removed the explicit check. The tests pass now

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/AnalysisCache.java
iloveeclipse: Stil I'm not convinced, may be too old school here.
The resulting stack trace will point to the code which does not exist, right? So given a bug report with some NPE we will need to guess which version of the code was used, because line numbers in the stack will show something no one can see in the code?

Also I'm worried about debugging in Eclipse - I guess the Traute compiler plugin isn't working if running Eclipse, because it looks like it is for javac only?

Means: running "same" code in Eclipse debugger (compiled by Eclipse ecj compiler) will show a different behavior? If yes, this is a real show stopper and I vote against this PR.

Maaartinus: > The resulting stack trace will point to the code which does not exist, right?

@iloveeclipse I guess, the stacktrace points to the method signature line, where there's no code, just the annotation. Someone might be confused one day, but the behavior is actually very logical and despite the possible confusion it's clear where to search for the culprit. So I don't think that this could be a problem (I'm not saying anything about the javac/ecj problem).

iloveeclipse:
@denis-zhdanov : I like the idea but I miss the suport for Eclipse development (means: ecj compiler plugin).
Without ecj support this PR will not add any benefit but only introduce pain for those who use Eclipse.
If you have a solution which makes Eclipse happy, feel free to reopen.

denis-zhdanov:
@iloveeclipse, fair enough, created a [ticket](https://github.com/denis-zhdanov/traute/issues/90) for that

PR #514, opened 2017-12-29
handle long constant values in the OpcodeStack

mebigfatguy:
processing of "J" (long value constants ) was not processed, so added. Also changed code to switch rather than if/else chain.
----


iloveeclipse:
Dave, few questions:
- do you have a use case for that, where we fail to interpret long constants?
- could you add a test case?
- looks like the change can go first to 3.1 branch, because it contains a fix only?

mebigfatguy:
I just found it thru inspection, but i can probably cook up a test case. and 3.1 is fine.

KengoTODA:
I changed base branch to `release-3.1`, please solve conflict.

KengoTODA:
@mebigfatguy ping

KengoTODA:
Created #688 to merge this change, please refer that

PR #528, opened 2018-01-11
497: Do not print exceptions for unsupported classpath files

briehman:
Assumes that any files thought to be a zip file in the classpath that
cannot be loaded as a zip file should be ignored. Many files may be
placed on the classpath but this does not guarantee they are valid.

Ignoring these files reduces error output and mimics the behavior of the
JVM which does not complain about invalid files on the classpath. These
files are simply ignored from processing.

----

- [X] Added an entry into `gradlePlugin/CHANGELOG.md` if you have changed Gradle plugin code


iloveeclipse:
Can you please rebase your PR on release-3.1 branch and also target to this branch?
I will try to review later.

briehman:
Will do.  I just realized that this only caught my initial case and missed
the required header in a new file. I should also be able to add some tests
around this change as well.

Issue #497, opened 2017-11-09
java.io.IOException: Wrong magic bytes of ...


BrainStone:
Hi. When running spotbugs on my test classes through gradle I get the following error:

```
> Task :spotbugsTest
Scanning archives (33 / 33)
2 analysis passes to perform
Pass 1: Analyzing classes (96 / 96) - 100% complete
Pass 2: Analyzing classes (7 / 7) - 100% complete
Done with analysis
The following errors occurred during analysis:
  Cannot open codebase filesystem:C:\Users\Yannick\Project-Creative\Arrestia\GitLab-Telegram-Bot\build\classes\java\test\META-INF\org\apache\logging\log4j\core\config\plugins\Log4j2Plugins.dat
    java.io.IOException: Wrong magic bytes of 1 for zip file C:\Users\Yannick\Project-Creative\Arrestia\GitLab-Telegram-Bot\build\classes\java\test\META-INF\org\apache\logging\log4j\core\config\plugins\Log4j2Plugins.dat of 108 bytes
      At edu.umd.cs.findbugs.classfile.impl.ZipFileCodeBase.<init>(ZipFileCodeBase.java:87)
      At edu.umd.cs.findbugs.classfile.impl.ZipCodeBaseFactory.makeZipCodeBase(ZipCodeBaseFactory.java:39)
      At edu.umd.cs.findbugs.classfile.impl.ClassFactory.createFilesystemCodeBase(ClassFactory.java:123)
      At edu.umd.cs.findbugs.classfile.impl.FilesystemCodeBaseLocator.openCodeBase(FilesystemCodeBaseLocator.java:77)
      At edu.umd.cs.findbugs.classfile.impl.ClassPathBuilder.processWorkList(ClassPathBuilder.java:620)
      At edu.umd.cs.findbugs.classfile.impl.ClassPathBuilder.build(ClassPathBuilder.java:230)
      At edu.umd.cs.findbugs.FindBugs2.buildClassPath(FindBugs2.java:674)
      At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:215)
      At com.github.spotbugs.internal.spotbugs.SpotBugsExecuter.runSpotbugs(SpotBugsExecuter.java:23)
      At sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
      At sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #529, opened 2018-01-12
497: Do not print exceptions for unsupported classpath files

briehman:
Assume that any files thought to be a zip file in the classpath that
cannot be loaded as a zip file should be ignored. Many files may be
placed on the classpath but this does not guarantee they are valid.

Prior to this change, an exception would be thrown which was presented
to the user regarding the invalid path. This did not cause a failure,
but was hard to discern from an actual exception. Ignoring these files
reduces unnecessary output and mimics the behavior of the JVM which does not
complain about invalid files on the classpath.

----

- [X] Added an entry into `gradlePlugin/CHANGELOG.md` if you have changed Gradle plugin code

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
@@ -19,6 +19,7 @@
 
 package edu.umd.cs.findbugs.classfile.impl;
 
+import edu.umd.cs.findbugs.FindBugs2;
 import java.io.File;
 import java.io.IOException;
 
@@ -107,11 +108,11 @@ public class ClassFactory implements IClassFactory {
         File file = new File(fileName);
 
         if (!file.exists() || !file.canRead()) {
-            return new EmptyCodeBase(codeBaseLocator);
+            return createEmptyCodeBase(codeBaseLocator, file);
         } else if (file.isDirectory()) {
             return new DirectoryCodeBase(codeBaseLocator, file);
         } else if (!file.isFile()) {
-            return new EmptyCodeBase(codeBaseLocator);
+            return createEmptyCodeBase(codeBaseLocator, file);
         } else if (fileName.endsWith(".class")) {
             return new SingleFileCodeBase(codeBaseLocator, fileName);
         } else if (fileName.endsWith(File.separator + "jrt-fs.jar")) {
@@ -120,9 +121,16 @@ public class ClassFactory implements IClassFactory {
             try {
                 return ZipCodeBaseFactory.makeZipCodeBase(codeBaseLocator, file);
             } catch (IOException e) {
-                return new EmptyCodeBase(codeBaseLocator);
+                return createEmptyCodeBase(codeBaseLocator, file);
             }
         }
+    }
+
+    private static IScannableCodeBase createEmptyCodeBase(FilesystemCodeBaseLocator codeBaseLocator, File file) {
+        if (FindBugs2.DEBUG) {
+            System.out.println("Ignoring unreadable or non-existent file " + file);
+        }
+        return new EmptyCodeBase(codeBaseLocator);
     }
 
     static IScannableCodeBase createNestedZipFileCodeBase(NestedZipFileCodeBaseLocator codeBaseLocator)
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
@@ -20,7 +20,6 @@
 package edu.umd.cs.findbugs.classfile.impl;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 
 import edu.umd.cs.findbugs.BugReporter;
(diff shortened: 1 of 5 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
```diff
@@ -107,20 +106,22 @@ static IScannableCodeBase createFilesystemCodeBase(FilesystemCodeBaseLocator cod
...
-            throw new FileNotFoundException("File " + file.getAbsolutePath() + " doesn't exist");
-        } else if (!file.canRead()) {
-            throw new IOException("File " + file.getAbsolutePath() + " not readable");
+        if (!file.exists() || !file.canRead()) {
+            return new EmptyCodeBase(codeBaseLocator);
         } else if (file.isDirectory()) {
             return new DirectoryCodeBase(codeBaseLocator, file);
         } else if (!file.isFile()) {
-            throw new IOException("File " + file.getAbsolutePath() + " is not a normal file");
+            return new EmptyCodeBase(codeBaseLocator);
```
KengoTODA: I think we need to log the absolute path of `file`, to help debugging.

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
```diff
@@ -107,20 +106,22 @@ static IScannableCodeBase createFilesystemCodeBase(FilesystemCodeBaseLocator cod
...
         } else if (fileName.endsWith(".class")) {
             return new SingleFileCodeBase(codeBaseLocator, fileName);
         } else if (fileName.endsWith(File.separator + "jrt-fs.jar")) {
             return new JrtfsCodeBase(codeBaseLocator, fileName);
         } else {
-            return ZipCodeBaseFactory.makeZipCodeBase(codeBaseLocator, file);
+            try {
+                return ZipCodeBaseFactory.makeZipCodeBase(codeBaseLocator, file);
+            } catch (IOException e) {
+                return new EmptyCodeBase(codeBaseLocator);
```
KengoTODA: I think we need to log the stack trace of `file`, to help debugging.  
We should pay attention when ignore thrown exception, I think.

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/EmptyCodeBase.java
```diff
@@ -0,0 +1,61 @@
...
+            @Override
+            public ICodeBaseEntry next() throws InterruptedException {
+                throw new NoSuchElementException();
+            }
+        };
+    }
+
+    @Override
+    public ICodeBaseEntry lookupResource(String resourceName) {
+        return null;
```
KengoTODA: confirmed that this method can return `null`.

https://github.com/spotbugs/spotbugs/blob/release-3.1/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/ICodeBase.java#L54-L62

spotbugs/src/test/java/edu/umd/cs/findbugs/classfile/impl/ClassFactoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private void assertHasNoCodeBase(IScannableCodeBase codeBase) throws Exception {
+        assertNotNull(codeBase);
+        assertFalse(codeBase.iterator().hasNext());
+    }
+
+    private void assertHasCodeBase(IScannableCodeBase codeBase) throws Exception {
+        assertNotNull(codeBase);
+        assertTrue(codeBase.iterator().hasNext());
+    }
+}
```
KengoTODA: Which test case reproduces the bug reported at #497?

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
briehman: Thanks for the review! I can definitely add some logging. I am not sure of the best way to add the logging. It seems that the `java.util.Logger` is used infrequently in the code and this doesn't fall under the use case of `IErrorLogger`.

Any recommendations for the best logging mechanism in this situation?

spotbugs/src/test/java/edu/umd/cs/findbugs/classfile/impl/ClassFactoryTest.java
briehman: Two of these could be represented by #497:  `ignoreUnreadableFile` and `ignoreUnknownNonClassFileFormat`. For the other two tests, an exception would be thrown which would generate a stacktrace in the output.

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
briehman: Hi @KengoTODA. Trying a ping to see if you have thoughts on how best to log in this situation. Thanks!

KengoTODA: Now SpotBugs uses `System.out` to log, you can refer [FindBugs2.java](https://github.com/spotbugs/spotbugs/blob/release-3.1/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java#L906-L908) for example.

note: When we log error, we can take `IErrorLogger` from `AnalysisContext` or `AnalysisCache`, but it doesn't support logging warning.

KengoTODA: If we're sure that this log is needless for normal use, we can log it only when `FindBugs2.DEBUG` is true.

briehman: Change made to log to `System.out` when `Findbugs2.DEBUG` is enabled.

PR #537, opened 2018-01-18
RFC: Add support for Checker Framework's nullable annotations

dpursehouse:


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/NullnessAnnotation.java
```diff
@@ -40,6 +40,8 @@ boolean match(@DottedClassName String className) {
                     || "com.google.common.base.Nullable".equals(className)
                     || "org.eclipse.jdt.annotation.Nullable".equals(className)
                     || "org.jetbrains.annotations.Nullable".equals(className)
+                    || "org.checkerframework.checker.nullness.qual.Nullable".equals(className)
```
iloveeclipse: I would remove all the `.Nullable".equals()` above and replace them all with `className.endsWith(".Nullable")`.

dpursehouse:
@iloveeclipse thanks for the quick feedback.  I will retarget this to the 3.1 branch and address the comments.

I wasn't sure if tests need to be added for this?

iloveeclipse:
> I wasn't sure if tests need to be added for this?

Good point, sorry I didn't mention this. Would be of course great if you could add a test.

philipl:
You also need to update spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java if you want the false positives in guava 23.6 to go away.

dpursehouse:
Superseded by #539

PR #538, opened 2018-01-18
Add support for errorprone's @CanIgnoreReturnValue

### LINKED ISSUE #463: RV_RETURN_VALUE_IGNORED should not be reported for methods annotated with '@com.google.errorprone.annotations.CanIgnoreReturnValue'
Methods annoted with `@com.google.errorprone.annotations.CanIgnoreReturnValue` generate an `RV_RETURN_VALUE_IGNORED` error.

For example `com.google.common.base.Stopwatch#start` causes this issue.

philipl:
Guava switched to using this errorprone annotation, which leads to RV_RETURN_VALUE_IGNORED false positives when using Guava.

Fixes #463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2017-??-??
 
+### Added
+
+* Support for errorprone @CanIgnoreReturnValue annotation ([#463](https://github.com/spotbugs/spotbugs/issues/463))
+
 ### Fixed
 
 * Lambda methods reported as missing classes ([#527](https://github.com/spotbugs/spotbugs/issues/527))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2017-??-??
 
+### Added
+
+* Support for errorprone @CanIgnoreReturnValue annotation
```
iloveeclipse: Can you please add a link to issue #463? See examples in the file below.

philipl:
Updated with both a test and actually fully implementing the annotation support. My first diff just supported the annotation in package-info but not on a class or method. Whoops. :-)

bentatham:
This still fails for me.  Running in eclipse (3.1.2.r201802250230-59118d9 - reports SpotBugs version 3.1.2 on the SpotBugs preferences "Plugin and misc. Settings" tab.  Also fails via maven using 3.1.2.
``Return value of com.google.common.collect.ImmutableList$Builder.add(Object) ignored in ca.nanometrics.ims.cd11.frames.DataFrameTest.createChannelConfigSet(BandId[]) [Scariest(3), 
 High confidence]``

Do I have to configure something to make this skip this warning?

PR #539, opened 2018-01-19
Add support for Checker Framework nullable annotations

dpursehouse:


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/NullnessAnnotation.java
```diff
@@ -40,6 +40,8 @@ boolean match(@DottedClassName String className) {
                     || "com.google.common.base.Nullable".equals(className)
                     || "org.eclipse.jdt.annotation.Nullable".equals(className)
                     || "org.jetbrains.annotations.Nullable".equals(className)
+                    || "org.checkerframework.checker.nullness.qual.Nullable".equals(className)
```
dpursehouse: FYI @iloveeclipse I didn't change this to use `endsWith` as you suggested in the other PR on master, because in the other places it's also using explicit class names and I thought it better to leave it consistent.  I will change it, however, if you still think it's better to.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2017-??-??
 
+* Added support for Checker Framework's Nullable annotations.
```
KengoTODA: Note that annotating type parameter isn't supported yet.

iloveeclipse:
Looks good, thanks David!

PR #554, opened 2018-02-05
Use StringBuilder instead of StringBuffer as it offers high performan…

reudismam:
…ce in single thread places as it is generally the case.

----


iloveeclipse:
Don't get me wrong, but the PR does not make sense because none of the changed places are performance critical in any way, and it just adds noise to the source history.

If you find a place where StringBuffer is used in a performance critical path, your patches are welcome.

Issue #580, opened 2018-03-09
False positive: 'parameter must be non null' even though parameter was checked by Guavas Preconditions.checkNotNull()


ewirch:
```    
int method(@Nullable String param) {
   Preconditions.checkNotNull(param);
   return param.length();
}
```

`checkNotNull()` will throw if `param` is null. So `param` is effectively non-null, when execution reaches the `return` statement.

Pull request containing test case follows....

Sovietaced:
Same issue here. Had to roll our own `Preconditions.checkNotNull` with an annotation to suppress the FB warnings.

KengoTODA:
See #581, you need to use `@NonNull` annotation instead of `@Nullable`.

PR #581, opened 2018-03-09
Add test to trigger 'parameter must be non null' when using Precondit…

ewirch:
…ions.checkNotNull

----

See issue #580

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullNullableTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullNullableTest.java
@@ -1,0 +1,37 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.BugCollection;
+import edu.umd.cs.findbugs.test.SpotBugsRule;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.nio.file.Paths;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.collection.IsEmptyIterable.emptyIterable;
+import static org.junit.Assume.assumeThat;
+
+public class PreconditionsCheckNotNullNullableTest {
+    @Rule
+    public SpotBugsRule spotbugs = new SpotBugsRule();
+
+    @Before
+    public void setUp() {
+        String property = System.getProperty("AUX_CLASSPATH");
+        assumeThat(property, notNullValue());
+        for (String path : property.split(",")) {
+            spotbugs.addAuxClasspathEntry(Paths.get(path));
+        }
+    }
+
+    @Test
+    public void testNullableIsSanitizedByCheckNonNull() {
+        BugCollection bugCollection = spotbugs.performAnalysis(Paths.get("../spotbugsTestCases/build/classes/java/main/bugPatterns/NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE_Guava_Preconditions.class"));
+        Assert.assertThat(bugCollection, is(emptyIterable()));
+    }
+
+}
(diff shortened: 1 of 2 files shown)


KengoTODA:
I'm afraid that this request is hard to support for current SpotBugs platform. Here is workaround:

```
int method(@Nullable String param) {
  return Preconditions.checkNotNull(param).length();
}
```

or:

```
int method(@Nullable String param) {
  param = Preconditions.checkNotNull(param);
  return param.length();
}
```

note: [`DefaultNullnessAnnotations.java`](https://github.com/spotbugs/spotbugs/blob/release-3.1/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/DefaultNullnessAnnotations.java) already treats the method parameter of `Preconditions#checkNotNull(Object)` as `@Nonnull`.

ewirch:
None of the workarounds work. (which makes them non-working-arounds? ;) ) The test still fails.

KengoTODA:
I found that this is an user error.

`@Nullable` means that argument can be null, but `Preconditions.checkNotNull(param)` does not permit to take null as 1st parameter. So if parameter is `@Nullable`, we should not use `Preconditions.checkNotNull(param)`.

If you want to use `Preconditions.checkNotNull(param)`, it means that you do not permit argument to be null, then you should use `@NonNull` instead of `@Nullable`.

ewirch:
While technically speaking this is correct, it makes the transition from `@Nullable` to `@NonNull` cumbersome and reduces the usefulness of `checkNotNull()` to nearly zero.

You often have to handle the transition between "nullable" and "possibly nullable" values to non-null values. Either because your code base is not completely nullity-annotated, or because you use library code, which does not use nullity annotations.

So let's pretend we use some unannotated library methods:
```
void method1() {
  Object obj1 = lib1.getValue();
  Object obj2 = lib1.getValue();
  process(obj1, obj2);
}

void method2() {
  Object obj1 = lib2.getName();
  Object obj2 = lib2.getName();
  process(obj1, obj2);
}

private void process(Object ob1, Object obj2) {
  use(obj1.toString(), obj2.toString());
}
```

Since the library methods are not annotated, we have to expect nullable values. The usual approach would be to allow to pass nullable values to `process()` and check the nullity there.

```
void method1() {
  Object obj1 = lib1.getValue();
  Object obj2 = lib1.getValue();
  process(obj1, obj2);
}

void method2() {
  Object obj1 = lib2.getName();
  Object obj2 = lib2.getName();
  process(obj1, obj2);
}

private void process(@Nullable Object ob1, @Nullable Object obj2) {
  use(
    Objects.checkNotNull(obj1).toString(),
    Objects.checkNotNull(obj2).toString()
  );
}
```

You suggest, that `process()` should not accept `@Nullable` values. So I have to check for nullity outside of `process()`. Additionally, I can't use `checkNotNull()` to do so, because the varables are `@Nullable` as well. This yields this code:

```
void method1() {
  Object obj1 = lib1.getValue();
  if (obj1 == null) {
    throw new NullPointerException();
  }

  Object obj2 = lib1.getValue();
  if (obj2 == null) {
    throw new NullPointerException();
  }

  process(obj1, obj2);
}

void method2() {
  Object obj1 = lib2.getName();
  if (obj1 == null) {
    throw new NullPointerException();
  }

  Object obj2 = lib2.getName();
  if (obj2 == null) {
    throw new NullPointerException();
  }

  process(obj1, obj2);
}

private void process(@NonNull Object ob1, @NonNull Object obj2) {
  use(
    Objects.checkNotNull(obj1).toString(),
    Objects.checkNotNull(obj2).toString()
  );
}
```

This makes the code verbose and cumbersome. Additionally `checkNotNull()` has neary zero value, because Spotbugs already checked, that only non-null values have been passed to `process()`.

If this does not convince you, this should: `java.util.Objects.requireNonNull(T)` is not annotated, and triggers the same behavior.

KengoTODA:
First of all, please understand that I'm not spec lead. What I can tell you is just about how FindBugs was implemented, based on my investigation.

> You suggest, that process() should not accept @Nullable values. So I have to check for nullity outside of process().

Yes, and it is what FindBugs/SpotBugs can help you. If argument is annotated as `@NonNull` but some codes give `null` or nullable value, then FindBugs/SpotBugs can alert it as wrong usage. Of course you can use `requireNonNull()` to make your code defensive.

Sovietaced:
> This makes the code verbose and cumbersome. Additionally checkNotNull() has neary zero value, because Spotbugs already checked, that only non-null values have been passed to process().

💯 

This is broken

Issue #587, opened 2018-03-16
Add a detector to check if placeholder correct when call log.[info|error|warn|debug|trace].


guoxiangbin:
I will add a detector to check this.

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class WrongPlaceholderOfSlf4j {
    public static void main(String[] args) {
        Logger logger = LoggerFactory.getLogger(WrongPlaceholderOfSlf4j.class);

        /* Shuld report a bug, have 2 placeholders but need 1 */
        logger.warn("hello {} {}", strParam);
    }
}

guoxiangbin:
This detector is duplicated with LO_INCORRECT_NUMBER_OF_ANCHOR_PARAMETERS in fb-contrib and SLF4J_PLACE_HOLDER_MISMATCH in findbugs-slf4j.

PR #588, opened 2018-03-16
Add CheckLogParamNum detector for slf4j. #587

guoxiangbin:
Add CheckLogParamNum detector for slf4j. #587 
Check if placeholder correct when call log.[info|error|warn|debug|trace].
When the first parameter is a formatted string, and the number of braces in it does not match the number of formatted parameters, report a bug.

---
This is the first time I submit my code on GitHub. If there is any mistake, I hope I can get your help.

----

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2018-??-??
 
+* Add a detector to check if placeholder correct for slf4j. ([#587](https://github.com/spotbugs/spotbugs/issues/587))
+
 ## 3.1.2 - 2018-02-24
 
 ### Added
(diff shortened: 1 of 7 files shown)


KengoTODA:
This detector is duplicated with `LO_INCORRECT_NUMBER_OF_ANCHOR_PARAMETERS` in fb-contrib and `SLF4J_PLACE_HOLDER_MISMATCH` in findbugs-slf4j. I personally recommend you to use them, because they are already tested during long time.

guoxiangbin:
Thank you for your reply. I can't search the two detectors you're talking about in the spotbugs project, and I can't detect the bug in my test case under the default configuration. I am very eager to get detailed instructions on the use of them.

KengoTODA:
Just install https://kengotoda.github.io/findbugs-slf4j/ or https://github.com/mebigfatguy/fb-contrib then they'll trigger additional detectors provided by plugins.

The procedure to install plugin depends on how you run spotbugs. For instance, if you're using gradle, check [this manual page](http://spotbugs.readthedocs.io/en/latest/gradle.html#introduce-spotbugs-plugin) for detail.

guoxiangbin:
Thank you very much. I'll try to use them. By the way, Is there a detector that detects the problem of synchronized on local variables? #590

guoxiangbin:
This detector is duplicated with LO_INCORRECT_NUMBER_OF_ANCHOR_PARAMETERS in fb-contrib and SLF4J_PLACE_HOLDER_MISMATCH in findbugs-slf4j.

PR #591, opened 2018-03-21
Fix resource leaks

KengoTODA:
Current implementation does not close `SourceFinder` even after analysis.
This PR also applies try-with-resources to handle zip files.

@iloveeclipse I'm not sure about the difference between `FindBugs2#dispose()` and `FindBugs2#clearCache()`, could you review? At least, can we destruct `project` instance by `clearCaches()` or not?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2018-??-??
 
+### Fixed
+
+* Close source file after analysis ([#591](https://github.com/spotbugs/spotbugs/issues/591))
+
 ## 3.1.2 - 2018-02-24
 
 ### Added
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
@@ -68,6 +68,7 @@ import edu.umd.cs.findbugs.config.AnalysisFeatureSetting;
 import edu.umd.cs.findbugs.config.UserPreferences;
 import edu.umd.cs.findbugs.detect.NoteSuppressedWarnings;
 import edu.umd.cs.findbugs.filter.FilterException;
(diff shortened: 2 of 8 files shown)


KengoTODA:
This PR still needs to add a fix for `PluginLoader`, see https://github.com/roberthubert/spotbugs/commit/7c471388c5e946cdc14abf21aea0a4320d43b3d3

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Handle annotation on `package-info.class` properly ([#592](https://github.com/spotbugs/spotbugs/issues/592))
 * Update asm to 6.1.1 to support Java 10
 * Update Apache BCEL to 6.2 to support Java 9 package & module reference
+* Close source file after analysis ([#591](https://github.com/spotbugs/spotbugs/issues/591))
```
KengoTODA: I'll move this to `Unreleased` part

KengoTODA:
@iloveeclipse please check and review:

> @iloveeclipse I'm not sure about the difference between FindBugs2#dispose() and FindBugs2#clearCache(), could you review? At least, can we destruct project instance by clearCaches() or not?

Note that we can ignore these two SonarQube warnings; method caller should be responsible to close resources, not these methods themselves.

iloveeclipse:
If I remember right, `clearCaches()` was there to allow Eclipse plugin to free memory after analyis while the rest of FB2 would still be reusable. `dispose()` is the end of life. I quickly tried the change, it seem no to break the plugin, but I have no time for a deep review.

KengoTODA:
OK then I will stop closing project in `clearCaches()`, and make sure that `FindBugs2.main(...)` invokes not only `clearCaches()` but also `dispose()`.

KengoTODA:
@iloveeclipse PR has succeeded. Please check. I think all the warning by spotbugs can be ignored, since they're for existing code, not the newly added one.

Please use 'squash and merge' to merge this PR, it seems that recently GitHub isn't so wise to handle conflict.

Issue #595, opened 2018-03-23
RANGE_ARRAY_LENGHT and RANGE_ARRAY_OFFSET false negative


yerayrodriguez:
If you invoke read/write operations using lenght = buffer.lenght + 1, RANGE_ARRAY_LENGHT is not detected. 
```
    private static final int BUFFER_LENGTH = 10;
    public void test() {
        File file = new File("temp.txt");
        try (FileReader fr = new FileReader(file)) {
            char[] buffer = new char[BUFFER_LENGTH];
            // False negative: Expected RANGE_ARRAY_LENGHT
            fr.read(buffer, 0, BUFFER_LENGTH + 1);
        } catch (Exception e) {
        }
    }
```
However, it works correctly with buffer.length + 2.

Same issue with RANGE_ARRAY_OFFSET. Also with some String constructors.

KengoTODA:
Closed by #596

PR #596, opened 2018-03-23
Fix for Issue 595

yerayrodriguez:
Test case and fix for [issue-595](https://github.com/spotbugs/spotbugs/issues/595)


yerayrodriguez:
TravisCI is having problem downloading JDKs

yerayrodriguez:
Is it necessary to do a new commit to force Travis CI to check again the pull request?

KengoTODA:
I've kicked build job, let's wait its result

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue595Test.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    @Test
+    public void testRangeArrayLength() {
+        performAnalysis("rangeArray/RangeArrayLengthAndOffsetBugExample.class");
+
+        BugInstanceMatcher bugTypeMatcherLength = new BugInstanceMatcherBuilder().bugType("RANGE_ARRAY_LENGTH").build();
+        BugInstanceMatcher bugTypeMatcherOffset = new BugInstanceMatcherBuilder().bugType("RANGE_ARRAY_OFFSET").build();
+
+        assertThat(getBugCollection(), containsExactly(4, bugTypeMatcherOffset));
+        assertThat(getBugCollection(), containsExactly(13, bugTypeMatcherLength));
```
KengoTODA: Are you sure that these all of 4 & 13 bugs are located at expected lines?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
```diff
@@ -208,13 +208,6 @@ public void sawOpcode(int seen) {
         }
     }
 
-    static int saturatingIncrement(int value) {
-        if (value == Integer.MAX_VALUE) {
-            return Integer.MAX_VALUE;
-        }
-        return value+1;
-    }
-
```
KengoTODA: I'm not sure why we can remove this logic safely. But the commit which introduces this method e09d217768d636b1eddcb90ca2317d62eacf331d doesn't explain its motivation so clearly... :sob:

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue595Test.java
KengoTODA: If not, it would be better to separate example class and test method

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
iloveeclipse: It is to avoid integer overflow and negative int value instead of a max int.

yerayrodriguez: That commit replace an increment operation with a call to incrementSaturating() to avoid integer overflow, but this increment operation is precisely the one that produces the bug. Moreover, since the increment operatation is not needed, there is not risk to overflow.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue595Test.java
yerayrodriguez: Example class lists the expected lines. Example:
```java
public void ioOperation5() {
    File file = new File("temp.txt");
    try (FileReader fr = new FileReader(file)) {
        char[] buffer = new char[BUFFER_LENGTH];
        // Expected RANGE_ARRAY_OFFSET #1
        // Expected RANGE_ARRAY_LENGHT #4
        fr.read(buffer, BUFFER_LENGTH + 1, 1);
    } catch (Exception e) {
    }
}
```
I did in this way because both rules are quite related and the bug affected both of them. If you prefer I can make an independent RANGE_ARRAY_LENGHT test class.

KengoTODA: Ya I know meaning of these comments, I just want to ensure that this comment matches actual analysis result or not.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
KengoTODA: You mean that this increment operation is totally needless, right? Does it mean that existing code misunderstood meaning of 3rd argument of `checkRange()` (e.g. it is 0-based index but we used it as 1-based one)?

yerayrodriguez: Yes, at least when checking RANGE_ARRAY_LENGTH and RANGE_ARRAY_OFFSET. Other rules, for instance RANGE_STRING_INDEX works fine (https://github.com/spotbugs/spotbugs/commit/e09d217768d636b1eddcb90ca2317d62eacf331d got rid of the increment operation in that rule)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue595Test.java
yerayrodriguez: I've rewritten the test class. Now it only checks one expected rule per method. I think it's clearer now.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Handle annotation on `package-info.class` properly ([#592](https://github.com/spotbugs/spotbugs/issues/592))
 * Update asm to 6.1.1 to support Java 10
 * Update Apache BCEL to 6.2 to support Java 9 package & module reference
+* RANGE_ARRAY_LENGHT and RANGE_ARRAY_OFFSET false negative ([#595](https://github.com/spotbugs/spotbugs/issues/595))
```
KengoTODA: Please merge latest `release-3.1`, solve conflict, and move this line to under `## Unreleased - 2018-??-??`.

KengoTODA:
Thanks @yerayrodriguez, now changes seems really good! Could you try to solve conflict, then I'll merge this PR.

yerayrodriguez:
Hi @KengoTODA, which conflict do you mean???

yerayrodriguez:
How strange, what I see is: This branch has no conflicts with the base branch.

KengoTODA:
Wmm, this is what I can see in my browser...

screenshot from 2018-04-26 17-00-33

yerayrodriguez:
Ok, I'll prepare a new PR

Issue #605, opened 2018-03-30
Unit test ClassFactoryTest.ignoreUnreadableFile fails on Windows


johaw:
The unit test ClassFactoryTest.ignoreUnreadableFile() fails when executed on Windows.  

The following is an example of failure when running './gradlew build' with windows git bash (a similar error is seen in eclipse).

```
:spotbugs:test

edu.umd.cs.findbugs.classfile.impl.ClassFactoryTest > ignoreUnreadableFile FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertFalse(Assert.java:64)
        at org.junit.Assert.assertFalse(Assert.java:74)
        at edu.umd.cs.findbugs.classfile.impl.ClassFactoryTest.assertHasNoCodeBase(ClassFactoryTest.java:103)
        at edu.umd.cs.findbugs.classfile.impl.ClassFactoryTest.ignoreUnreadableFile(ClassFactoryTest.java:63)
```

The problem is that file.setReadable(false) is not guaranteed to work on all JVM's and OS's.  The setReadable method does not throw an exception but fails silently and the return method needs to be checked.    

I will submit a PR to ignore/skip this test if the setReadable method fails.

johaw:
@KengoTODA Thank you for merging my PR.   I have rebased my forked release-3.1 branch to the upstream release-3.1 branch and I can confirm that this issue can now be closed.

Please let me know if you want me to close this issue or if you will. The only reason I'm not closing the issue is to check with you if some automated or manual process should close this issue after the PR was merged or if you want the original submitter to close the issue.

Thanks,
@johaw

KengoTODA:
Thanks, I will close this issue.

PR #606, opened 2018-03-31
BUG-605:  Ignore unit test on windows

johaw:
Ignore unit test if the required file permissions can not be set.  Changing
file permissions will likely work on *nix systems and fail on windows
systems.

----


johaw:
I don't think I have referenced the bug correctly, but this is for issue https://github.com/spotbugs/spotbugs/issues/605

PR #592, opened 2018-03-21
Fix 582: support error_prone's @CanIgnoreReturnValue with guava < 23.6-jre

KengoTODA:
To fix #582, we need to fix two problems:

* error_prone's `@CheckReturnValue` annotation is not supported (this is feature improvement)
* `CheckReturnAnnotationDatabase` does not handle annotation on `package-info.class` (this is bugfix)

PR #607, opened 2018-04-03
Fix 582: support error_prone's @CanIgnoreReturnValue with guava < 23.6-jre

KengoTODA:
This is rebased PR from #592


KengoTODA:
Thank you!

PR #613, opened 2018-04-18
Release 3.1.3

KengoTODA:
After we merge these commits, we do:

1. Tag 2bb1d9b as `3.1.3` or its rebased commit then Travis CI will release 3.1.3 to Sonatype Staging Nexus.
2. Use Sonatype Nexus to release staging artifacts to Maven Central.
3. Make sure that Eclipse Update Site is successfully updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased - 2018-??-??
+## 3.1.3 - 2018-04-18
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


KengoTODA:
`sudo apt-get update oracle-java8-installer` is failing, maybe we need to wait installer's update.

reference: https://stackoverflow.com/a/46815898/814928

iloveeclipse:
Kengo, can you please attach the zipped Eclipse update site from the release build? It is a convenient way for "offline" distribution of Eclipse plugins.

KengoTODA:
We made it possible by a5525ed673f9e7981ec64d4c5d664df4251a75d8, however current `build.gradle` has no configuration to do so. I will investigate when/why we stop deploying zip file to Maven central.

KengoTODA:
I found that I disabled it at #170, because Maven Central doesn't permit to upload artifacts which has no javadoc or sources.

> disable deploying eclipsePlugin jar file, which has no javadoc and sources (it cannot pass sonatype nexus verification)

Maybe we can tag the commit in `gh-pages` branch such as `3.1.3` then we can download zip file from GitHub. For instance:

* https://github.com/spotbugs/eclipse/archive/3.1.3.zip

I'll try to automate tagging `gh-pages`.

Issue #558, opened 2018-02-18
Add explanation about -pluginList option


KengoTODA:
[SpotBugs official document](https://spotbugs.readthedocs.io/) explains nothing about `-pluginList` option. It's better to add.

refs: https://stackoverflow.com/a/32992247/814928

PR #618, opened 2018-04-26
558 Added -pluginList and others to offcial doc

jpmartins:
Starting point was issue #558, used the -help and added the options that were missing.

----

- [n/a] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code


docs/running.rst
```diff
@@ -135,6 +135,10 @@ These options may be used with both the GUI and command-line interfaces.
   Enable analyses which increase precision and find more bugs, but which may require more memory and take more time to complete.
   See :doc:`effort`.
 
+-effort[:less|default|more]:
```
KengoTODA: better to use `effort:[less|default|more]`?

docs/running.rst
```diff
@@ -231,3 +235,110 @@ These options are only accepted by the Text User Interface.
...
+-release <release name>:
+  Set the release name of the analyzed application.
+
+-maxRank <rank>:
+  Only report issues with a bug rank at least as scary as that provided.
+
+-dontCombineWarnings:
+  Don't combine warnings that differ only in line number.
+
+-train[:outputDir]:
```
KengoTODA: maybe `train:[outputDir]` or `train <outputDir>`?

docs/running.rst
```diff
@@ -231,3 +235,110 @@ These options are only accepted by the Text User Interface.
...
+  Save training data (experimental); output dir defaults to '.'.
+
+-useTraining[:inputDir]:
```
KengoTODA: `useTraining:[inputDir]` or `useTraining <inputDir>`?

KengoTODA:
Really good, please change the target branch from `master` to `release-3.1`.

jpmartins:
Hi, sorry for the hold up, not sure for all questions, i will look them up, as for changing target, as soon as i can i will search how to do that. Thanks

docs/running.rst
jpmartins: It <is> optional, I found it [on the source].

```
class edu.umd.cs.findbugs.TextUICommandLine:
addSwitchWithOptionalExtraPart("-train", "outputDir", "Save training data (experimental); output dir defaults to '.'");
```

docs/running.rst
jpmartins: Not sure. The less (-precision) and more (+precision) are intermediary levels between min (--precision) and max (++precision). Not sure about default...

jpmartins:
Think i got it. Is it OK now?

docs/running.rst
KengoTODA: I just meant that the position of `:` seems strange. 

BTW default value should be `more`, I investigated and noted at http://spotbugs.readthedocs.io/en/latest/effort.html

KengoTODA: In other words, we have five levels: min, less, default, more and max. And for now default and more has no difference. See `FindBugs` class's `XXX_EFFORT` fields for detail.

docs/running.rst
KengoTODA: So `[:foo]` means optional parameter? It's first time for me to see this expression...

jpmartins: The The braces [ ] to indicate "zero or more", https://www.techrepublic.com/blog/software-engineer/what-is-bnf-and-why-should-developers-care 
Not sure about the colon : meaning yet, but guessing it is a variable indicator in some sort of notation.

docs/running.rst
jpmartins: About the position of the : 
In CLI the context help ouputs:
-effort[:min|less|default|more|max]      set analysis effort level
Will re-write and propose to put together in on entry.

docs/running.rst
KengoTODA: OK I see, you copied output from that command line help, then I can understand its meaning.

That `:` means necessary separator. For instance, for `-train[:outputDir]` means that we need to input like `-train:/tmp/foo`.

KengoTODA:
This PR includes many unexpected commits, please check. You may recreate new PR with rebased branch.

jpmartins:
Replaced by PR "Issue 558++". Not sure what i've done wrong to created unexpected commits, guessing it was because i started working on master.

Issue #623, opened 2018-05-03
Add PMD self-checks


ThrawnCA:
Adding selected PMD rules to the build will help us clean up the codebase and keep it clean. There's a Gradle plugin for PMD (https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.PmdExtension.html), so this shouldn't be hard.

JuditKnoll:
There were some PMD rules added, but I'm not sure whether all was checked. @ThrawnCA WDYT can this issue be closed?

PR #624, opened 2018-05-03
add PMD and start cleaning up its findings, #623

ThrawnCA:
Technically I have changed code, but since it was just things like removing unused imports etc, there should be nothing that would need to go in a changelog (if any changes are externally visible, then I made a mistake).
----

### DIFF
--- a/spotbugs/src/test/java/edu/umd/cs/findbugs/ba/generic/TestGenericObjectType.java
+++ b/spotbugs/src/test/java/edu/umd/cs/findbugs/ba/generic/TestGenericObjectType.java
@@ -19,7 +19,11 @@
 
 package edu.umd.cs.findbugs.ba.generic;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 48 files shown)


ThrawnCA:
:D Well, it looks like my cleanup wasn't clean *enough*.

Incidentally, I notice that `SourceDirectoryWizard` claims to be generated in some fashion, and that the fields shouldn't be altered...is that actually part of the build, or was it generated once and now maintained manually?

spotbugs/src/test/java/edu/umd/cs/findbugs/ba/generic/TestGenericObjectType.java
```diff
@@ -19,11 +19,7 @@
 
 package edu.umd.cs.findbugs.ba.generic;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
KengoTODA: I think we do not need to use wildcard to import, at least it's not enabled in [Eclipse setting](https://github.com/spotbugs/spotbugs/blob/release-3.1/gradle/.settings-templates/org.eclipse.jdt.ui.prefs).

ThrawnCA: Well, I can switch off that PMD rule, but it means writing a custom PMD rules file...

How would you feel about me introducing AssertJ? Then only one static import is needed.

KengoTODA: I personally don't care which tool we use.

iloveeclipse: I would avoid adding extra dependencies just to fix some cosmetic issue.

ThrawnCA: @iloveeclipse Fair, although AssertJ is actually quite a nice testing tool in its own right.

But I've written a custom rules file and disabled the static import limitation, anyway.

Issue #635, opened 2018-05-10
Need to choose a logging strategy


ThrawnCA:
The SpotBugs codebase includes examples of:

- JDK 1.4+ built-in logging, eg in `edu.umd.cs.findbugs.FindBugs` and `edu.umd.cs.findbugs.gui2.MainFrameComponentFactory`
- Multiple incompatible hand-rolled logging setups, based on `edu.umd.cs.findbugs.classfile.IErrorLogger` and `edu.umd.cs.findbugs.log.Logger`
- STDOUT/STDERR logging, all over the place. Note that the SpotBugsBot marks this as a major defect.

We should choose a preferred logging style and stick to it.

ThrawnCA:
Since we don't need complex logging setups given our use case, nor do we need to handle tremendous throughput, and we want maximum compatibility with a variety of codebases, I'm leaning toward JDK standard logging.

KengoTODA:
I personality prefer SLF4J over JUL (JDK standard logging), because JUL does not provide intuitive API. Maybe we can introduce it from v4.0.

ThrawnCA:
I agree that JUL has a very poor API for parameterised logging. Still, it's available without adding any dependencies. If we're OK with adding a dependency, then SLF4J is fine with me. My workplace uses it, it's a good API.

Might be worth starting to replace STDOUT/STDERR with JUL now, then converting to SLF4J later?

zolyfarkas:
+1 for JUL, adding a dependency has its costs, and I don't think is it worth it...

hazendaz:
If this was still not done, didn't look, I would also go with slf4j.  Nearly every OSS uses it that is popular so it only makes sense to follow the heard.  Jul logging sucks and I've never seen a production ready app in the 1000s in my daily life that make such usage.  They are all slf4j.  In fact, lets say we made spotbugs wrapped with spring boot for some fun reason, its going to be slf4j or log4j2.  More so slf4j by default.  A few repos I'm involved with tried years ago to be smarter than the loggers, those all end up causing issues.

PR #646, opened 2018-06-01
GitHub #635 standardise logging

ThrawnCA:
@KengoTODA @mebigfatguy @henrik242 @iloveeclipse How does this conversion look? I'll do similarly for other classes if we're happy with it.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
@@ -24,17 +24,18 @@ import java.lang.invoke.MethodHandles;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.jar.Manifest;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.annotation.CheckForNull;
+
+import static java.util.logging.Level.*;
 
 /**
  * Version number and release date information.
  */
 public class Version {
 
-    private static final Logger LOGGER = Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
+    private static final Logger LOG = Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
 
     /**
      * SpotBugs website.
@@ -105,22 +106,22 @@ public class Version {
         String arg = argv[0];
 
         if ("-release".equals(arg)) {
-            LOGGER.info(VERSION_STRING);
+            LOG.info(VERSION_STRING);
         } else if ("-plugins".equals(arg)) {
             DetectorFactoryCollection.instance();
             for(Plugin p : Plugin.getAllPlugins()) {
-                LOGGER.log(Level.INFO, "Plugin: {0}", p.getPluginId());
-                LOGGER.log(Level.INFO, "  description: {0}", p.getShortDescription());
-                LOGGER.log(Level.INFO, "     provider: {0}", p.getProvider());
+                LOG.log(INFO, "Plugin: {0}", p.getPluginId());
+                LOG.log(INFO, "  description: {0}", p.getShortDescription());
+                LOG.log(INFO, "     provider: {0}", p.getProvider());
                 String version = p.getVersion();
                 if (version != null && version.length() > 0) {
-                    LOGGER.log(Level.INFO, "      version: {0}", version);
+                    LOG.log(INFO, "      version: {0}", version);
                 }
                 String website = p.getWebsite();
                 if (website != null && website.length() > 0) {
-                    LOGGER.log(Level.INFO, "      website: {0}", website);
+                    LOG.log(INFO, "      website: {0}", website);
                 }
-                LOGGER.info("");
+                LOG.info("");
             }
         } else if ("-configuration".equals(arg)){
             printVersion(true);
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
```diff
@@ -20,28 +20,34 @@
...
+import java.util.logging.Logger;
 
 import javax.annotation.CheckForNull;
 
 /**
  * Version number and release date information.
  */
 public class Version {
+
+    private static final Logger LOGGER = Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
```
iloveeclipse: I personally prefer just LOG.

iloveeclipse: Why not just Version.class.getName()?

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
```diff
@@ -20,28 +20,34 @@
...
+
     /**
      * SpotBugs website.
      */
     public static final String WEBSITE = "https://spotbugs.github.io/";
 
-    public final static String VERSION_STRING;
+    public static final String VERSION_STRING;
```
iloveeclipse: I would avoid mixing unrelated changes, here and below too. even if this is due the style checks. PR's should not have side effects.

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
```diff
@@ -99,22 +105,22 @@ public static void main(String[] argv) throws InterruptedException {
         String arg = argv[0];
 
         if ("-release".equals(arg)) {
-            System.out.println(VERSION_STRING);
+            LOGGER.info(VERSION_STRING);
```
iloveeclipse: I've never used JUL API: is the INFO level enabled by default?

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
```diff
@@ -90,7 +96,7 @@ public static void registerApplication(String name, String version) {
         return applicationVersion;
     }
 
-    public static void main(String[] argv) throws InterruptedException {
+    public static void main(String[] argv) {
```
iloveeclipse: I'm not sure if main() the best place to introduce logging, as usually main() is called from the CLI and the output is (usually) expected to go to the STD OUT/ERR. Just assuming someone manages to redirect Java logging to a file, executing it will just don't do anything visible.

For the rest of the source code logging makes sense of course.

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
```diff
@@ -99,22 +105,22 @@ public static void main(String[] argv) throws InterruptedException {
...
         if ("-release".equals(arg)) {
-            System.out.println(VERSION_STRING);
+            LOGGER.info(VERSION_STRING);
         } else if ("-plugins".equals(arg)) {
             DetectorFactoryCollection.instance();
             for(Plugin p : Plugin.getAllPlugins()) {
-                System.out.println("Plugin: " + p.getPluginId());
-                System.out.println("  description: " + p.getShortDescription());
-                System.out.println("     provider: " + p.getProvider());
+                LOGGER.log(Level.INFO, "Plugin: {0}", p.getPluginId());
```
iloveeclipse: I would use  `static import Level.*`, to be able to write it like:
`LOG.log(INFO, "Plugin: {0}", p.getPluginId());`

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: LOG or LOGGER works. But the class name is Logger, so...eh.

The reason for using `MethodHandles` is so that this line can be copy-pasted to any other class without modification.

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: It's to keep SonarQube happy, yeah. Unrelated, but pretty trivial, so...

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: If logging is redirected to a file, then I would assume that that's the desired behavior, so I would hesitate to basically override it by forcing STDOUT here.

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: It's configurable in `jre/lib/logging.properties`, but yes, the default level in that file is INFO.

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: I guess it's a matter of style. Is there any risk of `Level` constants clashing with anything else we might want to static-import (I'm particularly thinking of ERROR)?

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
ThrawnCA: @iloveeclipse Which approach would you prefer? Second pull request on the same branch after the first one is merged? Separate Github issue just saying "cleanup for issue X"?

KengoTODA:
If this is for `master` branch, I think it's acceptable to introduce SLF4J. It's not so big, and really stable logging facade nowadays (I'm personally not sure Log4J 2's stability and usability).

I think JUL is just a temporal workaround to make logging logic easy to read without additional dependency.

ThrawnCA:
@iloveeclipse @KengoTODA Any consensus on JUL vs SLF4J for a first pass? And does this pull request look OK as a starting point? I want to migrate other classes that are using STDOUT and STDERR, but let's first settle on a direction.

iloveeclipse:
I'm OK with either one, but  I'm not a logging fetishist, it should just work.

ThrawnCA:
OK, I'll call that a vote for SLF4J plus an abstention. Will update the pull request to use SLF4J.

KengoTODA:
Thanks @ThrawnCA, I also vote for SLF4J as n author of [findbugs-slf4j](https://github.com/KengoTODA/findbugs-slf4j).

mebigfatguy:
slf4j +1

ThrawnCA:
Attempting to convert to SLF4J results in a problem like https://github.com/spotbugs/spotbugs-gradle-plugin/issues/11

ThrawnCA:
Looks like the needs-work review from @iloveeclipse takes priority over the approval. Do I need to make further changes, or are you OK with it as-is?

spotbugs/src/main/java/edu/umd/cs/findbugs/Version.java
iloveeclipse: Cleanup after the main fix. This way one can always do cherry picking without side effects. So yes, please next time try to not mix different changes in one commit.

PR #652, opened 2018-06-06
Remove codes using deprecated BCEL's Constants

wreulicke:
* `org.apache.bcel.Constants` is deprecated. so, I rewrite to codes which do not uses that. 
* `edu.umd.cs.findbugs.visitclass.Constants2` only implements `org.apache.bcel.Constants`.
  so, I rewrite to codes which do not uses `edu.umd.cs.findbugs.visitclass.Constants2`.
  but, I dont know why spotbugs have `edu.umd.cs.findbugs.visitclass.Constants2`...
 `edu.umd.cs.findbugs.visitclass.Constants2` is able to be removed?
 
----


KengoTODA:
I prefer using master branch for breaking changes like this.

iloveeclipse:
The change should be binary compatible, because compiler hard-code used constants. 
So if it will break, only the next *compilation* - and here I don't see a problem, because if you compile you can also change your code.

wreulicke:
base branch is changed to `release-3.1`.

wreulicke:
@iloveeclipse please check, again.

iloveeclipse:
Thanks.

PR #579, opened 2018-03-09
Add test to trigger 'return value ignored' on Preconditions.checkNotNull

ewirch:
- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

See issue #578.

PR #653, opened 2018-06-07
can ignore check not null return value

KengoTODA:
This PR is based on #579 made by @ewirch. I created a new branch in this repository, because I cannot push my code to his repository.

This fix treats `Preconditions.checkNotNull` as `CHECK_RETURN_VALUE_IGNORE` so SpotBugs will not alert even when code ignores returned value from it.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullCanIgnoreReturnValueTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullCanIgnoreReturnValueTest.java
@@ -1,0 +1,36 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.BugCollection;
+import edu.umd.cs.findbugs.test.SpotBugsRule;
+import org.junit.Before;
+import org.junit.Rule;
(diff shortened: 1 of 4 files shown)


KengoTODA:
@ThrawnCA please review. I think we can ignore sonarqube report, that is not related with this change.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullCanIgnoreReturnValueTest.java
```diff
@@ -0,0 +1,36 @@
...
+        String property = System.getProperty("AUX_CLASSPATH");
+        assumeThat(property, notNullValue());
+        for (String path : property.split(",")) {
+            spotbugs.addAuxClasspathEntry(Paths.get(path));
+        }
+    }
+
+    @Test
+    public void testDoNotWarnOnCanIgnoreReturnValue() {
+        BugCollection bugCollection = spotbugs.performAnalysis(Paths.get("../spotbugsTestCases/build/classes/java/main/bugPatterns/RV_RETURN_VALUE_IGNORED_Guava_Preconditions.class"));
```
ThrawnCA: Would be nice if we had some better way of locating test cases, rather than this relative path.

PR #662, opened 2018-06-12
Keep IO.close(Closeable) that was deleted by 3.1.4

### LINKED ISSUE #661: Need both of IO.close(Closeable) and IO.close(AutoCloseable)
5f8879483e685b6f0db816486da10fa97b4b3f6f removes `IO.close(Closeable)` so some user codes face problem. For example, sonar-findbugs cannot pass unit test:

```
-------------------------------------------------------------------------------
Test set: org.sonar.plugins.findbugs.FindbugsExecutorTest
-------------------------------------------------------------------------------
Tests run: 4, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 2.338 s <<< FAIL
URE! - in org.sonar.plugins.findbugs.FindbugsExecutorTest
canGenerateXMLReport(org.sonar.plugins.findbugs.FindbugsExecutorTest)  Time elap
sed: 1.477 s  <<< ERROR!
java.lang.IllegalStateException: Can not execute Findbugs
        at org.sonar.plugins.findbugs.FindbugsExecutorTest.canGenerateXMLReport(FindbugsExecutorTest.java:75)
Caused by: java.util.concurrent.ExecutionException: java.lang.NoSuchMethodError: edu.umd.cs.findbugs.io.IO.close(Ljava/io/Closeable;)V
        at org.sonar.plugins.findbugs.FindbugsExecutorTest.canGenerateXMLReport(FindbugsExecutorTest.java:75)
Caused by: java.lang.NoSuchMethodError: edu.umd.cs.findbugs.io.IO.close(Ljava/io/Closeable;)V

canGenerateXMLReportWithCustomConfidence(org.sonar.plugins.findbugs.FindbugsExecutorTest)  Time elapsed: 0.624 s  <<< ERROR!
java.lang.IllegalStateException: Can not execute Findbugs
        at org.sonar.plugins.findbugs.FindbugsExecutorTest.canGenerateXMLReportWithCustomConfidence(FindbugsExecutorTest.java:92)
Caused by: java.util.concurrent.ExecutionException: java.lang.NoClassDefFoundError: Could not initialize class com.h3xstream.findsecbugs.taintanalysis.TaintAnalysis
        at org.sonar.plugins.findbugs.FindbugsExecutorTest.canGenerateXMLReportWithCustomConfidence(FindbugsExecutorTest.java:92)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.h3xstream.findsecbugs.taintanalysis.TaintAnalysis
```

KengoTODA:
This PR fixes #661, that makes 3.1.3 non backward compatible.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
@@ -188,15 +188,7 @@ public class IO {
      *
      */
     public static void close(@CheckForNull Closeable c) {
-        if (c == null) {
-            return;
-        }
-
-        try {
-            c.close();
-        } catch (Exception e) {
-            // Ignore
-        }
+        close((AutoCloseable) c);
     }
 
     /**
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
@@ -35,6 +35,7 @@ package edu.umd.cs.findbugs.io;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -167,10 +168,26 @@ public class IO {
     }
 
     /**
-     * Close given InputStream, ignoring any resulting exception.
+     * Close given AutoCloseable instance, ignoring any resulting exception.
      *
      */
     public static void close(@CheckForNull AutoCloseable c) {
+        if (c == null) {
+            return;
+        }
+
+        try {
+            c.close();
+        } catch (Exception e) {
+            // Ignore
+        }
+    }
+
+    /**
+     * Close given Closeable instance, ignoring any resulting exception.
+     *
+     */
+    public static void close(@CheckForNull Closeable c) {
         if (c == null) {
             return;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2018-??-??
 
-## 3.1.4 - 2018-06-11
+### Fixed
+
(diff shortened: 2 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
```diff
@@ -167,7 +168,7 @@ public static long copy(Reader in, Writer out, long maxChars)
     }
 
     /**
-     * Close given InputStream, ignoring any resulting exception.
+     * Close given AutoCloseable instance, ignoring any resulting exception.
```
ThrawnCA: It doesn't actually have to be `AutoCloseable`, right? Just `Closeable`.

spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
```diff
@@ -182,6 +183,22 @@ public static void close(@CheckForNull AutoCloseable c) {
         }
     }
 
+    /**
+     * Close given Closeable instance, ignoring any resulting exception.
+     *
+     */
+    public static void close(@CheckForNull Closeable c) {
```
iloveeclipse: This could be written as `    public static void close(@CheckForNull Closeable c) {
        close((AutoCloseable) c);
    }`

spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
KengoTODA: `Closeable` is handled in 190 line.

Problem is that, I replaced `close(Closeable)` with `close(AutoCloseable)` because I thought that it keeps backward compatibility. Yes it keeps compatibility for compiler, but it breaks compatibility for compiled class files; Method invoker expect existence of `close(Closeable)` but there is no method so it throws Exception. See #661 and linked issues for detail.

spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
KengoTODA: possible, I will add a commit to fix.

PR #666, opened 2018-06-14
Improve deployment

KengoTODA:
Automatic deployment at Travis CI could be failed if `spotbugs:uploadArtifact` spends more than 10 minutes, due to limitation of Travis CI. To avoid this, we need to use `travis_wait` script.

And currently SNAPSHOT version from release-3.1 isn't deployed to Sonatype SNAPSHOT repository, it's better to deploy.

* https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,6 +103,14 @@ deploy:
     # https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
     script: travis_wait ./gradlew uploadArchives -PossrhUsername="$SONATYPE_USERNAME" -PossrhPassword="$SONATYPE_PASSWORD"
     on:
+      branch: release-3.1
+      jdk: oraclejdk8
+  - provider: script
+    skip_cleanup: true
+    # wait until uploading task finishes, because it may cost more than 10 minutes without any output
+    # https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
+    script: travis_wait ./gradlew uploadArchives -PossrhUsername="$SONATYPE_USERNAME" -PossrhPassword="$SONATYPE_PASSWORD"
+    on:
       branch: master
       jdk: oraclejdk8
   - provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,13 +99,17 @@ deploy:
       jdk: oraclejdk8
   - provider: script
     skip_cleanup: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -99,13 +99,25 @@ deploy:
       jdk: oraclejdk8
   - provider: script
     skip_cleanup: true
-    script: ./gradlew uploadArchives -PossrhUsername="$SONATYPE_USERNAME" -PossrhPassword="$SONATYPE_PASSWORD"
+    # wait until uploading task finishes, because it may cost more than 10 minutes without any output
+    # https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
ThrawnCA: Page anchor should start with lowercase 'b'

KengoTODA: That page actually uses upper case as `id`, you can have a try:

* https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

ThrawnCA:
I did try. Uppercase didn't work for me, lowercase did.

ThrawnCA:
I did try. Uppercase didn't work for me, lowercase did.

KengoTODA:
really... I guess it is your browser's problem, at least:

1. Google Chrome works with uppercase
2. Travis CI uses upper B in their site in both of `id` and `href` attributes

ThrawnCA:
Hmm...looks like it's specific to Firefox; Chrome works. However, when I look at the Travis page source, the anchor id uses lowercase.

KengoTODA:
I also checked that raw HTML uses lowercase, it's uppercase just in Chrome console.

I want to follow your guide, but one remaining problem is that the link with lowercase doesn't work in Chrome. But the link with uppercase works with Chrome, and Firefox 60.0.2 on my Ubuntu 18.04 LTS.

* https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Could you let me merge with uppercase, to let our comment works in major browsers? Actually it's the way used in [Travis CI's official document](https://docs.travis-ci.com/user/common-build-problems/)

.travis.yml
ThrawnCA: Well, the actual [standard](https://www.w3.org/TR/WD-html40-970708/struct/links.html) says that anchors are supposed to be case-insensitive, so I'll approve for merge.

PR #668, opened 2018-06-14
Rewrite codes using deprecated apis.

wreulicke:
All changes are because of deprecated API.


spotbugs/src/gui/edu/umd/cs/findbugs/gui2/PreferencesFrame.java
```diff
@@ -637,12 +637,11 @@ private JPanel createFilterPane() {
         gbc.insets = new Insets(5, 0, 0, 0);
         filterPanel.add(removeButton, gbc);
         removeButton.addActionListener(evt -> {
-            Object[] selected = filterCheckBoxList.getSelectedValues();
-            if (selected.length == 0) {
+            List<MatchBox> selected = filterCheckBoxList.getSelectedValuesList();
+            if (selected.size() == 0) {
```
ThrawnCA: Now that we have a collection, it's more efficient to call `isEmpty`.

wreulicke: Rewrited

PR #672, opened 2018-06-19
Avoid NPE in AnalysisRunner

KengoTODA:
The `cause` field in `AbstractBugReporter.Error` is nullable, so we cannot use suppressed error that makes log readable. Instead, use `reportQueuedErrors()` method that supports `AbstractBugReporter.Error` without `cause`.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
@@ -22,7 +22,6 @@ package edu.umd.cs.findbugs;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
-import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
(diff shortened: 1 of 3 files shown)


KengoTODA:
About report from SonarCloud: We can ignore the report for L122, but L107 should be fixed, I'll fix it in another PR.

https://sonarcloud.io/code?id=com.github.spotbugs.spotbugs&selected=com.github.spotbugs.spotbugs%3Atest-harness-core%3Asrc%2Fmain%2Fjava%2Fedu%2Fumd%2Fcs%2Ffindbugs%2Ftest%2FAnalysisRunner.java&branch=release-3.1

test-harness-core/src/main/java/edu/umd/cs/findbugs/test/AnalysisRunner.java
```diff
@@ -110,8 +110,7 @@ public BugCollectionBugReporter run(Path... files) {
             if (!bugReporter.getQueuedErrors().isEmpty()) {
                 AssertionError assertionError = new AssertionError(
```
ThrawnCA: No need for a local variable any more, just create and throw it after reporting bugs.

spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
```diff
@@ -35,6 +34,7 @@
 
 import org.dom4j.DocumentException;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
```
ThrawnCA: FindBugs `@Nullable` annotation doesn't mean anything, should use `@CheckForNull`

KengoTODA: Do we need to use it even for arguments? We accept null value then I think `@Nullable` is enough and intuitive.

Issue #683, opened 2018-07-04
Gradle build fails if project path contains a space


marzn:
`./gradlew build` fails with the following:

```
:spotbugs-tests:test

edu.umd.cs.findbugs.ba.Issue429Test > test FAILED
    java.lang.AssertionError: edu.umd.cs.findbugs.PluginException: Failed loading manifest for plugin jar: file:/home/user/software%2520projects/spotbugs-repo/spotbugs/build/libs/spotbugs.jar

        Caused by:
        edu.umd.cs.findbugs.PluginException: Failed loading manifest for plugin jar: file:/home/user/software%2520projects/spotbugs-repo/spotbugs/build/libs/spotbugs.jar

            Caused by:
            java.io.FileNotFoundException: /home/user/software%20projects/spotbugs-repo/spotbugs/build/libs/spotbugs.jar (No such file or directory)
```

SpotBugs seems to double encode the path. `%2520` is contained in the path in the first two exceptions, and `%20` in the last one.

Could be tracked down to https://github.com/spotbugs/spotbugs/blob/0bb77b347784b3b8288fc13894f0cecced6a993f/test-harness-core/src/main/java/edu/umd/cs/findbugs/test/AnalysisRunner.java#L154
which should be `return new File(url.toURI());`

PR #684, opened 2018-07-04
Fix issue 683

marzn:
This fixes issue #683 

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 
 * Potential NPE in test-harness-core ([#671](https://github.com/spotbugs/spotbugs/issues/671))
-* Support project path with spaces ([#683](https://github.com/spotbugs/spotbugs/issues/683))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 
 * Potential NPE in test-harness-core ([#671](https://github.com/spotbugs/spotbugs/issues/671))
+* Support project path with spaces ([#683](https://github.com/spotbugs/spotbugs/issues/683))
```
KengoTODA: Good. Could you add `in test-harness-core`, then reader can easily understand this fix is for that module.

KengoTODA:
If possible, please try to create an unit test case to reproduce problem.

marzn:
@KengoTODA because this is related to the absolute path of a project, i think this would be more a integration test. Of course we can check if `AnalysisRunner.createTempJar()` is producing the correct output, but this bug can occur in various methods, if the path is not decoded or encoded twice.

marzn:
Anyway, I had a look at the `AnalysisRunner`. Only for the `createTempJar()` unittest it would be a lot of code to setup the test. On the other hand this does not fulfill the requirements in my eyes. So I stopped working on the unittest.

PR #696, opened 2018-07-13
Issue a bug for BigDecimal(double) when the argument is not a constant

cpfeiffer:
The BigDecimal(double) detector only reported an issue when the argument
to the constructor was a constant whose value would cause rounding errors.
   
Now it also reports an issue if the value is *not* a constant.

- [x ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Processing of "J" (long value constants) was not processed in `OpcodeStack.Item(OpcodeStack.Item, String)`
 * Processing of "Z" (boolean value constants) was not processed in `OpcodeStack.Item(OpcodeStack.Item, String)`
 * Processing of Box classes like `java.lang.Integer` was not processed in `OpcodeStack.Item(OpcodeStack.Item, String)`
+* DMI_BIGDECIMAL_CONSTRUCTED_FROM_DOUBLE was only reported when the double argument was a constant
 
 ## 3.1.5 - 2018-06-15
 
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue696Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue696Test.java
@@ -1,0 +1,43 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2018, cpfeiffer
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+package edu.umd.cs.findbugs.detect;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+
+/**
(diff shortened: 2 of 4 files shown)


iloveeclipse:
@cpfeiffer: thanks for contribution, can you please add new *regression* test, see for example 3dc2a8d4baedabdb42da8f7f7709320c414ea552

cpfeiffer:
OK, but not right now, I'm in a hurry.

cpfeiffer:
I separated the test and rebased to current release-3.1 branch.

wrprice:
While I understand how this may catch some buggy cases that are out there, doesn't this also risk creating a bunch of false-positive noise?  Just because a value came as a non-constant `double` doesn't mean that's not _exactly_ what I want.  Maybe it came from a 3rd party library and I want to create a BigDecimal from it and do operations without losing any more accuracy.  That's _not a bug_, not unreasonable, and I shouldn't have to suppress it.  (And if you accept this change, I'd argue that the priority should be LOW, not normal.)

If you want to be smarter about non-constant values passed to the constructor, I think you'd have to be cautious and _only flag as bugs those things that are highly-likely to be non-exact_, which is a hard task.  At least with a constant value, Spotbugs tries to do it by looking at the "exact" string versus the "rounded" string of the constant value.  With a value that can come from nearly anywhere else, all bets are off.

This change presumes too much that the non-constant source of the `decimal` value is always under the control of the author using the `BigDecimal(double)` constructor.  And it also presumes that using the `BigDecimal(double)` constructor is always a likely bug.  It's not.

cpfeiffer:
It really depends on the case. It *may* be exactly what you want, *if* the double value really represents the accurate value that you're dealing with.

It may also be completely wrong to start with a double value at all and creating a BigDecimal from it would give a false sense of safety.

wrprice:
And that's my point, it _may_ be a bug, but it _may_ not be.  Spotbugs shouldn't flag everything that _could_ be a bug, its value is finding things that are _very likely_ to be bugs.  We're just disagreeing over the relative _likelihood_ in this case.  If that constructor were sufficiently indicative of buggy code, I would also hope that the JDK developers would mark it as `@Deprecated` like has been done to other API "mistakes"t in the past (e.g. Thread, java.util.Date, etc.).  They have not, and if I were to guess why it's because there are plenty of legitimate, non-buggy use cases.

Somewhat similarly, object reference equality using `==` _may_ be a bug, but it's also used sufficiently in cases where that's exactly what's desired that it shouldn't be flagged "just in case".  The [RC_REF_COMPARISON detector](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html?highlight=equals#rc-suspicious-reference-comparison-rc-ref-comparison) is cautious about flagging uses of `==`, for example, limiting to a suspicious set of `java.lang` wrapper classes where it's almost always wrong to do so.

I think this will introduce noise, and noise hurts adoption and buy-in for the tool.  It's also not hard to use an IDE and "find usages" in a project if a project maintainer wants to periodically ensure this constructor is never used in the project's code -- because they have _specialized knowledge_ that it's not right for _that project_.  I think this proposed change is too broad and should be opt-in, but I'll not harp on it any more.

cpfeiffer:
Thank you for your thoughtful comments!

If people think this should be low priority, so be it, I certainly don't demand that this should be normal priority.

FWIW, the [BigDecimal(double) javadoc](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#BigDecimal-double-) says this:
> The results of this constructor can be somewhat unpredictable.

and
> The String constructor, on the other hand, is perfectly predictable: writing new BigDecimal("0.1") creates a BigDecimal which is exactly equal to 0.1, as one would expect. Therefore, it is generally recommended that the String constructor be used in preference to this one.

So while it's not quite a "don't use this constructor" message, it goes in the same direction.

That said, I'm wondering if it wouldn't be possible to determine more values via data-flow analysis, when they are not passed as constants right into the constructor. Something along the lines of https://stackoverflow.com/questions/17799855/access-variable-constant-values-in-method-call

Does spotbugs do anything like this? (It's too late for me to check this myself, right now.)

wrprice:
Agreed, that the string constructor is the better option if you have the choice, and the javadoc speaks to that.  Particularly if you're initializing an instance from a constant (already checked) that's the proper guidance.  DFA would help catch "indirect" constants like in your test case example, and that _would_ be worthwhile.  (I don't know enough about Spotbugs' internals to speak to its feasibility.)

Also from the javadoc:

> When a double must be used as a source for a BigDecimal, note that this constructor provides an exact conversion; it does not give the same result as converting the double to a String using the Double.toString(double) method and then using the BigDecimal(String) constructor. To get that result, use the static valueOf(double) method.

So if you flag this as constructor a bug, there is no other mechanism provided by BigDecimal for preserving the _exact_ `double` value without introducing more (rounding) error.  The only guidance to developers that want the behavior will be to add suppressions at every location, introduce otherwise-unnecessary helper methods so that there's only "one" place to suppress, or ignore/disable Spotbugs.

Actually, as written, I can "fix" the problem by using the `BigDecimal(double, MathContext)` constructor. Was that intentional?  Maybe that's a compromise, to allow that to be _not-a-bug_ by showing intent beyond a "simple" conversion and make that the guidance to developers.

KengoTODA:

PR #699, opened 2018-07-17
Release 3.1.6

KengoTODA:
I think it's good timing to release this, we've spend one month to fix issues.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,8 +17,8 @@ import os
 
 html_context = {
   'version' : '3.1',
-  'full_version' : '3.1.5',
-  'maven_plugin_version' : '3.1.3.1',
+  'full_version' : '3.1.6',
+  'maven_plugin_version' : '3.1.5',
   'gradle_plugin_version' : '1.6.2'
 }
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -17,8 +17,8 @@
 
 html_context = {
   'version' : '3.1',
-  'full_version' : '3.1.5',
-  'maven_plugin_version' : '3.1.3.1',
+  'full_version' : '3.1.6',
+  'maven_plugin_version' : '3.1.5',
```
ThrawnCA: Maven plugin version 3.1.5 already exists. Are we releasing 3.1.6?

KengoTODA: We use the existing latest version at here. After we ship Maven plugin v3.1.6, we can change this part.

KengoTODA:
Currently the build at Travis CI is unstable, due to 404 error at download.oracle.com.
I've confirmed that 8b51a3b1fe9cc0f676e0909252502042d2020cbf cannot be the solution.

Instead of automatic build at Travis CI, I'll release SpotBugs from my PC. It uses the same gpg key to sign.

KengoTODA:
refer https://github.com/travis-ci/travis-ci/issues/9891

KengoTODA:
Travis problem has been fixed. Only Sonatype deployment failed so I've deployed from my clean environment manually.

PR #700, opened 2018-07-18
Add management for source filter using full source path (if available).

Irian81:
Initial PR for review to add support for full source paths on source filter (when available).

The primary bit I'm not sure about in this PR is the code in SourceLineAnnotation that retrieves an instance of the SourceFinder object.

Will also add a PR for the maven bit, these should probably be treated as a whole, and maybe support might be required also for other plugins.
----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
@@ -51,6 +51,7 @@ import edu.umd.cs.findbugs.internalAnnotations.DottedClassName;
 import edu.umd.cs.findbugs.visitclass.PreorderVisitor;
 import edu.umd.cs.findbugs.xml.XMLAttributeList;
 import edu.umd.cs.findbugs.xml.XMLOutput;
+import javax.annotation.CheckReturnValue;
 
 /**
  * A BugAnnotation that records a range of source lines in a class.
@@ -944,6 +945,13 @@ public class SourceLineAnnotation implements BugAnnotation {
         return sourcePath;
     }
 
+    /**
+     * Returns the complete path of the source file the analyzed class has been generated from,
+     * if this information is available to the current AnalysisContext, otherwise falls back to getSourcePath().
+     *
+     * @return the complete path of the source file the analyzed class has been generated from,
+     * otherwise falls back to getSourcePath().
+     */
     public String getRealSourcePath() {
         if (isSourceFileKnown()) {
             SourceFinder sourceFinder = getSourceFinder();
@@ -962,7 +970,7 @@ public class SourceLineAnnotation implements BugAnnotation {
         return getSourcePath();
     }
 
-    private SourceFinder getSourceFinder() {
+    private @CheckReturnValue SourceFinder getSourceFinder() {
         Project project = myProject.get();
         if (project != null) {
             return project.getSourceFinder();
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
@@ -946,30 +946,30 @@ public class SourceLineAnnotation implements BugAnnotation {
 
     public String getRealSourcePath() {
         if (isSourceFileKnown()) {
-            Project project = myProject.get();
-            if (project != null) {
+            SourceFinder sourceFinder = getSourceFinder();
+            if (sourceFinder != null)
+            {
                 try {
-                    SourceFinder mySourceFinder = project.getSourceFinder();
-                    return new File(mySourceFinder.findSourceFile(this).getFullFileName()).getCanonicalPath();
-
+                    return new File(sourceFinder.findSourceFile(this).getFullFileName()).getCanonicalPath();
                 } catch (IOException e) {
-                    assert true;
+                    AnalysisContext.logError("Error resolving Real SourcePath (only relative source path will be available) ", e);
                 }
             }
-            else {
-              AnalysisContext currentAnalysisContext = AnalysisContext.currentAnalysisContext();
-              if (currentAnalysisContext != null) {
-                try {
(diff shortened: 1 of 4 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,34 @@ public String getSourcePath() {
         return sourcePath;
     }
 
+    public String getRealSourcePath() {
+        if (isSourceFileKnown()) {
+            Project project = myProject.get();
+            if (project != null) {
```
ThrawnCA: "if not...else" is a double negative. Can you rearrange this?

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,34 @@ public String getSourcePath() {
...
+                try {
+                    SourceFinder mySourceFinder = project.getSourceFinder();
+                    return new File(mySourceFinder.findSourceFile(this).getFullFileName()).getCanonicalPath();
+
+                } catch (IOException e) {
+                    assert true;
```
ThrawnCA: Asserting `true` is pointless. A comment would be better.

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,34 @@ public String getSourcePath() {
...
+                }
+            }
+            else {
+              AnalysisContext currentAnalysisContext = AnalysisContext.currentAnalysisContext();
+              if (currentAnalysisContext != null) {
+                try {
+                  SourceFinder sourceFinder = currentAnalysisContext.getSourceFinder();
+                  return new File(sourceFinder.findSourceFile(this).getFullFileName()).getCanonicalPath();
+                } catch (IOException e) {
+                  e.printStackTrace(System.out);
```
ThrawnCA: `e.printStackTrace` is generally frowned upon. Can you use a `java.util.logging.Logger` instead?

Also, wrong indentation.

Irian81:
Fixed the issues you pointed out.

Let me know if you require anything else.

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,34 @@ public String getSourcePath() {
...
+            SourceFinder sourceFinder = getSourceFinder();
+            if (sourceFinder != null)
+            {
+                try {
+                    return new File(sourceFinder.findSourceFile(this).getFullFileName()).getCanonicalPath();
+                } catch (IOException e) {
+                    AnalysisContext.logError("Error resolving Real SourcePath (only relative source path will be available) ", e);
+                }
+            }
+            AnalysisContext.logError("No SourceFinder found (only relative source path will be available) ");
```
ThrawnCA: If a SourceFinder exists, but gets an error, the logic will fall through and still print this message.

It would be better to structure it as

    if (sourceFinder == null) {
        AnalysisContext.logError("No SourceFinder found (only relative source path will be available) ");
    } else {
        ...

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,36 @@ public String getSourcePath() {
...
+                }
+            }
+            else {
+                AnalysisContext.logError("No SourceFinder found (only relative source path will be available) ");
+            }
+        }
+        return getSourcePath();
+    }
+
+    private SourceFinder getSourceFinder() {
```
KengoTODA: Please annotate this method with `@CheckReturnValue`, to tell that returned value is nullable.

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +944,36 @@ public String getSourcePath() {
         return sourcePath;
     }
 
+    public String getRealSourcePath() {
```
KengoTODA: consider to add javadoc, to tell responsibility of this public method.

Irian81:
Added the requested annotation and javadoc.
Also added changelog.md that was left out.

Irian81:
any news on this PR status??

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
```diff
@@ -57,6 +57,18 @@ public boolean match(BugInstance bugInstance) {
         if (DEBUG) {
             System.out.println("Matching " + bugFileName + " with " + fileName + ", result = " + result);
         }
+        if (!result
+                && bugInstance.getPrimarySourceLineAnnotation().isSourceFileKnown()) {
+          String bugSourcePath = bugInstance.getPrimarySourceLineAnnotation().getRealSourcePath();
+
+          if(bugSourcePath == null || bugSourcePath.isEmpty()){
+            return result;
```
ThrawnCA: May as well just `return false` here.

Irian81:
I think there is still a requested change not marked as solved but it's really one of the first requests that were fixed.

Irian81:
Added a basic unit test to the existing SourceMatcherTest class with tests for the full path case.

Let me know if it's enough.

KengoTODA:

PhilMFischer:
@Irian81 Can you please create a new pull request on this one as mentioned in the comment of @KengoTODA ?

Irian81:
> 
> 
> @Irian81 Can you please create a new pull request on this one as mentioned in the comment of @KengoTODA ?

Unfortunately I'm so swamped right now that I've got no time at all for this; will definitely get back to it but not soon, feel free to create the PR yourself if it's really urgent for you.

PhilMFischer:
@Irian81 Thanks I will fork your work and set a  new pull request.

PR #710, opened 2018-07-23
Make sure BugInstance hash is consistent

gaddam1987:
Why:
   Instance hash is used for excluding bugs from Baseline report. The hash creation is strongly dependent on the order in which  bug annotation are added to BugInstance. Some plugins like find-sec-bugs are adding bug annotation  in different Order for same BugInstance on different runs as they use HashSet to store the order before populating to BugInstance. The changes were made To make hash creation independent of the order in which bug annotiations  added to BugInstance.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ExcludingHashesBugReporter.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ExcludingHashesBugReporter.java
@@ -62,8 +62,8 @@ public class ExcludingHashesBugReporter extends DelegatingBugReporter {
 
     @Override
     public void reportBug(@Nonnull BugInstance bugInstance) {
-        String instanceHash = bugInstance.getInstanceHash();
-        if (!excludedHashes.contains(instanceHash)) {
+        if (!excludedHashes.contains(bugInstance.getBrokenInstanceHashForBackwardCompatibility())
+                && !excludedHashes.contains(bugInstance.getInstanceHash())) {
             getDelegate().reportBug(bugInstance);
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2018-??-??
+* Make sure BugInstance hash is consistent
 
 ## 3.1.6 - 2018-07-18
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
@@ -552,7 +552,8 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
 
     private String getInstanceKeyNew() {
         StringBuilder buf = new StringBuilder(type);
-        for (BugAnnotation annotation : annotationList) {
+        Set<BugAnnotation> sortedAnnotationList = new TreeSet<>(annotationList);
+        for (BugAnnotation annotation : sortedAnnotationList) {
             if (annotation.isSignificant() || annotation instanceof IntAnnotation
                     || annotation instanceof LocalVariableAnnotation) {
                 buf.append(":");
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
@@ -34,6 +34,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.annotation.meta.When;
 
+import edu.umd.cs.findbugs.ba.bcp.LocalVariable;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.JavaClass;
 import org.apache.bcel.classfile.Method;
@@ -150,8 +151,13 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
 
     private static Set<String> missingBugTypes = Collections.synchronizedSet(new HashSet<String>());
 
+    public String getBrokenInstanceHashForBackwardCompatibility() {
+        return createHash(getInstanceKeyNew());
+    }
+
     public static class NoSuchBugPattern extends IllegalArgumentException {
         public final String type;
+
         public NoSuchBugPattern(String type) {
             super("Can't find definition of bug type " + type);
             this.type = type;
@@ -161,10 +167,8 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
     /**
      * Constructor.
      *
-     * @param type
-     *            the bug type
-     * @param priority
-     *            the bug priority
+     * @param type     the bug type
+     * @param priority the bug priority
      */
     public BugInstance(String type, int priority) {
         this.type = type.intern();
@@ -175,7 +179,7 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
 
         BugPattern p = DetectorFactoryCollection.instance().lookupBugPattern(type);
         if (p == null) {
-            if ( missingBugTypes.add(type)) {
+            if (missingBugTypes.add(type)) {
                 String msg = "Can't find definition of bug type " + type;
                 AnalysisContext.logError(msg, new NoSuchBugPattern(type));
             }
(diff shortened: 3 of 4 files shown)


ThrawnCA:
Would this potentially break existing exclusions?

gaddam1987:
Yes it is breaking the existing exclusions. 

Below is one of the bug instance which was part of the exclusions. 

`<BugInstance instanceOccurrenceMax="0" type="XSS_SERVLET" priority="2" category="SECURITY" abbrev="SECXSS2" rank="12" cweid="79" instanceHash="1a95e3133fb2f253b989fc0991db22ca" instanceOccurrenceNum="0">
   <ShortMessage>Potential XSS in Servlet</ShortMessage>
   <LongMessage>This use of javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V could be vulnerable to XSS</LongMessage>
   <Class primary="true" classname="somepackage.SomeServlet">
      <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="306" start="56">
         <Message>At SomeServlet.java:[lines 56-306]</Message>
      </SourceLine>
      <Message>In class somepackage.SomeServlet</Message>
   </Class>
   <Method primary="true" classname="somepackage.SomeServlet" name="processReq" signature="(Ljavax/servlet/http/HttpServletRequest;Ljavax/servlet/http/HttpServletResponse;)V" isStatic="false">
      <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage.SomeServlet.java" end="173" start="95" startBytecode="0" endBytecode="1575" />
      <Message>In method somepackage.SomeServlet.processReq(HttpServletRequest, HttpServletResponse)</Message>
   </Method>
   <SourceLine primary="true" classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage.SomeServlet.java" end="156" start="156" startBytecode="592" endBytecode="592">
      <Message>At SomeServlet.java:[line 156]</Message>
   </SourceLine>
   <String role="Sink method" value="javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V">
      <Message>Sink method javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V</Message>
   </String>
   <String role="Sink parameter" value="0">
      <Message>Sink parameter 0</Message>
   </String>
   <String role="Unknown source" value="no/bbs/bankid/common/exceptions/BusinessException.getErrorCode()Ljava/lang/String;">
      <Message>Unknown source no/bbs/bankid/common/exceptions/BusinessException.getErrorCode()Ljava/lang/String;</Message>
   </String>
   <String role="Unknown source" value="java/lang/StringBuffer.toString()Ljava/lang/String;">
      <Message>Unknown source java/lang/StringBuffer.toString()Ljava/lang/String;</Message>
   </String>
   <String role="Unknown source" value="java/lang/StringBuffer.append(Ljava/lang/String;)Ljava/lang/StringBuffer;">
      <Message>Unknown source java/lang/StringBuffer.append(Ljava/lang/String;)Ljava/lang/StringBuffer;</Message>
   </String>
   <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage.SomeServlet.java" end="155" start="155" startBytecode="581" endBytecode="581">
      <Message>At SomeServlet.java:[line 155]</Message>
   </SourceLine>
   <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage.SomeServlet.java" end="156" start="156" startBytecode="589" endBytecode="589">
      <Message>At SomeServlet.java:[line 156]</Message>
   </SourceLine>
</BugInstance>`

After couple of runs our build server starts reporting some failures and when we looked into the issue we see that spot bug is reporting same issue which we asked to ignore it. 

**Failed BugInstance after some successful runs**

`<BugInstance instanceOccurrenceMax="0" type="XSS_SERVLET" priority="2" category="SECURITY" abbrev="SECXSS2" rank="12" cweid="79" instanceHash="9393387a4548bfe7f647f46eef0e2cc5" instanceOccurrenceNum="0">
   <ShortMessage>Potential XSS in Servlet</ShortMessage>
   <LongMessage>This use of javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V could be vulnerable to XSS</LongMessage>
   <Class primary="true" classname="somepackage.SomeServlet">
      <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="306" start="56">
         <Message>At SomeServlet.java:[lines 56-306]</Message>
      </SourceLine>
      <Message>In class somepackage.SomeServlet</Message>
   </Class>
   <Method primary="true" classname="somepackage.SomeServlet" name="processReq" signature="(Ljavax/servlet/http/HttpServletRequest;Ljavax/servlet/http/HttpServletResponse;)V" isStatic="false">
      <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="173" start="95" startBytecode="0" endBytecode="1575" />
      <Message>In method somepackage.SomeServlet.processReq(HttpServletRequest, HttpServletResponse)</Message>
   </Method>
   <SourceLine primary="true" classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="156" start="156" startBytecode="592" endBytecode="592">
      <Message>At SomeServlet.java:[line 156]</Message>
   </SourceLine>
   <String role="Sink method" value="javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V">
      <Message>Sink method javax/servlet/ServletOutputStream.println(Ljava/lang/String;)V</Message>
   </String>
   <String role="Sink parameter" value="0">
      <Message>Sink parameter 0</Message>
   </String>
   <String role="Unknown source" value="java/lang/StringBuffer.append(Ljava/lang/String;)Ljava/lang/StringBuffer;">
      <Message>Unknown source java/lang/StringBuffer.append(Ljava/lang/String;)Ljava/lang/StringBuffer;</Message>
   </String>
   <String role="Unknown source" value="java/lang/StringBuffer.toString()Ljava/lang/String;">
      <Message>Unknown source java/lang/StringBuffer.toString()Ljava/lang/String;</Message>
   </String>
   <String role="Unknown source" value="no/bbs/bankid/common/exceptions/BusinessException.getErrorCode()Ljava/lang/String;">
      <Message>Unknown source no/bbs/bankid/common/exceptions/BusinessException.getErrorCode()Ljava/lang/String;</Message>
   </String>
   <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="155" start="155" startBytecode="581" endBytecode="581">
      <Message>At SomeServlet.java:[line 155]</Message>
   </SourceLine>
   <SourceLine classname="somepackage.SomeServlet" sourcefile="SomeServlet.java" sourcepath="somepackage/SomeServlet.java" end="156" start="156" startBytecode="589" endBytecode="589">
      <Message>At SomeServlet.java:[line 156]</Message>
   </SourceLine>
</BugInstance>`


And the difference is string annotation with role UnknownSource are in different order and it has created different hash

gaddam1987:
It will break existing exclusions the changes made by be are backward compatible. I Need to think how to avoid it if you have any suggestions i am happy to work on that

gaddam1987:
I did some analysis today, but i couldn't  find any solution to make it backward compatible.

wrprice:
First time I'm looking at the BugInstance class, but it appears to support an `oldInstanceHash` optional property.  It's set from SAXBugCollectionHandler from _either_ the `instanceHash` or `oldInstanceHash` XML attributes, in that order of preference.  I also couldn't see where SAXB.C.H. sets the actual instance hash, which would make sense as it should compute the current hash from whatever is added to the instance from the input XML.

So the object is capable of having both the current and the old hash value -- apparently to detect whether something has changed, as a quick Github search suggests the only external use of old vs. current is with the `isInstanceHashConsistent()` method called by  `edu.umd.cs.findbugs.workflow.Filter`.  I didn't notice if the old hash was exposed outside the instance class, but I don't think so.  I haven't done a dive into the Filter class to fully understand its purpose, yet, and it may turn out that this property can't be (ab)used for this purpose without breaking something else. In that case, a new field could be added if it's worth the effort for compatibility.

TL;DR -- one way to backwards-compatibly change the hash is to be able to (internally) compute the old way and the new way and accept/match either value for a given instance.  The risk of collisions between two hashes old/new is probably not significant, and then the cost of computing both hashes can be removed in a backwards-incompatible release, e.g. 4.0.0.

gaddam1987:
Hi ,
I added a check to verify the buginstance with oldHash. Generation of old hash only happens when baseline bug file is available for exclusion.

spotbugs/src/main/java/edu/umd/cs/findbugs/ExcludingHashesBugReporter.java
```diff
@@ -62,8 +62,8 @@ public static void addToExcludedInstanceHashes(Set<String> instanceHashesToExclu
 
     @Override
     public void reportBug(@Nonnull BugInstance bugInstance) {
-        String instanceHash = bugInstance.getInstanceHash();
-        if (!excludedHashes.contains(instanceHash)) {
+        if (!excludedHashes.contains(bugInstance.getBrokenInstanceHashForBackwardCompatibility())
+                || !excludedHashes.contains(bugInstance.getInstanceHash())) {
```
ThrawnCA: Shouldn't this be `&&` rather than `||`?

KengoTODA:
I'll close this stale PR. Feel free to reopen it.

PR #713, opened 2018-07-26
Fix wording in message

tobymurray:
"possibly to be incorrect" doesn't make sense, replacign it with "possibly incorrect"

----


iloveeclipse:
Thanks

PR #714, opened 2018-07-30
runMain: Don't print exit code information when -quiet is passed

philipl:
Today, it's impossible to run spotbugs where the exit code is set
and output is fully suppressed. It seems reasonable that -quiet
should suppress the exit code related output, which is not normally
very interesting.

----


KengoTODA:
Wmm GitHub's conflict management is still buggy... It still says "This branch cannot be rebased due to conflicts" even after my commit 72d807e.

Let me create a new PR from the latest `release-3.1` branch.

PR #727, opened 2018-08-08
Close #500 - Add missing visitINVOKEDYNAMIC override in TypeFrameModelingVisitor

### LINKED ISSUE #500: FindUnrelatedTypesInGenericContainer throws IAE on Java 9 ("Not a reference type: <bottom>")
Reduced my original class causing this error down to this MCVE:

```
package com.example;
import java.util.Map;

public class CrashesAnalysis {

    String literalOK(Map<String, String> map) {
        return map.get("const");
    }

    String refOK(Map<String, String> map, String key) {
        return map.get(key);
    }

    String nonConstConcatFails(Map<String, String> map, String key) {
        return map.get(key + "const");
    }
}
```

Throws:

<pre>
Scanning archives (149 / 149)
2 analysis passes to perform
Pass 1: Analyzing classes (338 / 338) - 100% complete
Pass 2: Analyzing classes (138 / 138) - 100% complete
Done with analysis
The following errors occurred during analysis:
  Exception analyzing com.example.CrashesAnalysis using detector edu.umd.cs.findbugs.detect.FindUnrelatedTypesInGenericContainer
    java.lang.IllegalArgumentException: Not a reference type: &lt;bottom&gt;
      At edu.umd.cs.findbugs.ba.generic.GenericUtilities.getTypeCategory(GenericUtilities.java:200)
      At edu.umd.cs.findbugs.detect.FindUnrelatedTypesInGenericContainer.compareTypes(FindUnrelatedTypesInGenericContainer.java:739)
      At edu.umd.cs.findbugs.detect.FindUnrelatedTypesInGenericContainer.analyzeMethod(FindUnrelatedTypesInGenericContainer.java:582)
      At edu.umd.cs.findbugs.detect.FindUnrelatedTypesInGenericContainer.visitClassContext(FindUnrelatedTypesInGenericContainer.java:259)
      At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
      At edu.umd.cs.findbugs.FindBugs2.analyzeApplication(FindBugs2.java:1079)
      At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:280)
      At com.github.spotbugs.internal.spotbugs.SpotBugsExecuter.runSpotbugs(SpotBugsExecuter.java:23)
</pre>

Environment:

<pre>
SpotBugs 3.1.0
SpotBugs Gradle plugin 1.6.0

java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)

------------------------------------------------------------
Gradle 4.2.1
------------------------------------------------------------

Build time:   2017-10-02 15:36:21 UTC
Revision:     a88ebd6be7840c2e59ae4782eb0f27fbe3405ddf

Groovy:       2.4.12
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          9 (Oracle Corporation 9+181)
OS:           Windows 10 10.0 amd64
</pre>

Note: there is no IAE thrown when compiled using JDK 8 (u152).  I strongly expect it relates to JDK 9 `javac` using [`invokedynamic` for string concatenation](http://openjdk.java.net/jeps/280), whereas JDK 8 uses a StringBuilder.  If I replace the concatenation with a single String literal, or the `key` reference, then the error no longer occurs on JDK 9.

wrprice:
Prevents underflowing the stack (reaching `<bottom>`) when encountering
INVOKEDYNAMIC.  Without the override, parent class logic was consuming
a stack frame but not pushing the return type from the INDY call.

Symptom was IllegalArgumentException ("Not a reference type: &lt;bottom&gt;")
from FindUnrelatedTypesInGenericContainer. See issue #500 

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 * Don't print exit code related output if '-quiet' is passed ([#714](https://github.com/spotbugs/spotbugs/pull/714))
+* Don't underflow the stack at INVOKEDYNAMIC when modeling stack frame types ([#500](https://github.com/spotbugs/spotbugs/issues/500))
 
 ### CHANGED
 * ASM_VERSION=ASM7_EXPERIMENTAL by default to support Java 11
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue500Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue500Test.java
@@ -1,0 +1,42 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2018, William R. Price
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+package edu.umd.cs.findbugs.detect;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+
+/**
+ * @author William R. Price
+ */
+public class Issue500Test extends AbstractIntegrationTest {
+
+    @Test
+    public void test() {
+        performAnalysis("lambdas/Issue500.class");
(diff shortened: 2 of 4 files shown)


wrprice:
@ThrawnCA you're technically correct, because the MCVE relies on under-the-hood `javac` changes in Java 9.  Did you mean that merely as an observation, or are you requesting an alternative/improvement?

As written, the test case verifies that there are no bugs found (or exceptions thrown) in the MCVE under _whatever version the build is using_, and that behavior should hold for _all_ Java >= 5.  Nothing in the MCVE _source_ requires Java 9 features.  Remove the new `@Override` in the production code and the test indeed fails on JDK 9.  (I wasn't able to build cleanly, prior to any changes, on JDK 10.  I assume that's expected?)

I suppose I can try to find an explicit lambda or something that would compile down to a similar INDY on Java 8 and hopefully trigger the same problem.  That would be a good additional test case, but I have lambdas and method references all over my production code and this was the only case (so far) where I've hit the bug.

By the way, I did consider putting the example class under `spotbugsTestCases/src/java9/`, but based on the Gradle project configuration those are only compiled in the case...

    if (System.getProperty("java.specification.version") == "9")

Lexical string ordering aside, maybe that was intended to be a `>=` relationship?  Or will these classes (and presumably their tests) stop working in the future when the project's CI no longer includes Java 9?

ThrawnCA:
Well, it would be nice to have a test that works on Java 8 too, but the change looks OK and it at least *has* a test (which many features lack), so I've approved it.

I'm not sure whether the `java9` test cases are supposed to be `>= java 9`, but probably.

wrprice:
Found a way to reproduce on Java 8, albeit with two lines of code that really warrant their own dodgy code detector.  Updated PR and squashed; please review again.

- Introduced second minimal example method into Issue500.java, uses a method reference.
- Above required moving Issue500.java from `ghIssues` package to `lambdas` because only certain packages are allowed to compile for target > 1.7
- Production fix is unchanged from prior rev.

Issue #728, opened 2018-08-09
Support Spotbugs Eclipse Preferences in userPrefs


dlwildschut:
The -userPrefs option for Spotbugs doesn't parse the .epf Preferences produced by the SpotBugs Eclipse Plugin.
The SpotBugs plugin preferences contain the prefix _/instance/com.github.spotbugs.plugin.eclipse/_, which is not removed in the `edu.umd.cs.findbugs.config.UserPreferences.read(@WillClose InputStream in)` method.
Please remove the prefix, which is already being done with _/instance/edu.umd.cs.findbugs.plugin.eclipse/_ from the Preferences produced by the old FindBugs Eclipse plugin.

iloveeclipse:
Do you want to contribute a patch?

dlwildschut:
I created a pull request https://github.com/spotbugs/spotbugs/pull/729, could you please review my changes?

PR #729, opened 2018-08-10
Added support for User Preferences exported from SpotBugs Eclipse Plugin

dlwildschut:
This should fix the issue #728.
UserPreferences now also replaces the prefix generated by the SpotBugs Eclipse Plugin when loading a userPrefs file.

----

### DIFF
--- a/spotbugs/src/test/java/edu/umd/cs/findbugs/config/UserPreferencesTest.java
+++ b/spotbugs/src/test/java/edu/umd/cs/findbugs/config/UserPreferencesTest.java
@@ -1,5 +1,9 @@
 package edu.umd.cs.findbugs.config;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


spotbugs/src/test/java/edu/umd/cs/findbugs/config/UserPreferencesTest.java
```diff
@@ -14,4 +18,44 @@ public void testClone() {
...
         Assert.assertEquals(sut.getClass(), clone.getClass());
     }
+
+    @Test
+    public void testReadEffortNoPrefix() throws IOException {
+        String testPrefsString = "effort=max";
+
+        UserPreferences userPrefs = UserPreferences.createDefaultUserPreferences();
+
+        userPrefs.read(new ByteArrayInputStream(testPrefsString.getBytes(StandardCharsets.ISO_8859_1)));
```
ThrawnCA: Why ISO-8859-1 instead of UTF-8?

dlwildschut: According to https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.InputStream- the Properties are assumed to be in ISO 8859-1, so I used that as the encoding.

PR #730, opened 2018-08-13
Basic ISSUE_TEMPLATE.md

h3xstream:
The issue template point to StackOverflow to general question and support.

The idea of using StackOverflow is to:
 - Increase the visibility of question to more users.
 - Encourage with points SpotBugs users to help others.
 - StackOverflow is a good archive for dev questions. (Good UI and good SEO)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,8 @@
+<!--
+Thank you for reporting an issue.
+
+If you have a question or if you need support, please ask a question on StackOverflow:
+https://stackoverflow.com/questions/ask?tags=spotbugs,java
+
+GitHub issues are mainly used for tracking enhancements and bugs.
+-->


ThrawnCA:
Do we really want to send people to Stack Overflow for this? I'm there from time to time, but is that the project's official stance?

h3xstream:
@ThrawnCA Do you have any objection to use Stackoverflow?

henrik242:
I don't know, really. There aren't many issues reported here, and personally I don't think I would check SO for spotbugs questions regularly. I'll let you guys decide.

iloveeclipse:
I'm definitely not monitoring stack overflow. Not sure how this template should work, but if it is directing people away from project to ask about project, it sounds somehow weird.

h3xstream:
I'm closing this PR. There is no consensus that SO has any advantage over GitHub issues.

PR #732, opened 2018-08-15
Update to Gradle 4.9 and spotbugs-maven-plugin 1.6.2 for jdk11 support

stevenschlansker:
Fixes `gradlew` invocation errors on jdk11
https://github.com/gradle/gradle/issues/4515
https://github.com/spotbugs/spotbugs-gradle-plugin/issues/22

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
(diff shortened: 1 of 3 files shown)


henrik242:
@stevenschlansker Did you take a look at the build problem?

henrik242:
@KengoTODA We should replace the jdk9 build with jdk10

stevenschlansker:
@henrik242 No, I cannot seem to find a jdk9 build - it is already EOL and there is no download on Oracle or OpenJDK that I can see.

Is jdk9 compatibility a goal over jdk10 / 11?  I do not need jdk9 support and so won't be spending much time to explore that.

wrprice:
FYI, JDK 9 builds archived here: http://jdk.java.net/archive/

Syquel:
I think the issue is caused by the `includeNoLocationClasses = true` jacoco flag in [jacoco.gradle](https://github.com/spotbugs/spotbugs/blob/build-on-jdk11/gradle/jacoco.gradle).
It has been explained in the [jacoco issues](https://github.com/jacoco/jacoco/issues/16#issuecomment-351943050) quite well.

I will try it out when I find the time.

KengoTODA:
This issue is closed by #784.

PR #747, opened 2018-09-04
Added bug pattern "OVERRIDING_METHODS_MUST_INVOKE_SUPER" for improved…

jpwilhelms:
… Detector "CbeckMustOverrideSuperAnnotation"

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## 3.1.6 - 2018-07-18
 
+### Added
+
+* Support for new annotation edu.umd.cs.findbugs.annotations.OverridingMethodsMustInvokeSuper
+
 ### Fixed
 
 * Potential NPE in test-harness-core ([#671](https://github.com/spotbugs/spotbugs/issues/671))
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/OverridingMethodsMustInvokeSuper.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/OverridingMethodsMustInvokeSuper.java
@@ -1,0 +1,34 @@
+/*
+ * FindBugs - Find Bugs in Java programs
+ * Copyright (C) 2005-2006, University of Maryland
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+package edu.umd.cs.findbugs.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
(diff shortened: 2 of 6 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CbeckMustOverrideSuperAnnotation.java
```diff
@@ -69,11 +61,22 @@ public void visit(Code code) {
...
         super.visit(code);
         if (!sawCallToSuper) {
-            bugReporter.reportBug(new BugInstance(this, "TESTING", NORMAL_PRIORITY).addClassAndMethod(this).addString(
+            bugReporter.reportBug(new BugInstance(this, "OVERRIDING_METHODS_MUST_INVOKE_SUPER", NORMAL_PRIORITY).addClassAndMethod(this).addString(
                     "Method must invoke override method in superclass"));
         }
     }
 
+    public String getEffectiveMethodSig() {
+        String methodSig = getMethodSig();
```
ThrawnCA: No point calling `getMethodSig` if we have a bridge. Can this call be deferred until we check that?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CbeckMustOverrideSuperAnnotation.java
```diff
@@ -36,29 +34,23 @@
 
 public class CbeckMustOverrideSuperAnnotation extends OpcodeStackDetector {
```
ThrawnCA: Is this misspelled?

jpwilhelms: I took that code (and that name) from a findbugs detector. You mean the "Cbeck" prefix? I don't know what that means. What would be a good name for this detector?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CbeckMustOverrideSuperAnnotation.java
jpwilhelms: OK, I'll change it.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CbeckMustOverrideSuperAnnotation.java
iloveeclipse: "Cbeck" -> "Check" ?

jpwilhelms: Oh, I see! Now I would call the detector "OverridingMethodsMustInvokeSuperDetector". The annotation suffix was misleading, too.

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/OverridingMethodsMustInvokeSuper.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * FindBugs - Find Bugs in Java programs
```
ThrawnCA: For new files, this should be SpotBugs, not FindBugs, and it doesn't belong to the University of Maryland.

@spotbugs/everyone I don't think we've actually discussed what copyright notice to use going forward...

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## 3.1.6 - 2018-07-18
 
+### Added
+
+* Support for new annotation edu.umd.cs.findbugs.annotations.OverridingMethodsMustInvokeSuper
```
KengoTODA: please move this entry to the bottom of `## Unreleased`.

KengoTODA:
it seems that this PR renames existing class, then I think the target branch must be `master` branch.

iloveeclipse:
I think changing (renaming) detector class is OK for maintenance branch. I doubt someone extended existing detector. BTW, I don't see changed config for the old one - was it not properly configured?

KengoTODA:
I'll close this stale PR. Feel free to reopen it.

PR #759, opened 2018-09-19
Simplify deployment script

KengoTODA:
Currently snapshot build from `release-3.1` is not deployed to [sonatype repo](https://oss.sonatype.org/content/repositories/snapshots/com/github/spotbugs/spotbugs/). This is because our usage of `travis_wait` script is invalid. We should wrap command by quote.

When we wrap command, its variable handling becomes complex, so I move secret variable handling from command-line parameter to `gradle.properties`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,6 +116,6 @@ deploy:
     # https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
     script: .travis/travis_wait "./gradlew uploadArchives"
     on:
-      jdk: oraclejdk8
       all_branches: true
-      condition: branch IN (master, "release-3.1") AND type = push
+      jdk: oraclejdk8
+      condition: (branch = master || branch = "release-3.1") && type = push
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -41,6 +40,11 @@ install:
   - rm -rf eclipse
   - tar xzvf deps/eclipse-SDK-4.6.3-linux-gtk-x86_64.tar.gz eclipse
   - echo eclipseRoot.dir=$(pwd)/eclipse > eclipsePlugin/local.properties
+  - echo ossrhUsername=eller86 >> gradle.properties
```
ThrawnCA: Are we certain that `gradle.properties` won't already exist?

KengoTODA: we're sure that we have it, because we commit it to this Git repo

https://github.com/spotbugs/spotbugs/blob/release-3.1/gradle.properties

KengoTODA:
deployment script still has problem, I'll close this PR temporally

KengoTODA:
I've confirmed that `3.1.8-SNAPSHOT` has been deployed onto https://oss.sonatype.org/content/repositories/snapshots/com/github/spotbugs/spotbugs/3.1.8-SNAPSHOT/ successfully!

PR #783, opened 2018-10-29
Fix warnings from LGTM

stweil:
LGTM reports:

    This array access might be out of bounds,
    as the index might be equal to the array length.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2018-??-??
 
+### Fixed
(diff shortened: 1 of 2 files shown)


KengoTODA:
Thanks for your PR!
Please consider to reproduce problem by unit test first, or we cannot judge this fix is necessary or not.

stweil:
I don't think that my commit fixes a real problem because those loops also test for an `INVALID` element at index `i` before accessing index `i+1`. Therefore it is not possible to write the desired unit test.

The changes make the upper bound clearer for human reviewers and for static code analyzers like LGTM. That's why I think they are useful.

Issue #763, opened 2018-09-26
Support Java 11


KengoTODA:
I'm working to build and test SpotBugs on OpenJDK 11, here is [a branch](https://github.com/spotbugs/spotbugs/compare/build-on-jdk11) for this trial so you can see detail.

Problem is that, [Java 11 does not provide methods like `System.runFinalizersOnExit()` and `Thread#destroy()`](https://travis-ci.org/spotbugs/spotbugs/jobs/433389109#L2201). So we cannot test SpotBugs on OpenJDK 11 for now.

In my opinion, we can remove `classesJava7` task in `spotbugsTestCases/build.gradle` since we've removed Java 7 support. If you have different opinion, let's discuss at here.

KengoTODA:
Technically possible. But it may make dev env construction complex. I'm not sure it pays enough benefit or not.

e.g. This plugin changes JDK to compile java code, we can refer its implementation. https://github.com/tbroyer/gradle-errorprone-plugin/blob/master/README.md#jdk-8-support

victorwss:
I tried to compile a Gradle project using SpotBugs that was ok into Java 10 and it failed miserably with Java 11 giving a `LambdaConversionException` raising out from the deep guts of gradle. However, it seems to be a problem specifically with SpotBugs:

```
* Exception is:
java.lang.BootstrapMethodError: bootstrap method initialization exception
        at com.github.spotbugs.SpotBugsPlugin.lambda$configureForSourceSet$7(SpotBugsPlugin.java:199)
        at org.gradle.util.GUtil.uncheckedCall(GUtil.java:459)
        at org.gradle.api.internal.ConventionAwareHelper$2.doGetValue(ConventionAwareHelper.java:82)
        at org.gradle.api.internal.ConventionAwareHelper$MappedPropertyImpl.getValue(ConventionAwareHelper.java:126)
        at org.gradle.api.internal.ConventionAwareHelper.getConventionValue(ConventionAwareHelper.java:109)
        at com.github.spotbugs.SpotBugsTask_Decorated.getClasses(Unknown Source)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #784, opened 2018-10-30
Update gradle and plugins to prepare for JDK 11

KengoTODA:
Before we introduce Gradle 5.0 and CI build on oraclejdk11, we need to bump up several tools:

* Gradle, to reduce difference between our current build and Gradle 5.0
* Mockito and jacoco, to prepare to handle Java 11 binaries
* spotbugs-gradle-plugin, to support latest Gradle

This PR is related with #763.
After we successfully merge this PR, I will create another that introduces Gradle 5.0 RC1 and oraclejdk11.


ThrawnCA:
Remind me again, which JDK version do we require? Because I suspect that the dependencies of Mockito 2.23 will choke on less than JDK 9.

KengoTODA:
As explained at #763, JDK 7 is not supported, then JDK 8 and above should be target.

Then I think we can focus on JDK 8 and 11. [Java 9 and 10 are already unsupported](https://www.oracle.com/technetwork/java/javase/eol-135779.html) so we can stop support on them.

KengoTODA:
> the dependencies of Mockito 2.23 will choke on less than JDK 9.

Could you share more info about this? I cannot find in [their release note](https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md#2230).

ThrawnCA:
> > the dependencies of Mockito 2.23 will choke on less than JDK 9.
> 
> Could you share more info about this? I cannot find in [their release note](https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md#2230).

Version 2.22.1 makes note of a Byte-Buddy update for JDK 11 support. I've tripped over this in other projects using JDK 8 and Jetty 9.2; when Jetty starts, it chokes on the newer Byte Buddy.

It might or might not be a problem here, just thought I'd bring it up.

wreulicke:
CI passed. 🎉

KengoTODA:
> It might or might not be a problem here, just thought I'd bring it up.

Got it. I think it is not be a problem in our case, at least test passes as usual.

KengoTODA:
@spotbugs/everyone please review and approve this PR, thanks in advance!

Issue #19, opened 2016-11-08
Support running under Java7


jsotuyod:
The JRT FS code for Java 9 uses lambdas, which forced to raise the min JRE from 7 to 8. However, all other needed APIs are present in Java 7.

Turning those lambdas back into anonymous classes would allow to keep SpotBugs running in JRE 7, 8 and 9 (of course, actually being able to load a JRT FS needs JRE9). This would leave no one behind if they want to switch to SpotBugs at least in the immediate future (FindBugs 3.0.1 requires JRE7+).

Future releases may take advantage of JRE 8, but for the time being JRE7 is probaly the sanest.

@iloveeclipse is there any other reason for the version bump I may be missing?

iloveeclipse:
Please stop here. Supporting Java 7 is definitely nothing worth the effort. If someone can't run latest FB on Java 8, there are still old FB versions. Maintaining code for this old Java release just cost time, and time is what we do not have.

jsotuyod:
I don't think this is the right approach to this conversation.

Before even thinking of effort / time, we should be considering what's the best for our users and the product as a whole.

The way I see it, keeping JRE7 support has some interesting advantages:

* It makes moving from FindBugs to SpotBugs a no brainer. No strings attached, no conditions, just drop in the replacement lib.
* It maximizes momentum to migrate. Allowing everyone to migrate now will probably see a better adoption than having people stuck on old projects and then having to remember to switch if they ever bump their JDK.
* It keeps in line with our proposal to remain binary-compatible with FindBugs, at least for version 3 of SpotBugs. Version 4 may be a different matter.
* It keeps providing value to Android users. Android has been stuck for quite some time on Java 6 + some Java 7 syntax sugar, meaning many developers are still using JRE7 (a battle I keep fighting within my organization for instance).
* It simplifies integration by third parties. Many tools that integrate FindBugs, [such as Gradle](https://docs.gradle.org/current/userguide/installation.html#sec:prerequisites) require JRE7. Keeping JRE7+ means they can actually just drop FindBugs for SpotBugs almost seamlessly. Unit and integration tests can probably remain the same. Dropping JRE7 support makes the whole matter a lot harder on them, which again, works against user adoption.

I believe these are strong arguments in favor of keeping JRE7 which can't be lightly dismissed.

Having said that, we may now approach the planification for such work.

[...]

PR #786, opened 2018-10-30
Build with oraclejdk11

### LINKED ISSUE #763: Support Java 11

KengoTODA:
This PR closes #763. It contains following changes:

1. Currently we have test code that needs JDK7 to compile, however [spotbugs does not support scanning Java7](https://github.com/spotbugs/spotbugs/issues/19). And JDK11 cannot compile with `target = 1.7` so I removed code under `spotbugsTestCases/src/java7`.
2. `javax.annotation` has been removed in JDK11, so added dependency on it explicitly.
3. JDK11 removed some API like `Thread#destroy()`, so move test depending on them into `spotbugsTestCases/src/java8` and skip them when we build with JDK11.
4. Use `oraclejdk11` to build. It makes test failed (see #756) so [allow failure](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail) as temporal solution.


KengoTODA:
@spotbugs/everyone please review this PR, thank you!

.travis.yml
```diff
@@ -49,7 +51,7 @@ install:
 script:
   - jdk_switcher use oraclejdk8
   - ./gradlew classes testClasses -S
-  - jdk_switcher use $TRAVIS_JDK_VERSION
+  - if [[ $TRAVIS_JDK_VERSION == "oraclejdk11" ]]; then export JAVA_HOME=$HOME/oraclejdk11; fi
```
ThrawnCA: What provides this JDK?

KengoTODA: TravisCI installs it by [install-jdk.sh](https://github.com/sormuras/bach#install-jdksh).

Here is [official documentation about how to switch JDK 10 and up](https://docs.travis-ci.com/user/languages/java/#switching-jdks-to-java-10-and-up-within-one-job).

iloveeclipse: Just wondering that we first switch to jdk 8 but next line check if the jdk version is 11? How is this possible?

KengoTODA: 9 and 10 have already achieved their EOL. Recently Java changed their release policy.

iloveeclipse: I was not wondering where 9 and 10 are, I was trying to understand why the script switches to java 8 but a line after tests if it will be 11? I simply can't understand this part of the script. If I switch to 8 I expect 8 to be used, not 11.

KengoTODA: By this change, we'll have two build jobs [like this](https://travis-ci.org/spotbugs/spotbugs/builds/448196622).

One will **build and test** spotbugs with JDK8,  
another will **build** spotbugs with JDK8 and **test** it with JDK11.

Its background is:

1. We distribute binary built by JDK8, so build part should always use JDK8 (realized by `jdk_switcher use oraclejdk8` in `.travis.yml`)
2. Target classes to analyse can be compiled by different JDK in users' environment, and JVM that runs SpotBugs also can be different JDK. So test part should have several patterns. (realized by `if [[ $TRAVIS_JDK_VERSION == "oraclejdk11" ]]; then ...` in `.travis.yml`)

In other words, our build matrix is like this:

|#|JDK to compile|JDK to run test|
|-|----|----|
|1|JDK8|JDK8|
|2|JDK8|JDK11|

> If I switch to 8 I expect 8 to be used, not 11.

That's right. For the 1st pattern, we never use JDK11. We use it only for the 2nd pattern.

iloveeclipse: Thanks Kengo. This now makes sense, but I beat next time I will see this code I will ask the same question again :-).

spotbugsTestCases/src/java7/sfBugs/Bug3458246.java
```diff
@@ -1,37 +0,0 @@
-package sfBugs;
```
iloveeclipse: The bug is https://sourceforge.net/p/findbugs/bugs/1003/, shouldn't we move this to Java 8 directory instead of deleting it?

KengoTODA: OK I will.

KengoTODA:
Sorry but please review this PR again, thanks @ThrawnCA !

spotbugsTestCases/src/java/sfBugs/Bug3458246.java
```diff
@@ -0,0 +1,37 @@
+package sfBugs;
+
+import java.util.logging.Logger;
```
ThrawnCA: Is SLF4J fixed yet? There's supposed to be a Gradle fix.

KengoTODA: It is old test code, not related with that trouble.

PR #790, opened 2018-11-05
Upgrade Gradle to 5.0 RC, to introduce SLF4J for logging

KengoTODA:
The milestone-1 has a bug related with Groovy 2.5.2, and it makes eclipse build failed.
The 5.0-rc-1 also has a problem, that triggers [an error like this](https://gist.github.com/KengoTODA/ee0f4b1c8f8281855a17373f0374caad).
So we should use nightly build instead of milestone build that is newer than `5.0-20181030000033+0000`.

This change should fix [SLF4J related classloader issue](https://github.com/gradle/gradle/issues/2657).
@ThrawnCA do you have a commit that introduces SLF4J to spotbugs? Then we can cherry-pick to this branch to have a try.


KengoTODA:
SoanrQube scanner has problem, but the local build should have no problem.
I'll investigate about this problem later.

ThrawnCA:
@KengoTODA I don't have such a commit yet, but I'll push one.

Are we OK with relying on a snapshot build of Gradle? Or is this just to test that it will work?

KengoTODA:
https://twitter.com/gradle/status/1062005368794968064?s=19
5.0-rc2 is out.

KengoTODA:
We can successfully build SpotBugs on Gradle 5.0 RC2.

@ThrawnCA When we launch SpotBugs, it prints warning message like below. Maybe we need to add SLF4J bridge or logback?

```
java -jar lib/spotbugs.jar 
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

ThrawnCA:
Hmm.

We *can* provide an SLF4J implementation, but that kind of defeats the point of using a facade.

How easy is it to bundle an implementation inside the executable JAR, but not eg the Maven plugin?

I suppose it would also be OK to make the Maven plugin depend on an implementation, which the end user could then exclude if they want something else.

KengoTODA:
> How easy is it to bundle an implementation inside the executable JAR, but not eg the Maven plugin?

I realized it by 3147c44. in generated `pom.xml`, we use `<scope>test</scope>` for binding. So it won't affect users like spotbugs-maven-plugin.

However current version prints following warning. Better to add config file for logback.

```
$ java -jar spotbugs.jar
ERROR StatusLogger No Log4j 2 configuration file found. Using default configuration (logging only errors to the console), or user programmatically provid
ed configurations. Set system property 'log4j2.debug' to show Log4j 2 internal initialization logging. See https://logging.apache.org/log4j/2.x/manual/co
nfiguration.html for instructions on how to configure Log4j 2
```

We also need to add LICENSE file under `spotbugs/licenses/` directory.

KengoTODA:
OK I will merge this, and propose another PR to add `log4j2.yaml` into our distributions.

jimshowalter:
RC 4 has a similar-looking exception:

com.foo.bar.FactoryTest > shouldLogDebugInCorrectFormatToMessageLog FAILED
    java.lang.ClassCastException: class org.gradle.internal.logging.slf4j.OutputEventListenerBackedLoggerContext cannot be cast to class ch.qos.logback.classic.LoggerContext (org.gradle.internal.logging.slf4j.OutputEventListenerBackedLoggerContext and ch.qos.logback.classic.LoggerContext are in unnamed module of loader 'app')
        at com.foo.bar.LoggerFactory.getOrCreateLogger(IFactory.java:23)

This showed up when upgrading from Gradle 3 to Gradle 5.

PR #795, opened 2018-11-09
Remove superfluous empty line from about dialog

Bananeweizen:
There is an unexpected empty line in the about dialog of the feature.
The properties file entry should not end with a line break nor with a line continuation escape character.


KengoTODA:
@Bananeweizen thanks, could you add an entry into `CHANGELOG.md`?

Issue #705, opened 2018-07-20
New performance rule for using String replace instead of replaceAll


blindspeed90:
People mistakenly use replaceAll as a literal replacement, and will get the overhead of doing a regexp.  Possibly the code could check the parameter for the existence of any special regexp characters as its' test.

lasselindqvist:
Based on https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html it would make sense to warn if the replaceAll does not contain any letter from "\\.[]{}()<>*+-=?^$|"

PR #797, opened 2018-11-13
Implement issue #705

lasselindqvist:
Implement new detector IRA_INEFFICIENT_REPLACEALL. Finds occurrences of replaceAll(String regex, String replacement) without any special regex characters.

----
This is my first implementation pull request here, so most likely there are some things I missed.
The rule has been marked disabled="true" which I assumed disabled it by default, and experimental="true" which could be removed in the future? ITA_INEFFICIENT_TO_ARRAY served as the main example I looked at.

----

I did benchmark the change from replaceAll to replace with JMH with a simple case the throughputs were 2307481 for replaceAll and 2827110 for replace, so I believe the optimization is valid.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 * Dataflow generates too much log ([#601](https://github.com/spotbugs/spotbugs/issues/601))
 
+### ADDED
+* Add new detector IRA\_INEFFICIENT\_REPLACEALL for detecting usage of String.replaceAll where no regex is being used ([#705](https://github.com/spotbugs/spotbugs/issues/705))
+
 ## 3.1.9 - 2018-11-20
 
 ### Fixed
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
@@ -1,0 +1,118 @@
+/*
+ * Contributions to SpotBugs
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+package edu.umd.cs.findbugs.detect;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+
+import org.apache.bcel.Const;
+import org.apache.bcel.classfile.Code;
+import org.apache.bcel.classfile.Method;
+
+import edu.umd.cs.findbugs.BugAccumulator;
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.StatelessDetector;
+import edu.umd.cs.findbugs.ba.ClassContext;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import edu.umd.cs.findbugs.classfile.MethodDescriptor;
+
+/**
+ * Find occurrences of replaceAll(String regex, String replacement) without any
+ * special regex characters.
+ *
+ * @author Lasse Lindqvist
+ */
+public class InefficientReplaceAll extends OpcodeStackDetector implements StatelessDetector {
+
+    private static final Logger LOGGER = Logger.getLogger(InefficientReplaceAll.class.getName());
+
+    private static final List<MethodDescriptor> methods = Collections.singletonList(
+            new MethodDescriptor("", "replaceAll", "(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;"));
+
+    private final BugAccumulator bugAccumulator;
+
+    private static final Pattern REGEX_CHARS_PATTERN = Pattern.compile("[" + Pattern.quote("-.\\[]{}()<>*+=?^$|") + "]");
+
+    public InefficientReplaceAll(BugReporter bugReporter) {
+        this.bugAccumulator = new BugAccumulator(bugReporter);
+    }
+
+    @Override
+    public void visitClassContext(ClassContext classContext) {
+        if (hasInterestingMethod(classContext.getJavaClass().getConstantPool(), methods)) {
+            classContext.getJavaClass().accept(this);
+        }
+    }
+
+    @Override
+    public void visit(Method obj) {
+        LOGGER.log(Level.FINE, "------------------- Analyzing {0} ----------------", obj.getName());
+        super.visit(obj);
+    }
+
+    @Override
+    public void visit(Code obj) {
+        super.visit(obj);
+        bugAccumulator.reportAccumulatedBugs();
+
+    }
+
+    @Override
+    public void sawOpcode(int seen) {
+        LOGGER.log(Level.FINE, "Opcode: {0}", Const.getOpcodeName(seen));
+        if (((seen == Const.INVOKEVIRTUAL) || (seen == Const.INVOKEINTERFACE))
+                && ("replaceAll".equals(getNameConstantOperand()))
+                && ("(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;".equals(getSigConstantOperand()))
+                && hasConstantArguments()) {
+            Optional<String> firstArgument = getFirstArgument();
+            if (firstArgument.isPresent() && !REGEX_CHARS_PATTERN.matcher(firstArgument.get()).find()) {
+                bugAccumulator.accumulateBug(
+                        new BugInstance(this, "IRA_INEFFICIENT_REPLACEALL", LOW_PRIORITY).addClassAndMethod(this),
+                        this);
+            }
+        }
+    }
+
+    /**
+     * @return first argument of the called method if it's a constant
+     */
+    private Optional<String> getFirstArgument() {
+        Object value = getStack().getStackItem(getNumberArguments(getMethodDescriptorOperand().getSignature()) - 1)
+                .getConstant();
+        return Optional.ofNullable(Objects.toString(value, null));
+    }
+
+    /**
+     * @return true if only constants are passed to the called method
+     */
+    private boolean hasConstantArguments() {
+        int nArgs = getNumberArguments(getMethodDescriptorOperand().getSignature());
+        for (int i = 0; i < nArgs; i++) {
+            if (getStack().getStackItem(i).getConstant() == null) {
+                return false;
+            }
+        }
+        return true;
+    }
+}
--- a/spotbugsTestCases/src/java/ghIssues/Issue705.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue705.java
@@ -1,0 +1,66 @@
+/*
+ * Contributions to SpotBugs
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+package ghIssues;
+
+public class Issue705 {
+    
+    public static String bug() {
+        String replaceable = "replaceable";
+        String replacement = "replacement";
+        String target = "target";
+        return target.replaceAll(replaceable, replacement);
+    }
+
+    public static String nonBug() {
+        String replaceable = "(replaceable)";
+        String replacement = "replacement";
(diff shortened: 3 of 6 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,143 @@
...
+    public void visit(Code obj) {
+        super.visit(obj);
+        bugAccumulator.reportAccumulatedBugs();
+
+    }
+
+    @Override
+    public void sawOpcode(int seen) {
+        if (DEBUG) {
+            System.out.println("Opcode: " + Const.getOpcodeName(seen));
```
ThrawnCA: Can this use a logger? `java.util.logging.Logger` for now, and we'll convert it later to SLF4J (#635)

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+        if (((seen == Const.INVOKEVIRTUAL) || (seen == Const.INVOKEINTERFACE)) && ("replaceAll".equals(getNameConstantOperand()))
+                && ("(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;".equals(getSigConstantOperand()))
+                && hasConstantArguments()) {
+            String firstArgument = getFirstArgument();
+            boolean found = false;
+            for (String regexChar : regexCharList) {
```
ThrawnCA: It would likely be more efficient to implement this as a regex, instead of iterating through a list. Something like

    private static final Pattern REGEX_CHARS_PATTERN = Pattern.compile("[-.\\\\[\\]{}()<>*+=?^$|]");
    ...
    if (REGEX_CHARS_PATTERN.matcher(firstArgument).find()) {
        ...

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+    
+    /**
+     * @return first argument of the called method if it's a constant
+     */
+    private String getFirstArgument() {
+        Object value = getStack().getStackItem(getNumberArguments(getMethodDescriptorOperand().getSignature()) - 1)
+                .getConstant();
+        return value == null ? null : value.toString();
```
ThrawnCA: This could be simplified with `Objects.toString` and perhaps returning an `Optional`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public void sawOpcode(int seen) {
+        if (DEBUG) {
+            System.out.println("Opcode: " + Const.getOpcodeName(seen));
+        }
+        if (((seen == Const.INVOKEVIRTUAL) || (seen == Const.INVOKEINTERFACE)) && ("replaceAll".equals(getNameConstantOperand()))
+                && ("(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;".equals(getSigConstantOperand()))
+                && hasConstantArguments()) {
+            String firstArgument = getFirstArgument();
```
ThrawnCA: `getFirstArgument` can return `null`, which will cause an NPE a few lines later.

ThrawnCA:
@lasselindqvist Thanks for the pull request. I'm passing it back due to the potential NPE, but the concept seems sound.

ThrawnCA:
Hmm...while you're in this space, what do you think about producing a second detector that would recognise when `replaceAll` is called with a literal string that contains regex characters but will fail to compile (`Pattern.compile` throws PatternSyntaxException)?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
lasselindqvist: Accidentally not in this case because hasConstantArguments returns false on null arguments. But it is anyway good to refactor these two methods so that getFirstArgument uses Optional. Both concerns are now fixed.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final boolean DEBUG = SystemProperties.getBoolean("ira.debug");
+
+    private static final Logger LOGGER = Logger.getLogger(InefficientReplaceAll.class.getName());
+
+    private static final List<MethodDescriptor> methods = Collections.singletonList(
+            new MethodDescriptor("", "replaceAll", "(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;"));
+
+    private final BugAccumulator bugAccumulator;
+
+    private static final Pattern REGEX_CHARS_PATTERN = Pattern.compile("[\\-\\.\\\\\\[\\]\\{\\}\\(\\)\\<\\>\\*\\+\\=\\?\\^\\$\\|]");
```
ThrawnCA: Wow, that's a lot of backslashes.

You really don't need most of them inside a character class. The only characters that retain their significance in that context are the closing square bracket `]`, the backslash itself, and the hyphen - but the hyphen is still OK when it's at the start of the character class as this one is. Even an opening square bracket is taken as a literal.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,128 @@
...
+    public void sawOpcode(int seen) {
+        if (DEBUG) {
+            LOGGER.log(Level.INFO, "Opcode: " + Const.getOpcodeName(seen));
+        }
+        if (((seen == Const.INVOKEVIRTUAL) || (seen == Const.INVOKEINTERFACE))
+                && ("replaceAll".equals(getNameConstantOperand()))
+                && ("(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;".equals(getSigConstantOperand()))
+                && hasConstantArguments()) {
+            Optional<String> firstArgument = getFirstArgument();
+            if (firstArgument.isPresent()) {
```
ThrawnCA: These two `if` statements can be combined, `if (firstArgument.isPresent() && !REGEX_CHARS_PATTERN...`

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    /**
+     * @return first argument of the called method if it's a constant
+     */
+    private Optional<String> getFirstArgument() {
+        Object value = getStack().getStackItem(getNumberArguments(getMethodDescriptorOperand().getSignature()) - 1)
+                .getConstant();
+        if (value == null) {
```
ThrawnCA: Simpler:

    return Optional.ofNullable(Objects.toString(value, null));

spotbugsTestCases/src/java/ghIssues/Issue705.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public static String nonBug() {
+        String replaceable = "(replaceable)";
+        String replacement = "replacement";
+        String target = "target";
+        return target.replaceAll(replaceable, replacement);
+    }
+    
+    public static String nonBug2() {
```
ThrawnCA: Nice coverage :+1: 

Might be easier to make an array of the metacharacters and loop through that.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,128 @@
...
+    public void visitClassContext(ClassContext classContext) {
+        if (hasInterestingMethod(classContext.getJavaClass().getConstantPool(), methods)) {
+            classContext.getJavaClass().accept(this);
+        }
+    }
+
+    @Override
+    public void visit(Method obj) {
+        if (DEBUG) {
+            LOGGER.log(Level.INFO, "------------------- Analyzing " + obj.getName() + " ----------------");
```
ThrawnCA: You don't need string concatenation here. The syntax you want is:

    LOGGER.log(Level.INFO, "------------------- Analyzing {0} ----------------", obj.getName());
￼
This way, if the logging level is set higher than INFO, then the system doesn't have to perform the string concatenation at all.

lasselindqvist: I will change the logging level. The flag usage was just copied from another detector.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
lasselindqvist: I will change that.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
lasselindqvist: I will change that.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
lasselindqvist: The reason was that the suggested regex broke because, if I remember correctly, the dot required backslashes. But I will test all of those independently so there will be no extra backslashes.

spotbugsTestCases/src/java/ghIssues/Issue705.java
lasselindqvist: The problem with that might be that the arguments are no longer recognized as constants. I don't know what the full capabilities of Spotbugs in those situations are.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
lasselindqvist: This is now using
`Pattern.compile("[" + Pattern.quote("-.\\[]{}()<>*+=?^$|") + "]")`
which is quite clear.

spotbugsTestCases/src/java/ghIssues/Issue705.java
ThrawnCA: Oh, of course. You're right, it would probably fail to trip the detector for the wrong reasons.

I shouldn't try to make bug test cases elegant :laughing:

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Fix some out-of-bounds reports from LGTM
 * Update asm to 7.0 for better Java 11 support ([#785](https://github.com/spotbugs/spotbugs/pull/785))
 * Ignore @FXML annotated fields in UR\_UNIT\_READ ([#702](https://github.com/spotbugs/spotbugs/issues/702))
+* Add new detector IRA\_INEFFICIENT\_REPLACEALL for detecting usage of String.replaceAll where no regex is being used ([#705](https://github.com/spotbugs/spotbugs/issues/705))
```
KengoTODA: This is not bugfix, so please use `Added` instead of `Fixed`. refer https://keepachangelog.com/en/1.0.0/ for detail

lasselindqvist: This actually says "Add". Should it be "Added"? I just looked at row 10 for an example. I can change all of the lines use that if that is the preferred form.

ThrawnCA: @lasselindqvist I think he means that it's in the wrong section of the changelog. It's filed under `### Fixed` instead of `### Added`.

lasselindqvist: Done. I will squash the changes together later today.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,118 @@
...
+public class InefficientReplaceAll extends OpcodeStackDetector implements StatelessDetector {
+
+    private static final Logger LOGGER = Logger.getLogger(InefficientReplaceAll.class.getName());
+
+    private static final List<MethodDescriptor> methods = Collections.singletonList(
+            new MethodDescriptor("", "replaceAll", "(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;"));
+
+    private final BugAccumulator bugAccumulator;
+
+    private static final Pattern REGEX_CHARS_PATTERN = Pattern.compile("[" + Pattern.quote("-.\\[]{}()<>*+=?^$|") + "]");
```
ThrawnCA: Only the backslash and the closing square brace really need to be quoted (since it's inside a character class), but no big deal.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InefficientReplaceAll.java
```diff
@@ -0,0 +1,118 @@
...
+        return Optional.ofNullable(Objects.toString(value, null));
+    }
+
+    /**
+     * @return true if only constants are passed to the called method
+     */
+    private boolean hasConstantArguments() {
+        int nArgs = getNumberArguments(getMethodDescriptorOperand().getSignature());
+        for (int i = 0; i < nArgs; i++) {
+            if (getStack().getStackItem(i).getConstant() == null) {
```
ThrawnCA: I think that there is some quirkiness around inline arrays having a constant value equal to their length, but it should be OK in this case since `replaceAll` doesn't take an array argument.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### CHANGED
 * Allow parallel workspace builds in Eclipse with Spotbugs installed
+* Add new detector IRA\_INEFFICIENT\_REPLACEALL for detecting usage of String.replaceAll where no regex is being used ([#705](https://github.com/spotbugs/spotbugs/issues/705))
```
KengoTODA: this is new feature, so please use `### ADDED` instead of `### CHANGED`.

lasselindqvist: Done

ThrawnCA:
@lasselindqvist Github keeps reporting that you're using force-pushes; that's a hazardous practice. Is there something interfering with regular pushes?

KengoTODA:
please merge latest `release-3.1` branch to your branch, and resolve conflict around CHNAGELOG.md and others.

lasselindqvist:
> 
> 
> please merge latest `release-3.1` branch to your branch, and resolve conflict around CHNAGELOG.md and others.

This is now done.

FoxyBOA:
Does that detector presents in spotbugs 4.1.4? Can't spot it in an Eclipse plugin...

PR #814, opened 2018-11-29
Change OPCODE_NAMES[opcode] to getOpcodeName(opcode) in comments

TysonAndre:
This saves time if there is a need to locally re-enable these debugging
statements, and makes it easier to base code off of these methods.

There is no expected impact - These are all commented out and the constant doesn't exist any more.

I'm not sure which target branch to use (checked CONTRIBUTING.md) - is 3.1 correct?

----

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code (only comments were changed)


KengoTODA:
> I'm not sure which target branch to use (checked CONTRIBUTING.md) - is 3.1 correct?

Yes.

KengoTODA:
Merged, thanks for your PR!

Issue #470, opened 2017-10-23
Add support for Java 1.8 TYPE_USE, TYPE_PARAMETER to @Nonnull, @Nullable


unix-junkie:
Description copied from the corresponding _FindBugs_ [issue](https://sourceforge.net/p/findbugs/feature-requests/349/).

Currently `@NonNull` and `@Nullable` annotations don't specify any 1.8+ `@Target`'s. Adding a `@Target` with `ElementType.TYPE_USE` and `ElementType.TYPE_PARAMETER` among others, would allow constructs like this:

```java
@Nullable List<@NonNull String> foo = ...;
```

or 

```java
class Foo<@NonNull T> {
    T doSmth(Comparable<T> bar) {
        // ...
    }
}
```

This would ease null analysis in IDEs which support this (particularly, _Eclipse_ and _IDEA_).

See code listings for more examples:
* http://www.oracle.com/technetwork/articles/java/ma14-architect-annotations-2177655.html
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=403917
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=403816

KengoTODA:
related with #419

KengoTODA:
Consider to support #484 too. e.g. `CompletableFuture<@NonNull String>`

demonfiddler:
IMO this ought to be a priority given that JavaSE-1.8 was released in March 2014 and was superseded by JavaSE-9 in September 2017. Anyone who's using nullability annotations with 1.8/9 will undoubtedly be using type annotations, only to discover that SpotBugs null analyis doesn't work and produces a bunch of false positives.

KengoTODA:
SpotBugs really welcome your patches!

kashike:
Be aware of https://bugs.openjdk.java.net/browse/JDK-8144185.

Issue #743, opened 2018-08-28
NP_NONNULL_PARAM_VIOLATION with Checker Framework @Nullable


skitt:
I'm aware this is supposed to be fixed by #539, and I have also tried to create this as a failing test in spotbugs (without success). However, building the following code:

```
import com.google.common.util.concurrent.Futures;
import java.util.concurrent.Future;

public class Test {
    private static final Future NULL_FUTURE = Futures.immediateFuture(null);
}
```

with the following POM:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.sk2</groupId>
    <artifactId>spotbugs-guava-test</artifactId>
    <version>0.0.1-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>26.0-jre</version>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>com.github.spotbugs</groupId>
                <artifactId>spotbugs-maven-plugin</artifactId>
                <version>3.1.6</version>
                <configuration>
                    <failOnError>true</failOnError>
                </configuration>
                <executions>
                    <execution>
                        <id>analyze-compile</id>
                        <phase>compile</phase>
                        <goals>
                            <goal>check</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

results in

```
[INFO] --- spotbugs-maven-plugin:3.1.6:check (analyze-compile) @ spotbugs-guava-test ---
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
[ERROR] Null passed for non-null parameter of com.google.common.util.concurrent.Futures.immediateFuture(Object) in org.sk2.spotbugs.Test.<static initializer for Test>() [org.sk2.spotbugs.Test] At Test.java:[line 7] NP_NONNULL_PARAM_VIOLATION
```

even though ``immediateFuture`` is declared as

```
public static <V> ListenableFuture<V> immediateFuture(@Nullable V value)
```

[...]

Issue #643, opened 2018-05-25
Support for JSR308 type annotations in nullness checker


PhilippWendler:
The nullness checker of SpotBugs ignores the JSR308 `TYPE_USE` annotations like `org.checkerframework.checker.nullness.qual.Nullable`. This specific annotation is known to SpotBugs since https://github.com/spotbugs/spotbugs/commit/dd436ff56abf0273a080ad228fcf5abd8ee3a097, but it is not detected, because it seems that SpotBugs looks only for conventional annotations, not type annotations.

I am not suggesting a full analysis of type annotations that supports something like `List<@Nullable String>`, I am only suggesting that `@Nullable Object var = ...` should be supported if `@Nullable` is a type annotation instead of a regular annotation.

Test case (tested with SpotBugs 3.1.3):
```
@org.sosy_lab.common.annotations.ReturnValuesAreNonnullByDefault
public abstract class TestNullable {

  private static @org.checkerframework.checker.nullness.qual.Nullable Object getFoo() {
    return null;
  }
}
```

I do not know about the format of class files, but I suspect that type annotations are stored in a different place than regular annotations.
(As a further explanation: type annotations are also available in a different place when using Java's reflection API: for regular annotations you call e.g. `Parameter.getAnnotations()`, for type annotations you call `Parameter.getAnnotatedType().getAnnotations()`.)

Guava 25.1 has switched to using this annotation, so I suspect this will be a more common problem in the future. It has already been mentioned as a problem here: https://github.com/spotbugs/spotbugs/issues/621#issuecomment-391686616.

Note that this is not the same as #419 and #470, which suggest adding `TYPE_USE` to SpotBugs' annotations.

slovdahl:
Is there any kind of workaround available for this problem? Is there any timeline for a SpotBugs 4.0 release?

slovdahl:
Version 3.1.10 seems to have a fix for this, am I interpreting that right?

olegshtch:
I got it in 4.2.0  version.

ankilosaurus:
I am using spotbugs version 4.8.6.6 and hitting this issue with jakarta.annotation.Nullable

Issue #616, opened 2018-04-26
Regression in 3.1.3: false alarm for "Parameter must be non-null but is marked as nullable"


PhilippWendler:
For this code:
```
import com.google.common.base.Preconditions;
import javax.annotation.Nullable;

public enum TestClass {
  TEST("one", "two", null);

  public final String one;
  public final String two;
  @Nullable public final String three;

  private TestClass(String one, String two, @Nullable Object three) {
    this.one = Preconditions.checkNotNull(one);
    this.two = Preconditions.checkNotNull(two);
    this.three = three == null ? null : three.toString();
  }
}
```

SpotBugs 3.1.3 outputs this warning:

```
Parameter must be non-null but is marked as nullable

In class TestClass
In method new TestClass(String, int, String, String, Object)
Parameter one
At TestClass.java:[lines 11-15]
one must be non-null but is marked as nullable
```

SpotBugs 3.1.2 does not show this warning. The code is compiled for Java 8.

This has to be a bug, because if I remove either parameter `two` or `three`, the warning about parameter `one` goes away.

KengoTODA:
I cannot reproduce this reported bug by unit test, see 605fc5a for detail.

PhilippWendler:
Thanks for investigating this!

In general, we have a `package-info.java` with annotations, but not in the default package where I put the test class.

I have uploaded my `.class` file here: [Issue616.zip](https://github.com/spotbugs/spotbugs/files/1965950/Issue616.zip)
Can you reproduce it with this?

Is the class path relevant for this kind of problem?
I have (among others) Guava 23.6-jre, spotbugs-annotations 3.1.1 and jsr305 3.0.2 from com.google.code.findbugs on the auxiliary classpath.

FH3095:
Hello,

I today encountered the same problem as PhilippWendler . To me it seems to be relevant to mark the enum members NonNull to trigger the false positive. So when I modify your Issue616.java to the following
```
package com.example;

import java.util.Objects;

import edu.umd.cs.findbugs.annotations.NonNull;
import edu.umd.cs.findbugs.annotations.Nullable;

public enum Issue616 {
	TEST("one", "two", null);

	public final @NonNull String one;
	public final @NonNull String two;
	@Nullable
	public final String three;

	private Issue616(@NonNull String one, @NonNull String two, @Nullable Object three) {
		this.one = Objects.requireNonNull(one);
		this.two = Objects.requireNonNull(two);
[...]

PR #816, opened 2018-11-30
Detect method parameter type annotations

bhavanki:
This change causes Spotbugs to treat type annotations on formal method parameters just like parameter annotations. The main purpose is to get Spotbugs to see @Nullable and @NonNull annotations in code compiled under Java 8, where the compiler slots them as type annotations.

This might not be a "correct" way of handling the issue, because it doesn't make Spotbugs understand type annotations. So, feel free to reject for that reason (or others!) or use as a basis for some better way. It does at least illuminate, for example, how moving to newer Guava libraries causes false positives - Guava started being built under Java 8.

Should fix issues #470 in some aspects, #616, #643 in some aspects, and #743.
----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/ClassParserUsingASM.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/ClassParserUsingASM.java
@@ -34,6 +34,8 @@ import org.objectweb.asm.Handle;
 import org.objectweb.asm.Label;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.TypePath;
+import org.objectweb.asm.TypeReference;
 
 import edu.umd.cs.findbugs.ba.SignatureParser;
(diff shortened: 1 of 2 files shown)


ThrawnCA:
Can you target the 3.1 branch instead of master?

bhavanki:
Done!

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/ClassParserUsingASM.java
```diff
@@ -490,6 +492,19 @@ public void visitEnd() {
             mBuilder.addParameterAnnotation(parameter, desc, value);
             return value.getAnnotationVisitor();
         }
+
+        @Override
+        public org.objectweb.asm.AnnotationVisitor visitTypeAnnotation(int typeRef, TypePath typePath,
+                String desc, boolean visible) {
+            TypeReference typeRefObject = new TypeReference(typeRef);
+            if (typeRefObject.getSort() == TypeReference.METHOD_FORMAL_PARAMETER) {
```
ThrawnCA: It would be more readable if this were reversed: "if not the right sort then return `null`". No big deal.

PR #803, opened 2018-11-21
Sign Eclipse plugins 

KengoTODA:
Sign eclipse plugin jar files by certificate provided by Let's Encrypt.
We can verify that jar file is certainly signed by the following command:

```
$ jarsigner -verify -verbose -certs eclipsePlugin/build/site/eclipse/plugins/com.github.spotbugs.plugin.eclipse_3.1.9.201811211256-95f25b7.jar

s      10109 Wed Nov 21 12:57:06 CST 2018 META-INF/MANIFEST.MF

      [entry was signed on 11/21/18 12:57 PM]
      X.509, CN=spotbugs.kengo-toda.jp
      [certificate is valid from 11/20/18 2:42 PM to 2/18/19 2:42 PM]
      [ExtendedKeyUsage extension does not support code signing]
      X.509, CN=Let's Encrypt Authority X3, O=Let's Encrypt, C=US
      [certificate is valid from 3/18/16 12:40 AM to 3/18/21 12:40 AM]
      [CertPath not validated: Path does not chain with any of the trust anchors]
```

PR #818, opened 2018-12-04
Update document

KengoTODA:
Same with #803, but improved documentation about how to build Eclipse plugin.


KengoTODA:
@iloveeclipse they can refer this PR when they need. it has no magic, just a common signjar execution.

KengoTODA:
confirmed that I can install plugin without warning at the following site:
https://spotbugs.github.io/eclipse-stable-latest/

sewe:
@KengoTODA Eclipse really accepts a plug-in signed with a Let's Encrypt cert as signed? Let's Encrypt [explicitly](https://letsencrypt.org/docs/faq/#does-let-s-encrypt-issue-certificates-for-anything-other-than-ssl-tls-for-websites) does not issue code signing certificates:

> Email encryption and code signing require a different type of certificate that Let’s Encrypt does not issue.

So to me it looks like you accidentally have found an Eclipse bug.

PR #831, opened 2018-12-18
fix CHANGELOG about 3.1.10

KengoTODA:
The changelog of 3.1.10 is broken, we added one entry into not 3.1.10 but 3.1.9.


KengoTODA:
Hi @spotbugs/core-devs, could you review and merge this PR? Thanks in advance!

PR #836, opened 2018-12-24
Treat annotations on synthetic parameter properly

### LINKED ISSUE #772: False positive: parameter must be non-null in inner class constructor
Spotbugs 3.1.6 shows an issue for rule findbugs:NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE that an inner class constructor parameter must be non-null, even if the parameter is already annotated as NonNull. Code example:

```
package com.example;

import edu.umd.cs.findbugs.annotations.NonNull;
import edu.umd.cs.findbugs.annotations.Nullable;

public class NonNullParameterOfInnerClassConstructor {

  private class Inner {
    private final String a;
    private final Object b;

    Inner(@NonNull String a, @Nullable Object b) {
      this.a = a.toLowerCase(); // <--- false positive: a must be non-null but is marked as nullable
      this.b = b;
    }
  }
}

KengoTODA:
Currently `ClassParserUsingASM` uses parameter index that doesn't consider the synthetic parameter like the reference from non-static inner class to outer class. This means that, SpotBugs treats annotation on the 1st parameter __in source code__ (`String` in the test case) as annotation on the 1st parameter __in bytecode__ (`NonNullParameterOfInnerClassConstructor` in the test case).

Unfortunately at the timing of `visitParameterAnnotation()`, the `mBuilder` doesn't know that the parameter is synthetic or not. So we need to count the number of synthetic parameters by `number of actual parameters - number of annotatable parameters` dynamically.

This PR fixes #772. It also closes #773.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2019-??-??
 
+### Fixed
+* False positive: parameter must be non-null in inner class constructor ([#772](https://github.com/spotbugs/spotbugs/issues/772))
+
 ## 3.1.10 - 2018-12-19
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


KengoTODA:
@spotbugs/core-devs please review this change, thank you!

PR #851, opened 2019-01-16
Fix typos in documentation

Bananeweizen:


KengoTODA:
LGTM, but we also need to fix [docs/locale/ja/LC_MESSAGES/filter.po](https://github.com/spotbugs/spotbugs/blob/release-3.1/docs/locale/ja/LC_MESSAGES/filter.po) file for Japanese language users. Can you handle it? If it's too messy, I will handle it.

KengoTODA:
I've confirmed that po file needs no update, sorry for confusing you

PR #853, opened 2019-01-18
Util#getXMLType refactoring

inponomarev:
Refactored `Util#getXMLType` method based on [static analysis finding](https://www.viva64.com/en/b/0603/#ID0E5LDK). `counter` variable had no chance to be equal to or greater than 4 anyway.

Since this code worked unchanged since initial commit 70bd1096fd325c44eaad92e5e6a9c029d183c4aa in 2007, I think there is no point in 'correcting' it by actually limiting the search of opening tag by 4 lines. 'Limited search' will not work, e.g. in case of extensive comment or just empty lines in the beginning of the file.

Since this is not changing the code behaviour, do we need changelog entry?

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/Util.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/Util.java
@@ -385,14 +385,11 @@ public class Util {
         }
 
         in.mark(5000);
-        BufferedReader r = null;
         try {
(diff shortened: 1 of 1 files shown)


KengoTODA:
I feel this is not critical problem, then let me merge to 4.0 release later.

KengoTODA:

inponomarev:
👌

PR #900, opened 2019-03-05
bump up Gradle to the latest 5.2.1

KengoTODA:
It's better to use the latest Gradle instead of RC version.


KengoTODA:
This change does not affect the build result, it is just a build tool, not a part of product.

PR #854, opened 2019-01-18
fix FindNonShortCircuit#reportBug

inponomarev:
Fixed `FindNonShortCircuit#reportBug` method based on [static analysis finding](https://www.viva64.com/en/b/0603/#ID0E61EK). `sawNullTestVeryOld` variable had no impact on `priority`.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #902, opened 2019-03-05
fixed FindNonShortCircuit#reportBug

inponomarev:
This is a reincarnation of #854.

Fixed `FindNonShortCircuit#reportBug` method based on [static analysis finding](https://www.viva64.com/en/b/0603/#ID0E61EK). `sawNullTestVeryOld` variable had no impact on `priority`.

----

### DIFF
--- a/spotbugs/src/test/java/edu/umd/cs/findbugs/detect/FindNonShortCircuitTest.java
+++ b/spotbugs/src/test/java/edu/umd/cs/findbugs/detect/FindNonShortCircuitTest.java
@@ -20,23 +20,23 @@ public class FindNonShortCircuitTest {
                 AnalysisContext.setCurrentAnalysisContext(new AnalysisContext(new Project()));
             }
             FindNonShortCircuit check = new FindNonShortCircuit(bugReporter);
-            BugInstance bug = check.getBugInstance();
+            BugInstance bug = check.createBugInstance();
             assertEquals(FindNonShortCircuit.NS_NON_SHORT_CIRCUIT, bug.getType());
             assertEquals(Priorities.LOW_PRIORITY, bug.getPriority());
 
             check.sawDangerOld = true;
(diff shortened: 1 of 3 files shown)


inponomarev:
@KengoTODA done as you requested

spotbugs/src/test/java/edu/umd/cs/findbugs/detect/FindNonShortCircuitTest.java
```diff
@@ -0,0 +1,46 @@
...
+import edu.umd.cs.findbugs.Project;
+import edu.umd.cs.findbugs.ba.AnalysisContext;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class FindNonShortCircuitTest {
+
+    @Test
+    public void testBugTypeAndPriority() {
```
ThrawnCA: Ideally this name should be more self-documenting, especially for such a long unit test; what behavior does it expect?

inponomarev: Will  `testBugTypeAndPriorityDependOnSawBooleanFlags` name do?

I know this test is indecently long. 

But look at the terrible thing I had to do with static `AnalysisContext` just to run the test. So I decided to do it only once instead of four times. Having static state makes things non-testable, I think this is why I don't see any other non-trivial unit tests in `edu.umd.cs.findbugs.detect` package.

ThrawnCA: Have you tried JUnit's `@Before` and `@After` annotations?

inponomarev: @ThrawnCA Well you're right, I split the test.

Issue #694, opened 2018-07-10
SourceFilter does not work with the full file path


Irian81:
Curently "spotbugs/spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java" only works against plain source file name, no path only the file name, this makes the whole point of using it way less powerful than what it should be IMHO it should also consider the full file path.

Following my use case:

I was planning to exclude a set of automatically generated files that live in my maven project \target\generated-sources, from analysis, I wanted to create a generic rule when only files present in this folder and any subfolders are excluded.

To do so i created an exclude file with the following content:

&lt;FindBugsFilter&gt;
  &lt;Match&gt;
    &lt;Source name="~.&#42;generated-sources.&#42;.java"/&gt;
  &lt;/Match&gt;
&lt;/FindBugsFilter&gt;

This does not work as the SourceMatcher class only processes the file name, not the actual path.

I've attached an example maven project that demonstrates the issue, ideally I would like to exclude generated sources but retain the same warning that exists in the manually written classes.

I know I could define a package specific for generated classes, but this forces me to change the exclusion logic every time I add another package to the auto generation so not really ideal.

[testspotbugexclude.zip](https://github.com/spotbugs/spotbugs/files/2181862/testspotbugexclude.zip)

Irian81:
I've had a bit of spare time and managed to get something working for my specific use case, 4 files modified 2 in main spotbugs (SourceMatcher, and SourceLineAnnotation), and 2 in spotbugs-maven-plugin (SpotBugsMojo, and the pom.xml to depend on 4.0.0-SNAPSHOT).

Would you be willing to review a pull request for this??

I've gone ahead and forked the 2 repos so if you want you can look at the modifications there.

To get the sample project to work it should be sufficient to add the new <addSourceDirs>true</addSourceDirs> param and up the spotbugs-maven-plugin version to a 3.1.6-SNAPSHOT built with the "patch" in place.

ThrawnCA:
> Would you be willing to review a pull request for this??

I think we're always happy to review :).

Irian81:
Added PR #700 for spotbugs and PR #71 for maven.

these will most probably require checking for the SourceLineAnnotation modification sanity, and at least a new test case to be added, and potentially some documentation.

Let me know if you require something else to be added

KengoTODA:
closed by #973

sfc-gh-akazantsev:
[...]

PR #903, opened 2019-03-06
Add management for source filter using full source path (if available).

PhilMFischer:
Fixes Issue #694. 

Based on the work of the initial Pull Request #700 by @Irian81

Please be aware that there is another pull request of Irian for the maven part spotbugs/spotbugs-maven-plugin#71


CHANGELOG.md
```diff
@@ -113,6 +114,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * False positive: 'return value ignored' on Guavas Preconditions.checkNotNull() ([#578](https://github.com/spotbugs/spotbugs/issues/578))
 * spotbugs-ant Ant dependency in wrong scope ([#655](https://github.com/spotbugs/spotbugs/issues/655))
 
+* Add management for source filter using full source path, if available ([#694](https://github.com/spotbugs/spotbugs/issues/694))
+
```
KengoTODA: Remove this needless CHANGELOG entry

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
```diff
@@ -57,6 +57,18 @@ public boolean match(BugInstance bugInstance) {
         if (DEBUG) {
             System.out.println("Matching " + bugFileName + " with " + fileName + ", result = " + result);
```
KengoTODA: It is better to add your logic BEFORE this debug log, then you don't need to copy-and-paste it at Line 68~71.

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
```diff
@@ -118,6 +122,45 @@ public void match() throws Exception {
...
+        String sourceDir = "src/test/java/edu/umd/cs/findbugs/filter";
+        testProject.addSourceDirs(Collections.singletonList(sourceDir));
+
+        // add test project to SourceLineAnnotation
+        SourceLineAnnotation.generateRelativeSource(new File(sourceDir), testProject);
+
+        // regexp match source folder with project
+        SourceMatcher sm = sm = new SourceMatcher("~.*findbugs.*.java");
+        assertTrue(sm.match(bug));
+        sm = sm = new SourceMatcher("~.*notfound.*.java");
```
KengoTODA:
```suggestion
        sm = new SourceMatcher("~.*notfound.*.java");
```

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
```diff
@@ -118,6 +122,45 @@ public void match() throws Exception {
...
+        // setup a testing project with source directory, as of right now the source directory should really exist!!
+        Project testProject = new Project();
+        String sourceDir = "src/test/java/edu/umd/cs/findbugs/filter";
+        testProject.addSourceDirs(Collections.singletonList(sourceDir));
+
+        // add test project to SourceLineAnnotation
+        SourceLineAnnotation.generateRelativeSource(new File(sourceDir), testProject);
+
+        // regexp match source folder with project
+        SourceMatcher sm = sm = new SourceMatcher("~.*findbugs.*.java");
```
KengoTODA:
```suggestion
        SourceMatcher sm = new SourceMatcher("~.*findbugs.*.java");
```

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
```diff
@@ -118,6 +122,45 @@ public void match() throws Exception {
...
+        // clear test project from SourceLineAnnotation
+        SourceLineAnnotation.clearGenerateRelativeSource();
+
+        // setup test analysis context
+        AnalysisContext.setCurrentAnalysisContext(new AnalysisContext(testProject));
+
+        // regexp match source folder with analysis context
+        sm = new SourceMatcher("~.*findbugs.*.java");
+        assertTrue(sm.match(bug));
+        sm = sm = new SourceMatcher("~.*notfound.*.java");
```
KengoTODA:
```suggestion
        sm = new SourceMatcher("~.*notfound.*.java");
```

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
```diff
@@ -944,6 +945,43 @@ public String getSourcePath() {
...
+        }
+        return getSourcePath();
+    }
+
+    private @CheckReturnValue SourceFinder getSourceFinder() {
+        Project project = myProject.get();
+        if (project != null) {
+            return project.getSourceFinder();
+        }
+        AnalysisContext currentAnalysisContext = AnalysisContext.currentAnalysisContext();
```
ThrawnCA: I would probably write this as

    return Optional.ofNullable(AnalysisContext.currentAnalysisContext())
        .map(AnalysisContext::getSourceFinder)
        .orElse(null);

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
```diff
@@ -118,6 +122,45 @@ public void match() throws Exception {
         assertFalse(sm.match(bug));
     }
 
+    @Test
+    public void testRealPathMatchRegexp() throws Exception {
```
ThrawnCA: A more descriptive name would be good. What behavior do we expect?

Also, this test is very long and should probably be split.

CHANGELOG.md
PhilMFischer: Agree, fixed that.

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
PhilMFischer: Reordered Code as suggested.

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
PhilMFischer: Obvious one, fixed it.

spotbugs/src/main/java/edu/umd/cs/findbugs/SourceLineAnnotation.java
PhilMFischer: I  Like that one, added to code.

spotbugs/src/test/java/edu/umd/cs/findbugs/filter/SourceMatcherTest.java
PhilMFischer: Well, tried to split the test case. Not super happy with it. Any suggestions for it?

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
```diff
@@ -54,6 +55,16 @@ public boolean match(BugInstance bugInstance) {
             return false;
         }
         boolean result = fileName.match(bugFileName);
+        
+        // if no result try to match with the full path as well
+        if (!result && bugInstance.getPrimarySourceLineAnnotation().isSourceFileKnown()) {
+          bugFileName = bugInstance.getPrimarySourceLineAnnotation().getRealSourcePath();
+          if(bugFileName == null || bugFileName.isEmpty()){
+            return false;
```
KengoTODA: When we achieve this line, the debug log will be skipped. better to use:

```java
if(bugFileName != null && !bugFileName.isEmpty()){
  result = fileName.match(bugFileName);
}

Comment (PhilMFischer) [reply spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java]: I actually refactored the whole method. The issue you pointed out, was applicable in two places. I think it should be betetr now.

**Comment by KengoTODA** [comment]
Could you resolve this conflict? @PhilMFischer

**Comment by PhilMFischer** [comment]
@KengoTODA Merged with Master to add Logger support instead of System.out for debugging purposes.

File: CHANGELOG.md [hunk 9]
```diff
@@ -16,6 +16,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Changed
 
+* Add management for source filter using full source path, if available ([#694](https://github.com/spotbugs/spotbugs/issues/694))
```
KengoTODA: Please move this entry to the latest `## Unreleased` section.

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
```diff
@@ -47,15 +47,25 @@ public SourceMatcher(String fileName) {
...
         String bugFileName = primaryClassAnnotation.getSourceFileName();
-        if(bugFileName == null || bugFileName.isEmpty()){
-            return false;
+        if(bugFileName != null && !bugFileName.isEmpty()){
+            result = fileName.match(bugFileName);
+            
+            // if no result try to match with the full path as well
+            if (!result && bugInstance.getPrimarySourceLineAnnotation().isSourceFileKnown()) {
+                bugFileName = bugInstance.getPrimarySourceLineAnnotation().getRealSourcePath();
+                result = fileName.match(bugFileName);
```
KengoTODA: Maybe we can omit this line? We overwrite the result at the 68th line.

CHANGELOG.md
PhilMFischer: Rebased and moved entry to unreleased

spotbugs/src/main/java/edu/umd/cs/findbugs/filter/SourceMatcher.java
PhilMFischer: Will have a look to it.

PhilMFischer: I removed the one from line #68 because of the following points:
1. the result is false by initialization. Only in case we have a valid bugfile name it will change.
2. than the result is written. in case it is negative we proceed checking with the full path of the file
3. finally the result is written again (overwritten).

If we remove the one in #64 and leave #68 we do an unneccesary check.

So i think, if there is no bug file name the result should be always false.

I updated the comments in the code, to help understand the logic.

KengoTODA:
Wmm the change in this PR contain old commits in the `master` branch...? At least it contains needless commits in `.travis.yml`. Could you check changed file?

PhilMFischer:
What exactly do you mean by old commits? I am confused by the changes displayed by GitHub. If i compare your master with my branch manually the .travis.yml files are identical :-/

PhilMFischer:
It seems that there is something very inconsistant. I reckomend not to merge this branch. I will investigate in the mean time and get back to you.

PhilMFischer:
Fixed issues with incorrect merge in PR #973.

PR #907, opened 2019-03-15
Replace usage of sysout with SLF4J

KengoTODA:
Keep replacing usage of sysout.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/BugRanker.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/BugRanker.java
@@ -30,6 +30,9 @@ import java.util.List;
 
 import javax.annotation.CheckForNull;
 
+import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/BugRanker.java
```diff
@@ -90,7 +93,7 @@
     /** Minimum value for user visible ranks (most relevant) */
     public static final int VISIBLE_RANK_MIN = 1;
 
-    static final boolean PLUGIN_DEBUG = Boolean.getBoolean("bugranker.plugin.debug");
+    private static final Logger LOG = LoggerFactory.getLogger(BugRanker.class);
```
ThrawnCA: Note that you can use `LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());` to get a class-agnostic logger statement, easier to copy and paste.

PR #932, opened 2019-04-29
Use openjdk instead of oraclejdk in travisci

wreulicke:
Use openjdk instead of oraclejdk because of license.

Ref: 
* https://askubuntu.com/questions/1136104/get-error-when-install-oracle-jdk-8-on-ubuntu-18-04

----

- [] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ dist: trusty
 
 jdk:
   - openjdk8
-  - oraclejdk11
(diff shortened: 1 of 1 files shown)


KengoTODA:
It's still free for development usage, the build could be broken due to its installer's instability. I'll merge this.

PR #933, opened 2019-04-30
Debug reset on exception

topolik:
Hello,

there's a problem with static DataAnalysis.DEBUG flag that is turned on when data analysis reaches MAX_ITERS iterations. In the end an exception is thrown which doesn't reset back the flag and ongoing analysis generates tons of data.

Reached the problem during aspectj-weaver.jar analysis (`Too many iterations TaintAnalysis on org.aspectj.weaver.AjcMemberMaker.interConstructor(org.aspectj.weaver.ResolvedType, org.aspectj.weaver.ResolvedMember, org.aspectj.weaver.UnresolvedType)`).

This fix is to reset the DEBUG back to it's original value when terminating the method analysis so other methods analyses have again DEBUG disabled.

Thank you.

----


topolik:
Thanks guys.

PR #939, opened 2019-05-04
Use Map.computeIfAbsent instead of Map.get and Map.put

wreulicke:
Use Map.computeIfAbsent instead of Map.get and Map.put

----


wreulicke:
Fixed conflicts.

@ThrawnCA 
Could you review again?

PR #941, opened 2019-05-09
Use standard charsets

wreulicke:
Use standard charsets

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/UTF8.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/UTF8.java
@@ -35,6 +35,7 @@ import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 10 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/UTF8.java
```diff
@@ -47,11 +48,6 @@
      *
      */
     private static final String UTF_8 = "UTF-8";
```
ThrawnCA: Could also drop this and use `StandardCharsets.UTF_8.name`

wreulicke: Oh, Thanks!

wreulicke: Fixed! 😄

wreulicke:
@ThrawnCA 
Sorry, I pushed because the CI is failed.
Could you review again?
Thanks.

wreulicke:
@ThrawnCA Thanks for your review! :smile:

PR #951, opened 2019-05-21
Release 4.0.0 beta2

KengoTODA:
See [the changelog](https://github.com/spotbugs/spotbugs/blob/cacf3d27950af50b7a95ae0ebfd75caeb58b3945/CHANGELOG.md) about changes in this release.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

[//]: # (rtdbot-start)

URL of RTD documents:
en: https://spotbugs.readthedocs.io/en/release-4.0.0-beta2/
ja: https://spotbugs.readthedocs.io/ja/release-4.0.0-beta2/

[//]: # (rtdbot-end)

PR #960, opened 2019-05-28
Provide better support for CheckerFramework annotations

badgersow:
Spotbugs doesn't recognize TYPE_USE annotations on method return values. In other words, if you use CheckerFramework annotation in the following context, the annotation is ignored on class parsing phase.

```java
@Nullable
String foo() {
  return null;
}
```

This PR adds a clause to the bytecode parser to check type annotations on method return types. It is done in the same way as similar change for method formal parameters [here](https://github.com/spotbugs/spotbugs/pull/816)

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2019-??-??
 
+### Added
+
+* Provide support for CheckerFramework @NonNull annotation
+* Recognize CheckerFramework type annotations on method return values ([#960](https://github.com/spotbugs/spotbugs/pull/960))
+
 ## 4.0.0-beta2 - 2019-05-21
 
 ### Fixed
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
@@ -73,6 +73,10 @@ public class TypeQualifierResolver {
 
     static final ClassDescriptor checkerFrameworkNullableDecl = DescriptorFactory.createClassDescriptor("org/checkerframework/checker/nullness/compatqual/NullableDecl");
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
+## 3.1.13
```
KengoTODA: do not add version like this, instead, add entries under `## Unreleased`.

badgersow: I've switched to master branch and filled my changes to existing `## Unreleased` section

badgersow:
@KengoTODA Thanks for quick and detailed feedback!

1. Done that.
2. I've added `TestCheckerFrameworkTypeAnnotations` which fails before my changes and passes after they've been done.
3. Thanks for the tip. I didn't changed this part and CheckerFramework `@Nullable` annotation is continuing to be treated as `@CheckForNull` annotation.
4. I've fixed `CHANGELOG.md` entries per your suggestion.

I've addressed all your comments. Could you please do another round of code review?

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
```diff
@@ -150,7 +154,9 @@ private static void resolveTypeQualifierNicknames(AnnotationValue value, LinkedL
                         || annotationClass.equals(androidxNonNull)
                         || annotationClass.equals(eclipseNonNull)
                         || annotationClass.equals(eclipseNonNullByDefault)
-                        || annotationClass.equals(intellijNotNull)) {
+                        || annotationClass.equals(intellijNotNull)
```
badgersow: This is required to recognize `@NonNull` annotation. For some reason we had support for `@Nullable`, but not for `@NonNull`

badgersow:
Thanks everyone!
When can I expect these changes to be available in maven central repository? Is there some way to trigger a milestone build?

KengoTODA:
I personally has an objective to release SpotBugs monthly. However feel free to raise a PR like #951 to release new version, then I'll merge and tag for release. Here is our release procedure, you may check:

* https://github.com/spotbugs/spotbugs/blob/master/RELEASE_PROCEDURE.md

PR #967, opened 2019-06-10
Fixing 'Inefficient use of String.indexOf(String)' in BugInfoView

dvmarcilio:
Hello,

I found two instances of **IIO_INEFFICIENT_INDEX_OF** in `BugInfoView`. Fortunately, the fix is very easy!

----

Should I add an entry into the changelog for this?


KengoTODA:
Good! Thank you.

PR #970, opened 2019-06-15
Experimental multi threading

KengoTODA:
Related with #249. This PR adds an experimental multi-thread analysis. It's deactivated by default because I already know many problems in current implementation such as:

1. `Profiler`, `ClassContext`, `BugReporter` and more classes are not thread-safe.
2. Some default detectors (`LoadOfKnownNullValue`, `Dataflow` and more) expect it runs right after other detector (so we cannot run them at the same time).
3. BCEL is not thread-safe.
4. Dependency among classes are complicated (f93c21a983bd6b8a3c87a59ab1eb7b613c268b09).

This PR just make it possible to run in parallel, to ease debugging.

## Basic approach

I follow the existing `ExecutionPlan` mechanism that orders detectors within each `AnalysisPass` based on ordering constraints specified in the plugin descriptor(s). For detail:

a. Each detector run on a single thread (so detector does not have to be thread-safe).
b. Run detectors in one `AnalysisPass` in parallel, to run database construction before running detectors.
c. Detectors that belongs to another `AnalysisPass` won't run at the same time.

## Remaining TODOs

I will work on following tasks in different PRs:

- [ ] Make `Profiler`, `ClassContext` and others thread-safe, or stop sharing them among threads (to solve 1). 8f332523e5107dc1e8fa2e554b8a6d2e9e98ba0e is rough but not complete idea.
- [ ] Update plugin descriptor to solve existing detector dependency issue (to solve 2).

I hope the problem number 3 (BCEL issue) cannot be blocker, because we do not use BCEL's `Repository` directly. We use an adapter named `AnalysisCacheToRepositoryAdapter` so we can take a lock in this class.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 * Provide support for CheckerFramework @NonNull annotation
 * Recognize CheckerFramework type annotations on method return values ([#960](https://github.com/spotbugs/spotbugs/pull/960))
+* The feature toggle `spotbugs.experimental.multiThread` for experimental multi-thread analysis
 
 ## 4.0.0-beta2 - 2019-05-21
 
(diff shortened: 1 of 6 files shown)


KengoTODA:
The coverage analysis has a problem, I will solve #971 first then reopen this PR.

PR #972, opened 2019-06-15
Experimental multi threading

KengoTODA:
Related with #249. This PR adds an experimental multi-thread analysis. It's deactivated by default because I already know many problems in current implementation such as:

1. `Profiler`, `ClassContext`, `BugReporter` and more classes are not thread-safe.
2. Some default detectors (`LoadOfKnownNullValue`, `Dataflow` and more) expect it runs right after other detector (so we cannot run them at the same time).
3. BCEL is not thread-safe.
4. Dependency among classes are complicated (f93c21a983bd6b8a3c87a59ab1eb7b613c268b09).

This PR just make it possible to run in parallel, to ease debugging.

## Basic approach

I follow the existing `ExecutionPlan` mechanism that orders detectors within each `AnalysisPass` based on ordering constraints specified in the plugin descriptor(s). For detail:

a. Each detector run on a single thread (so detector does not have to be thread-safe).
b. Run detectors in one `AnalysisPass` in parallel, to run database construction before running detectors.
c. Detectors that belongs to another `AnalysisPass` won't run at the same time.

## Remaining TODOs

I will work on following tasks in different PRs:

- [ ] Make `Profiler`, `ClassContext` and others thread-safe, or stop sharing them among threads (to solve 1). 8f332523e5107dc1e8fa2e554b8a6d2e9e98ba0e is rough but not complete idea.
- [ ] Update plugin descriptor to solve existing detector dependency issue (to solve 2).

I hope the problem number 3 (BCEL issue) cannot be blocker, because we do not use BCEL's `Repository` directly. We use an adapter named `AnalysisCacheToRepositoryAdapter` so we can take a lock in this class.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
@@ -1237,6 +1237,8 @@ public class FindBugs2 implements IFindBugsEngine, AutoCloseable {
             // Away we go!
 
             FindBugs.runMain(findBugs, commandLine);
+        } finally {
+            service.shutdown();
         }
     }
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
@@ -43,7 +43,6 @@ import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
-import org.apache.bcel.classfile.ClassFormatException;
 import org.dom4j.DocumentException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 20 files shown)


KengoTODA:
I will activate this PR after #941 because this PR contains the same changes with that. SonarQube marks this PR as failed, but actually I didn't changed the code in `FindBugs2.java` but moved it into stream API, so I think it is OK to merge...

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -1062,33 +1095,40 @@ private void analyzeApplication() throws InterruptedException {
...
-                            }
-                            if (isHuge && !FirstPassDetector.class.isAssignableFrom(detector.getClass())) {
-                                continue;
-                            }
-                            if (DEBUG) {
-                                System.out.println("Applying " + detector.getDetectorClassName() + " to " + classDescriptor);
-                                // System.out.println("foo: " +
-                                // NonReportingDetector.class.isAssignableFrom(detector.getClass())
-                                // + ", bar: " + detector.getClass().getName());
+                        Collection<Callable<Void>> tasks = Arrays.stream(detectorList).map(detector ->
```
Charmik: sorry, @KengoTODA don't you need to write thread-safe AnalysisCache for it too?
What will happen if 2 detectors ask for the same analysis from cache?

KengoTODA: Yes you're right. As documented in the description of this PR, this change isn't complete to run detectors in parallel.

Charmik: > Yes you're right. As documented in the description of this PR, this change isn't complete to run detectors in parallel.

do you  have any idea how to finish this task with situation about BCEL? as I understand moving to  ASM is not really possible in reality...so how to solve it? I think it's the main problem about it.

raghav-deepsource: Sorry to necro, I'm wondering about the reasons it is not easy to migrate from BCEL to ASM. Are there any docs regarding why this isn't possible?

iloveeclipse: > Sorry to necro, I'm wondering about the reasons it is not easy to migrate from BCEL to ASM.

It is simple: these API's are completely different - one id DOM based, another one is visitor based (the "fast" ASM API), so porting BCEL based code to ASM is almost a full rewrite. Of course one could use ASM tree API that is somewhat similar to BCEL code except missing ConstantPool access, but as far as I remember we will still need a replacement for class repository from BCEL that of course doesn't exists at all in ASM API.

PR #977, opened 2019-06-23
Apply formatter to spotbugs/test, spotbugs/xml, spotbugs/workflow, spotbugs/visitclass, spotbugs/util

wreulicke:
Apply formatter below.

* spotbugs/testjava
* spotbugs/main/java/edu/umd/cs/findbugs/xml
* spotbugs/main/java/edu/umd/cs/findbugs/workflow
* spotbugs/main/java/edu/umd/cs/findbugs/visitclass
* spotbugs/main/java/edu/umd/cs/findbugs/util

This PR is cherry-picked from https://github.com/wreulicke/spotbugs/tree/feature/apply-spotless


spotbugs/src/main/java/edu/umd/cs/findbugs/util/Util.java
```diff
@@ -338,7 +341,7 @@ public static String getFileExtensionIgnoringGz(File f) {
 
     public static boolean isPowerOfTwo(int i) {
         return i > 0
-            && (i | (i - 1)) + 1 == 2 * i;
```
ThrawnCA: I don't really agree with the extra indentation here, but eh, if it means we can apply consistent formatting, it's a win.

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/AnnotationVisitor.java
```diff
@@ -143,8 +143,8 @@ public void visitSyntheticParameterAnnotation(int p, boolean runtimeVisible) {
         }
         return values;
     }
-
-
+    
+
```
ThrawnCA: Should we just delete a line here?

wreulicke: I do not think so. It is fine like this.
IMO, To remove comment-out lines is better than to delete a line.

I would like to separate issue.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/Util.java
wreulicke: > I don't really agree with the extra indentation here
I think so, but i seem that extra indentations is contained in code base.
so, I will merge for consistent formatting.

PR #980, opened 2019-06-23
Apply formatter to edu/umd/cs/findbugs 

wreulicke:
Apply formatter to below

* edu/umd/cs/findbugs/classfile
* edu/umd/cs/findbugs


ThrawnCA:
For some reason, SonarQube says that this change will reduce test coverage, but that seems very odd for changes that are mostly whitespace.

PR #981, opened 2019-06-24
introduce spotless for consistent formatting

wreulicke:
Introduce spotless for consitsent formatting.

We should use formatter for consistently.
We can use `spotlessApply` task for formatting

We ignore spotbugsTestCases for test.

----


KengoTODA:
Changes seems really nice. Thanks for your contribution!
But why don't we call `spotlessCheck` task in CI process?

wreulicke:
I will add `spotlessCheck` task in CI process

wreulicke:
@ThrawnCA thanks

iloveeclipse:
I'm not sure I'm the only one using Eclipse & Windows,  but I can't `gradle build` on Windows due spotless errors on spotbugs Eclipse code. After "fixing" the code via `./gradlew spotlessApply` git shows 1333 checked out files, most of them due CRLF conversion. Is this expected?

KengoTODA:
Git can manage line ends so could you visit [this page](https://help.github.com/en/articles/dealing-with-line-endings) to check your Git is properly configured?

iloveeclipse:
I'm not sure what is this all about. I don't want to convert source line ending to windows style, I don't want to set `git config core.autocrlf` to true, because it is wrong, I'm only trying to build spotbugs on Windows, which worked flawlessly before, and isn't working now. I'm sorry, this is not something *I* should configure - if the spotless silly enough to not work on Windows, it should be either disabled or configured *for all* in the repository.

iloveeclipse:
I've created #994 and #995 should fix it.

PR #986, opened 2019-06-29
Optimisation : replace Collection.addAll by constructor

Dichotomia:
----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2019-??-??
 
-* Replace Collection addAll by constructor
 ## 4.0.0-beta3 - 2019-06-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2019-??-??
 
+* Replace Collection addAll by constructor
 ## 4.0.0-beta3 - 2019-06-24
 
 ### Added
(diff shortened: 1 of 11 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -742,8 +742,7 @@ private void buildReferencedClassSet() throws InterruptedException {
         }
         Set<String> referencedPackageSet = new HashSet<>();
 
-        LinkedList<ClassDescriptor> workList = new LinkedList<>();
-        workList.addAll(appClassList);
+        LinkedList<ClassDescriptor> workList = new LinkedList<>(appClassList);
```
ThrawnCA: While this is being cleaned up, you could probably also change the variable to use the interface type, `List`, instead of the concrete type.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2019-??-??
 
+* Replace Collection addAll by constructor
```
ThrawnCA: I'm not sure that this really needs to be in the changelog since it's invisible from the outside.

Dichotomia: I just followed the template. I am going to remove this changelog.

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
Dichotomia: Yes indeed it would be much better. But i would like to do such things by regrouping modification by type.

Dichotomia:
@ThrawnCA I have removed the Changelog section

PR #1000, opened 2019-08-14
958: default.xsl declares it is a 2.0 stylesheet, but it appears to h…

chrismiceli:
…ave issues with a 2.0 processor

fix conflict per error message:
SEPM0009 omit-xml-declaration is set to yes, and standalone has a value other than omit.
Fix tab/space mixed together in default.xsl.

----

- [x ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/spotbugs/src/xsl/default.xsl
+++ b/spotbugs/src/xsl/default.xsl
@@ -50,8 +50,8 @@
 	method="xml"
 	indent="yes"
 	omit-xml-declaration="yes"
-	standalone="yes"
-    doctype-system="http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"
+	standalone="omit"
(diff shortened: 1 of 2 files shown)


spotbugs/src/xsl/default.xsl
```diff
@@ -50,8 +50,8 @@
 	method="xml"
 	indent="yes"
 	omit-xml-declaration="yes"
-	standalone="yes"
-    doctype-system="http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"
+	standalone="omit"
```
chrismiceli: This is the only actual change. The rest is spaces to tabs to align with the rest of the majority of the file.

chrismiceli:
Is there anything else left of me before this can be merged?

PR #1008, opened 2019-09-04
fix: update relaxing nullness check for nullable annotation

LupusUmbrae:
----

- [/ ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 
 * default.xsl declares it is a 2.0 stylesheet, but it appears to have issues with a 2.0 processor ([#958](https://github.com/spotbugs/spotbugs/issues/958))
+* update relaxing nullness check for nullable annotation
 
 ## 4.0.0-beta3 - 2019-06-24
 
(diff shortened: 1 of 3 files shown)


KengoTODA:
I'm not sure we can relax this spec (implementation).

FindBugs and JSR305 has no clear documentation around this spec, but we've already got used to work with existing rules, so I'm personally not so positive to merge this. I wish other teammates will give their opinion in this ticket.

LupusUmbrae:
It seems odd to me to relax the null checking rule for `@CheckForNull` but then not to do so for the `@Nullable` which seems very much like the same use case.

Currently we have to have a locally modified version of Spotbugs as we use `@Nullable` and without this change it would raise new errors :(

ThrawnCA:
@LupusUmbrae No, `@Nullable` is not the same as `@CheckForNull`. `@Nullable` is supposed to be equivalent to having no annotation at all.

LupusUmbrae:
@ThrawnCA No part of the current documentation on `@Nullable` say it's equivalent to having no annotation. If you don't use it then why have it at all?

From our point of view, the annotations should be read from the caller's perspective, so `@Nullable` for parameters, and `@CheckForNull` for return values.

Is there any documentation from your point of view what each of the annotations mean. Another from our team pointed to https://checkerframework.org/manual/#findbugs-nullable which implies `@Nullable` with spotbugs means something different

ThrawnCA:
@LupusUmbrae Documented at http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/Nullable.html

LupusUmbrae:
The documentation:

> The annotated element could be null under some circumstances. In general, this means developers will have to read the documentation to determine when a null value is acceptable and whether it is necessary to check for a null value. When this annotation is applied to a method it applies to the method return value.

Are you saying that

> The annotated element could be null under some circumstances

Is what means it shouldn't be checked in this case? The code seems to be treating `@Nullable` the same as `@NonNull` which seems wrong. One example of this failing is for equals methods:

```
public boolean equals(@Nullable Object obj) {..}
```

The above fails the check with `NP_METHOD_PARAMETER_TIGHTENS_ANNOTATION`

PR #1010, opened 2019-09-10
Add checksum-dependency-plugin for verification of plugin/dependency checksums

vlsi:
`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin
See https://medium.com/@vladimirsitniko/dependency-verification-checksum-vs-pgp-582e76207019?sk=7485298b76eaf9f935b899b002f4c3b5

Note: upcoming versions of ASM would publish PGP signatures. See https://gitlab.ow2.org/asm/asm/issues/317878

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,13 +55,13 @@ install:
 
 script:
   - jdk_switcher use openjdk8
-  - ./gradlew classes testClasses -S
+  - ./gradlew classes testClasses -S -PchecksumPrint -PchecksumFailOn=build_finish
   - if [[ $TRAVIS_JDK_VERSION == "openjdk11" ]]; then export JAVA_HOME=$HOME/openjdk11; fi
-  - ./gradlew -v --no-daemon
-  - ./gradlew build smoketest -x signArchives -S --no-daemon
+  - ./gradlew -v --no-daemon -PchecksumPrint -PchecksumFailOn=build_finish
+  - ./gradlew build smoketest -x signArchives -S --no-daemon -PchecksumPrint -PchecksumFailOn=build_finish
   - jdk_switcher use openjdk8
-  - if [[ $TRAVIS_JDK_VERSION == "openjdk8" ]]; then ./gradlew sonarqube -S; fi
-  - if [[ $TRAVIS_JDK_VERSION == "openjdk8" ]]; then ./gradlew spotlessCheck -S; fi
+  - if [[ $TRAVIS_JDK_VERSION == "openjdk8" ]]; then ./gradlew sonarqube -S -PchecksumPrint -PchecksumFailOn=build_finish; fi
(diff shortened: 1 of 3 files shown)


KengoTODA:
For me, this plugin seems needless. Does anyone have motivation to merge this PR?

ThrawnCA:
I suppose the use case would be preventing an attacker from spoofing/infiltrating repositories and serving tampered libraries.

KengoTODA:
Yes but [this plugin is not so widely used](https://github.com/search?q=com.github.vlsi.gradle%3Achecksum-dependency-plugin&type=Code) and I think HTTPS is enough in most cases. I can understand [this post](https://medium.com/@vladimirsitniko/dependency-verification-checksum-vs-pgp-582e76207019) but if we do not trust Maven Central we have more parts to handle with care (e.g. Travis, GitHub and my cert file to sign Eclipse plugin). For a small and volunteer product like us, I think it's a kind of overkill... just my personal opinion.

vlsi:
Hi. I hope you are doing well. You won't hurt me (much) if you close the PR :)
However, I have something new for you.

> if we do not trust Maven Central we have more parts to handle with care

You are right. However, in practice, it is never enough safety :)

You might be interested that [Gradle 6.2 introduces in-core dependency verification](https://github.com/gradle/gradle/issues/10443#issuecomment-568741472)

The documentation can be reviewed here: https://github.com/gradle/gradle/pull/11755


Here are some findings by Cédric Champeau: https://twitter.com/CedricChampeau/status/1209432835842617347 , https://twitter.com/CedricChampeau/status/1209432067097997313, https://twitter.com/CedricChampeau/status/1209143476740403201

So, for now, the options are like "upgrade to Gradle 6.2 and use in-core verification" or "use checksum-dependency-plugin.

From what I know Gradle would cover more cases when compared with `checksum-dependency-plugin`. For instance, it will be able to verify `pom.xml` which are implicitly fetched by Gradle when resolving transitive dependencies and probably other cases.

Some bits can be previewed in the current release candidates, release nightly builds and master nightly builds (see https://gradle.org/releases/ )

It would be nice if you could preview the feature and provide your feedback.

KengoTODA:
I prefer the standard way, then I vote for waiting Gradle 6.2. :)

hazendaz:
closing as 4 years without applying.  http is blocked now to central and only https now.  Think we also don't need this at least for this project.  This would seem more appropriate for downstream project usage that does not trust their sources.  From our perspective here, even if we had some highjacked item on usage, what exactly would that cause here?  Nothing really, we are not the final usage.

At any rate, if there is a need to explore again, please refresh from latest base and reopen new PR.

PR #1022, opened 2019-10-07
Introduce reproducible-builds

KengoTODA:
This change introduces [reproducible builds](https://reproducible-builds.org/) and stabilizes  distribution files (zip, tgz). For instance, in this branch the md5 hash of `spotbugs/build/distributions/spotbugs-4.0.0-SNAPSHOT.zip` is always 1d00579fbd62cf5c13959e7657f2e4c1. I've confirmed it in my macOS, and will check even in Windows and Ubuntu.

This change lets third-party verify that our official distribution is not cracked and secure.

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -6,3 +6,11 @@ tasks.withType(JavaCompile) {
 tasks.withType(Javadoc) {
     options.encoding = 'UTF-8'
 }
+
+// make the build reproducible https://reproducible-builds.org/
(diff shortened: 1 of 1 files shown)


KengoTODA:
> I've confirmed it in my macOS, and will check even in Windows and Ubuntu.

Confirmed that Ubuntu 18.04.3 (Linux 4.15.0-64-generic amd64) generates the same binary.

PR #1024, opened 2019-10-16
Closes #522 - Supported annotations are not documented

### LINKED ISSUE #522: Supported annotations are not documented?
I cannot find any equivalent to http://findbugs.sourceforge.net/manual/annotations.html for SpotBugs. 

Can such a page be added to the SpotBugs documentation?

jscancella:
This should close issue #522 - Supported annotations are not documented. I used the information from https://javadoc.io/doc/com.github.spotbugs/spotbugs-annotations/4.0.0-beta4 to populate the descriptions.

### DIFF
--- a/docs/annotations.rst
+++ b/docs/annotations.rst
@@ -1,0 +1,108 @@
+Annotations
+===========
+
+SpotBugs supports several annotations to express the developer's intent so that SpotBugs can issue warnings more appropriately.
+Annotations for SpotBugs (mostly deprecated except for SuppressFBWarnings).
+
+CheckForNull
+---------------------------
+The annotated element might be null, and uses of the element should check for null.
+
+CheckReturnValue
+---------------------------
+This annotation is used to denote a method whose return value should always be checked when invoking the method.
+
+CleanupObligation
+---------------------------
+Mark a class or interface as a resource type requiring cleanup.
+
+CreatesObligation
+---------------------------
+Mark a constructor or method as creating a resource which requires cleanup.
+
+DefaultAnnotation
+---------------------------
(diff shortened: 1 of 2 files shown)


docs/annotations.rst
```diff
@@ -0,0 +1,108 @@
...
+---------------------------
+The annotated element must not be null.
+
+NoWarning (Deprecated)
+---------------------------
+The annotation based approach is useless for lambdas.
+
+Nullable
+---------------------------
+The annotated element could be null under some circumstances.
```
ThrawnCA: This should mention that it is treated the same way as not being annotated.

jscancella: Updated based on your feedback.

KengoTODA:
I'm building the latest code in my local, but it takes time due to bad network performance... wait a moment, sorry.

KengoTODA:
still building :sob: I'll check it tomorrow...

jscancella:
> still building 😭 I'll check it tomorrow...

Man, and I thought my old Intel Celeron laptop was slow... What are you using to build it? 
I did notice that the dockerfile does a lot of setup steps in it, maybe we could reduce them?
instead of: pull docker image -> run ubuntu updates -> run other setup/install commands -> build documentation
we create a standard docker image (or use someone else's) and it becomes: pull docker image -> build documentation.

KengoTODA:
Confirmed that we have no error during the build. Thanks for your contribution!

About the build speed, the best is having an official image of sphinx, and [there is a thread for it](https://groups.google.com/forum/#!msg/sphinx-users/MwneIM7OSGM/wTe-vM-hEAAJ) but we have no progress recently. I will push them later.

PR #1042, opened 2019-11-27
Add failing test case for UWF_UNWRITTEN_FIELD false negative

granthenke:
Adds a test case to show a false negative for UWF_UNWRITTEN_FIELD
when `assert` is used or when Slf4j’s `LoggerFactory.getLogger(Foo.class)`
is used.

----


granthenke:
This isn't intended to be committed, but just to illustrate the issue.

granthenke:
Digging in a little, adding `assert` prevents `unwrittenField` from being added to `declaredFields` [here](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java#L868) because `xFactory.isReflectiveClass(classDescriptor)` is `true`. 

The same thing happens when `LoggerFactory.getLogger(UnreportedUnwrittenField.class);` is used.

granthenke:
It looks like it is marked as a reflective class [here](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ReflectiveClasses.java#L53) because of an LDC (18) op code in the case of `assert` and `LoggerFactory.getLogger(UnreportedUnwrittenField.class);`.

FWIW the test passes after removing the `!xFactory.isReflectiveClass(classDescriptor)` check [here](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java#L867).

granthenke:
In case it's helpful this behavior was introduced in this commit: https://github.com/spotbugs/spotbugs/commit/9ae2687e771dbd927e11b6a19ad23de3419fe7c8#diff-bfcd2c82604fc6cb72facf4b143d8394

wreulicke:
@granthenke Thank you for your PR.
I seem that CI is failing by formatting problem.
Could you apply `./gradlew spotlessApply` in your terminal for code format and commit the changes.

PR #1080, opened 2020-02-08
Replace Travis CI with GitHub Actions for stable deployment

KengoTODA:
Recently the publish task from Travis CI is quite unstable, it makes 4~5 staging registries and sometimes lack necessary files on Sonatype Central Nexus.

When I use a node on Google Cloud the publish task usually becomes stable, so hope that using another cloud service solves the problem.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,37 @@
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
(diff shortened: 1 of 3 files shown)


KengoTODA:
LF chars in GPG key is removed by GitHub when I store it onto Secret.
I will raise another PR to fix build failure like https://github.com/spotbugs/spotbugs/runs/435452563

PR #1081, opened 2020-02-10
Improve GPG key decryption phase to ship artifacts

KengoTODA:
Current way to decrypt GPG key file has problem, so replace it with the way introduced by [GitHub official document](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#limits-for-secrets).

refs https://github.com/spotbugs/spotbugs/pull/1080#issuecomment-583969836


KengoTODA:
Thanks for your review! Artifacts have been successfully deployed to https://oss.sonatype.org/content/repositories/snapshots/com/github/spotbugs/spotbugs/4.0.0-SNAPSHOT/

Issue #1077, opened 2020-02-03
XSL issue in 4.0.0-RC1


davidburstrom:
While using 4.0.0-RC1, every SpotBugs task emits the following warning when HTML reports are used. The build is running Gradle 6.1.1 and SpotBugs Gradle plugin 3.0.0:
```
> Task :spotbugsMain
Warning at xsl:variable on line 348 column 57 of default.xsl:
  SXWN9001: A variable with no following sibling instructions has no effect
Warning at xsl:variable on line 351 column 57 of default.xsl:
  SXWN9001: A variable with no following sibling instructions has no effect
```

davidburstrom:
I should add that this happens on JDK 8-13.

cmuchinsky:
Seeing the same with the GA version (JDK 8).

renatomefi:
Still happens on JDK 13:
```
> Task :spotbugsMain
Warning at xsl:variable on line 348 column 57 of default.xsl:
  SXWN9001: A variable with no following sibling instructions has no effect
Warning at xsl:variable on line 351 column 57 of default.xsl:
  SXWN9001: A variable with no following sibling instructions has no effect
```

g-dx:
@renatomefi - the fix is in `master` but hasn’t been released yet.

paulk-asert:
Any eta for next release?

PR #1088, opened 2020-02-17
Correct XSL variable declaration & assignment

g-dx:
Resolves Saxon warning `SXWN9001: A variable with no following sibling instructions has no effect`

----


g-dx:
Fixes #1077

PR #1098, opened 2020-03-09
Allow any ant task or command-line user to specify large values or many arguments

mkienenb:
Originally submitted as

Ant task parses stdin as argv
https://github.com/findbugsproject/findbugs/pull/61

Here's a maintainable way forward to allow any findbugs ant task (or command-line user) to specify large values or many arguments. I started with the idea used by -auxclasspathFromInput and expanded it to be more generic so that it could work for any option. Options are parsed from the command-line and if -parseStdinAsOptions is specified, options are then parsed from standard input, which each line considered as a separate argument. Since sourcePath was the option which was not working for me in the findbugs ant task, I updated both it and auxClasspath to use this new mechanism from the ant task. It is a simple matter to apply the same change to other options. Candidates could include class, visitors, and my proposed excludePath and includePath elements. This change maintains backward compatibility although the use of -parseStdinAsOptions will be incompatible with anything else using standard input.

### DIFF
--- a/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
+++ b/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
@@ -95,6 +95,8 @@ public abstract class AbstractFindBugsTask extends Task {
 
     public String execResultProperty = "edu.umd.cs.findbugs.anttask.AbstractFindBugsTask" + "." + RESULT_PROPERTY_SUFFIX;
 
+    private StringBuilder inputStringBuilder = null;
+
     /**
      * Constructor.
      */
@@ -388,8 +390,23 @@ public abstract class AbstractFindBugsTask extends Task {
      * Sets the given string to be piped to standard input of the FindBugs JVM
      * upon launching.
      */
+    protected void appendToInputString(String input) {
+        if (null == inputStringBuilder) {
+            inputStringBuilder = new StringBuilder();
+        }
+        inputStringBuilder.append(input);
+    }
+
+    protected void appendToInputString(char input) {
+        if (null == inputStringBuilder) {
+            inputStringBuilder = new StringBuilder();
+        }
+        inputStringBuilder.append(input);
+    }
+
     protected void setInputString(String input) {
-        findbugsEngine.setInputString(input);
+        inputStringBuilder = new StringBuilder();
+        inputStringBuilder.append(input);
     }
 
     /**
@@ -404,6 +421,10 @@ public abstract class AbstractFindBugsTask extends Task {
         configureFindbugsEngine();
 
         beforeExecuteJavaProcess();
+
+        if (null != inputStringBuilder) {
+            findbugsEngine.setInputString(inputStringBuilder.toString());
+        }
 
         if (getDebug()) {
             log(getFindbugsEngine().getCommandLine().describeCommand());
--- a/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
+++ b/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
@@ -901,10 +901,24 @@ public class FindBugsTask extends AbstractFindBugsTask {
             addArg(excludeFile.getPath());
         }
         if (excludePath != null) {
-            String[] result = excludePath.toString().split(java.io.File.pathSeparator);
-            for (String element : result) {
-                addArg("-exclude");
-                addArg(element);
+            String excludePathValue = excludePath.toString();
+            if (!excludePathValue.isEmpty()) {
+                if (excludePathValue.length() > 100) {
+                    addArg("-parseStdinAsOptions");
+                    String[] result = excludePathValue.split(java.io.File.pathSeparator);
+                    for (int x = 0; x < result.length; x++) {
+                        appendToInputString("-exclude");
+                        appendToInputString('\n');
+                        appendToInputString(result[x]);
+                        appendToInputString('\n');
+                    }
+                } else {
+                    String[] result = excludePathValue.split(java.io.File.pathSeparator);
+                    for (int x = 0; x < result.length; x++) {
+                        addArg("-exclude");
+                        addArg(result[x]);
+                    }
+                }
             }
         }
         if (includeFile != null) {
@@ -912,10 +926,24 @@ public class FindBugsTask extends AbstractFindBugsTask {
             addArg(includeFile.getPath());
         }
         if (includePath != null) {
-            String[] result = includePath.toString().split(java.io.File.pathSeparator);
-            for (String element : result) {
-                addArg("-include");
-                addArg(element);
+            String includePathValue = includePath.toString();
+            if (!includePathValue.isEmpty()) {
+                if (includePathValue.length() > 100) {
+                    addArg("-parseStdinAsOptions");
+                    String[] result = includePathValue.split(java.io.File.pathSeparator);
+                    for (int x = 0; x < result.length; x++) {
+                        appendToInputString("-include");
+                        appendToInputString('\n');
+                        appendToInputString(result[x]);
+                        appendToInputString('\n');
+                    }
+                } else {
+                    String[] result = includePathValue.split(java.io.File.pathSeparator);
+                    for (int x = 0; x < result.length; x++) {
+                        addArg("-include");
+                        addArg(result[x]);
+                    }
+                }
             }
         }
         if (visitors != null) {
--- a/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
+++ b/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
@@ -101,6 +101,11 @@ import edu.umd.cs.findbugs.ExitCodes;
 
 public class FindBugsTask extends AbstractFindBugsTask {
 
+    /**
+     * Arguments longer than this value will be passed via STDIN
+     */
+    private static final int MAXIMUM_DIRECTLY_PASSED_ARGUMENT_LENGTH = 100;
+
     private String effort;
 
     private boolean conserveSpace;
@@ -903,7 +908,7 @@ public class FindBugsTask extends AbstractFindBugsTask {
         if (excludePath != null) {
             String excludePathValue = excludePath.toString();
             if (!excludePathValue.isEmpty()) {
-                if (excludePathValue.length() > 100) {
+                if (excludePathValue.length() > MAXIMUM_DIRECTLY_PASSED_ARGUMENT_LENGTH) {
                     addArg("-parseStdinAsOptions");
                     String[] result = excludePathValue.split(java.io.File.pathSeparator);
                     for (int x = 0; x < result.length; x++) {
@@ -928,7 +933,7 @@ public class FindBugsTask extends AbstractFindBugsTask {
         if (includePath != null) {
             String includePathValue = includePath.toString();
             if (!includePathValue.isEmpty()) {
-                if (includePathValue.length() > 100) {
+                if (includePathValue.length() > MAXIMUM_DIRECTLY_PASSED_ARGUMENT_LENGTH) {
                     addArg("-parseStdinAsOptions");
                     String[] result = includePathValue.split(java.io.File.pathSeparator);
                     for (int x = 0; x < result.length; x++) {
@@ -969,9 +974,13 @@ public class FindBugsTask extends AbstractFindBugsTask {
                 String unreadReference = auxClasspath.toString();
                 String auxClasspathString = auxClasspath.toString();
                 if (!auxClasspathString.isEmpty()) {
-                    if (auxClasspathString.length() > 100) {
-                        addArg("-auxclasspathFromInput");
-                        setInputString(auxClasspathString);
+                    if (auxClasspathString.length() > MAXIMUM_DIRECTLY_PASSED_ARGUMENT_LENGTH) {
+                        addArg("-parseStdinAsOptions");
+
+                        appendToInputString("-auxclasspath");
+                        appendToInputString('\n');
+                        appendToInputString(auxClasspathString);
+                        appendToInputString('\n');
                     } else {
                         addArg("-auxclasspath");
                         addArg(auxClasspathString);
@@ -982,8 +991,20 @@ public class FindBugsTask extends AbstractFindBugsTask {
             }
         }
         if (sourcePath != null) {
-            addArg("-sourcepath");
-            addArg(sourcePath.toString());
+            String sourcePathValue = sourcePath.toString();
+            if (!sourcePathValue.isEmpty()) {
+                if (sourcePathValue.length() > MAXIMUM_DIRECTLY_PASSED_ARGUMENT_LENGTH) {
+                    addArg("-parseStdinAsOptions");
+
+                    appendToInputString("-sourcepath");
+                    appendToInputString('\n');
+                    appendToInputString(sourcePathValue);
+                    appendToInputString('\n');
+                } else {
+                    addArg("-sourcepath");
+                    addArg(sourcePathValue);
+                }
+            }
         }
         if (outputFileName != null) {
             addArg("-outputFile");
--- a/spotbugs/src/doc/manual.xml
+++ b/spotbugs/src/doc/manual.xml
@@ -928,6 +928,18 @@ These options are only accepted by the Text User Interface.
   </varlistentry>
 
   <varlistentry>
+  <term><command>-parseStdinAsOptions</command> </term>
+  <listitem>
+    <para>
+    Read additional options from from standard input, with each new line as a separate argument.
+    This option is incompatible with other arguments using standard input,
+    such as -auxclasspathFromInput and reading files or data from standard input.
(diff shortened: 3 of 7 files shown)


spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
```diff
@@ -388,8 +390,23 @@ protected void addArg(String arg) {
...
      * upon launching.
      */
+    protected void appendToInputString(String input) {
+        if (null == inputStringBuffer) {
+            inputStringBuffer = new StringBuffer();
+        }
+        inputStringBuffer.append(input);
+    }
+
+    protected void appendToInputString(char input) {
```
ThrawnCA: Perhaps this method should return `this` so it can be chained?

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
```diff
@@ -95,6 +95,8 @@ public String getValue() {
 
     public String execResultProperty = "edu.umd.cs.findbugs.anttask.AbstractFindBugsTask" + "." + RESULT_PROPERTY_SUFFIX;
 
+    private StringBuffer inputStringBuffer = null;
```
ThrawnCA: If we need synchronization, then the methods that initialise this field should be synchronized.

If we don't need synchronization, it's cheaper to use a `StringBuilder`.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
```diff
@@ -342,9 +354,33 @@ private int parse(String argv[], boolean dryRun) throws IOException, HelpRequest
...
+        while (true) {
+            String s = in.readLine();
+            if (s == null) {
+                break;
+            }
+            stdinArgvList.add(s);
+        }
+        in.close();
+
+        String[] stdinArgv = stdinArgvList.toArray(new String[stdinArgvList.size()]);
```
ThrawnCA: No need to define a local variable when we can just return it.

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
mkienenb: I suspect that my original thinking was that, because these are not private methods, there was no way of knowing if some pre-existing caller might need these to be synchronized.   But yes, you are correct that I missed synchronizing the initialization.

Reconsidering it now, it makes more sense to push the synchronization to the caller if that is needed.   There's nothing formally declared that any of these methods are safe to call from multiple threads.  I'll change it to StringBuilder.

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
mkienenb: It's not something that I've needed.  I'm not opposed to it, but it's a backward-compatible change from `void` to `this` later.  Changing it from `this` to `void` would not be.  I'd prefer to leave that change to someone who needed it.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
```diff
@@ -342,9 +354,32 @@ private int parse(String argv[], boolean dryRun) throws IOException, HelpRequest
...
         return arg;
     }
 
+    private String[] readStdinAsArgv() throws IOException {
+        List<String> stdinArgvList = new ArrayList<String>();
+
+        BufferedReader in = UTF8.bufferedReader(System.in);
+        while (true) {
+            String s = in.readLine();
+            if (s == null) {
```
ThrawnCA: Should this handle (ie skip) blank lines?

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/AbstractFindBugsTask.java
ThrawnCA: Or perhaps there could instead be a `getAppendableInputString` method that simply returns the field as an `Appendable`, initialising it if needed, and then the caller can use the regular `append` methods.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: Handling blank lines is probably a good idea since someone might be manually entering data into stdin or using a script.

Your comment about using Appendable and getAppendableInputString is already marked resolved so I can't answer there, but I think you're over-designing it.   This is the API for setting up the ant task.  It's rarely going to be used, and what's already there works well enough.

KengoTODA:
I'm not sure in which kind of cases we need this feature.
Have you tried `-analyzeFromFile` and `-auxclasspathFromFile`? I expect that they can cover most cases which needs large set of options.

mkienenb:
@KengoTODA Every large project using the ant spotbugs task needs this feature.

Without it:
```
> SPOTBUGS_HOME=[...]/spotbugs-4.0.0 ant spotbugs-main

spotbugs-main:
 [spotbugs] Executing SpotBugs FindBugsTask from ant task
 [spotbugs] Running SpotBugs...
 [spotbugs] java.io.IOException: Cannot run program "/home/mkienenb/temp/jdk1.8.0_60/jre/bin/java" (in directory "/home/mkienenb/workspaces/gvea-oxygen/Ebpp-struts/building"): error=7, Argument list too long
 [spotbugs] 	at org.apache.tools.ant.taskdefs.Java.fork(Java.java:861)
```

With it:
```
> SPOTBUGS_HOME=[...]/spotbugs-4.0.1-SNAPSHOT/ ant spotbugs-main

spotbugs-main:
 [spotbugs] Executing SpotBugs FindBugsTask from ant task
 [spotbugs] Running SpotBugs...
 [spotbugs] The following classes needed for analysis were missing:
 [spotbugs]   apply
 [spotbugs] Java Result: 1
```

The error shows up unexpectedly once the project grows to a certain size.   The user hasn't done anything wrong, but suddenly the ant task stops working.

Are there workarounds to this bug?  In some cases..

However, no similar ant task requires such workarounds.   I don't have to switch my `java` ant task to reading the source file list from a file after my project reaches a certain size.

So call it the use case of "Works out of the box" or "Principle of lease surprise".

spotbugs/src/doc/manual.xml
```diff
@@ -927,6 +927,18 @@ These options are only accepted by the Text User Interface.
...
+  <listitem>
+    <para>
+    Read additional options from from standard input, with each new line as a separate argument.
+    This option is incompatible with other arguments using standard input,
+    such as -auxclasspathFromInput and reading files or data from standard input.
+    -auxclasspathFromInput can be replaced with -auxclasspath followed by a classpath line.
+    </para>
+  </listitem>
+  </varlistentry>
+
```
KengoTODA: Please update files under `/docs` directory too.

mkienenb: @KengoTODA 

While `exclude` takes a single value, my `excludePath` is quite large.   Each module of the project has its own separately-maintained exclusions.

Note that this patch merely takes the approach that `auxclasspathFromInput` started and expands it to include other options.

A mcve is probably not reasonable, considering that large amounts of files are what triggers the situation.

You can find my ant task call here in the spotbugs/discuss project:

XTTE1020 error w/ html report; missing class "apply" #90 
https://github.com/spotbugs/discuss/issues/90

I will see about providing a sanitized example of what argument passing occurs in my project.

mkienenb: Here is a heavily-sanitized version of the ant -v -d output with this patch in place.

```
 [spotbugs] Running SpotBugs...
Class java.util.UUID loaded from parent loader (parentFirst)
 [spotbugs] Executing '[...]/jdk1.8.0_60/jre/bin/java' with arguments:
 [spotbugs] '-Dfindbugs.hostApp=FBAntTask'
 [spotbugs] '-Dspotbugs.home=/home/user/java/spotbugs-4.0.1-SNAPSHOT'
 [spotbugs] '-classpath'
 [spotbugs] '/home/user/java/spotbugs-4.0.1-SNAPSHOT/lib/spotbugs.jar'
 [spotbugs] 'edu.umd.cs.findbugs.FindBugs2'
 [spotbugs] '-pluginList'
 [spotbugs] '/home/user/workspaces/xyz-oxygen/application-framework/building/lib-build/findsecbugs-plugin-1.4.6.jar'
 [spotbugs] '-effort:max'
 [spotbugs] '-sortByClass'
 [spotbugs] '-timestampNow'
 [spotbugs] '-html'
 [spotbugs] '-low'
 [spotbugs] '-parseStdinAsOptions'
 [spotbugs] '-parseStdinAsOptions'
 [spotbugs] '-parseStdinAsOptions'
 [spotbugs] '-outputFile'
 [spotbugs] '../deploying/dist/application-spotbugs-report-2020-03-11-11-15-16.html'
 [spotbugs] '-exitcode'
 [spotbugs] '/home/user/workspaces/xyz-oxygen/application-framework/WEB-INF/classes'
 [spotbugs] 
 [spotbugs] The ' characters around the executable and arguments are
 [spotbugs] not part of the command.
 [spotbugs] Using input "-exclude

 [spotbugs] -exclude
 [spotbugs] [...]/findbugs-[...]-exclude-filter.xml
**[REPEATS 10 times]**
 
 [spotbugs] -auxclasspath
 [spotbugs] **[36 ITEMS LISTED]**
 [spotbugs] -sourcepath
 [spotbugs] **[1442 ITEMS LISTED]**
 [spotbugs] "
Execute:Java13CommandLauncher: Executing '[...]/jdk1.8.0_60/jre/bin/java' with arguments:
'-Dfindbugs.hostApp=FBAntTask'
'-Dspotbugs.home=/home/user/java/spotbugs-4.0.1-SNAPSHOT'
'-classpath'
'/home/user/java/spotbugs-4.0.1-SNAPSHOT/lib/spotbugs.jar'
'edu.umd.cs.findbugs.FindBugs2'
'-pluginList'
'/home/user/workspaces/xyz-oxygen/application-framework/building/lib-build/findsecbugs-plugin-1.4.6.jar'
'-effort:max'
'-sortByClass'
'-timestampNow'
'-html'
'-low'
'-parseStdinAsOptions'
'-parseStdinAsOptions'
'-parseStdinAsOptions'
'-outputFile'
'../deploying/dist/application-spotbugs-report-2020-03-11-11-15-16.html'
'-exitcode'
'/home/user/workspaces/xyz-oxygen/application-framework/WEB-INF/classes'

The ' characters around the executable and arguments are
not part of the command.

```

mkienenb: I noticed that `-parseStdinAsOptions` shows up once for each kind of option passed in which supports it.  For now there are only four, but I should fix it to output only once if we are going forward with this patch.

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
```diff
@@ -901,21 +901,49 @@ protected void configureFindbugsEngine() {
...
             addArg(includeFile.getPath());
         }
         if (includePath != null) {
-            String[] result = includePath.toString().split(java.io.File.pathSeparator);
-            for (String element : result) {
-                addArg("-include");
-                addArg(element);
+            String includePathValue = includePath.toString();
+            if (!includePathValue.isEmpty()) {
+                if (includePathValue.length() > 100) {
```
KengoTODA, line 931: Could you make a constant representing this `100`? Magic number makes code confusing :sob:.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
```diff
@@ -342,9 +354,35 @@ private int parse(String argv[], boolean dryRun) throws IOException, HelpRequest
...
+                break;
+            }
+            if (s.trim().isEmpty()) {
+                break;
+            }
+            stdinArgvList.add(s);
+        }
+        in.close();
```
KengoTODA, line 381: Use try-with-resources to close this `BufferedReader`.
(personally not so sure that we can close `System.in` or not.)

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
```diff
@@ -342,9 +354,35 @@ private int parse(String argv[], boolean dryRun) throws IOException, HelpRequest
...
+
+        BufferedReader in = UTF8.bufferedReader(System.in);
+        while (true) {
+            String s = in.readLine();
+            if (s == null) {
+                break;
+            }
+            if (s.trim().isEmpty()) {
+                break;
+            }
```
KengoTODA, line 378: SpotBugs depends on commons-lang3, so we can shorten these lines like below:

```suggestion
            if (null == org.apache.commons.lang3.StringUtils.trimToNull(s)) {
                break;
            }
```

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
```diff
@@ -342,9 +354,35 @@ private int parse(String argv[], boolean dryRun) throws IOException, HelpRequest
...
+            String[] stdinArgv = readStdinAsArgv();
+            if (0 != stdinArgv.length) {
+                parse(stdinArgv, true, dryRun);
+            }
+        }
+
         return arg;
     }
 
+    private String[] readStdinAsArgv() throws IOException {
```
KengoTODA, line 367: This implementation loads all data on memory, then its memory-footprint should be large.
Can we reduce this footprint?

spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FindBugsTask.java
mkienenb: Will do, but I copied the existing code from auxclasspath.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: Redesigning how the ant task parses command-line arguments is outside of the scope of this PR.

I haven't had any issues with memory usage.  I have > 2000 application java classes and 40 external jars.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: Done.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: I was also not sure if we can/should close System.in or not.  I copied this code from -auxclasspathFromInput in TextUICommandLine

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: If someone wanted to go about implementing this, I suppose they would need to implement an "ArgumentProvider.getNextArgument()" interface, then have each class which passes arguments to the spotbugs parser implement it.   Then the stdin line reader could read in arguments as needed.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
mkienenb: I went ahead and updated it with try-with-resources.

spotbugs/src/main/java/edu/umd/cs/findbugs/config/CommandLine.java
KengoTODA: Wmm, for me, it seems that this PR just moved the huge memory footprint from the Ant task to the SpotBugs.

Does this change give increase the load for java heap in the SpotBugs side?
If it's limited, then we can put this problem aside for a while, and merge this PR to release this change as a part of 4.1.0.

mkienenb: I have no way to compare a run of "it-works-with-this-patch" against "it-crashes-without-this-patch" in order to determine what the possible java heap impact would be.   And honestly, it really doesn't matter to me since a spotbugs run with an potential increased java heap load is far better than a spotbugs run that crashes.

The best fix for ant would be for someone to rewrite the spotbugs engine so that the ant task doesn't have to communicate by passing arguments through the commandline.   That won't help people who run spotbugs from the commandline, though.

ThrawnCA:
@mkienenb Please try to avoid force-pushing since it breaks the history.

mkienenb:
@ThrawnCA Sorry about that.  The last project for which I submitted PRs required that I rebase and force-push a clean PR each time.

ThrawnCA:
> @ThrawnCA Sorry about that. The last project for which I submitted PRs required that I rebase and force-push a clean PR each time.

That...doesn't sound like a very good policy. If they really wanted a clean commit history, then they could probably achieve that using a squash commit.

mkienenb:
You can go ahead and close this.   We've moved on to alternative tooling.

Issue #1025, opened 2019-10-17
Could not generate HTML output (XPathException) [4.0.0 beta 4]


grrosminet:
Hi,

I tried 4.0.0-beta-4 with same parameters I'm using for 3.1.12 and got this error at the end of analysis :

The following errors occurred during analysis:
  Could not generate HTML output
    net.sf.saxon.trans.XPathException: A sequence of more than one item is not allowed as the @select attribute of xsl:sort (@classname="infass.util.base.data.IRecord", @classname="infass.util.importation.mapper.MapperRecord", @cl
assname="infass.util.base.data.FlatIRecord", ...)
      At net.sf.saxon.expr.Expression.typeError(Expression.java:1464)
      At net.sf.saxon.expr.SingletonAtomizer.evaluateItem(SingletonAtomizer.java:221)
      At net.sf.saxon.expr.SingletonAtomizer.evaluateItem(SingletonAtomizer.java:31)
      At net.sf.saxon.expr.sort.SortExpression.evaluateSortKey(SortExpression.java:394)
      At net.sf.saxon.expr.sort.SortedIterator.buildArray(SortedIterator.java:215)
      At net.sf.saxon.expr.sort.SortedIterator.doSort(SortedIterator.java:231)
      At net.sf.saxon.expr.sort.SortedIterator.next(SortedIterator.java:148)
      At net.sf.saxon.om.FocusTrackingIterator.next(FocusTrackingIterator.java:73)
      At net.sf.saxon.trans.Mode.applyTemplates(Mode.java:455)
      At net.sf.saxon.expr.instruct.ApplyTemplates.apply(ApplyTemplates.java:300)
      At net.sf.saxon.expr.instruct.ApplyTemplates.processLeavingTail(ApplyTemplates.java:255)
      At net.sf.saxon.expr.instruct.Block.processLeavingTail(Block.java:735)
      At net.sf.saxon.expr.instruct.Instruction.process(Instruction.java:132)
      At net.sf.saxon.expr.instruct.ElementCreator.processLeavingTail(ElementCreator.java:352)
      At net.sf.saxon.expr.instruct.ElementCreator.processLeavingTail(ElementCreator.java:299)
      At net.sf.saxon.expr.instruct.Block.processLeavingTail(Block.java:735)
      At net.sf.saxon.expr.instruct.NamedTemplate.expand(NamedTemplate.java:243)
[...]

PR #1102, opened 2020-03-11
Resolved XPathException generating html report if BugInstance contains multiple Class elements

mkienenb:
Resolved XPathException generating html report if BugInstance contains multiple Class elements reported in issue #1025 

net.sf.saxon.trans.XPathException: A sequence of more than one item is not allowed as the @select attribute of xsl:sort (@ClassName=...)

Found a solution on stackoverflow which explicitly sets the first child node match to be used in the sort, which seems reasonable.
https://stackoverflow.com/questions/45601621/what-does-1-mean-in-an-xslt-sort

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 
 * Resolved Saxon warning ([#1077](https://github.com/spotbugs/spotbugs/issues/1077))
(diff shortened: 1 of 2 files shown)


paulk-asert:
I created a very similar PR before seeing yours #1108

iloveeclipse:
I've cherry-picked your commit, updated commit message and log entry & pushed a new PR with another fix, see #1112

Issue #1105, opened 2020-03-18
Report generation still fails using spotbugs 4.0.1 and gradle plugin 4.0.2


paulk-asert:
Steps to reproduce (shown for `groovy-json` module but similar for numerous other modules in the Groovy project - some modules work fine though):
```
> git clone https://github.com/apache/groovy.git
> ./gradlew :groovy-json:spotBugsMain
```
Output:
```
Task :groovy-json:spotbugsMain
Type error evaluating ((./Class)/@classname) in xsl:sort/@select on line 285 column 41 of default.xsl:
  XTTE1020: A sequence of more than one item is not allowed as the @select attribute of
  xsl:sort (@classname="groovy.json.JsonTokenType$1", @classname="groovy.lang.Closure")
at template generateWarningTable on line 275 of default.xsl:
     invoked by xsl:call-template...
```
using the `fancy-hist.xsl` stylesheet works.

paulk-asert:
Possibly duplicates #1025

iloveeclipse:
> Possibly duplicates #1025

Yep

PR #1108, opened 2020-03-20
Avoid Saxon "XPTY0004: A sequence of more than one item is not allowe…

paulk-asert:
…d as ..." errors as per #1105

----


paulk-asert:
This seemed to fix the issue for me with `default.xsl`. The same adjustment is likely needed for other stylesheets though `fancy-hist.xml` seems okay as-is wrt this issue.

iloveeclipse:
This PR basically duplicates #1102 and misses another fix for fancy.xsl. I've cherry-picked original fix and added fancy.xsl fix via #1112.

PR #1111, opened 2020-03-29
Update some libraries, fixed asm version numbers throughout, fixed xml issue with escapes

hazendaz:
----

- [X ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/spotbugs/src/test/java/edu/umd/cs/findbugs/util/StringsTest.java
+++ b/spotbugs/src/test/java/edu/umd/cs/findbugs/util/StringsTest.java
@@ -21,13 +21,14 @@ package edu.umd.cs.findbugs.util;
 
 import static org.junit.Assert.assertEquals;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 public class StringsTest {
 
     public static String[] escapedStrings = {
         // mixed entities/unicode escape sequences
-        "a b c 1 2 3 &amp; &lt; &gt; &quot; &apos; \\u0005 \\u0013 &#955; \\\\u0007",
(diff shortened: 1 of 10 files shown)


hazendaz:
This started out as an effort of seeing commons-lang still used in the spot bugs maven plugin coming from here.  Also comes from maven too, but felt the need to come over here and fix it.  Through that process, I found a number of other things in same area of code and had to spend a good deal of time looking at that encoding issue.  If 'html' was used for encoding that does work but clearly it is xml.  Further commons-lang2 is really old.  I couldn't find anything that definatively said it was needed to escape in xml, just html.

Other notes, github actions / travis both fail when coming from a fork.  While not a huge concern here, probably something that should be looked at to resolve so forks are not trying to do all the same things.

spotbugs/src/test/java/edu/umd/cs/findbugs/util/StringsTest.java
```diff
@@ -21,13 +21,14 @@
...
 import static org.junit.Assert.assertEquals;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 public class StringsTest {
 
     public static String[] escapedStrings = {
         // mixed entities/unicode escape sequences
-        "a b c 1 2 3 &amp; &lt; &gt; &quot; &apos; \\u0005 \\u0013 &#955; \\\\u0007",
```
KengoTODA: is it necessary change? 🤔

hazendaz:
Yes.  The escape value doesn't apply to XML.  The character is valid.  Greek lambda.  It only applies to HTML.  You can trust by flipping time HTML encode which will show the difference.

Get Outlook for Android<https://aka.ms/ghei36>

PR #1112, opened 2020-03-30
Issue 1025

iloveeclipse:
----

### DIFF
--- a/spotbugs/src/xsl/default.xsl
+++ b/spotbugs/src/xsl/default.xsl
@@ -282,7 +282,7 @@
 		<xsl:copy-of select="$bugTableHeader"/>
 		<xsl:apply-templates select="$warningSet">
 			<xsl:sort select="@abbrev"/>
-			<xsl:sort select="Class/@classname"/>
+			<xsl:sort select="Class[1]/@classname"/>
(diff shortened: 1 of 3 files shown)


spotbugs/src/xsl/default.xsl
```diff
@@ -282,7 +282,7 @@
 		<xsl:copy-of select="$bugTableHeader"/>
 		<xsl:apply-templates select="$warningSet">
 			<xsl:sort select="@abbrev"/>
-			<xsl:sort select="Class/@classname"/>
+			<xsl:sort select="Class[1]/@classname"/>
```
ThrawnCA: Does this still work if there is only one element?

iloveeclipse: I think yes. At least I see in our build that both single / multiple matches are working now.
I have not much experience with that, but stack overflow suggests this as a viable approach:
https://stackoverflow.com/questions/16928121/how-to-select-first-node-only-in-xslt

iloveeclipse: @mkienenb , @paulk-asert : WDYT?

paulk-asert: +1

PR #1129, opened 2020-05-11
Avoid setting the custom SecurityManager when running under IntelliJ IDEA

amaembo:
When SpotBugs is launched under IDEA plugin changing the SecurityManager affects the whole IntelliJ IDEA caused strange issues. See:
https://github.com/JetBrains/spotbugs-intellij-plugin/issues/10
https://youtrack.jetbrains.com/issue/IDEA-230268

This functionality is already excluded for the Eclipse plugin. This patch extends the exclusion to IntelliJ IDEA plugin as well. I don't see how we can fix this without patching SpotBugs itself. The only possibility is to update reflectively static final field, but this is too dirty.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/SystemProperties.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/SystemProperties.java
@@ -42,8 +42,15 @@ public class SystemProperties {
 
     public final static boolean ASSERTIONS_ENABLED;
 
-    public final static boolean RUNNING_IN_ECLIPSE = SystemProperties.class.getClassLoader().getClass().getCanonicalName()
-            .startsWith("org.eclipse.osgi");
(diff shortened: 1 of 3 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/SystemProperties.java
```diff
@@ -42,8 +42,12 @@
 
     public final static boolean ASSERTIONS_ENABLED;
 
-    public final static boolean RUNNING_IN_ECLIPSE = SystemProperties.class.getClassLoader().getClass().getCanonicalName()
```
KengoTODA, line 45: Please keep public API compatible.

amaembo:
> Thanks for your PR! Please update `CHANGELOG.md` to tell your changes to users.

Done.

Issue #1155, opened 2020-06-08
False positive: NP_NULL_ON_SOME_PATH cannot occur because of thrown exception.


nielsbasjes:
In this project https://github.com/nielsbasjes/yauaa/blob/f4d7d06b88f6784ba1a202a92ad92f13fef0d28f/analyzer/src/main/java/nl/basjes/parse/useragent/AbstractUserAgentAnalyzerDirect.java#L644 I got an NP_NULL_ON_SOME_PATH that cannot occur.

The minimal reproduction I could create using spotbugs-maven-plugin 4.0.0 + spotbugs 4.0.3 is this (silly code just to prove the problem)

```
package nl.basjes.bugreports;

import java.util.List;

public class FalsePositiveNPEError {
    public static void require(boolean condition, String error) {
        if (!condition) {
            throw new IllegalStateException(error);
        }
    }

    public int demoOfInvalidNPEBeingReported(List<String> theList, String findValue) {
        String found = null;
        for (String value : theList) {
            if (value.contains(findValue)) {
                found = value;
                break;
            }
        }
        require(found != null, "Should have found it");
        return found.length();
    }
}
```

I get the error reported on the return statement because the `found` may be null and thus a dereference would fail.

What happens here is that the case of the parameter being null is caught by the static `require` method. Because it is static it cannot be overridden or changed (i.e. same as final). In case of the null value (which CAN occur) the exception will break the normal application flow and thus never reach the reported NPE location (in this case the return statement).

nielsbasjes:
**Important observation:** 
Apparently the problem changes with the NAME of the static method.
A name like 
-   `require` or `check` will trigger the invalid error.
-   `failIf` does NOT trigger this invalid error.

PR #1159, opened 2020-06-09
Reproduce issue #1155: Invalid NP_NULL_ON_SOME_PATH if static method throws exception

nielsbasjes:
This currently is ONLY the reproduction of the problem I reported in issue #1155.
So this must fail the build.

----


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1155Test.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.hamcrest.Matchers.is;
+
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/1155">GitHub issue</a>
+ */
+public class Issue1155Test extends AbstractIntegrationTest {
+    @Test
+    public void testIssue1155() {
+        performAnalysis("ghIssues/Issue1155.class");
+        assertThat(getBugCollection(), is(emptyIterable()));
```
JuditKnoll, line 17: Please assert the specific bugtypes instead of all bugtypes.

JuditKnoll (changes requested):

PR #1166, opened 2020-06-17
plain.xsl declares it is a 2.0 stylesheet, but it appears to have iss…

jqyp:
…ues with a 2.0 processor

The same problem with default.xsl was solved here: https://github.com/spotbugs/spotbugs/pull/1000

----


KengoTODA:
Thanks for suggesting this PR! I'll release next version later.

PR #1180, opened 2020-06-22
Use GitHub Secrets only when they are accessible

pzygielo:


pzygielo:
@KengoTODA - thanks for checking. But now I see that on pushed master just `release` build was fired and not `build` (modified here). I understand that sonar is expected to run only on other branches pushed to this repository. Is that correct?

PR #1183, opened 2020-06-24
* Added bug pattern "OVERRIDING_METHODS_MUST_INVOKE_SUPER"

jpwilhelms:
* changed class name from CbeckMustOverrideSuperAnnotationTest to OverridingMethodsMustInvokeSuperDetector.

----


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/OverridingMethodsMustInvokeSuperDetector.java
```diff
@@ -27,53 +27,64 @@
...
+            bugReporter.reportBug(new BugInstance(this, "OVERRIDING_METHODS_MUST_INVOKE_SUPER", NORMAL_PRIORITY)
+                    .addClassAndMethod(this).addString("Method must invoke override method in superclass"));
         }
     }
 
+    private String getEffectiveMethodSig() {
+        final String methodSig;
+        XMethod bridgeFrom = getXMethod().bridgeFrom();
+
+        if (bridgeFrom != null) {
```
ThrawnCA, line 79: "if not...else" is a double negative; it would be more readable if this were written positively.

PR #1184, opened 2020-06-30
Experimental support for the SARIF 2.1.0 report

KengoTODA:
This PR introduce an experimental `BugReporter` implementation which generates report in SARIF 2.1.0 format.
refs https://github.com/spotbugs/discuss/issues/95

### Note for reviewers

* Apply [patch.txt](https://github.com/spotbugs/spotbugs/files/4848835/patch.txt) to try SNAPSHOT version in your local.
* Check [main.sarif](https://github.com/spotbugs/spotbugs/files/4857510/main.sarif.txt) as sample output.

### Known issues

I will find solution for following issues in another PR.

* `version` is empty, even though SpotBugs should read it from `META-INF/MANIFEST.MF`.
* The spotbugs-gradle-plugin cannot use Worker API (cannot find `SarifException` class from its CLASSPATH), so feature flag is necessary like: `./gradlew spotbugs:spotbugsMain -Pcom.github.spotbugs.snom.worker=false`
* `messageString` contains placeholders like `{2.givenClass}` that cannot be handled by SARIF client.
* Need to find how to add location data to `frames` in `exception`.
* Tell where the `srcRoot` is.


spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyser.java
```diff
@@ -0,0 +1,59 @@
...
+        bugCollection.forEach(bug -> {
+            String type = bug.getType();
+            int index = typeToIndex.computeIfAbsent(type, (t) -> processRule(bug.getBugPattern()));
+
+            processResult(index, bug);
+        });
+    }
+
+    JSONArray getRules() {
+        JSONArray array = new JSONArray();
```
ThrawnCA, line 30: There is a JSONArray constructor that takes a collection directly; why not use it?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @see <a href="https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317670">3.28 location object</a>
+ */
+class Location {
+    @NonNull
+    private final List<PhysicalLocation> physicalLocations;
```
ThrawnCA, line 27: Why make this a list when we only ever add one element?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Message.java
```diff
@@ -0,0 +1,31 @@
...
+    @NonNull
+    final List<String> arguments;
+
+    Message(@NonNull List<String> arguments) {
+        this.arguments = Collections.unmodifiableList(Objects.requireNonNull(arguments));
+    }
+
+    JSONObject toJSONObject() {
+        JSONObject jsonObject = new JSONObject().put("id", "default");
+        if (!arguments.isEmpty()) {
```
ThrawnCA, line 24: No need for this condition when we're about to iterate over it anyway. A null check might make sense, except that the field is non-null.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Notification.java
```diff
@@ -0,0 +1,49 @@
...
+        if (exception != null) {
+            result.put("exception", exception.toJSONObject());
+        }
+        return result;
+    }
+
+    static Notification fromError(@NonNull AbstractBugReporter.Error error) {
+        String id = String.format("spotbugs-error-%d", error.getSequence());
+        Throwable cause = error.getCause();
+        if (cause != null) {
```
ThrawnCA, line 43: "if not...else" is a double negative, please rearrange.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Result.java
```diff
@@ -0,0 +1,32 @@
...
+    Result(@NonNull String ruleId, int ruleIndex, Message message, List<Location> locations) {
+        this.ruleId = Objects.requireNonNull(ruleId);
+        this.ruleIndex = ruleIndex;
+        this.message = Objects.requireNonNull(message);
+        this.locations = Collections.unmodifiableList(Objects.requireNonNull(locations));
+    }
+
+    JSONObject toJSONObject() {
+        JSONObject result = new JSONObject().put("ruleId", ruleId).put("ruleIndex", ruleIndex).put("message", message.toJSONObject());
+        locations.forEach(location -> result.append("locations", location.toJSONObject()));
```
ThrawnCA, line 29: A slightly streamier approach:

    locations.stream().map(Location::toJSONObject).forEach(location -> result.append("locations", location))

lgolding:
Hi @KengoTODA, thanks for attaching the sample SARIF output above. I'll take a look. @michaelcfanning FYI.

KengoTODA:
PhysicalLocation is not set well, I'll fix it.

KengoTODA:
@lgolding I fixed `PhysicalLocation`, and added `tags` and `fullDescription` for rules. Here is new generated SARIF file for reference: [main.sarif.txt](https://github.com/spotbugs/spotbugs/files/4864008/main.sarif.txt)

@ThrawnCA I'm sorry but could you review this again? After I pass you review, I'm going to merge this.
I haven't updated `TextUICommandLine.java` so this PR won't affect spotbugs' behaviour, just add `org.json` to its dependency. I also changed scope of some existing members, but it keeps backward compatibilities.

PR #1215, opened 2020-07-13
Move notifications from tool.driver to invocations

KengoTODA:
Based on https://github.com/microsoft/sarif-sdk/issues/1952#issuecomment-657254642 current place is not good to put reported errors during parse, so move it to proper place.

Also fix one build problem: c70db23 solved duplicated slf4j-api problem, but it introduced missing log4j-api problem.
Because [log4j binding depends on not only slf4j-api but also log4j-api](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-slf4j18-impl/2.13.1).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,14 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2020-??-??
 ### Added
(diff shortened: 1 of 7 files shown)


KengoTODA:
I'll close this PR for a while. I will reopen after I understand the discussion at https://github.com/microsoft/sarif-sdk/issues/1952

PR #1221, opened 2020-07-16
Use URI for files instead of converting string to URI each time.

olegshtch:
Fix java.net.URISyntaxException: Illegal character in opaque part at index 2: C:\... on WIndows
Fixes some test by comparing URIs because File and Path returns URIs
with different String representation https://stackoverflow.com/a/46634310

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ZipSourceFileDataSource.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ZipSourceFileDataSource.java
@@ -22,6 +22,7 @@ package edu.umd.cs.findbugs.ba;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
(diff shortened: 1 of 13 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ZipSourceFileDataSource.java
```diff
@@ -33,11 +34,14 @@
...
     private final String entryName;
 
+    private final URI entryURI;
+
     private final ZipEntry zipEntry;
 
     public ZipSourceFileDataSource(ZipFile zipFile, String entryName) {
         this.zipFile = zipFile;
         this.entryName = entryName;
+        this.entryURI = URI.create(entryName);
```
KengoTODA, line 44: I'm not sure this really works or not.
Both ZipFile and [ZipEntry](https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html) provide no method returning `URI`, so I think that we need to use `FileSystems` like https://stackoverflow.com/q/37936627

KengoTODA: Hi @olegshtch, could you confirm this? Thank you!

olegshtch: I've implemented it.

PR #1222, opened 2020-07-18
Allow private methods to inherit default annotations from package or class scope

### LINKED ISSUE #374: @ParametersAreNonnullByDefault ignored for private methods if set on package or class level.
We have a problem here where spotbugs does not apply the "@ParametersAreNonnullByDefault" annotation to private methods if the annotation has been set on package or class level. If you annotate the private method directly however it works.

How to reproduce:
1) Unzip the attached maven project.
2) Run "mvn clean verify"

Expected Behavior:
Spotbugs should find the same bug in all three example classes, as all classes have the same code, but simply get their @ParametersAreNonnullByDefault annotation from a different level (package/class/method).

Actual Behavior:
Spotbugs only finds the bug in "TestNonnullByDefaultFromMethodAnnotation.java"

Workaround:
Change all private methods to protected or public and spotbugs will correctly complain about all three classes.

[spotbugs.zip](https://github.com/spotbugs/spotbugs/files/1298881/spotbugs.zip)

cwbriones:
Resolves #374.

Thanks @RitChan for the reproduction test cases attached to the Issue.

From the comment in the code, the original decision to omit private methods seems very much intentional. But since the issue is classified as a bug it seemed safe to just remove that check.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## Unreleased - 2020-??-??
 ### Fixed
 * [A meaningless exception data from `SAXBugCollectionHandler`](https://lgtm.com/projects/g/spotbugs/spotbugs/rev/a77ab08634687b7791e902636996ab6184462693)
+* Allow private methods to inherit default annotations from package or class scope. ([#374](https://github.com/spotbugs/spotbugs/issues/374))
 
 ## 4.1.1 - 2020-07-31
 ### Fixed
@@ -19,7 +20,6 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 * Fixed not working detector 'CbeckMustOverrideSuperAnnotation' and renamed to 'OverridingMethodsMustInvokeSuperDetector'
-* Allow private methods to inherit default annotations from package or class scope. ([#374](https://github.com/spotbugs/spotbugs/issues/374))
 
 ### Changed
 * Bump commons-lang3 from 3.10 to 3.11 ([#1231](https://github.com/spotbugs/spotbugs/pull/1231))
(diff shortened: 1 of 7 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue374Test.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2020, kengo
```
KengoTODA, line 3: Use your name, please :)

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierApplications.java
```diff
@@ -986,11 +986,6 @@ private static TypeQualifierAnnotation computeEffectiveTypeQualifierAnnotation(T
             // default annotations
         }
 
-        /** private methods don't inherit from class or package scope */
-        if (xmethod.isPrivate()) {
-            stopAtMethodScope = true;
-        }
-
```
KengoTODA, line 993: I'm still not sure about the intention behind this code. 
But commit message of b1fcbae0c11852eb42e5e2f7e1eca61003aa4a7b tells nothing, so I agree to change the behavior to make it intuitive.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 * Fixed not working detector 'CbeckMustOverrideSuperAnnotation' and renamed to 'OverridingMethodsMustInvokeSuperDetector'
+* Allow private methods to inherit default annotations from package or class scope. ([#374](https://github.com/spotbugs/spotbugs/issues/374))
```
KengoTODA, line 13: Please merge the latest `origin/master` to your branch, and move your entry to under `## Unreleased`.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue374Test.java
RitChan: I just copied from another place and didn't notice that XD. 
Can you correct this copyright owner in your repository @cwbriones? You can use "ritchan". Thanks!

PR #1230, opened 2020-07-23
Bump up ASM to 9.0-beta

### LINKED ISSUE #1229: Fails with OpenJDK 15-ea: Unsupported class file major version 60
It looks like the Maven plug-in needs an update for JDK 15. I get the following error when switching the JDK:

```
     [java] The following errors occurred during analysis:
     [java]   Error scanning java/lang/Object for referenced classes
     [java]     java.lang.IllegalArgumentException: Unsupported class file major version 60
     [java]       At org.objectweb.asm.ClassReader.<init>(ClassReader.java:196)
     [java]       At org.objectweb.asm.ClassReader.<init>(ClassReader.java:177)
     [java]       At org.objectweb.asm.ClassReader.<init>(ClassReader.java:163)
     [java]       At edu.umd.cs.findbugs.asm.FBClassReader.<init>(FBClassReader.java:35)
```

KengoTODA:
close #1229

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 * Fixed not working detector 'CbeckMustOverrideSuperAnnotation' and renamed to 'OverridingMethodsMustInvokeSuperDetector'
 
+### Changed
+* Upgrade ASM to 9.0-beta which supports Java 15
+
 ## 4.0.6 - 2020-06-23
 ### Fixed
 * Use method call instead of reflection to get BCEL frame type ([#1176](https://github.com/spotbugs/spotbugs/issues/1176))
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -9,7 +9,7 @@ configurations {
 }
 
 ext {
-  asmVersion = '8.0.1'
+  asmVersion = '9.0-beta'
 }
 
 sourceSets {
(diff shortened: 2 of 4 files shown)


KengoTODA:
> Are we ok with a beta version?

I think so, because its changeset has less changes (just 15 support and three bugfix).
I'll keep this PR open for a while, to welcome opinions from community.

iamatgit:
Won't you need to update spotbugs/jnlp/findbugs.jnlp and spotbugs/src/sampleXml/analysisResultsWithFilterAndUserAnnotations.xml as well?

iamatgit:
Also, the class file version 60 is for JDK 16 (https://hg.openjdk.java.net/jdk/jdk/rev/74571a9cede2).

May I suggest to call update to ASM 9_0_BETA to support JDK 16 (than JDK 15) to be consistent with class file version change and release notes from ASM (https://asm.ow2.io/versions.html). 

Also, please consider to update https://github.com/spotbugs/spotbugs/blob/91e26e6f0fbf2110deedcf5de5a99cdb85bf4ef4/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/asm/FindBugsASM.java#L31 to Opcodes.ASM10_EXPERIMENTAL

KengoTODA:
Support for JNLP is deprecated #904 so we won't update related files.

KengoTODA:
To use `Opcodes.ASM9_EXPERIMENTAL`, we need to compile our code with `--enable-preview` option.
This option is introduced from Java 11, so we cannot use this experimental feature because we build our code with Java 8.

Let's wait the public stable release for ASM9. I'll find a way to build our solution with Java 11.

PR #1242, opened 2020-08-02
Resolve issues reported by LGTM.com

KengoTODA:
This PR fixes some of reported issues listed at https://lgtm.com/projects/g/spotbugs/spotbugs/alerts/?mode=list

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/XFactory.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/XFactory.java
@@ -84,6 +84,10 @@ public class XFactory {
 
     private final Set<XField> emptyArrays = new HashSet<>();
 
+    /**
+     * @deprecated This field is not updated by any code in the project. Will be removed in 5.x release.
+     */
+    @Deprecated
     private final Set<String> calledMethodSignatures = new HashSet<>();
 
     private final Set<MethodDescriptor> functionsThatMightBeMistakenForProcedures = new HashSet<>();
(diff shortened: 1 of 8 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/XFactory.java
```diff
@@ -223,6 +227,10 @@ private boolean isCalledDirectlyOrIndirectly(@CheckForNull ClassDescriptor clazz
 
     }
 
+    /**
+     * @deprecated This method does not work as expected. Will be removed in 5.x release.
```
KengoTODA, line 231: According to IntelliJ IDEA, we have no code depending on this method.
But it's a public method, so I'll keep this method at this moment, and will remove it in the next major update.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/type/TypeAnalysis.java
```diff
@@ -888,8 +888,6 @@ private ExceptionSet computeThrownExceptionTypes(BasicBlock basicBlock) throws D
                     Type throwType = frame.getTopValue();
                     if (throwType instanceof ObjectType) {
                         exceptionTypeSet.addExplicit((ObjectType) throwType);
-                    } else if (throwType instanceof ExceptionObjectType) {
-                        exceptionTypeSet.addAll(((ExceptionObjectType) throwType).getExceptionSet());
```
KengoTODA, line 892: One more idea is that, switch the order of if-blocks like below. It could be similar to the intent of the author the commit 3b31f36597c84e4dfe9cb0bd1e631c309e6bc660.

```java
                    if (throwType instanceof ExceptionObjectType) {
                        exceptionTypeSet.addAll(((ExceptionObjectType) throwType).getExceptionSet());
                    } else if (throwType instanceof ObjectType) {
                        exceptionTypeSet.addExplicit((ObjectType) throwType);
                    } else {
```

However, the behavior of `addAll()` is not the same as `addExplicit()`, so I put a higher priority on keeping the current behavior.

Issue #1244, opened 2020-08-04
Build with JDK14


KengoTODA:
Build the project with the latest JDK (14 in this timing) with https://github.com/java9-modularity/gradle-modules-plugin
to make sure we can run tests with Java 9 or later. refs #1136 for example.

This change won't affect the version of .class files (we keep using Java 8 to compile), however, the JPMS support could be updated in this change from using `META-INF` to `package-info.java`.

refs #1123

iloveeclipse:
Could we please stay on Java 11 LTS? Non-LTS Java is not available on RHEL releases.

KengoTODA:
SpotBugs users won't requested to use Java 14, they still can use Java 8 to run. 
This is the proposal to use latest `javac` to build, to let SpotBugs developers to add unit test cases which rely on Java 9 or later like #1136.

iloveeclipse:
Sure, this is clear, but as a developer on RHEL I have no Java 14. Java 11 can also run Java 9+ examples, so that should be fine too.

KengoTODA:
OK first I'll introduce Java 11 the LTS. Later I'll introduce Java 14 to support cases which needs Java 12 or later.

KengoTODA:
#996 is the blocker of this task. Even when I build with `--release 8` option, Java 11 makes different `.class` files and it triggers this issue.

I also tried to add `module-info.java` to each project in 1f671d2, and found that we have several problems to solve:

* Multiple `test-harness` projects provide the same package. Need to rename package to avoid this conflict.
* `jaxen` should be bumped up to `1.2.0` which contains [a fix for jigsaw](https://github.com/jaxen-xpath/jaxen/pull/15/)

KengoTODA:
#996 has been solved. The JPMS problem still exists.

I tried to build the project with Java 14 as #1270, but the build performance was not changed.

KengoTODA:
Now we build with Java 15.
https://github.com/spotbugs/spotbugs/commit/50cb283f3d97bd6de3e1578688eae87d998157e1#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721

Issue #996, opened 2019-07-27
Build on master / Java 11 fails due test fail in testAllRegressionFilesJavac


iloveeclipse:
```
 Task :spotbugs-tests:test

edu.umd.cs.findbugs.DetectorsTest > testAllRegressionFilesJavac FAILED
    java.lang.AssertionError: Unexpected bugs (3):
    unexpected RCN At Bug1169.java:[line 69]
    unexpected RCN At Bug1169.java:[line 47]
    unexpected RCN At Bug1169.java:[line 32]
        at org.junit.Assert.fail(Assert.java:88)
        at edu.umd.cs.findbugs.DetectorsTest.checkForUnexpectedBugs(DetectorsTest.java:142)

57 tests completed, 1 failed, 2 skipped

> Task :spotbugs-tests:test FAILED

```

iloveeclipse:
Same on the travis: https://travis-ci.org/spotbugs/spotbugs/jobs/564310590.

So it is Java 11 issue with the test.

KengoTODA:
fixed by #1248

PR #1248, opened 2020-08-04
fix: redundant RCN false positive to build SpotBugs with Java 11

### LINKED ISSUE #259: false positive RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE on try-with-resources
FindBugs (eclipse plugin v3.0.1) reports a redundant null check on the closing brace of the following try-with-resources statement:
  
    public void falsePositive() {
      try (X x = new X(); X y = new X()) {
        throw new IllegalStateException();
      } // Redundant nullcheck of x, which is known to be non-null in test.Bug1169.falsePositive()
    }
  
    private static class X implements AutoCloseable {
      @Override
      public void close() {
      }
    }

The finding seems to be reported when more than one AutoClosable is used and the try block ends with a thrown exception.

KengoTODA:
To reproduce bugs with recent Java and its compilers, it's nice to make this project ready to build with them.
Java 14 is the latest version, but 11 is the latest LTS so this PR uses Java 11 to build SpotBugs. refs #1244.
Gradle uses `--release` option to make sure the version of `.class` files are Java 8, so this change won't affect users.

At this moment we won't introduce `module-info.java` because [some of our submodules provide the same package and it's not supported by JPMS](https://github.com/spotbugs/spotbugs/issues/1244#issuecomment-668872019).

## Fix for RCN false positive

To pass the build with Java 11, this PR fixes #259 and #756 by ignoring `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE` next to the end of try-block which catches `java/lang/Throwable`. See 7700cfc for detail.
This will fix reported false positives, but introduce false negatives in the following case, because we have no way to judge the bytecode to handle exception is generated by `javac` or not.

```java
@NonNull
SomethingClosable s = ...;
try {
  ...
} catch (Throwable t) { // when the catch clause catches Throwable instance,
  // and when we have null check at the beginning of catch block,
  if (s != null) { // this is redundant null check but cannot be detected by SpotBugs any more
    ...
  }
}
```

I hope this false negative is acceptable to us. The redundant null check itself has less impact on runtime, I believe.


spotbugsTestCases/build.gradle
```diff
@@ -32,6 +32,11 @@ dependencies {
 tasks.withType(JavaCompile).all {
   options.compilerArgs << '-Xlint:none'
   options.encoding = 'UTF-8'
+  if (it.name == 'classesJava11') {
+    options.compilerArgs.addAll(['--release', '11'])
+  } else if (it.name != 'comoileJava') {
```
danielfn, line 37: Typo

KengoTODA: Nice catch! Thanks

KengoTODA:
Found that not #996 but #259 is the root cause. 

Problem is that the java compiler adds redundant null-check in the generated code:
https://gist.github.com/KengoTODA/8011f3860ee3d66397b9b0e247ba82c4

KengoTODA:
@spotbugs/everyone please review this PR when you have time. Thanks in advance!

henrik242 (approved):
lgtm.

jscancella:
LGTM

wreulicke (approved):

Issue #390, opened 2017-09-21
Show warning for assertTrue(xxx instanceof SomeClass) in junit tests


trancexpress:
I run into numerous tests which make assumptions for specific types of objects by casting to that type. Before a lot of those casts, an `assertTrue` is used to validate that the assumption is correct.

This assertion however is far less verbose than the class cast exception _that would occur without the assertion_. Namely, when the assertion fails it will state that _false is not true_. As opposed to _class XXX cannot be casted to ZZZ_, without the assertion. In some cases, such as very sporadic test failures, the actual type of the object can give vital clues for the nature of the fail.

It would be great if a warning issue is shown on the line, explaining why such assertions are not a good practice in tests.

trancexpress:
Please add enhancement label as I'm not able to do so.

WarpspeedSCP:
I'd like to implement a detector for this.

@KengoTODA, should I do a pr to fb-contrib or to spotbugs for this?

KengoTODA:
Both of them are OK. If you want to run your detector with FindBugs or old SpotBugs, fb-contrib will be a good choice.

PR #1256, opened 2020-08-14
Implement  #390

WarpspeedSCP:
I've implemented #390 as a detector, `DontAssertInstanceofInTests`. 

This only covers usages of `Assert.assertTrue(obj instanceof Class)` which are present in the body of an `@Test` annotated method.

I would like to add usages such as `assertEquals(true, <instanceof expression>)` but I am not completely sure of the best way to go about it.

### DIFF
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -1664,8 +1664,7 @@ factory pattern to create these objects.</p>
     <Details>
       <![CDATA[
             <p>Detector for patterns in JUnit tests where the type of an object
-               is checked by asserting the instanceof operator before an object
-               is cast to a certain type. </p>
+               is checked by asserting the instanceof operator.</p>
             <p>
                This should be avoided as the ClassCastException that would result
                from an improper cast may provide more information regarding the
@@ -1695,6 +1694,8 @@ factory pattern to create these objects.</p>
             Asserting the type before casting would instead result in a less informative <code>"false is not true"</code>
             message.</p>
 
+            <p>If JUnit is used with hamcrest, the <a href="https://junit.org/junit4/javadoc/latest/index.html?org/hamcrest/core/IsInstanceOf.html"><code>IsInstanceOf</code></a>
+            class from hamcrest could be used instead.</p>
             ]]>
     </Details>
   </BugPattern>
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -1660,11 +1660,45 @@ factory pattern to create these objects.</p>
       ]]>
       </Details>
   </Detector>
+  <Detector class="edu.umd.cs.findbugs.detect.DontAssertInstanceofInTests">
+    <Details>
+      <![CDATA[
+            <p>Detector for patterns in JUnit tests where the type of an object
+               is checked by asserting the instanceof operator before an object
+               is cast to a certain type. </p>
+            <p>
+               This should be avoided as the ClassCastException that would result
+               from an improper cast may provide more information regarding the
+               cause of the error than a "false is not true" message which would
+               result from asserting the result of the instanceof operator.
+            </p>
+			<p>It is a fast detector</p>
+            ]]>
+    </Details>
+  </Detector>
   <!--
   **********************************************************************
   BugPatterns
   **********************************************************************
    -->
+
+  <BugPattern type="JUA_DONT_ASSERT_INSTANCEOF_IN_TESTS">
+    <ShortDescription> Asserting value of instanceof in tests is not recommended. </ShortDescription>
+    <LongDescription> Assertion of type {0} in {2} at {3} may hide useful information about why a cast may have failed.</LongDescription>
+    <Details>
+      <![CDATA[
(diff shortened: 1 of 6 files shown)


spotbugs/etc/messages.xml
```diff
@@ -1660,11 +1660,45 @@ factory pattern to create these objects.</p>
...
+      <![CDATA[
+            <p>Asserting type checks in tests is not recommended as a class cast exception message could better indicate
+            the cause of an instance of the wrong class being used than an instanceof assertion.</p>
+
+            <p>When debugging tests that fail due to bad casts, it may be more useful to observe the output of the
+            resulting ClassCastException which could provide information about the actual encountered type.
+            Asserting the type before casting would instead result in a less informative <code>"false is not true"</code>
+            message.</p>
+
+            ]]>
```
KengoTODA, line 1698: Do we have suggestion about how to fix the bug? 

For JUnit4 with hamcrest, I guess we can use [isInstanceOf(Class)](https://junit.org/junit4/javadoc/latest/org/hamcrest/core/IsInstanceOf.html) as replacement.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontAssertInstanceofInTests.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public void visitCode(Code obj) {
+        if (isTest)
+            super.visitCode(obj);
+    }
+
+    @Override
+    public void sawOpcode(int seen) {
```
KengoTODA, line 47: I'll confirm that we need to call `super.sawOpcode(seen)` or not.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontAssertInstanceofInTests.java
```diff
@@ -0,0 +1,76 @@
...
+        for (AnnotationEntry entry : obj.getAnnotationEntries()) {
+            String entryType = entry.getAnnotationType();
+            isTest |= "Lorg/junit/Test;".equals(entryType);
+        }
+    }
+
+    @Override
+    public void visitCode(Code obj) {
+        if (isTest)
+            super.visitCode(obj);
```
KengoTODA, line 43: If the purpose is skipping `sawOpcode(int)`, it could be better to implement [`beforeOpcode(int)`](https://javadoc.io/static/com.github.spotbugs/spotbugs/4.1.1/edu/umd/cs/findbugs/visitclass/DismantleBytecode.html#beforeOpcode-int-) instead.

WarpspeedSCP: does `beforeOpcode` control if we want to skip processing an entire method? The intention was to avoid processing any methods that weren't annotated as tests.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontAssertInstanceofInTests.java
WarpspeedSCP: I noticed that the super method was empty, so I didn't bother.

spotbugs/etc/messages.xml
WarpspeedSCP: I've added a suggestion to use it now.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontAssertInstanceofInTests.java
KengoTODA: confirmed. thanks

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontAssertInstanceofInTests.java
KengoTODA: I think so. The javadoc describes its responsibility as `return false if we should skip calling sawOpcode`, and the implementation is like below:

https://github.com/spotbugs/spotbugs/blob/c195f7e6632c078bcb8bbf8d3036476d5956f2f1/spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java#L877-L879

WarpspeedSCP: from my understanding, beforeOpcode would run on every opcode of a method.
Now that I look, there's a `shouldVisitCode` method in BytecodeScanningDetector which would do exactly what I was trying to do, so I'll just override that instead of that `visitCode` hack.

rydenius:
Following the advice of the documentation and using the instanceOf matcher, then proceeding the test by casting the object to that type in order to inspect it will then instead trigger a `BC_UNCONFIRMED_CAST_OF_RETURN_VALUE` error, something that does not happen after a regular instanceof call. So I need to introduce an exclude pattern for my test, or introduce a not that elegant if statement. Can you recommend a better way that was intended?

PR #1270, opened 2020-08-28
Use Java 14 to boost the performance of compilation

KengoTODA:
The performance of javac is improved continuously, then try to apply the latest release to shorten the build time.
This change will not affect the version of .class files in our artifacts, because we have `--release` option to specify the class file version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,10 +12,10 @@ jobs:
       - uses: actions/checkout@722adc6
         with:
           fetch-depth: 0
-      - name: Set up JDK 11
+      - name: Set up JDK 14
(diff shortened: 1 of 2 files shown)


KengoTODA:
[The build with Java 14](https://github.com/spotbugs/spotbugs/runs/1039514100) takes `12m20s`, and [the build with Java 11](https://github.com/spotbugs/spotbugs/runs/1010214110) takes `11m10s`. It's really rough comparation, but seems less enhancement in our case, so I'll close this PR.

iloveeclipse:
Unfortunately with every new release compiler gets slower because they have more cases to check, especially with -- release option. JLS changes constantly, so it bloats the compiler logic...

PR #1278, opened 2020-09-03
Remove dependencies on deprecated Gradle feature

KengoTODA:
The `compileOnly` configuration is not expected to refer from other projects, so load properties file in `eclipsePlugni-junit` and `eclipsePlugin-test` projects directly.

[This build scan](https://scans.gradle.com/s/7kyeies6rngmm) proves that we have no deprecated feature in the current build.

Also bump up Gradle to the latest version 6.6.1.

### DIFF
--- a/eclipsePlugin-junit/build.gradle
+++ b/eclipsePlugin-junit/build.gradle
@@ -1,11 +1,33 @@
 apply from: "$rootDir/gradle/checkstyle.gradle"
 apply from: "$rootDir/gradle/jacoco.gradle"
 
+def localProps = new Properties()
+localProps.load(new FileInputStream("$rootDir/eclipsePlugin/local.properties"))
+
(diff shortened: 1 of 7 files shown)


eclipsePlugin-junit/build.gradle
```diff
@@ -1,11 +1,33 @@
 apply from: "$rootDir/gradle/checkstyle.gradle"
 apply from: "$rootDir/gradle/jacoco.gradle"
 
+def localProps = new Properties()
+localProps.load(new FileInputStream("$rootDir/eclipsePlugin/local.properties"))
+
+def eclipseRootDir = new File(localProps.getProperty('eclipseRoot.dir'))
```
wreulicke, line 7: Can we use just `file(...)` instead of `new File(...)`?

```suggestion
def eclipseRootDir = file(localProps.getProperty('eclipseRoot.dir'))
```

ThrawnCA (approved):

PR #1284, opened 2020-09-08
fix for issue 1161

stoty:


spotbugs/src/main/java/edu/umd/cs/findbugs/BugCollection.java
```diff
@@ -371,4 +371,8 @@
     public void setApplySuppressions(boolean applySuppressions);
 
     public void bugsPopulated();
+
+    public boolean remove(BugInstance bugInstance);
```
ThrawnCA, line 375: I'm uncertain about adding to an interface. @KengoTODA Would this require a new major release?

KengoTODA: Hmm, not so sure. But usually, plugin developers won't implement this interface.
I think the patch release isn't suitable but the minor release can provide such change which has no impact for tool users.

stoty:
This seems to be fixed by #1436

PR #1302, opened 2020-09-22
Resolved XPathException generating html report with plain.xsl

jqyp:
See issue #1025: we had troubles to transorm if BugInstance contained multiple Class elements. This commit  fixes plain.xsl.

----

### DIFF
--- a/spotbugs/src/xsl/plain.xsl
+++ b/spotbugs/src/xsl/plain.xsl
@@ -205,7 +205,7 @@
 				<xsl:apply-templates select="$warningSet">
 					<xsl:sort select="@priority"/>
 					<xsl:sort select="@abbrev"/>
-					<xsl:sort select="Class/@classname"/>
+					<xsl:sort select="Class[1]/@classname"/>
 				</xsl:apply-templates>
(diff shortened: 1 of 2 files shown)


jqyp:
I'm not good at XSLT, but I hope I did all right.

ThrawnCA (approved):

spotbugs/src/xsl/plain.xsl
```diff
@@ -205,7 +205,7 @@
 				<xsl:apply-templates select="$warningSet">
 					<xsl:sort select="@priority"/>
 					<xsl:sort select="@abbrev"/>
-					<xsl:sort select="Class/@classname"/>
+					<xsl:sort select="Class[1]/@classname"/>
```
ThrawnCA, line 208: Does this still work if there is only one bug instance?

KengoTODA: According to W3C, `[1]` is the first element, so it should work.

* https://www.w3.org/TR/xslt20/
* https://www.w3schools.com/xml/xpath_syntax.asp

I'll check it later.

Issue #1311, opened 2020-09-28
./gradlew eclipse generates broken spotbugs/spotbugs/.classpath file


iloveeclipse:
I'm back to patching Spotbugs again, I haven't built it for a long time now, and it looks like the "eclipse" task is broken.

The generated spotbugs/spotbugs/.classpath file is not valid, I see following two errors after a "successful" execution on a spotbugs-spotbugs project:

```
git clean -fdxq
echo "eclipseRoot.dir=/data/eclipse_zoo/eclipse_463" > eclipsePlugin/local.properties
./gradlew build
./gradlew eclipse
```
Eclipse shows:

```
Build path contains duplicate entry: '.libs/asm-commons-8.0.1.jar' for project 'spotbugs-spotbugs'
Project 'spotbugs-spotbugs' is missing required library: '.libs/AppleJavaExtensions-1.4.jar'
```
In the eclipsePlugin/.classpath I see that *all* entries (not just `.libs/asm-commons-8.0.1.jar`) are duplicated now, there are entries that look like this:
```
	<classpathentry sourcepath="/home/aloskuto/.gradle/caches/modules-2/files-2.1/org.ow2.asm/asm-commons/8.0.1/ba5ccc27aa05831a74c01c17cb8058b9c2bfdb26/asm-commons-8.0.1-sources.jar" kind="lib" path=".libs/asm-commons-8.0.1.jar" exported="true">
		<attributes>
			<attribute name="javadoc_location" value="jar:file:/home/aloskuto/.gradle/caches/modules-2/files-2.1/org.ow2.asm/asm-commons/8.0.1/f85e00e7d62f2bfe18294551d8e78f1298336e8f/asm-commons-8.0.1-javadoc.jar!/"/>
			<attribute name="gradle_used_by_scope" value="gui,guiTest,main,test"/>
		</attributes>
	</classpathentry>
```
and much simpler entries that look like this:
```
	<classpathentry kind="lib" path=".libs/asm-commons-8.0.1.jar" exported="true">
		<attributes>
			<attribute name="gradle_used_by_scope" value="gui,guiTest,main,test"/>
		</attributes>
	</classpathentry>
```

So we have two problems:

1. *All* classpath entries in generated .classpath are duplicated
2. `.libs/AppleJavaExtensions-1.4.jar` is missing

KengoTODA:
Solved by #1312

PR #1312, opened 2020-09-28
Fixed issue #1311

iloveeclipse:
According to [1], project.configurations.testRuntime is default for
Eclipse (I assume it contains everything). So including compileClasspath
& guiCompileClasspath shouldn't be needed (and avoids that duplicated
classpath entries are generated in the .classpath file).

Removed "testImplementation sourceSets.gui.output" because it caused
generated class files to be added too the project classpath in Eclipse.

Added "appleExtensions" to allow Eclipse compilation see apple jar
without referencing "guiCompileClasspath" that added duplicated entries
to the classpath too.

[1] https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.eclipse.model.EclipseClasspath.html,

----

### DIFF
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -6,6 +6,8 @@ apply from: "$rootDir/gradle/javadoc.gradle"
 configurations {
   // used only in distribution. It is not listed in pom.xml, so users like maven plugin don't use this dependency.
   logBinding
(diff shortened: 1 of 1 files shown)


KengoTODA (approved):
I've confirmed that the change fix #1311. We have the apple extension in `.classpath` and no entry is duplicated. Nice catch, thank you!

Issue #1316, opened 2020-09-29
spotbugsTestCases is not compilable anymore in Eclipse


iloveeclipse:
spotbugsTestCases project is not compilable anymore in Eclipse, because of Java 11 "modules" crap that doesn't allow to compile Java 11 code if the modulepath contain two "modules" that ship same package. In current case we have Java 11 SDK that has `org.w3c.dom` declared in `java.xml` package, and *some other* dependency:

```
The package org.w3c.dom is accessible from more than one module: <unnamed>, java.xml
```

If I manually disable two dependencies below from .classpath, "accessible from more than one module" errors disappear: 
```
spotbugs-spotbugs
jaxen-1.1.6.jar
```

I have to grock who and how brings which dependency to the module path that has `org.w3c.dom` included.

iloveeclipse:
OK, looks like the problem is coming from the spotbugs dependency on `dom4j` that in turn optionally depends on `jaxen`
In the 1.1.6 version of jaxen we import there is `org.w3c.dom` package defined. I think updating to 1.2.0 should fix the problem.

Issue #1315, opened 2020-09-29
Spotbugs test projects have compilation errors if imported in Eclipse


iloveeclipse:
I've just started to look at spotbugs issues again, and wanted to open test projects, but I see that most of them aren't compilable anymore in Eclipse.
```
spotbugsTestCases
spotbugs-tests
test-harness
test-harness-core
test-harness-jupiter
```
I will try to push gradle build file patches that should fix that, I need clean workspace before starting to work on other issues.

iloveeclipse:
I've fixed all but spotbugsTestCases problems, for that one we need some more work, I will open an extra issue.

Issue #1254, opened 2020-08-13
spotbugs reports VR_UNRESOLVABLE_REFERENCE when code is compiled with Java 11


trancexpress:
We have some code with nested inner classes calling private methods in the outer class. As we are trying to compile our product with Java 11, we notice a ton of `VR_UNRESOLVABLE_REFERENCE` reports from `spotbugs`.

This seems to boil down to JEP-181: https://openjdk.java.net/jeps/181

Synthetic accessors seem to be no longer necessary; the compiler does not generate them. From what I can tell, when code is compiled with Java 11, `ResolveAllReferences.visit()` falls into the `ConstantCP` case for method calls (`CONSTANT_Methodref`) that previously needed synthetic accessors (but now no longer do). The method is not found, as its private (see how `defined` is filled in `ResolveAllReferences.addAllDefinitions()`). `spotbugs` then reports `VR_UNRESOLVABLE_REFERENCE`, which is **false**; our product tests run just fine with the Java 11 compiled code.

Example:

```
...:/home/sandreev$ /usr/lib/jvm/java-11/bin/javac Foo.java
...:/home/sandreev$ javap Foo.class
Compiled from "Foo.java"
public class Foo {
  public Foo();
  public static void main(java.lang.String[]);
  public void bar();
}
...:/home/sandreev$ /usr/lib/jvm/java-1.8.0/bin/javac Foo.java
...:/home/sandreev$ javap Foo.class
Compiled from "Foo.java"
public class Foo {
  public Foo();
  public static void main(java.lang.String[]);
  public void bar();
  static void access$000(Foo);
}
...:/home/sandreev$ cat Foo.java 
public class Foo {

    public static void main(String[] args) {
        new Foo().bar();
    }

    public void bar() {
        java.lang.Runnable r = new java.lang.Runnable() {
            @java.lang.Override
            public void run() {                                                                                                                                                                                                              
                foo();                                                                                                                                                                                                                       
            }                                                                                                                                                                                                                                
        };                                                                                                                                                                                                                                   
        r.run();                                                                                                                                                                                                                             
    }                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
    private void foo() {                                                                                                                                                                                                                     
        System.out.println("Foo.foo()");                                                                                                                                                                                                     
    }                                                                                                                                                                                                                                        
}
```

See my comment below for a reproducer.

trancexpress:
Snippet to reproduce:

```
public class Foo {

[...]

PR #1317, opened 2020-09-29
Test project compilation fixes for Eclipse

iloveeclipse:
This fixes issues #1316 and #1315.
----

### DIFF
--- a/test-harness/build.gradle
+++ b/test-harness/build.gradle
@@ -4,7 +4,7 @@ apply from: "$rootDir/gradle/maven.gradle"
 
 dependencies {
   compileOnly 'junit:junit:4.13'
-  compileOnly project(':spotbugs')
+  testImplementation project(':spotbugs')
   api project(':test-harness-core')
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Changed
 * Bump jaxen from 1.1.6 to 1.2.0 supporting Java 11 compilation ([#1316](https://github.com/spotbugs/spotbugs/issues/1316))
 * Fixed error dialog on cancelling SpotBugs job in Eclipse ([#1314](https://github.com/spotbugs/spotbugs/issues/1314))
+* Fixed IllegalArgumentException in OpcodeStack.constantToInt ([#893](https://github.com/spotbugs/spotbugs/issues/893))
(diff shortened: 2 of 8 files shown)


test-harness/build.gradle
```diff
@@ -4,7 +4,7 @@ apply from: "$rootDir/gradle/maven.gradle"
 
 dependencies {
   compileOnly 'junit:junit:4.13'
-  compileOnly project(':spotbugs')
+  testImplementation project(':spotbugs')
```
KengoTODA, line 7: Could you consider to add explanations in `CHANGELOG` for this change?  
I think users of `test-harness` and `test-harness-jupiter` needs to this change (from compile-only scope to compile scope).

KengoTODA (changes requested):

test-harness/build.gradle
iloveeclipse: I'm not sure how the changed compilation setting here can affect users. Dependencies itself aren't changed. That should be fully transparent for clients.

iloveeclipse: @KengoTODA : anything else is missing?

iloveeclipse: Ping...

trancexpress: Hi @KengoTODA ,

to work on https://github.com/spotbugs/spotbugs/issues/1254, I had to cherry-pick the changes here. Can the changes get merged so I can work with actual ```master```?

Best regards and thanks,
Simeon

wreulicke: @iloveeclipse 
I think changes in `test-harness` affect its dependency.
That dependency will be changed.
`api` configuration makes `test-harness` dependent on that.

iloveeclipse: @wreulicke : not sure what do you mean by that. How this change can affect *clients* of test-harness? I'm not the gradle user, I have no idea why there are multiple dependencies chains invented by gradle, and how that, once a module is compiled, can affect its clients - I really can't understand that. And btw, let assume test-harness clients would be now also depend on spotbugs - what's wrong by that? How that would affect them, they had spotbugs on classpath already anyway?

My problem is that current state on master is not usable with Eclipse, so contributing is a pain, nothing compiles. That is way more severe issue as a theoretical possibility to somehlw break one or two test-harness clients that may exists in the world.

KengoTODA:
Sorry for my late reaction, recently it's a long national holiday in China.

I'll submit another PR to make `CHANGELOG` better.

PR #1319, opened 2020-10-01
Run test for issue408 properly

wreulicke:
We do not run Issue408Test because we use java11 to build.
We can remove `assumeThat` for java 9 now.

----

  - This PR does not change spotbugs codes for users.


KengoTODA (approved):
Yes, this is necessary. Nice catch & thank you!

PR #1334, opened 2020-10-13
fix for issue 1254, support nested private member access from JEP 181

trancexpress:
JEP 181 defines relaxed access rules for private members of nested
classes, removing the need for synthetic methods to access private
members of inner/outer classes. Instead, byte code contains e.g. a
method reference constant and the relaxed access rules make the method
call legal.

This changes makes spotbugs aware of JEP 181, so that in case of nested
private access spotbugs does not report unresolvable members/types, or
unused private members.

----


iloveeclipse (changes requested):
The rest looks good

iloveeclipse (approved):

iloveeclipse:
Next time please write `Fixes issue #1254` in commit or PR, so github could automatically close the right issue (issue #1254 in that case).

Issue #79, opened 2016-12-07
False positive OBL_UNSATIFIED_OBLIGATION with try with resources


jsotuyod:
Local issue to track https://github.com/findbugsproject/findbugs/issues/99

jsotuyod:
The issue can be reproduced and is confirmed. I'll be working on a fix.

jsotuyod:
This one is going to be harder to fix than expected. The detector seems to be deeply broken as it's failing on lot of scenarios were it shouldn't.

The database should recognize that closing a `Statement` closes the `ResultSet`, 'though the opposite is not currently true:

```java
database.addEntry(new MatchMethodEntry(new SubtypeTypeMatcher(BCELUtil.getObjectTypeInstance("java.sql.Statement")),
        new ExactStringMatcher("close"), new ExactStringMatcher("()V"), false, ObligationPolicyDatabaseActionType.DEL,
        ObligationPolicyDatabaseEntryType.STRONG, statement, resultSet));
database.addEntry(new MatchMethodEntry(new SubtypeTypeMatcher(BCELUtil.getObjectTypeInstance("java.sql.ResultSet")),
         new ExactStringMatcher("close"), new ExactStringMatcher("()V"), false, ObligationPolicyDatabaseActionType.DEL,
         ObligationPolicyDatabaseEntryType.STRONG, resultSet));
```

However, this is failing even without the use-with-resources construct:

```java
public class Issue79 {
  private static final String QUERY = "";

  public void f(Connection cnx) throws SQLException {
    PreparedStatement st = null;
    ResultSet rs = null;
    try {
      st = cnx.prepareStatement(QUERY);
      rs = st.executeQuery();
      while (rs.next()) {
        System.out.println(rs.getString("ID"));
      }
    } finally {
      /*
       * The statement closes the result set, and there is no scenario where st may be null
       * but not the resultset, however an unsatisfied obligation is reported on the resultset
      */
      if (st != null) {
    	st.close();
      }
    }
  }
}
```

Also of note, the current implementation can't track which result set is associated with which statement, so when using more than one it can produce lots of false positives and false negatives...

We should probably drop this one for release 4.0.0 and think of a deep refactor for this detector.

PR #1347, opened 2020-10-21
Fix issue #79

gloNelson:
As @jsotuyod stated,  "The database recognizes that closing a Statement closes the ResultSet". 

However,  the ResultSet obligation also should be deleted when removing the statement obligation from all states.

The test case was added for review.
----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/obl/ObligationAnalysis.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/obl/ObligationAnalysis.java
@@ -215,6 +215,17 @@ public class ObligationAnalysis extends ForwardDataflowAnalysis<StateSet> {
                             + edge.getSource().getLastInstruction());
                 }
                 fact.deleteObligation(comparedObligation, edge.getTarget().getLabel());
+
+                // closing a Statement closes the ResultSet
+                Obligation statement = database.getFactory().getObligationByName("java.sql.Statement");
+                if (comparedObligation.equals(statement)) {
+                    Obligation resultSet = database.getFactory().getObligationByName("java.sql.ResultSet");
+                    fact.deleteObligation(resultSet, edge.getTarget().getLabel());
+                    if (DEBUG_NULL_CHECK) {
+                        System.out.println("Deleting " + resultSet.toString() + " on edge from comparison "
+                                + edge.getSource().getLastInstruction());
+                    }
+                }
             }
         }
     }
--- a/spotbugsTestCases/src/java/ghIssues/Issue79.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue79.java
@@ -1,0 +1,30 @@
+package ghIssues;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public class Issue79 {
+    private static final String QUERY = "";
+
+    public void f(Connection cnx) throws SQLException {
+        PreparedStatement st = null;
+        ResultSet rs = null;
+        try {
+            st = cnx.prepareStatement(QUERY);
+            rs = st.executeQuery();
+            while (rs.next()) {
+                System.out.println(rs.getString("ID"));
+            }
+        } finally {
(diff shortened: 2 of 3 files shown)


KengoTODA (changes requested):
Thanks for your contribution! The change seems LGTM, but the build fails due to unexpected code format.
Please pull the latest topic branch, and run `./gradlew spotlessApply` to format your code.

KengoTODA (changes requested):
The change in the `CHANGELOG.md` is missing. Please add one entry to describe your change.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/obl/ObligationAnalysis.java
```diff
@@ -215,6 +215,17 @@ public void edgeTransfer(Edge edge, StateSet fact) throws DataflowAnalysisExcept
...
                 }
                 fact.deleteObligation(comparedObligation, edge.getTarget().getLabel());
+
+                // closing a Statement closes the ResultSet
+                Obligation statement = database.getFactory().getObligationByName("java.sql.Statement");
+                if (comparedObligation.equals(statement)) {
+                    Obligation resultSet = database.getFactory().getObligationByName("java.sql.ResultSet");
+                    fact.deleteObligation(resultSet, edge.getTarget().getLabel());
+                    if (DEBUG_NULL_CHECK) {
+                        System.out.println("Deleting " + resultSet.toString() + " on edge from comparison "
```
ThrawnCA, line 225: Can this be a proper logging statement?

spotbugsTestCases/src/java/ghIssues/Issue79.java
```diff
@@ -0,0 +1,30 @@
...
+    private static final String QUERY = "";
+
+    public void f(Connection cnx) throws SQLException {
+        PreparedStatement st = null;
+        ResultSet rs = null;
+        try {
+            st = cnx.prepareStatement(QUERY);
+            rs = st.executeQuery();
+            while (rs.next()) {
+                System.out.println(rs.getString("ID"));
```
ThrawnCA, line 18: Can this be a proper logging statement?

raghav-deepsource: I'd suggest using sl4j loggers instead. println is not fun to deal with when debugging with log statements.

raghav-deepsource: I'd say using sl4j loggers instead would be better.

wreulicke: @raghav-deepsource 
It seems that this is a test fixture. So, I think we don't need use slf4j loggers here.
But, It looks that we can omit this block.

```suggestion
```

KengoTODA:
@gloNelson could you check this PR too? I'm going to close stale PRs later.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/obl/ObligationAnalysis.java
gloNelson: I just copied from line 214 to line 215 [ObligationAnalysis.java ](https://github.com/spotbugs/spotbugs/blob/3b4da42bdb627c0bf97fc97629a563edb880e44b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/obl/ObligationAnalysis.java#L214-L215). Maybe all `system.out.println()` in this file should be converted to proper logging statement to keep consistency?

spotbugsTestCases/src/java/ghIssues/Issue79.java
gloNelson: @wreulicke, I have modified the code according to your suggestion.  Thanks :)

gloNelson: I modified it based on @ThrawnCA's suggestion.
If this block is omitted, SpotBugs will detect `DLS: Dead storage for local variables` bug instance and the test will fail.

gloNelson:
new pr is #1479

Issue #749, opened 2018-09-08
Class search (in the GUI's class name filter) should be case-insensitive


cyril-briquet:
(1) Searching for "MyClass", "myclass", "Myclass" should all return the same results.

If more than one class matches (e.g. if close syntactical variants do exist), all the matching classes should be returned.
(And this should be flagged as a bug as well.)

(2) Even better: make class search typo-insensitive, by e.g. implementing fuzzy search based on the edit distance (or Levenshtein distance).

(3) Dream: automatic completion of class names?

andipabst:
I'd be willing to work on this. 
@KengoTODA Is it possible to use the Apache Commons Text Library, since it would include an implementation of the Levenshtein distance?

dharshiniramu:
hey can I work on this issue?

hazendaz:
@dharshiniramu It seems all but 3rd item where completed based on #1348.   Feel free to work on it.

MayureshRajesh:
Hi @hazendaz, 

Since this had been inactive for a while, I went ahead and implemented the third item (class name auto-completion in the GUI class search filter) along with unit tests and GUI verification. 

I've opened PR #4280 for review. This is my first open-source contribution to SpotBugs, so any feedback or suggestions on the implementation would be greatly appreciated!

PR #1348, opened 2020-10-21
Case-Insensitivity and Levenshtein-distance for the Class Name Filter in the GUI

andipabst:
This PR introduces case-insensitive and fuzzy searching for the class name filter in the GUI. This is the first half for issue #749. 
----


spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -265,13 +268,34 @@ public static boolean matchedPrefixes(String[] classSearchStrings, @DottedClassN
...
-            if (p.length() > 0 && className.indexOf(p) >= 0) {
+            if (p.length() > 0 && (containsIgnoreCase(className, p) || fuzzyMatch(className, p))) {
                 return true;
             }
         }
 
         return false;
+    }
+
+    private static boolean containsIgnoreCase(String str, String searchStr) {
```
ThrawnCA, line 279: If we're using commons-text, then that means that Apache Commons StringUtils is available.

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
andipabst: Thanks, I did not notice that. I have corrected it to use StringUtils#containsIgnoreCase now!

andipabst:
@ThrawnCA I have modified the PR as per your suggestion, could you please have another look? Thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -265,13 +267,26 @@ public static boolean matchedPrefixes(String[] classSearchStrings, @DottedClassN
...
+    /**
+     * Perform a fuzzy matching, by comparing the Levenshtein distance of the
+     * simple class name and the search string.
+     * 
+     * @param className the full class name
+     * @param searchString the search string
+     * @return true, if the strings are similar, false otherwise
+     */
+    private static boolean fuzzyMatch(String className, String searchString) {
+        // compare to a maximum Levenshtein distance of 3
```
ThrawnCA, line 287: What is Levenshtein distance 3? How does it compare to other distances? Please describe or link to a description.

spotbugs/src/test/java/edu/umd/cs/findbugs/util/ClassNameTest.java
```diff
@@ -65,4 +64,21 @@ public void testGetPrimitiveType() {
     public void testExtractClassNameBad() {
         ClassName.extractClassName("L[Ljava/lang/Integer;");
     }
+
+    @Test
+    public void testMatchedPrefix() {
```
ThrawnCA, line 69: It might be worth considering a parameterised test for this - although JUnit doesn't have as much support for that as TestNG does. Still, iterating through a collection of test cases might be more readable and maintainable than having a repeated line for each one.

andipabst:
@ThrawnCA I added a description for the Levenshtein distance and an iteration for the different test cases. Is this okay for you?

ThrawnCA (approved):
Looks good. Please try to avoid force-pushing.

KengoTODA (changes requested):
Really cool, thanks for your contribution!

Please consider to `git fetch && git merge --no-ff origin/master` before you add yours, because other PR like #1358 may add other entries to this CHANGELOG.

andipabst:
@ThrawnCA Sorry, I was used to force-pushing from other projects with a rule of one commit per pull request.
@KengoTODA Thanks for the hint, I fetched and merged the other commits, reworded the changelog and bumped the version.

build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
   id "com.github.spotbugs" version "4.5.1"
 }
 
-version = '4.1.5-SNAPSHOT'
+version = '4.2.0-SNAPSHOT'
```
ThrawnCA, line 8: We don't necessarily want to increment the version number yet.

ThrawnCA (changes requested):

build.gradle
KengoTODA: Sorry, but I suggested to do so at https://github.com/spotbugs/spotbugs/pull/1348#discussion_r515011220
I think this is new feature for SpotBugs, isn't it? 🤔

ThrawnCA (approved):

KengoTODA (approved):

PR #1352, opened 2020-10-23
Fixing issues/performance when using Sarif flag

### LINKED ISSUE #1281: SARIF report missing physical location information
A SARIF report from SpotBugs 4.1.2 contains SARIF results with no physical location information, only logical location information. For example:
```json
      "results": [
        {
          "ruleIndex": 0,
          "level": "note",
          "locations": [
            {
              "logicalLocations": [
                {
                  "kind": "type",
                  "name": "ClassReader",
                  "fullyQualifiedName": "aj/org/objectweb/asm/ClassReader.java:[line 3072]"
                }
              ]
            }
          ],
          "ruleId": "UC_USELESS_CONDITION",
          "message": {
            "arguments": [
              "frameType >= 248 (0xf8)"
            ],
            "id": "default"
          }
        },
```
When I was working with @KengoTODA on validating the SARIF report, the sample I saw _did_ have physical location information, for example:
```json
      "results": [
        {
          "ruleIndex": 0,
          "level": "note",
          "locations": [
            {
              "physicalLocation": {
                "region": {
                  "endLine": 166,
                  "startLine": 166
                },
                "artifactLocation": {
                  "uri": "edu/umd/cs/findbugs/AnalysisError.java",
                  "uriBaseId": "-176137563"
                }
              },
              "logicalLocation": [
                {
                  "kind": "type",
                  "name": "AnalysisError",
                  "fullyQualifiedName": "edu/umd/cs/findbugs/AnalysisError.java:[line 166]"
                }
              ]
            }
          ],
```

Is this a regression? Is full SARIF support scheduled for a later version than 4.1.2?

Thanks,
Larry

@boandron @michaelcfanning @lukadlet

### LINKED ISSUE #1349: How to use SpotBugs with SARIF?
Hi All,

I have been trying to compile and run the solution to test the sarif output.

After using the gradlew script, i was able to build the .jar file, but when I double click the .jar, the interface isn't launching.

My questions:
1. should the double click work in Windows?
2. how can I test and generate a sarif locally?

What I did to test:
```
spotbugs\build\libs> java .\spotbugs-4.1.5-SNAPSHOT.jar
Error: Could not find or load main class .\spotbugs-4.1.5-SNAPSHOT.jar

spotbugs\build\libs> java -jar .\spotbugs-4.1.5-SNAPSHOT.jar -gui
no main manifest attribute, in .\spotbugs-4.1.5-SNAPSHOT.jar
```

Followed this tutorial: https://spotbugs.readthedocs.io/en/latest/running.html

Thank you!

eddynaka:
Resolves #1349 , Resolves #1281 

----


spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -178,9 +191,8 @@ JSONObject toJSONObject() {
 
         Region(int startLine, int endLine) {
             assert startLine > 0;
-            assert endLine > 0;
             this.startLine = startLine;
-            this.endLine = endLine;
+            this.endLine = startLine == endLine ? 0 : endLine;
```
eddynaka, line 195: if startline is equal endline, we should not print, since they have the same information.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -192,7 +204,13 @@ JSONObject toJSONObject() {
         }
 
         JSONObject toJSONObject() {
-            return new JSONObject().put("startLine", startLine).put("endLine", endLine);
+            JSONObject json = new JSONObject().put("startLine", startLine);
+
+            if (endLine != 0) {
```
eddynaka, line 209: based on the change in the constructor, I'm just checking if endline is different from zero. If yes, we have to concat the endline to the json

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -270,14 +291,13 @@ static LogicalLocation fromStackTraceElement(@NonNull StackTraceElement element)
...
             ClassAnnotation primaryClass = bugInstance.getPrimaryClass();
-            SourceLineAnnotation sourceLine = bugInstance.getPrimarySourceLineAnnotation();
             return bugInstance.getAnnotations().stream().map(annotation -> {
                 String kind = findKind(annotation);
                 if (kind == null) {
                     return null;
                 }
                 String name = annotation.format("givenClass", primaryClass);
-                return new LogicalLocation(name, null, kind, sourceLine.format("full", primaryClass), null);
+                return new LogicalLocation(name, null, kind, null, null);
```
eddynaka, line 300: for now, i'm removing the fullyQualified property for one reason: that property should point to the method/parameter/etc. Right now, it is almost the same as the physical location, which is wrong.

When I was taking a look, I saw that the annotations have the content that we need, but I couldn't figure out yet how annotation to choose to add here. Because, some times, we have a few:

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -41,11 +41,13 @@
     }
 
     JSONObject toJSONObject() {
-        JSONObject messageStrings = new JSONObject().put("default", new JSONObject().put("text", defaultText));
+        String textEndsWithPeriod = !defaultText.endsWith(".") ? defaultText + "." : defaultText;
```
eddynaka, line 44: We have a pattern that each text should end with a period. So this check is doing that.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -137,6 +137,19 @@ JSONObject toJSONObject() {
                     throw new UncheckedIOException(e);
                 }
             });
+
+            if (location.isPresent()) {
+                return location;
+            }
+
+            try {
+                String path = bugAnnotation.format("full", classAnnotation);
```
eddynaka, line 146: added a fallback in case we don't have a location yet. This is printing the same thing that were being used in logicalLocation.fullyqualifiedname

eddynaka:
Hi @KengoTODA ,

I finished the updates and added some comments.

Locally I can build and test fine. In the pipeline, it's showing build failed, but I couldn't figure out what is causing it. Can you help me?

Thank you!

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Location.java
```diff
@@ -269,16 +292,50 @@ static LogicalLocation fromStackTraceElement(@NonNull StackTraceElement element)
...
-            }).filter(Objects::nonNull).findFirst();
+            }
+
+            if (annotation == null) {
+                return null;
+            }
+
+            String kind = findKind(annotation);
+            String name = annotation.format("givenClass", classAnnotation);
+            return Optional.of(new LogicalLocation(name, null, kind, fullyQualifiedName, null));
```
eddynaka, line 338: I changed the logic to be more precise.
If the bugAnnotation has only class in the annotations, so that is the name
if the bugAnnotation has method, so that is the name
if the bugAnnotation has field, so it is class + field
if the bugAnnotation has variable, so it is method + field

jhutchings1:
👋🏻  Hi folks, love seeing this PR because it should make it so that SpotBugs can plug into GitHub's code scanning platform going forward too. Thanks @eddynaka ! I'd love to get that integration work scheduled on the GitHub side, but I think it's blocked on this. Is there an ETA for this PR review? That'll help me set expectations internally 🙂

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
ThrawnCA: "if not...else" is a double-negative; can this be reversed to be more easily readable?

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
eddynaka: just updated! thanks for the review!

ThrawnCA (approved):

KengoTODA (approved):
Great, thank you! And sorry for my lazy reaction!

Issue #566, opened 2018-02-23
Rank matcher and its documentation do not match


matthiasjordan:
According to the documentation, the Rank matcher "matches warnings with a particular bug rank". According to its implementation, it matches if the bug's rank is >= the matcher's rank. This is in contradiction of the documentation. Also, I don't see the use case for "let's break the build if any bug is found that is less important than rank X".

There is a workaround using <Not> and <And> but it makes the filter code ridiculous (matching exactly rank 4 would need to match "4 and not 5").

KengoTODA:
closed by #1359

PR #1359, opened 2020-10-30
Rank matcher and its documentation do not match - #566

jeff303:
Updating docs to use "at least" terminology for rank matcher

----

- [N/A] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -50,7 +50,7 @@ Same as ``<Confidence>``, exists for backward compatibility.
 <Rank>
 ^^^^^^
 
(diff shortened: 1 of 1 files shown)


ThrawnCA (approved):

matthiasjordan:
Thanks for fixing this. But what about the semantics of this rule now? According to the documentation, 1 is scariest and 20 is least scary. When I add a rule to match bugs relative to rank 5, e.g., then the build breaks even at the most silly bug, but not, if severe bugs are found. This seems a bit odd, since I guess most users would like to break the build if severe bugs are found and might want to skip those less troublesome. Such a rule is now only possible to implement using a complex AND construct. It would be better to have the semantics of the rule reversed - i.e. rank 5 matches everything at rank 5 or more severe, i.e. 1, 2, 3, 4, 5, so >= rank, not <= rank.

PR #1424, opened 2021-02-10
Make UWF_NULL_FIELD reporting the line number

### LINKED ISSUE #1368: UWF_NULL_FIELD doesn't report line number
Hi :)
Thank you for this wonderful tool :)
We noticed at @codacy that the rule `UWF_NULL_FIELD` doesn't report any line number.
We're using Spotbugs API to get results line numbers using `bugInstance.getPrimarySourceLineAnnotation.getStartLine`
For this rule it reports `-1`

KengoTODA:
The reported problem caused because we clear `data.fieldAccess` when we detect access on field.
I guess it's designed for `URF_UNREAD_FIELD` which does not need `data.fieldAccess` for accessed fields.

This PR changes the behaviour to keep  `data.fieldAccess` even when we detect access on field.
Instead, we clear it only when the field is updated. This change should increase memory footprint, so I clear `data.fieldAccess` after the report is created by this detector.

close #1368

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2021-??-??
 
+### Fixed
+* `UWF_NULL_FIELD` doesn't report line number ([#1368](https://github.com/spotbugs/spotbugs/issues/1368))
+
 ### Changed
 * Bump ASM from 9.0 to 9.1 supporting JDK17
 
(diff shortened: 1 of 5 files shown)


KengoTODA:
Not sure why the `:spotbugs:spotbugsTest` fails even though we set `spotbugsTest.ignoreFailures = true`.

https://github.com/spotbugs/spotbugs/runs/1869042533#step:6:351

ThrawnCA (approved):

Issue #1427, opened 2021-02-11
Bug ID vs Bug name


eddynaka:
Hi,

Looking at the examples, I saw that the bug contains a type / name that looks similar:
```
CNT: Rough value of known constant found (CNT_ROUGH_CONSTANT_VALUE)
```

When we implemented the SARIF integration, the Rule.Id should be opaque, which means something like "SPOT0001".

What do you think about this? Is it possible to set an ID so we can use in the sarif converter?

Thank you!

KengoTODA:
Yes, currently we simply map the Bug ID to SARIF's rule id.
https://github.com/spotbugs/spotbugs/blob/f579c4b4f63dbbf8023385fab13ec84923519999/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java#L72

> which means something like "SPOT0001".

One idea is that, SpotBugs bug patterns have an `abbrev` so we can generate opaque ID like `SPOTBUGS_${abbrev}_${number}` that covers not only rules in SpotBugs core but also rules in 3rd party plugins.

But still not sure how we need to generate the `number` part. Better to use Bug ID's hashcode? Any idea? 🤔

eddynaka:
Hi @KengoTODA ,

well, for, example, I have two examples:
- binskim: started from 2001 for a specific set and 3001 for another
- sarif-sdk: started from 1001 for a specific set and 2001 for another

besides that, it's a random start and we fixed those numbers "forever".

From the spec:
```
A reportingDescriptor object SHALL contain a property named id whose value is a string. In the case of a rule, id SHALL contain a stable identifier for the rule and SHOULD be opaque. In the case of a notification, id does not need be a stable, opaque identifier; it MAY be a user-readable identifier.

EXAMPLE: "id": "CA2101"

NOTE 1: Rule identifiers must be stable for two reasons:

·         So build automation scripts can refer to specific checks, for example, to disable them, without the risk of a script breaking if a rule id changes.

·         So result management systems can compare results from one run to the next, without erroneously designating results as “new” because a rule id has changed.

[...]

PR #1429, opened 2021-02-12
Update SARIF output Rule Id to Opaque Rule Id

yongyan-gh:
As described in issue #1427

* Changed Rule Id value from "id" to "SPOTBUGS_HashCode(id)"
* Added name property to Rule, value is original rule "id" (BugPattern.type)
* Changed Result RuleId value to "SPOTBUGS_HashCode(ruleId)" accordingly


yongyan-gh:
Attached sample sarif logs generated before/after the change for comparing. 
[sarif_result.zip](https://github.com/spotbugs/spotbugs/files/5974036/sarif_result.zip)

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Result.java
```diff
@@ -20,7 +20,7 @@
     final Level level;
 
     Result(@NonNull String ruleId, int ruleIndex, Message message, List<Location> locations, @NonNull Level level) {
-        this.ruleId = Objects.requireNonNull(ruleId);
+        this.ruleId = String.format("SPOTBUGS_%d", Objects.requireNonNull(ruleId).hashCode());
```
KengoTODA, line 23: Not sure the `rule id` can contain underscore or not 🤔 .
The [official doc](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#def_rule_id) says nothing about acceptable characters.

KengoTODA: Note for test: hashcode could be negative.

KengoTODA (changes requested):
Please consider removing the .vs directory. I'll test and confirm that generated id is valid and well-formatted.

yongyan-gh:
> Please consider removing the .vs directory. I'll test and confirm that generated id is valid and well-formatted.

Removed, checked-in by mistake.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Result.java
yongyan-gh: The rule id can have underscore, e.g. "EC_UNRELATED_TYPES" is the rule id without the change in this PR.
The negative hash code is valid hash code, if change it e.g. math.abs may increase the possibility of a collision. 
cc @michaelcfanning @eddynaka what do you think about the opaque rule id?

michaelcfanning: These ids need work. The prefix is too long. We shouldn't have underscores or hyphens in the name. Ideally, an opaque id starts with a well-known initial value, 1001 is typical, and then monotonically increases as each rule is added. So, something like this SB1001, SB1002, SB1003, would work best.

A useful convention is to retain rule ids of 0001 - 0999 for internal tool/engine errors.

I can appreciate that my idea requires a more intrusive change, we may need to consider creating a formal place for the rule id in the object model and to populate SARIF from this.

yongyan-gh: In this case we probably need to assign the fixed Ids in meta data in SpotBugs and use these Ids as rule id, and the id always points to same rule.
E.g. in findbugs.xml
<BugPattern **ruleId="SB1001"** abbrev="SBSC" type="SBSC_USE_STRINGBUFFER_CONCATENATION" category="PERFORMANCE"/>
<BugPattern **ruleId="SB1002"**  abbrev="IIL" type="IIL_ELEMENTS_GET_LENGTH_IN_LOOP" category="PERFORMANCE" experimental="true"/>
...
Let me know what do you think?

yongyan-gh:
close this PR since hash code id is not a good solution. Will work on the fix of another approach.

PR #1434, opened 2021-02-18
Add opaque rule id for Sarif log

yongyan-gh:
Related to issue #1427

* Assigned opaque rule Id values to BugPattern metadata 
* Changed Sarif Result.ruleId/Rule.id value to opaque rule id defined in metadata.
* Added name property to Sarif Rule, value is original rule "id" (BugPattern.type)

Attached Sarif log generated before/after this change. Please review.
[newSarifLog.zip](https://github.com/spotbugs/spotbugs/files/5999555/newSarifLog.zip)


spotbugs/src/main/java/edu/umd/cs/findbugs/BugPattern.java
```diff
@@ -93,13 +97,20 @@ public BugPattern(String type, String abbrev, String category, boolean experimen
...
             "Unknown warning; core bug patterns not found", "Unknown warning BUG_PATTERN in {1}; core bug patterns not found",
             "<p>A warning was recorded, but SpotBugs can't find the description of this bug pattern "
                     + "and so can't describe it. This should occur only in cases of a bug in SpotBugs or its configuration, "
                     + "or perhaps if an analysis was generated using a plugin, but that plugin is not currently loaded. "
                     + "</p>", null, 0);
 
+    /**
+     * Get the Rule Id
+     */
+    public String getRuleId() {
```
KengoTODA, line 110: If we allow plugins to not providing the `ruleId` attribute, annotate this method with [`@CheckForNull`](https://spotbugs.readthedocs.io/en/stable/annotations.html#edu-umd-cs-findbugs-annotations-checkfornull).

spotbugs/src/main/java/edu/umd/cs/findbugs/BugPattern.java
```diff
@@ -79,9 +82,10 @@
      * @param bugsUrl
      *            URL of web-page containing bug descriptions or null if there's no such page.
      */
-    public BugPattern(String type, String abbrev, String category, boolean experimental, String shortDescription,
+    public BugPattern(String ruleId, String type, String abbrev, String category, boolean experimental, String shortDescription,
             String longDescription, String detailText, String bugsUrl, int cweid) {
 
+        this.ruleId = ruleId;
```
KengoTODA, line 88: If we do not allow plugins to provide no `ruleId` attribute, it's better to check null-ness of given value and throw a `NullPointerException` with a user-friendly message. Also it's better to annotate the parameter with [`@NonNull`](https://spotbugs.readthedocs.io/en/stable/annotations.html#edu-umd-cs-findbugs-annotations-nonnull) (even though other existing parameters aren't annotated so :<).

KengoTODA (commented):
I understand the motivation to set `ruleId` explicitly in the XML, however, we need to consider how to support plugins.

Do we need to enforce them to add `ruleId` attribute? If so, we need to ask each plugin owner to update.

KengoTODA:
The broken build has been fixed in the `master` branch, please merge the latest `master` to your topic branch or simply rebase on it.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugPattern.java
yongyan-gh: @KengoTODA thank you for pointing out the plugin mechanism, you are right we should keep it backward compatible. 
in case plugin doesn't pass rule id we can use the BugPattern type as rule id, which is currently used as rule id. and provide new API interface so plugin owners can add the opaque rule id in future.

@michaelcfanning @eddynaka what do you think?

KengoTODA: Thanks.

If we need to follow the [definition of 'stable value'](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#def_stable_value) and cannot make this attribute optional, it is also possible to release a major version to break compatibility. Of course we need to help plugin authors to migrate, though.

michaelcfanning: Ideally, we would break plug-in authors on update so that the opportunity to provide a valid opaque id is very obvious, but also provide a very easy way to fall back to the legacy rule id if they can't stop to make a fix. @yongyan-gh, please go read the SARIF spec on the `deprecatedIds` property for SARIF `reportingDescriptor` objects. If we are emitting the `rules` table to Spotbugs SARIF, this data can be populated. If we do not emit this data, obviously, we can't persist the deprecated ids. Doing this is not a high priority, as I doubt there are SARIF readers out there who consume this information, though. :) People worried about compatibility will just consult the readable name, which maps to the old rule id value.

yongyan-gh: @KengoTODA in latest change we added an overload constructor to wrap the new constructor and marked with deprecated annotation. 
In this way it wont break existing plugins. Plugins call deprecated constructor will get a warning to use new constructor and provide opaque rule id. 
Please let me know what do you think?

spotbugs/src/main/java/edu/umd/cs/findbugs/BugPattern.java
```diff
@@ -36,6 +37,8 @@
  * @see BugInstance
  */
 public class BugPattern implements Comparable<BugPattern> {
+    final private String ruleId;
```
KengoTODA, line 40: Consider to add a documentation comment, especially a `@see` tag with an anchor tag to http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html#def_rule_id

spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
```diff
@@ -928,6 +928,7 @@ private void loadPluginComponents()
         // Create BugPatterns
         List<Node> bugPatternNodeList = XMLUtil.selectNodes(pluginDescriptor, "/FindbugsPlugin/BugPattern");
         for (Node bugPatternNode : bugPatternNodeList) {
+            String ruleId = bugPatternNode.valueOf("@ruleId");
```
KengoTODA, line 931: This `ruleId` can be an empty string, right? Then I think here we need a consideration for (old) plugins.

Consider to test your implementation with a plugin. You can create distributed files by `./gradlew assemble`. Unpack the generated .tar or .zip and [run it in your local](https://spotbugs.readthedocs.io/en/stable/running.html#quick-start).

I suggest you to use find-sec-bugs to run with, but if you need a small project for faster build, you may try [findbugs-slf4j](https://github.com/kengoTODA/findbugs-slf4j) instead. It's also possible to [create a plugin project by your own](https://github.com/spotbugs/spotbugs-archetype/tree/a3e126806686bc7f280dc1e294fbf98d21903468#how-to-use).

KengoTODA (changes requested):

KengoTODA:
@yongyan-gh sorry but please resolve conflicts later.

yongyan-gh:
@KengoTODA no problem, merged changes from remote and resolved the conflicts.

KengoTODA:
Sorry but again we have a conflict, please merge the latest `origin/master` to your topic branch and resolve it.
You also need to update `CHANGELOG.md` to move your changes under `## Unreleased`. 🙇

yongyan-gh (commented):
> Sorry but again we have a conflict, please merge the latest `origin/master` to your topic branch and resolve it.
> You also need to update `CHANGELOG.md` to move your changes under `## Unreleased`. 🙇

Resolved & merged, updated the change log.

1 thing we can also improve for the opaque rule id: we may divide the ids into couple segments based on bug pattern/rule category etc. so that each segment has spare ids can be assigned to new bug patterns/rules in the future, and don't need to change existing id.

PR #1136, opened 2020-05-27
False positive SA_LOCAL_SELF_COMPARISON when using instanceof pattern matching

schosin:
This PR adds a failing test when using instanceof pattern matching in Java 14 (preview feature).

In this code, SA_LOCAL_SELF_COMPARISON is incorrectly reported for `number`:
```
Number number = 1f;
if (number instanceof Float f) {
    System.out.println("number is a float: " + f);
}
```

Edit: According to [JEP 375](https://openjdk.java.net/jeps/375) this feature will be going into a second preview with no changes relative to the first preview introduced in JDK 15. This PR should be changed to use JDK 15 as soon as it is released and gradle/other dependencies support JDK 15.

Issue #1134, opened 2020-05-23
Incorrect ModifiedDate (2/1/1980) for files in zip and files in JARs


blindspeed90:
Building locally matched the pre-built spotbugs-4.0.3.zip, where all files within it are stamped with 2/1/1980 for the ModifiedDate.  Additionally, JARs within the zip such as asm*.jar and spotbugs*.jar have files within them stamped with 2/1/1980 also.

KengoTODA:
<img width="882" alt="Screen Shot 2020-05-27 at 10 48 11" src="https://user-images.githubusercontent.com/505751/82972415-95671f00-a007-11ea-8cc1-839af1f43d13.png">

KengoTODA:
4.0.4 still has this problem.

KengoTODA:
I found that this is intentional one. We set [preserveFileTimestamps](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/bundling/AbstractArchiveTask.html#setPreserveFileTimestamps-boolean-) property, to [make the build reproducible](https://reproducible-builds.org/).

https://github.com/spotbugs/spotbugs/blob/815b0f743e2601d01fc9f47cc32ff40960c73517/gradle/java.gradle#L12

PR #1435, opened 2021-02-18
False positive SA_LOCAL_SELF_COMPARISON when using instanceof pattern matching

KengoTODA:
Recreated from #1136 to build with the latest change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,14 +9,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@5a4ac90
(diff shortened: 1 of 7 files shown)


KengoTODA:
Hi @spotbugs/everyone , please review this PR when you have time. Thanks in advance!

iloveeclipse:
The PR references some wrong issue #1134? 
Also it seem not to have a fix, but only a test case, or it doesn't explicitly explain the problem & solution?

KengoTODA:
Good catch @iloveeclipse, it should be #1136.

ThrawnCA (approved):

Issue #1279, opened 2020-08-24
Please consider `org.json:json` alternative


DPUkyle:
Hi @KengoTODA I see that spotbugs/discuss#95 was merged and released with tool v4.1.0 - that's great!

I have one question about the [license](https://github.com/stleary/JSON-java/blob/20200518/LICENSE) of the `org.json:json` library - it's unconventional and is actually banned by ASF: http://apache.org/legal/resolved.html#category-x -> see "Nonsensical licenses".

Debian lists some alternatives but I'm sure this list is not very current, nor comprehensive: https://wiki.debian.org/qa.debian.org/jsonevil

My organization specifically objects to the license's use of the language "The Software shall be used for Good, not Evil.", as such I'd encourage you to please consider alternative implementations to produce the SARIF report. As a consequence, we will be unable to use spotbugs v4.1.0 or higher ☹️ 

FYI @lgolding @michaelcfanning

KengoTODA:
Working on this issue. Gson seems smaller than Jackson, so I'm trying to play with Gson.

The Gradle plugin isn't working well to generate both annotation and additionalProperties, so reported the issue as https://github.com/eirnym/js2p-gradle/issues/10
If problem will not be resolved, I'll try to code without annotation.

sabberworm:
> My organization specifically objects to the license's use of the language "The Software shall be used for Good, not Evil."

@DPUkyle I think you should quit working for this company ASAP…

h-vetinari:
Any update on this? This is blocking (much of) the hadoop ecosystem from moving to Java 11+, see https://issues.apache.org/jira/browse/HADOOP-17269

michaelcfanning:
@eddynaka @yongyan-gh

 @KengoTODA, want any help with this problem? We're just about to propose a new SARIF contribution that provides for SARIF-friendly rule ids. Maybe we could finish whatever work you've started with gson as well?

Glad to help if you want it.

KengoTODA:
The current blocker for me is [json2pojo does not generate fields of `MessageStrings` as expected](https://github.com/microsoft/sarif-sdk/issues/1952#issuecomment-687668869). 

For the current implementation, I manually coded `MessageStrings` and other classes. But it's better to generate it based on JSON schema. This is why I want to use json2pojo via https://github.com/eirnym/js2p-gradle

michaelcfanning:
ok! we'll take a look and see if we can get the SARIF schema through json2pojo today. More soon...

[...]

PR #1437, opened 2021-02-19
Replace 'org.json:json' lib with 'com.google.code.gson' lib

yongyan-gh:
This is for issue #1279
* Some of Gson classes/methods doesn't support fluent syntax now, has to refactor a little bit.
* org.json:json has append() / putOpt() can handle empty/null object/array, which need to manually handle in Gson.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Extension.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Extension.java
@@ -4,6 +4,7 @@ import edu.umd.cs.findbugs.Plugin;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import com.google.gson.JsonObject;
+import org.apache.commons.lang3.StringUtils;
 
 import java.net.URI;
 import java.util.Objects;
@@ -20,7 +21,7 @@ class Extension {
     @Nullable
     final URI informationUri;
     @Nullable
-    final Object organization;
+    final String organization;
 
     Extension(@NonNull String version, @NonNull String name, @Nullable String shortDescription, @Nullable String fullDescription,
             @Nullable URI informationUri, @Nullable String organization) {
@@ -48,8 +49,8 @@ class Extension {
         if (informationUri != null) {
             extensionJson.addProperty("informationUri", informationUri.toString());
         }
-        if (organization != null) {
-            extensionJson.addProperty("organization", String.valueOf(organization));
+        if (!StringUtils.isEmpty(organization)) {
+            extensionJson.addProperty("organization", organization);
         }
         return extensionJson;
     }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Extension.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Extension.java
@@ -3,7 +3,7 @@ package edu.umd.cs.findbugs.sarif;
 import edu.umd.cs.findbugs.Plugin;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import org.json.JSONObject;
+import com.google.gson.JsonObject;
 
 import java.net.URI;
 import java.util.Objects;
@@ -32,16 +32,26 @@ class Extension {
         this.organization = organization;
     }
 
-    JSONObject toJSONObject() {
+    JsonObject toJsonObject() {
         // TODO put 'fullDescription' with both of text and markdown representations
-        JSONObject desc = null;
+        JsonObject desc = null;
         if (shortDescription != null) {
-            desc = new JSONObject().put("text", shortDescription);
+            desc = new JsonObject();
+            desc.addProperty("text", shortDescription);
         }
-        return new JSONObject().put("version", version).put("name", name)
-                .putOpt("shortDescription", desc)
-                .putOpt("informationUri", informationUri)
-                .putOpt("organization", organization);
+        JsonObject extensionJson = new JsonObject();
+        extensionJson.addProperty("version", version);
+        extensionJson.addProperty("name", name);
+        if (desc != null) {
+            extensionJson.add("shortDescription", desc);
+        }
+        if (informationUri != null) {
+            extensionJson.addProperty("informationUri", informationUri.toString());
+        }
+        if (organization != null) {
+            extensionJson.addProperty("organization", String.valueOf(organization));
+        }
+        return extensionJson;
     }
 
     static Extension fromPlugin(@NonNull Plugin plugin) {
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Level.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Level.java
@@ -2,14 +2,13 @@ package edu.umd.cs.findbugs.sarif;
 
 import edu.umd.cs.findbugs.BugRankCategory;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import org.json.JSONString;
 
 /**
  * An enum representing {@code level} property.
  *
  * @see <a href="https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Toc16012604">3.27.10 level property</a>
  */
-enum Level implements JSONString {
+enum Level {
     /**
      * The rule specified by ruleId was evaluated and a problem was found.
      */
@@ -27,10 +26,7 @@ enum Level implements JSONString {
      */
     NONE;
 
-    @Override
-    public String toJSONString() {
-        return String.format("\"%s\"", name().toLowerCase());
-    }
+    public String toJsonString() { return name().toLowerCase(); }
 
     @NonNull
     static Level fromBugRank(int bugRank) {
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Level.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Level.java
@@ -26,7 +26,9 @@ enum Level {
      */
     NONE;
 
-    public String toJsonString() { return name().toLowerCase(); }
+    public String toJsonString() {
+        return name().toLowerCase();
+    }
 
     @NonNull
     static Level fromBugRank(int bugRank) {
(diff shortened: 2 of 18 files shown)


yongyan-gh:
Attached the Sarif log generated with this change (using com.google.code.gson) 
also Sarif log generated using current SpotBugs (org.json:json) for a comparison reference.
[GsonSarifLog.zip](https://github.com/spotbugs/spotbugs/files/6009688/GsonSarifLog.zip)

@KengoTODA  pls review.
cc @michaelcfanning @eddynaka

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Extension.java
```diff
@@ -32,16 +32,26 @@
...
+        extensionJson.addProperty("version", version);
+        extensionJson.addProperty("name", name);
+        if (desc != null) {
+            extensionJson.add("shortDescription", desc);
+        }
+        if (informationUri != null) {
+            extensionJson.addProperty("informationUri", informationUri.toString());
+        }
+        if (organization != null) {
+            extensionJson.addProperty("organization", String.valueOf(organization));
```
KengoTODA, line 52: Is are any reason to prefer `String.valueOf( ... )` over `organization.toString()`? If so, I want to make the intention clear.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Level.java
```diff
@@ -27,9 +26,8 @@
      */
     NONE;
 
-    @Override
-    public String toJSONString() {
-        return String.format("\"%s\"", name().toLowerCase());
+    public String toJsonString() {
+        return name().toLowerCase();
```
KengoTODA, line 30: Do we need this method? The old one was provided to configure the org.json's serializer.
So I assume what we need is not this method but similar way to configure the serialier, like [SerializedName](https://stackoverflow.com/a/24621292).

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -17,29 +19,42 @@ public SarifBugReporter(Project project) {
...
-            jsonWriter.key("version").value("2.1.0").key("$schema").value(
+            JsonWriter jsonWriter = new JsonWriter(outputStream);
+            jsonWriter.beginObject();
+            jsonWriter.name("version").value("2.1.0").name("$schema").value(
                     "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json");
             processRuns(jsonWriter);
             jsonWriter.endObject();
             getBugCollection().bugsPopulated();
+        } catch (IOException e) {
+            e.printStackTrace(System.err);
```
KengoTODA, line 30: I assume it's better to throw [UncheckedIOException](https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html) instead?
At least, we cannot ignore this exception like this, because we surely have problem in generated SARIF report file.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -17,29 +19,42 @@ public SarifBugReporter(Project project) {
...
+            analyser.getResults().forEach((result) -> gson.toJson(result, jsonWriter));
+            jsonWriter.endArray();
+
+            jsonWriter.name("originalUriBaseIds");
+            gson.toJson(analyser.getOriginalUriBaseIds(), jsonWriter);
+
+            jsonWriter.endObject().endArray(); // end "runs", end "runs" array
+        } catch (IOException e) {
+            throw e;
+        }
```
KengoTODA, line 54: This try-block seems needless.

```suggestion
        Gson gson = new Gson();
        jsonWriter.name("runs").beginArray().beginObject();
        BugCollectionAnalyser analyser = new BugCollectionAnalyser(getBugCollection());
        processTool(jsonWriter, analyser.getRules());
        processInvocations(jsonWriter, analyser.getBaseToId());

        jsonWriter.name("results").beginArray();
        analyser.getResults().forEach((result) -> gson.toJson(result, jsonWriter));
        jsonWriter.endArray();

        jsonWriter.name("originalUriBaseIds");
        gson.toJson(analyser.getOriginalUriBaseIds(), jsonWriter);

        jsonWriter.endObject().endArray(); // end "runs", end "runs" array
```

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -59,27 +74,49 @@ private void processInvocations(JSONWriter jsonWriter, @NonNull Map<URI, String>
...
-        jsonWriter.key("invocations").array().value(invocation.toJSONObject());
-        jsonWriter.endArray();
+        try {
+            Gson gson = new Gson();
+            jsonWriter.name("invocations").beginArray();
+            gson.toJson(invocation.toJsonObject(), jsonWriter);
+            jsonWriter.endArray();
+        } catch (IOException e) {
+            throw e;
+        }
```
KengoTODA, line 84: This try-block seems needless.
```suggestion
        Gson gson = new Gson();
        jsonWriter.name("invocations").beginArray();
        gson.toJson(invocation.toJsonObject(), jsonWriter);
        jsonWriter.endArray();
```

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -59,27 +74,49 @@ private void processInvocations(JSONWriter jsonWriter, @NonNull Map<URI, String>
...
-        jsonWriter.key("driver").object();
-        jsonWriter.key("name").value("SpotBugs");
-        // Eclipse plugin does not follow the semantic-versioning, so use "version" instead of "semanticVersion".
-        jsonWriter.key("version").value(Version.VERSION_STRING);
-        // SpotBugs refers JVM config to decide which language we use.
-        jsonWriter.key("language").value(Locale.getDefault().getLanguage());
-        jsonWriter.key("rules").value(rules);
-        jsonWriter.endObject().endObject();
+    private void processTool(@NonNull JsonWriter jsonWriter, @NonNull JsonArray rules) throws IOException {
+        try {
```
KengoTODA, line 88: This try-block is also needless. Simply throw `IOException` to the caller side.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -59,27 +74,49 @@ private void processInvocations(JSONWriter jsonWriter, @NonNull Map<URI, String>
...
+            Gson gson = new Gson();
```
KengoTODA, line 89: Is it OK to generate `Gson` instance on demand? Not sure its cost, but if it takes much resources (CPU or Java heap), consider to create it once and reuse by providing it via method parameter.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -59,27 +74,49 @@ private void processInvocations(JSONWriter jsonWriter, @NonNull Map<URI, String>
...
+            // Eclipse plugin does not follow the semantic-versioning, so use "version" instead of "semanticVersion".
+            jsonWriter.name("version").value(Version.VERSION_STRING);
+            // SpotBugs refers JVM config to decide which language we use.
+            jsonWriter.name("language").value(Locale.getDefault().getLanguage());
+
+            jsonWriter.name("rules").beginArray();
+            rules.forEach((rule) -> gson.toJson(rule, jsonWriter));
+            jsonWriter.endArray();
+
+            jsonWriter.endObject().endObject(); // end "driver", end "tool"
```
KengoTODA, line 104: Good comment to keep the code readable! 👍

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
```diff
@@ -59,27 +74,49 @@ private void processInvocations(JSONWriter jsonWriter, @NonNull Map<URI, String>
...
+            throw e;
+        }
     }
 
-    private void processExtensions(@NonNull JSONWriter jsonWriter) {
-        jsonWriter.key("extensions").array();
-        DetectorFactoryCollection.instance().plugins().stream().map(Extension::fromPlugin).map(Extension::toJSONObject).forEach(jsonWriter::value);
-        jsonWriter.endArray();
+    private void processExtensions(@NonNull JsonWriter jsonWriter) throws IOException {
+        try {
```
KengoTODA, line 111: this try-block seems needless.

KengoTODA (changes requested):
Thanks for your contribution! Please check my comments for maintainability.

Also, please put https://github.com/google/gson/blob/gson-parent-2.8.6/LICENSE into `spotbugs/licenses` directory to distribute.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
KengoTODA: Here we do not need to invoke `e.printStackTrace(System.err);`. Who catches the error will print it.

And when we need to log, we use SLF4J as a logging facade so please do not use `System.err` directly.

KengoTODA (changes requested):
Cool. Just one change is necessary, please check.

KengoTODA (approved):
LGTM. Thanks for your contribution!

KengoTODA:
@spotbugs/everyone if possible, please double-check this PR. I'm recently little bit careless to make a high quality review. 🙇‍♂️

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/SarifBugReporter.java
yongyan-gh: Removed printStackTrace to let caller to log the error.

yongyan-gh (commented):
> LGTM. Thanks for your contribution!

Thank you for reviewing!

yongyan-gh:
Attached the Sarif log generated by SpotBugs with the Gson changes. 
Except some name/value pairs are in different order, its identical to original Sarif log file generated using org.json
[gsonResult.zip](https://github.com/spotbugs/spotbugs/files/6026408/gsonResult.zip)

yongyan-gh:
@KengoTODA
I ran all tests all passed and adhoc tests it works as expected. Can you pls merge this change if no more issue?
Also pls review #1434.

Thanks!

michaelcfanning:
Fantastic!

Issue #1464, opened 2021-03-12
Should issue warning for SecureRandom object  created and used only once


Mary904:
I found a false negative when spotbugs-maven-plugin checked [the following line in eclipse/jetty.project](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-keystore/src/main/java/org/eclipse/jetty/keystore/KeystoreGenerator.java#L59).

```java
BigInteger serial = BigInteger.valueOf(new SecureRandom().nextLong());
```

According to the description of [DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once):

> You should strongly consider using a `java.security.SecureRandom` instead (and **avoid allocating a new SecureRandom for each random number needed**).

Spotbugs should generate a warning because the code line `BigInteger serial = BigInteger.valueOf(new SecureRandom().nextLong());` created and used SecureRandom object only once. As noted in the bug description, it should **"avoid allocating a new SecureRandom for each random number"**. However, in the demo, this is allowed by Spotbugs without giving any warning so I think that this is a false negative.

The plugin version used is as follow:

```xml
            <plugin>
                <groupId>com.github.spotbugs</groupId>
                <artifactId>spotbugs-maven-plugin</artifactId>
                <version>4.1.4</version>
                <configuration>
                    <effort>Default</effort>
                    <threshold>Low</threshold>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>com.github.spotbugs</groupId>
                        <artifactId>spotbugs</artifactId>
                        <version>4.1.4</version>
                    </dependency>
                </dependencies>
            </plugin>
```

### Demo

Here is a small demo that could be used as test case. See https://github.com/spotbugs/spotbugs/pull/1475

```java
import java.security.SecureRandom;
import java.util.Random;
import edu.umd.cs.findbugs.annotations.ExpectWarning;

public class DMI_RANDOM_USED_ONLY_ONCE {
    @ExpectWarning("DMI_RANDOM_USED_ONLY_ONCE")
    long m1(){
        return new SecureRandom().nextLong();  // This test case will fail now
    }

    @ExpectWarning("DMI_RANDOM_USED_ONLY_ONCE")
    long m2(){
        return new Random().nextLong();  // This will pass
    }
}
```

[...]

PR #1475, opened 2021-03-25
Should detect SecureRandom Object used only once

Mary904:
reproduce #1464


ThrawnCA:
Indentation uses 3 spaces instead of 4.

ThrawnCA (changes requested):
Please update to pass Spotless checks.

PR #1476, opened 2021-03-25
Fix #1464, and issue a warning for a SecureRandom new created and used only once

### LINKED ISSUE #1464: Should issue warning for SecureRandom object  created and used only once

gloNelson:
Hi,
Fixes #1464 
Merge and upload test-cases present in #1475.
Please help me review @KengoTODA, thanks :)

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
@@ -490,17 +490,25 @@ public class DumbMethods extends OpcodeStackDetector {
 
         @Override
         public void sawOpcode(int seen) {
-            if (seen == Const.INVOKEVIRTUAL && ("java/util/Random".equals(getClassConstantOperand()) || "java/security/SecureRandom".equals(
-                    getClassConstantOperand()))
-                    && (freshRandomOnTos || freshRandomOneBelowTos)) {
-                accumulator.accumulateBug(new BugInstance(DumbMethods.this, "DMI_RANDOM_USED_ONLY_ONCE", HIGH_PRIORITY)
-                        .addClassAndMethod(DumbMethods.this).addCalledMethod(DumbMethods.this), DumbMethods.this);
+            if (seen == Const.INVOKEVIRTUAL) {
+                String classConstantOperand = getClassConstantOperand();
+                if ("java/util/Random".equals(classConstantOperand) || "java/security/SecureRandom".equals(
+                        classConstantOperand)
+                        && (freshRandomOnTos || freshRandomOneBelowTos)) {
+                    accumulator.accumulateBug(new BugInstance(DumbMethods.this, "DMI_RANDOM_USED_ONLY_ONCE", HIGH_PRIORITY)
+                            .addClassAndMethod(DumbMethods.this).addCalledMethod(DumbMethods.this), DumbMethods.this);
 
+                }
             }
-            freshRandomOneBelowTos = freshRandomOnTos && isRegisterLoad();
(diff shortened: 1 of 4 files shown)


KengoTODA (approved):
LGTM, I'll wait for other members' review to merge this.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
```diff
@@ -490,14 +490,16 @@ public void initMethod(Method method) {
 
         @Override
         public void sawOpcode(int seen) {
-            if (seen == Const.INVOKEVIRTUAL && "java/util/Random".equals(getClassConstantOperand())
+            if (seen == Const.INVOKEVIRTUAL && ("java/util/Random".equals(getClassConstantOperand()) || "java/security/SecureRandom".equals(
```
ThrawnCA, line 493: Perhaps `getClassConstantOperand()` should be extracted to a local variable.

ThrawnCA (approved):

ThrawnCA:
I've approved, but with a suggestion for possible improvement.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
gloNelson: Done. Thanks for your good suggestion :)

ThrawnCA (approved):

PR #1474, opened 2021-03-24
Fix #1472 SA_LOCAL_SELF_COMPUTATION bug

gloNelson:
Hi,
Fixes #1472
I modified the problematic code line in `spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java`.
Test-cases mentioned in #1472 are added in `spotbugsTestCases/src/java/ghIssues/Issue1472.java` and `spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1472Test.java`.  

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #1478, opened 2021-03-28
Fix #1472 SA_LOCAL_SELF_COMPUTATION bug

### LINKED ISSUE #1472: SA_LOCAL_SELF_COMPUTATION bug
Hi,
I read [SA_LOCAL_SELF_COMPUTATION](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#sa-nonsensical-self-computation-involving-a-variable-e-g-x-x-sa-local-self-computation) on SpotBugs Bug Description page, which mentions that this method performs a nonsensical computation of a local variable with another reference to the same variable (e.g., `x&x` or `x-x`).
So, I wrote test cases as below.
```java
int testSA_LOCAL_SELF_COMPUTATION(){
      int flags = 1;
      return flags ^ flags; // warning SA_LOCAL_SELF_COMPUTATION
  }

  int testSA_LOCAL_SELF_COMPUTATION1(){
      int flags = 1;
      return flags & flags; // no warning (False negative)
  }

  int testSA_LOCAL_SELF_COMPUTATION2(){
      int flags = 1;
      return flags - flags; // no warning (False negative)
  }

  int testSA_LOCAL_SELF_COMPUTATION3(){
      int flags = 1;
      return flags | flags; // no warning (False negative)
  }

```

It can be seen from the results that only `^` will trigger SpotBugs to report **SA_LOCAL_SELF_COMPUTATION**, while other operators will not. 

I tried to search test cases of **SA_LOCAL_SELF_COMPUTATION** in SpotBugs source code and found several test cases of  **SA_FIELD_SELF_COMPUTATION** which detect self computation of **field** instead of **local variables**.
Similar test-cases for **SA_FIELD_SELF_COMPUTATION** executed and all four output **SA_FIELD_SELF_COMPUTATION** errors report.
```java
int x = 1;
  int testSA_FIELD_SELF_COMPUTATION(){
      return x ^ x; // warning SA_FIELD_SELF_COMPUTATION
  }

  int testSA_FIELD_SELF_COMPUTATION1(){
      return x & x; // warning SA_FIELD_SELF_COMPUTATION
  }

  int testSA_FIELD_SELF_COMPUTATION2(){
      return x - x; // warning SA_FIELD_SELF_COMPUTATION
  }

  int testSA_FIELD_SELF_COMPUTATION3(){
      return x | x; // warning SA_FIELD_SELF_COMPUTATION
  }
```

Finally, I realized that the fault is in line **325** of [FindSelfComparison.java](https://github.com/spotbugs/spotbugs/blob/a6f9acb2932b54f5b70ea8bc206afb552321a222/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java#L325).
```java
else if (opCode == Const.IXOR && item0.sameValue(item1)) {
```
The code shows that SpotBugs reports **SA_FIELD_SELF_COMPUTATION** only when `opCode` equals `Const.IXOR`(`^`).

It is so confusing that self computation detection of `local variables` only including `^` operator while the detection of `field` supports much more. Is it a bug?

gloNelson:
Sorry, I accidentally deleted the original clone repo in #1474 . So I have to create a new pr.
I have resolved the conflict, please help me review this :) @KengoTODA 

----


KengoTODA (approved):
I'll wait for a review from another teammate, to merge this PR.

ThrawnCA (approved):

PR #1529, opened 2021-04-26
Refactor and improve `FindReturnRef` to find indirect cases

baloghadamsoftware:
Currently, the detector `FindReturnRef` which is responisble for bugs
`EI_EXPOSE_REP`, `EI_EXPOSE_REP2`, `MS_EXPOSE_REP` and
`EI_EXPOSE_STATIC_REP2` tracks the data on the top of the stack.
However, this is not necessary in the current version of SpotBugs
because the infrastructure already does this job more accurately. This
patch removes the redundant tracking the the top of the stack and
relies on the already existing infrastructure instead. This way the
detector is able to find more true positives where returning of the
private mutable attribute happens indirectly, e.g. by first assigning it
to a local variable and then returning that local.

----


baloghadamsoftware:
This pull request depends on #1525 which is already accepted by @ThrawnCA but not merged yet.

baloghadamsoftware:
#1525 is merged now. Could someone (@KengoTODA  or @ThrawnCA), please start the review on this PR?

spotbugsTestCases/src/java/FindReturnRefNegativeTest.java
```diff
@@ -0,0 +1,207 @@
...
+        return dateCopy;
+    }
+
+    @NoWarning("EI")
+    public Hashtable<Integer, String> getValues() {
+        return (Hashtable<Integer, String>) ht.clone();
+    }
+
+    @NoWarning("MS")
+    public static Hashtable<Integer, String> getSaticValues() {
```
ThrawnCA, line 101: Mispelled "static"

spotbugsTestCases/src/java/FindReturnRefNegativeTest.java
```diff
@@ -0,0 +1,207 @@
+import edu.umd.cs.findbugs.annotations.NoWarning;
+
+import java.util.Date;
+import java.util.Hashtable;
```
ThrawnCA, line 4: Why not `Map`?

ThrawnCA (changes requested):

spotbugsTestCases/src/java/FindReturnRefNegativeTest.java
baloghadamsoftware: I copied the original test from the SEI CERT rule, but now I changed it to HashMap.

ThrawnCA (dismissed):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -152,4 +154,19 @@ private boolean isPotentialCapture(OpcodeStack.Item top) {
         return top.getRegisterNumber() != parameterCount - 1;
 
     }
+
+    private boolean isFieldOf(XField field, ClassDescriptor klass) {
```
ThrawnCA, line 158: The usual convention is "clazz", but okay.

ThrawnCA (approved):

KengoTODA:
It seems that this branch contains the change from #1542 that is already merged to master.
Not sure why it happens, but I suggest you to recreate branch and PR to make the change focused on your intention.

Issue #1538, opened 2021-05-05
OpcodeStack.Item.sameValue() has redundant condition


baloghadamsoftware:
The function sameValue() of nested class OpcodeStack.Item has the following condition:

`this.equals(that) && (this.registerNumber != -1 && this.registerNumber == that.registerNumber || this.fieldLoadedFromRegister != -1)`.

As we can see, the first member of this expression is a call to `equals()` and everything is and-ed to it. However, the condition of `equals()` also compares fields `registerNumber` and `fieldLoadedFromRegister` and both of them must be equal:

`...
&& this.registerNumber == that.registerNumber && this.flags == that.flags
&& this.fieldLoadedFromRegister == that.fieldLoadedFromRegister`

Thus `sameValue()` does exactly the same as `equals()` just it spends a bit of extra time if the two items are equal. I assume that this is not intentional. What is exactly the purpose of `sameValue()`? I would assume that it should check whether two stack items are the same, independently whether one of them is stored in a register. Am I right?

baloghadamsoftware:
@KengoTODA?

JuditKnoll:
This change was added in https://github.com/spotbugs/spotbugs/commit/e6bd133948630f037e8f8f0b94f969d471fe8c3e . From that it looks like to me, that it's intentional, and the goal was to handle the edge case of -1 register value meaning unknown source.

PR #1541, opened 2021-05-07
Add new rule REFL_REFLECTION_INCREASES_ACCESSIBILITY_OF_CLASS

baloghadamsoftware:
SEI CERT rule SEC05-J forbids the use of reflection to increase
accessibility of classes, methods or fields. If a class in a package
provides a public method which takes an instance of java.lang.Class as
its parameter and calls its newInstance() method then it increases
accessibility of classes in the same package without public constructors.
An attacker code may call this method and pass such class to create an
instance of it. This should be avoided by either making the method
non-public or by invoking SecurityManager.checkPackageAccess on the
package. A third possiblity is to use the java.beans.Beans.instantiate()
method instead of java.lang.Class.newInstance() which checks whether the
Class object being received has any public constructors.

This patch adds a new detector to check this rule.

Change-Id: I1881502512848f8a04b7ee6be95539f5fe88c6b3

----


ThrawnCA (changes requested):

baloghadamsoftware:
Sorry for force pushing but there was a git mess which I had to sort out.

baloghadamsoftware:
We tested this on some open-source projects and found some true positives. (I mean that they are violations of the SEI CERT rule.)

[MATSim](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)
[NanoHTTPD](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20NanoHTTPD%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20NanoHTTPD%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)

On the rest we did not found anything. Thus no false positives at all and no crashes:
[SpotBugs itself](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20SpotBugs%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20SpotBugs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)
[Jenkins](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20Jenkins%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20Jenkins%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)
[Bt](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20Bt%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20Bt%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)
[Ttorrent](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?run=SpotBugs%20on%20Ttorrent%20%28user%3A%20spotbugs,%20branch%3A%20master%29&newcheck=SpotBugs%20on%20Ttorrent%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Reflection%29)

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ReflectionIncreaseAccessibility.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * FindBugs - Find bugs in Java programs
```
ThrawnCA, line 2: New files should use a SpotBugs copyright notice, not FindBugs.

ThrawnCA (changes requested):

ThrawnCA (dismissed):

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Bump Saxon-HE from 10.3 to 10.5 ([#1513](https://github.com/spotbugs/spotbugs/pull/1513))
 * Function `mutableSignature()` improved and factored out from the `MutableStaticFields` detector
 
+### Added
+* New rule `REFL_REFLECTION_INCREASES_ACCESSIBILITY_OF_CLASS` to detect public methods instantiating a class they get in their parameter. This rule based on the SEI CERT rule *SEC05-J. Do not use reflection to increase accessibility of classes, methods, or fields*. ([#SEC05-J](https://wiki.sei.cmu.edu/confluence/display/java/SEC05-J.+Do+not+use+reflection+to+increase+accessibility+of+classes%2C+methods%2C+or+fields))
+
```
KengoTODA, line 15: Please move these line to the bottom of `## Unreleased`, then I'll merge this PR.

KengoTODA (changes requested):

CHANGELOG.md
KengoTODA: `spotbugs/etc/findbugs.xml` and `spotbugs/etc/messages.xml` also have conflicts

baloghadamsoftware:
@KengoTODA Rebasing done, also moved line in CHANGELOG.md to ###Unreleased.

KengoTODA (dismissed):

baloghadamsoftware:
@KengoTODA It seems that you approved this one. Are there any obstacles in merging it?

KengoTODA:
@baloghadamsoftware thanks for reminding me. Could you merge `origin/master` again and move your CHANGELOG entry to `## Unreleased`?

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ReflectionIncreaseAccessibility.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            JavaClass cls = obj.getJavaClass();
+            if (cls == null) {
+                return;
+            }
+            XMethod met = getXMethodOperand();
+            if (!securityCheck && obj.isInitialParameter() && obj.getXField() == null &&
+                    "java.lang.Class".equals(cls.getClassName()) &&
+                    "newInstance".equals(met.getName()) &&
+                    "()Ljava/lang/Object;".equals(met.getSignature())) {
```
KengoTODA, line 73: As [I commented previously](https://github.com/spotbugs/spotbugs/pull/1541#discussion_r735461270https://github.com/spotbugs/spotbugs/pull/1541#discussion_r735461270), it is necessary to check visibility of target class and current method. For instance, if both are public then there is no vulnerability.

KengoTODA (changes requested):

spotbugs/etc/messages.xml
```diff
@@ -8676,4 +8717,5 @@ after the call to initLogging, the logger configuration is lost
   <BugCode abbrev="DL">Unintended contention or possible deadlock due to locking on shared objects</BugCode>
   <BugCode abbrev="JUA">Problems in JUnit Assertions</BugCode>
   <BugCode abbrev="EOS">Bad End of Stream check</BugCode>
+  <BugCode abbrev="REFL">Reflection increasing accessibility of classes, methods and fields</BugCode>
```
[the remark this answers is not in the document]
baloghadamsoftware: Do you mean that I should use two different bug codes for the two kinds of bugs? I changed them now to REFLC and REFLF.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ReflectionIncreaseAccessibility.java
```diff
@@ -0,0 +1,110 @@
...
+            AnalysisContext.reportMissingClass(e);
+        }
+    }
+
+    @Override
+    public void afterOpcode(int seen) {
+        super.afterOpcode(seen);
+
+        if (fieldNameUnderGet) {
+            stack.getStackItem(0).setUserValue(true);
```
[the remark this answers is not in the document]
baloghadamsoftware: I did it now. However, originally I did not intend to use `OpcodeStack.Item#userValue` but to store `OpcodeStack.Item`s in a container and compare them to the actual `OpcodeStack.Item` later on. The problem is that `OpcodeStack.Item` contains the register number as well so the comparison fails even if we have the same object. There is a function `OpcodeStack.Item.sameValue()` which is probably to compare `OpcodeStack.Item`s without the register number but it does something else, see #1538.

spotbugs/etc/messages.xml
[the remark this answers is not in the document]
baloghadamsoftware: Do you mean that I should separate the bug codes? I did it, now we have REFLC and REFLF.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ReflectionIncreaseAccessibility.java
baloghadamsoftware: Sorry, I was probably not clear enough when I [tried to explain the purpose of the detector](https://github.com/spotbugs/spotbugs/pull/1541#issuecomment-950813189). We cannot check the visibility of the target class here because we do not have the attacker code. The issue is that our function (which is public and is defined in a public class) takes the class as a parameter thus any class can be passed to it and our function instantiates them. This is a security vulnerability. Thus we do not know the value of the parameter (and would not know it even if the attacker code is present during the analysis [but it is of course not] because SpotBugs is not interprocedural in this way) at the time of the analysis we only know that there is risk that someone passes a class to the function which may not have a public constructor thus it is not intended to be instantiated by anyone. The same applies to fields as well which may be private but our function sets their value or sets the accessibility to true.

KengoTODA (approved):
LGTM, thanks for your contribution! I will keep this PR open for a while, to get more feedback from the team.

ThrawnCA (approved):

spotbugs/etc/messages.xml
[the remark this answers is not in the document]
KengoTODA: I mean you need to add `<BugCode>` element in another file. Both `messages.xml` and `findbugs.xml` have `<BugCode>` element.

KengoTODA:
@baloghadamsoftware Please resolve conflict, probably around `CHANGELOG.md`.

baloghadamsoftware:
> 
> 
> @baloghadamsoftware Please resolve conflict, probably around `CHANGELOG.md`.

Done.

KengoTODA (approved):

PR #1551, opened 2021-05-18
Mutable array clones

baloghadamsoftware:
SEI-CERT rule OBJ05-J warns that cloning a private array before returning may not be sufficient if the elements of the array are of a mutable type. The clone() method of arrays only does a shallow-copy thus exposing the elements of the private array to all other classes. This patch improves the FindReturnRef detector to find such issues. The improved detector also warns for methods taking an array parameter of a mutable type and cloning it to a private array: the elements of the array are not copied, thus the caller can later modify them which results in modification of the elements of the private array as well.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * Bump Saxon-HE from 10.3 to 10.5 ([#1513](https://github.com/spotbugs/spotbugs/pull/1513))
 * Function `mutableSignature()` improved and factored out from the `MutableStaticFields` detector
 
+### Added
+*  `MS_EXPOSE_REP`, `EI_EXPOSE_REP`, `EI_EXPOSE_STATIC_REP2` and `EI_EXPOSE_REP2` now report for shallowly copied arrays (using clone()) of mutable objects
+
 ## 4.2.3 - 2021-04-12
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2021-??-??
 
+### Fixed
+- `MS_EXPOSE_REP` and `EI_EXPOSE_REP` are now reported for code returning a reference to a mutable object indirectly (e.g. via a local variable)
+
 ### Changed
 * Bump Saxon-HE from 10.3 to 10.5 ([#1513](https://github.com/spotbugs/spotbugs/pull/1513))
 * Function `mutableSignature()` improved and factored out from the `MutableStaticFields` detector
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
@@ -31,29 +31,19 @@ import edu.umd.cs.findbugs.LocalVariableAnnotation;
 import edu.umd.cs.findbugs.OpcodeStack;
 import edu.umd.cs.findbugs.ba.AnalysisContext;
 import edu.umd.cs.findbugs.ba.XField;
+import edu.umd.cs.findbugs.classfile.CheckedAnalysisException;
+import edu.umd.cs.findbugs.classfile.ClassDescriptor;
 import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
 import edu.umd.cs.findbugs.util.MutableClasses;
 
 public class FindReturnRef extends OpcodeStackDetector {
     boolean check = false;
 
-    boolean thisOnTOS = false;
-
-    boolean fieldOnTOS = false;
-
     boolean publicClass = false;
 
     boolean staticMethod = false;
 
-    String nameOnStack;
-
-    String classNameOnStack;
-
-    String sigOnStack;
-
     int parameterCount;
-
-    boolean fieldIsStatic;
 
     private final BugAccumulator bugAccumulator;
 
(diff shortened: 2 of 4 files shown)


baloghadamsoftware:
After #1529 was accepted by @ThrawnCA, this is the final patch of the series (from my side).

baloghadamsoftware:
Sorry for force pushing but there was a git mess which I had to sort out.

baloghadamsoftware:
I had to force push this again, but no it should be no problem because the review did not start yet.

baloghadamsoftware:
With tihs patch spotbugs found a [new error on MATSim](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Mutable_Array_Clones%29).

iloveeclipse:
> With tihs patch spotbugs found a [new error on MATSim](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Mutable_Array_Clones%29).

I'm really not sure this is a bug. Let's assume we have not array clone but a Collection copy - there are gazillion methods returning copies of internal collections without cloning every single element - that is not worth a warning IMHO. Why should it be different for array clones?

baloghadamsoftware:
> 
> 
> > With tihs patch spotbugs found a [new error on MATSim](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Mutable_Array_Clones%29).
> 
> I'm really not sure this is a bug. Let's assume we have not array clone but a Collection copy - there are gazillion methods returning copies of internal collections without cloning every single element - that is not worth a warning IMHO. Why should it be different for array clones?

Actually, according to the philosophy behind the SEI CERT rule OBJ05-J we should warn for every Collection shallow copy if the Collection contains mutable elements, the member is private and the method is public. However, finding all of them is difficult so we decided to start with arrays which are explicitly mentioned in the rule description at the second non-compliant code example: "The method fails to make a defensive copy of the array before returning it. Because the array contains references to Date objects that are mutable, a shallow copy of the array is insufficient because an attacker can modify the Date objects in the array."

iloveeclipse:
> > > With tihs patch spotbugs found a [new error on MATSim](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Mutable_Array_Clones%29).
> > 
> > 
> > I'm really not sure this is a bug. Let's assume we have not array clone but a Collection copy - there are gazillion methods returning copies of internal collections without cloning every single element - that is not worth a warning IMHO. Why should it be different for array clones?
> 
> Actually, according to the philosophy behind the SEI CERT rule OBJ05-J we should warn for every Collection shallow copy if the Collection contains mutable elements, the member is private and the method is public.

I believe in that case this detector should be disabled by default, or be part of https://find-sec-bugs.github.io/
Detector for such kind of "issues" would produce false positives in 99% of cases.

baloghadamsoftware:
> I believe in that case this detector should be disabled by default, or be part of https://find-sec-bugs.github.io/
> Detector for such kind of "issues" would produce false positives in 99% of cases.

Do you mean the whole detector here? The detector indeed produces many reports but this particular patch only adds the detection of returned clone of private arrays of mutable elements in a public method. If you click the link, you can see yourself that we got only one single new detection when applying this patch in a single project and no detection in the rest.

Sorry, I disagree with you that returning such a clone is 99% false positive. On the contrary, I strongly believe that returning a clone of an array of mutable elements is at least a code smell. There are very few reasons to protect the array itself by only returning its clone and not the array itself but not protecting the elements of the array. (I cannot think an example myself where this is the intended behavior.) In most of the cases the programmer intends to protects the elements as well but forgets that clone() does not make a deep copy (or forgets that the array stores references). This is a typical programming error, committed by novice Java developers.

baloghadamsoftware:
What would be the best way forward? Should I maybe introduce of a new error message (still with prefix "EI" and "MS") for these cases. This way we could better explain the user this particular bug. Furthermore, it also enables the user to distinguish between the different "EI" and "MS" messages, maybe filtering some kinds. If you think it is the appropriate way then I can make the new message with low priority so it is disabled by default.

This patch was tested on the following projects: MATSim, SpotBugs itself, Jenkins, NanoHTTPD, Bt, Ttorrent. It did not give any new reports, only 1 new finding on the MATSim project.  

Thanks for your review. I am looking forward to your suggestions on how to proceed.

baloghadamsoftware:
@iloveeclipse I changed the priority of the new error to low but I did not introduce a new bug type for now. Is this OK?

ThrawnCA (approved):

PR #1561, opened 2021-06-01
Try to build with JDK 17-ea

KengoTODA:
[17-ea build 24](https://github.com/openjdk/jdk/compare/jdk-17%2B23...jdk-17%2B24) now contains support for the [Pattern Matching for switch (preview)](https://openjdk.java.net/jeps/406), so I want to confirm how SpotBugs reacts on it.

I just worry that it may reproduce a previous problem at JDK 11 b7, #756 in our issues.
Java 11 changed a way to generate bytecode for try-with-resources and it affects SpotBugs, JaCoCo, and other tools. It's better to confirm the behavior by Early Access build before the GA for the coming LTS.

note: Gradle now uses Groovy 3.0.7, that depends on ObjectWeb ASM 9.0 which doesn't support JDK 17.
I guess we need to bump up the Groovy to 3.0.8 with [this change](https://github.com/apache/groovy/commit/d27994ea1f516894423980c958b65c8c02fb9395), but not sure how at this moment.
If the build in GitHub Actions failed, I will contact Gradle community to grasp how.

Update: [posted to Gradle Community Slack](https://gradle-community.slack.com/archives/CAHSN3LDN/p1622511901123100)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,10 +12,15 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Set up JDK 16
-        uses: actions/setup-java@v1
+      - uses: sormuras/download-jdk@v1
         with:
-          java-version: '16.0.x'
+          feature: 17
+      - name: Set up JDK 17-ea
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ env.JDK_VERSION }}
+          distribution: jdkfile
+          jdkFile: ${{ env.JDK_FILE }}
       - uses: actions/cache@v2
         with:
           path: ~/.gradle/caches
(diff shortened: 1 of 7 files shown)


ThrawnCA (approved):

KengoTODA:
It seems JEP 406 isn't ready in the current 17-ea (build 24):
https://github.com/spotbugs/spotbugs/pull/1561/checks?check_run_id=2723285422#step:8:417


```
/home/runner/work/spotbugs/spotbugs/spotbugsTestCases/src/java14/PatternMatchingForSwitch.java:19: error: : or -> expected

            case Integer i -> String.format("int %d", i);
> Task :spotbugsTestCases:classesJava14 FAILED
                        ^
/home/runner/work/spotbugs/spotbugs/spotbugsTestCases/src/java14/PatternMatchingForSwitch.java:19: error: not a statement
            case Integer i -> String.format("int %d", i);
```

I will wait for the coming EA release.

ThrawnCA (dismissed):

KengoTODA:
note: I've confirmed that [JEP 306: Restore Always-Strict Floating-Point Semantics](https://openjdk.java.net/jeps/306) does not affect SpotBugs detectors. There is no logic related to `strictfp`, except for [IClassConstants.ACC_STRICT](https://github.com/spotbugs/spotbugs/blob/51e586bed98393e53559a38c1f9bd15f54514efa/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IClassConstants.java#L84). There is no code depending on this `IClassConstants.ACC_STRICT`.

KengoTODA:
The build 26 contains enough feature to build switch with pattern match. :tada:

Issue #600, opened 2018-03-28
redundant null check doesn't recognize new pattern generated by javac in JDK 11


dougxc:
When the code below is compiled by javac from a JDK 11 build that includes [JDK-8191708](https://bugs.openjdk.java.net/browse/JDK-8191708), SpotBugs generates a warning:
```
> java -jar ~/spotbugs-3.1.1/lib/spotbugs.jar -textui bin/Nullcheck.class
M C RCN: Nullcheck of os at line 9 of value previously dereferenced in Nullcheck.test(OutputStream)  At Nullcheck.java:[line 8]
Warnings generated: 1
```
This is not something that can be addressed changing the code so SpotBugs should detect this pattern and suppress warnings related to it.

```
import java.io.*;
import java.net.*;

public class Nullcheck {
    protected boolean test(OutputStream os) {
        try {
            try (OutputStream outputStream = os) {
                outputStream.write(0);
            }
            return true;
        } catch (IOException ex) {
        }
        return false;
    }
}
```

dougxc:
Note that the `jdk-11-ea+5` build at http://jdk.java.net/11 does *not* include JDK-8191708 so to reproduce will require building the JDk from source or waiting for the next early access build to be made available.

jlahoda:
Doug meant:
https://bugs.openjdk.java.net/browse/JDK-8194978

The patch changes how try-with-resources are desugared, the new output is much smaller than the JDK7/8 one. Please see the review thread for details on the desugaring (although it is not mentioning the exact bytecode, only desugared Java source code):
http://mail.openjdk.java.net/pipermail/compiler-dev/2018-March/011753.html

Should there be questions, I'll try to answer them as best as I can.

dougxc:
Sorry for the wrong link - I fixed it in situ.

JuditKnoll:
As far as I can see this was fixed in https://github.com/spotbugs/spotbugs/pull/1575.

PR #1455, opened 2021-03-04
false positive: try-with-resource on JDK11 and later might fail with RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE

lars-sh:
#1338 Test that calling a method call when initializing a try-with-reource variable doesn't result in redundant nullcheck of nonnull value.

Issue #1338, opened 2020-10-17
false positive: try-with-resource on JDK11 and later might fail with RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE


lars-sh:
It seems, that RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE is still behaving different based on the JDK used in some situations. I expected my problems to be solved with #259, #647, #1248 or #1250, but the latest Spotbugs release didn't solve them.

I'm using Spotbugs 4.1.4 with low threshold via the Maven Spotbugs Plugin 4.1.3 and receive the error message `Redundant nullcheck of inputStream, which is known to be non-null in de.larssh.pulls.IssuesSpotbugs.loadKeyStore(Path) [de.larssh.pulls.IssuesSpotbugs] Redundant null check at IssuesSpotbugs.java:[line 10] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE`.

You can find an issue showcase at https://github.com/lars-sh/parent/tree/issues-spotbugs . Just in case the links do not work in future, here's the essential code snippet:
```
	public static KeyStore loadKeyStore(final Path path) throws Exception {
		try (InputStream inputStream = Files.newInputStream(path)) {
			// final KeyStore keyStore = KeyStore.getInstance("JKS");
			// keyStore.load(inputStream, password.toCharArray());
			// return keyStore;
			return KeyStore.getInstance("JKS");
		}
	}
```

The related GitHub Action builds can be found at https://github.com/lars-sh/parent/actions?query=branch%3Aissues-spotbugs . They show, that Spotbugs do **not** raise a RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE for JDK 8 to 10, but raise one when compiling with JDK 11 and later.

EDIT: Migrated from Travis CI to GitHub Actions

Zlika:
I confirm the problem.
It seems that the fix for #259 in spotbugs 4.1.3 only fixes "simple" try with resources statements, but it does not work if the resource is created by calling a method inside the try with resource statement.

A simple test case to reproduce the problem:

```
public class RcnRedundantNullCheckOfNonNullValueBug
{
    public static void main(String[] args) throws IOException
    {
        // OK, no RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE here with spotbugs 4.1.4
        try (Socket socket = new Socket(InetAddress.getLocalHost(), 80))
        {
[...]

PR #1575, opened 2021-06-10
#600: Fix false positves for RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE…

elruwen:
… in Java 11

This should fix issue #600 and #1338. I also fixed the broken tests for #259.

This is my very first change to spotbugs. I am not 100% sure about the java11 and java14 test classes folder. Are those only run for those JDKs? My code should work for all of them...

Cheers
Ruwen

----


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -1822,6 +1826,97 @@ private boolean isGeneratedCodeInCatchBlock(@NonNull ConstantPool cp, int pc) {
...
+        // - at least one assignment
+        // - at least one call of addSuppressed
+        // - at least one throws statement
+        //
+        // The two calls to the close method need to have the same line number as the end of a throwables catch block.
+        // There must be also another catch of Throwable, but the line number does not need to match.
+        //
+        // To understand the code, please have a look at the test Issue600Test.
+        // TryWithResources* are the try-with-resources examples, ExplicitNullCheck* is TryWithResources* compiled and then decompiled.
+        if (lineNumberTable != null) {
```
ThrawnCA, line 1850: This method is getting quite long; perhaps move this block out to a helper?

spotbugsTestCases/src/java/ghIssues/issue600/MyAutocloseable.java
```diff
@@ -0,0 +1,16 @@
+package ghIssues.issue600;
+
+public class MyAutocloseable implements AutoCloseable {
+    @Override
+    public void close() throws Exception {
+        System.out.println("close");
```
ThrawnCA, line 6: Perhaps a logger instead of STDOUT?

elruwen: That class belongs to a testcase. Why use a logger? I prefer to keep them as basic as possible. I just didn't want to leave the method empty to avoid it being optimised away or something like that.

ThrawnCA: Loggers are inherently more flexible and configurable than STDOUT. Consider a situation where you might want to capture test outputs and email them somewhere; it would be easier to configure that with a logger.

Also, as a result, tools like PMD will tend to flag usage of System.out.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
elruwen: Done!

spotbugsTestCases/src/java/ghIssues/issue600/MyAutocloseable.java
elruwen: Sorry I should have been more precise.

This is one of the classes the unit tests are analysing. This code never actually runs. That's why it is in the spotbugsTestCases project. Have a look, there are a lot of System.out.printlns in that project.

While I am in general agree with you on using loggers in production and unit test, in this case I try to keep the code as minimal as possible and in order to achieve that, I avoid any dependencies. java.util.logging is not part of the Java base module, I don't want to include a module dependency for no good reason. I can't use java.lang.System.Logger since it doesn't exist in Java 8. I just need a dummy statement and System.out.println is easily understood by everyone.

KengoTODA:
This PR will close #1455

ThrawnCA (approved):

elruwen:
@ThrawnCA I noticed I had a conflict with the Readme. That is fixed. Can you please approve the PR again? Thanks!

ThrawnCA (approved):

elruwen:
@ThrawnCA can you also please approve the workflow? I guess that is the build. They don't run automatically anymore for first-time contributors like me (see https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). 
And once the build passed, can you merge it?

elruwen:
I had to merge again since the changelog was changed. Please review and merge :)

ThrawnCA (approved):

dmivankov:
looks like it creates new false -positives in 4.4.0, places like
```java
    try (Connection connection = dataSource.getConnection()) {
      try (PreparedStatement preparedStatement = connection.prepareStatement("SELECT DATABASE()")) { // HERE
        try (ResultSet result = preparedStatement.executeQuery()) {    
```
or
```java
    try (Connection conn = dataSource.getConnection()) {
      Statement stmt = conn.createStatement(); // HERE, off-by-one?
```
or
```java
    try (...something...) {
       ....
    } catch (IOException e) { // HERE?
```

elruwen:
Hmm I just gave it a try with the following code:
```
public class TryWithResourcesSimple {
    public void checkedException() throws Exception {
        try (InputStream is = inputStream()) {
            System.out.println(is.available());
        } catch (IOException e) {
            e.printStackTrace();
        }
    }

    public InputStream inputStream() {
        try {
            return Files.newInputStream(Paths.get("/dummy"));
        } catch (IOException e) {
        }
    }


}

```
which matches your last example?

Works for me on Java 8, 11 and 16.

Can you provide a complete example?

dougxc:
Great to see this fixed - thanks! Is there a timeline for when this will show up in a release?

KengoTODA:
This fix has been released as 4.4.0. You can confirm it in the `CHANGELOG.md`.

dougxc:
> This fix has been released as 4.4.0. You can confirm it in the CHANGELOG.md.

Thanks. I did not check the CHANGELOG but instead just tried using 4.4.2 and saw there were still false positives (#1791) so assumed this PR was not part of 4.4.2. Sorry for the confusion.

PR #1580, opened 2021-06-14
Extend FindReturnRef to warn for exposing buffers

baloghadamsoftware:
SEI CERT rule FIO05-J requires that buffers or their backing arrays
should not be exposed by methods to untrusted code. This patch
implements the check for such behavior in the already existing detector
FindReturnRef. To make the error easier to understand for the user
we did not reuse the existing messages but instead added new
ones: EI_EXPOSE_BUF, MS_EXPOSE_BUF, EI_EXPOSE_BUF2 and
EI_EXPOSE_STATIC_BUF.

----

### DIFF
--- a/spotbugs/etc/findbugs.xml
+++ b/spotbugs/etc/findbugs.xml
@@ -436,7 +436,7 @@
           <Detector class="edu.umd.cs.findbugs.detect.FindNakedNotify" speed="fast"
                     reports="NN_NAKED_NOTIFY"/>
           <Detector class="edu.umd.cs.findbugs.detect.FindReturnRef" speed="fast"
-                    reports="MS_EXPOSE_REP,EI_EXPOSE_REP,EI_EXPOSE_REP2,EI_EXPOSE_STATIC_REP2"/>
+                    reports="MS_EXPOSE_REP,EI_EXPOSE_REP,EI_EXPOSE_REP2,EI_EXPOSE_STATIC_REP2,MS_EXPOSE_BUF,EI_EXPOSE_BUF,EI_EXPOSE_STATIC_BUF2,EI_EXPOSE_BUF2"/>
           <Detector class="edu.umd.cs.findbugs.detect.FindRunInvocations" speed="fast"
                     reports="RU_INVOKE_RUN"/>
           <Detector class="edu.umd.cs.findbugs.detect.SwitchFallthrough" speed="fast"
@@ -864,9 +864,13 @@
           <BugPattern abbrev="IS" type="IS2_INCONSISTENT_SYNC" category="MT_CORRECTNESS"/>
           <BugPattern abbrev="NN" type="NN_NAKED_NOTIFY" category="MT_CORRECTNESS"/>
           <BugPattern abbrev="MS" type="MS_EXPOSE_REP" category="MALICIOUS_CODE"/>
+          <BugPattern abbrev="MS" type="MS_EXPOSE_BUF" category="MALICIOUS_CODE"/>
           <BugPattern abbrev="EI" type="EI_EXPOSE_REP" category="MALICIOUS_CODE" cweid="374"/>
+          <BugPattern abbrev="EI" type="EI_EXPOSE_BUF" category="MALICIOUS_CODE" cweid="374"/>
           <BugPattern abbrev="EI2" type="EI_EXPOSE_REP2" category="MALICIOUS_CODE" cweid="374"/>
           <BugPattern abbrev="MS" type="EI_EXPOSE_STATIC_REP2" category="MALICIOUS_CODE"/>
+          <BugPattern abbrev="EI2" type="EI_EXPOSE_BUF2" category="MALICIOUS_CODE" cweid="374"/>
+          <BugPattern abbrev="MS" type="EI_EXPOSE_STATIC_BUF2" category="MALICIOUS_CODE"/>
           <BugPattern abbrev="Ru" type="RU_INVOKE_RUN" category="MT_CORRECTNESS" cweid="572"/>
           <BugPattern abbrev="SP" type="SP_SPIN_ON_FIELD" category="MT_CORRECTNESS"/>
           <BugPattern abbrev="NS" type="NS_NON_SHORT_CIRCUIT" category="STYLE"/>
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -3540,18 +3540,18 @@ Thus, having a mutable instance field generally creates race conditions.
 <![CDATA[
   <p> Returning a reference to a buffer (java.nio.*Buffer) which wraps an array stored in one of the object's fields
   exposes the internal representation of the array elements because the buffer only stores a reference to the array
-  instead of copying its content. Simliarly, returning a shallow-copy of such a buffer (using its duplicate() method
+  instead of copying its content. Simlarly, returning a shallow-copy of such a buffer (using its duplicate() method)
   stored in one of the object's fields also exposes the internal representation of the buffer.&nbsp;
    If instances are accessed by untrusted code, and unchecked changes to
    the array would compromise security or other
    important properties, you will need to do something different.
   Returning a read-only buffer (using its asReadOnly() method) or copying the array to a new buffer (using its put()
-  method) is better approach in many situations.</p>
+  method) is a better approach in many situations.</p>
 ]]>
     </Details>
   </BugPattern>
   <BugPattern type="EI_EXPOSE_BUF2">
-    <ShortDescription>May expose internal representation by creating a buffer which incorporating reference to array</ShortDescription>
+    <ShortDescription>May expose internal representation by creating a buffer which incorporates reference to array</ShortDescription>
     <LongDescription>{1} may expose internal representation by creating a buffer which contains an external array into {2.givenClass}</LongDescription>
     <Details>
 <![CDATA[
@@ -3561,7 +3561,7 @@ Thus, having a mutable instance field generally creates race conditions.
    are accessed by untrusted code, and unchecked changes to
    the array would compromise security or other
    important properties, you will need to do something different.
-  Storing a copy of the array is better approach in many situations.</p>
+  Storing a copy of the array is a better approach in many situations.</p>
 ]]>
     </Details>
   </BugPattern>
@@ -3575,7 +3575,7 @@ Thus, having a mutable instance field generally creates race conditions.
    If unchecked changes to
    the array would compromise security or other
    important properties, you will need to do something different.
-  Storing a copy of the array is better approach in many situations.</p>
+  Storing a copy of the array is a better approach in many situations.</p>
 ]]>
     </Details>
   </BugPattern>
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -3520,6 +3520,65 @@ Thus, having a mutable instance field generally creates race conditions.
 ]]>
     </Details>
   </BugPattern>
+  <BugPattern type="MS_EXPOSE_BUF">
+    <ShortDescription>May expose internal representation by returning a buffer sharing non-public data</ShortDescription>
+    <LongDescription>{1} may expose internal representation by returning {2.givenClass}</LongDescription>
+    <Details>
+<![CDATA[
+  <p> A public static method either returns a buffer (java.nio.*Buffer) which wraps an array that is part of the
+  static state of the class by holding a reference only to this same array or it returns a shallow-copy of a buffer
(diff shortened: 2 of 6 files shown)


baloghadamsoftware:
If you think it is better then I can reuse the EI_EXPOSE_REP, MS_EXPOSE_REP, EI_EXPOSE_REP2 and EI_EXPOSE_STATIC_REP error messages for this kind of error as well since this one is similar to the original one. Just tell me which way to go.

spotbugs/etc/findbugs.xml
```diff
@@ -864,9 +864,13 @@
           <BugPattern abbrev="IS" type="IS2_INCONSISTENT_SYNC" category="MT_CORRECTNESS"/>
           <BugPattern abbrev="NN" type="NN_NAKED_NOTIFY" category="MT_CORRECTNESS"/>
           <BugPattern abbrev="MS" type="MS_EXPOSE_REP" category="MALICIOUS_CODE"/>
+          <BugPattern abbrev="MS" type="MS_EXPOSE_BUF" category="MALICIOUS_CODE"/>
```
ThrawnCA, line 867: Aren't abbreviations supposed to be unique?

spotbugs/etc/messages.xml
```diff
@@ -3517,6 +3517,65 @@ Thus, having a mutable instance field generally creates race conditions.
...
+    </Details>
+  </BugPattern>
+  <BugPattern type="EI_EXPOSE_BUF">
+    <ShortDescription>May expose internal representation by returning a buffer sharing non-public data</ShortDescription>
+    <LongDescription>{1} may expose internal representation by returning {2.givenClass}</LongDescription>
+    <Details>
+<![CDATA[
+  <p> Returning a reference to a buffer (java.nio.*Buffer) which wraps an array stored in one of the object's fields
+  exposes the internal representation of the array elements because the buffer only stores a reference to the array
+  instead of copying its content. Simliarly, returning a shallow-copy of such a buffer (using its duplicate() method
```
ThrawnCA, line 3543: Misspelled "similarly".

Also, opening bracket needs a closing bracket.

spotbugs/etc/messages.xml
```diff
@@ -3517,6 +3517,65 @@ Thus, having a mutable instance field generally creates race conditions.
...
+  stored in one of the object's fields also exposes the internal representation of the buffer.&nbsp;
+   If instances are accessed by untrusted code, and unchecked changes to
+   the array would compromise security or other
+   important properties, you will need to do something different.
+  Returning a read-only buffer (using its asReadOnly() method) or copying the array to a new buffer (using its put()
+  method) is better approach in many situations.</p>
```
ThrawnCA, line 3549: "is better approach" -> "is a better approach"

spotbugs/etc/messages.xml
```diff
@@ -3517,6 +3517,65 @@ Thus, having a mutable instance field generally creates race conditions.
...
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="EI_EXPOSE_BUF2">
+    <ShortDescription>May expose internal representation by creating a buffer which incorporating reference to array</ShortDescription>
```
ThrawnCA, line 3554: "which incorporating" -> "which incorporates"

spotbugs/etc/messages.xml
```diff
@@ -3517,6 +3517,65 @@ Thus, having a mutable instance field generally creates race conditions.
...
+    <LongDescription>{1} may expose internal representation by creating a buffer which contains an external array into {2.givenClass}</LongDescription>
+    <Details>
+<![CDATA[
+  <p> This code creates a buffer which stores a reference to an external array or the array of an external buffer into
+  the internal representation of the object.&nbsp;
+   If instances
+   are accessed by untrusted code, and unchecked changes to
+   the array would compromise security or other
+   important properties, you will need to do something different.
+  Storing a copy of the array is better approach in many situations.</p>
```
ThrawnCA, line 3564: "is better approach" -> "is a better approach"

spotbugs/etc/messages.xml
```diff
@@ -3517,6 +3517,65 @@ Thus, having a mutable instance field generally creates race conditions.
...
+    <ShortDescription>May expose internal static state by creating a buffer which stores an external array into a static field</ShortDescription>
+    <LongDescription>{1} may expose internal static state by creating a buffer which stores an external array into static field {2}</LongDescription>
+    <Details>
+<![CDATA[
+  <p> This code creates a buffer which stores a reference to an external array or the array of an external buffer into
+  a static field.
+   If unchecked changes to
+   the array would compromise security or other
+   important properties, you will need to do something different.
+  Storing a copy of the array is better approach in many situations.</p>
```
ThrawnCA, line 3578: "is better approach" -> "is a better approach"

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -109,88 +112,167 @@ public void sawOpcode(int seen) {
...
-            thisOnTOS = true;
-            fieldOnTOS = false;
-            return;
+        if (seen == Const.ARETURN) {
+            OpcodeStack.Item item = stack.getStackItem(0);
+            XField field = item.getXField();
+            boolean isBuf = false;
+            if (field == null) {
+                field = bufferFieldDuplicates.get(item);
+                isBuf = true;
```
ThrawnCA, line 158: This line will execute even if a buffer wasn't actually found, which seems error-prone.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -109,88 +112,167 @@ public void sawOpcode(int seen) {
...
-            sigOnStack = getSigConstantOperand();
-            fieldIsStatic = false;
-            return;
+        if (seen == Const.INVOKEVIRTUAL) {
+            MethodDescriptor method = getMethodDescriptorOperand();
+            OpcodeStack.Item item = stack.getStackItem(0);
+            XField field = item.getXField();
+            if (method == null || !"duplicate".equals(method.getName())
+                    || !method.getSignature().matches("\\(\\)Ljava/nio/[A-Za-z]+Buffer;")
+                    || !method.getClassDescriptor().getSignature().matches("Ljava/nio/[A-Za-z]+Buffer;")) {
```
ThrawnCA, line 186: Perhaps this regex could be precompiled?

ThrawnCA (changes requested):

spotbugs/etc/findbugs.xml
baloghadamsoftware: Not at all. There are lots of "MS" bugs from various detectors. Also "Dm", "Eq" etc. The more interesting question for me is that the original bugs from FindReturnRef are quite inconsistent: EI_EXPOSE_REP (abbr. EI) for returning a private instance level member. MS_EXPOSE_REP (abbr. MS) for doing the same with a static one. However: EI_EXPOSE_REP2 (abbr. EI2) for storing an external mutable at instance level but EI_EXPOSE_STATIC_REP (abbr. MS!!!) for doing the same in a static method. Also the bug descriptions are somewhat inconsistent: for the same kind of bug there are totally different descriptions for static and non-static versions. I tried to follow this way with the "buffer" error messages because I do not know the reason behind. However, I am not sure if we need separate messages for the buffer versions at all.

baloghadamsoftware:
Here are our results (merged for all 6 projects): [link](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&run=SpotBugs%20on%20Jenkins%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&run=SpotBugs%20on%20SpotBugs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&run=SpotBugs%20on%20Ttorrent%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&run=SpotBugs%20on%20Bt%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&run=SpotBugs%20on%20NanoHTTPD%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Simplify_FindReturnRef%29&newcheck=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29&newcheck=SpotBugs%20on%20SpotBugs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29&newcheck=SpotBugs%20on%20Jenkins%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29&newcheck=SpotBugs%20on%20Ttorrent%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29&newcheck=SpotBugs%20on%20Bt%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29&newcheck=SpotBugs%20on%20NanoHTTPD%20%28user%3A%20baloghadamsoftware,%20branch%3A%20Expose_Buffers%29). You can see that we have exactly one buffer returned and one array taken as parameter and used as buffer backing array. The return is intentional in this case, thus false positive, although it is quite non-standard. The other one is a constructor which takes several parameters causing several EI2 bugs anyway.

spotbugs/etc/messages.xml
```diff
@@ -3540,18 +3540,18 @@ Thus, having a mutable instance field generally creates race conditions.
 <![CDATA[
   <p> Returning a reference to a buffer (java.nio.*Buffer) which wraps an array stored in one of the object's fields
   exposes the internal representation of the array elements because the buffer only stores a reference to the array
-  instead of copying its content. Simliarly, returning a shallow-copy of such a buffer (using its duplicate() method
+  instead of copying its content. Simlarly, returning a shallow-copy of such a buffer (using its duplicate() method)
```
ThrawnCA, line 3543: Misspelled "Similarly".

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -60,6 +62,12 @@
 
     private final BugAccumulator bugAccumulator;
 
+    private static final Matcher bufferClassMatcher = Pattern.compile("Ljava/nio/[A-Za-z]+Buffer;").matcher("");
```
ThrawnCA, line 65: Class constants should use SCREAMING_SNAKE_CASE.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -182,8 +194,8 @@ public void sawOpcode(int seen) {
             OpcodeStack.Item item = stack.getStackItem(0);
             XField field = item.getXField();
             if (method == null || !"duplicate".equals(method.getName())
-                    || !method.getSignature().matches("\\(\\)Ljava/nio/[A-Za-z]+Buffer;")
-                    || !method.getClassDescriptor().getSignature().matches("Ljava/nio/[A-Za-z]+Buffer;")) {
+                    || !duplicateMethodSignatureMatcher.reset(method.getSignature()).matches()
```
ThrawnCA, line 197: This is a curious approach. Normally I would just create a constant `Pattern` and then generate a new `Matcher` on each use. Is this more efficient?

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
baloghadamsoftware: Yes, that was also my intention. But then I met this: https://www.baeldung.com/java-regex-pre-compile. In this study there is a benchmark which claims that creating a matcher instead and resetting it on each use is almost 2 time faster. I think it is no wonder, since creation of a whole new Matcher class is definitely more expensive than just replacing its string reference to another string. However, if you think that this is too hard to understand for other developers then I can change it to pattern-only.

ThrawnCA (approved):

baloghadamsoftware:
@KengoTODA please note that this one cannot be merged together with #1551. One of them (preferably #1551) should be merged first and then I have to rebase the other one (this one, #1580).

PR #1582, opened 2021-06-17
Mutable array clones

baloghadamsoftware:
SEI-CERT rule OBJ05-J warns that cloning a private array before returning may not be sufficient if the elements of the array are of a mutable type. The clone() method of arrays only does a shallow-copy thus exposing the elements of the private array to all other classes. This patch improves the FindReturnRef detector to find such issues. The improved detector also warns for methods taking an array parameter of a mutable type and cloning it to a private array: the elements of the array are not copied, thus the caller can later modify them which results in modification of the elements of the private array as well.
----

Reopened version of #1551

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
@@ -51,14 +51,21 @@ public class FindReturnRef extends OpcodeStackDetector {
 
     int parameterCount;
 
+    private XField fieldUnderClone = null;
+    private OpcodeStack.Item paramUnderClone = null;
+    private XField fieldCloneUnderCast = null;
+    private OpcodeStack.Item paramCloneUnderCast = null;
     private XField bufferFieldUnderDuplication = null;
     private OpcodeStack.Item bufferParamUnderDuplication = null;
     private XField fieldUnderWrapToBuffer = null;
     private OpcodeStack.Item paramUnderWrapToBuffer = null;
+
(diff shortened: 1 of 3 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -132,7 +143,7 @@ public void sawOpcode(int seen) {
             if (capture != CaptureKind.NONE) {
                 bugAccumulator.accumulateBug(
                         new BugInstance(this, "EI_EXPOSE_STATIC_" + (capture == CaptureKind.BUF ? "BUF2" : "REP2"),
-                                capture == CaptureKind.BUF ? LOW_PRIORITY : NORMAL_PRIORITY)
+                                (capture == CaptureKind.REP) ? NORMAL_PRIORITY : LOW_PRIORITY)
```
ThrawnCA, line 146: The extra brackets aren't strictly necessary.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -255,15 +315,21 @@ private boolean nonPublicFieldOperand() {
...
+                        return CaptureKind.NONE;
+                    }
                 }
+                kind = CaptureKind.BUF;
+            } else {
+                kind = CaptureKind.ARRAY_CLONE;
             }
             top = newTop;
-            kind = CaptureKind.BUF;
+            System.err.println("This is a clone:" + top);
```
ThrawnCA, line 332: Should have a space after the colon.

ThrawnCA (approved):

ThrawnCA (approved):

Issue #1601, opened 2021-07-05
EI_EXPOSE_REP and EI_EXPOSE_REP2 occur if using Guava Immutable objects


callum-kilby:
We are using [Guava Immutable objects](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/ImmutableCollection.html) in our project, and from SpotBugs 4.3.0 we have SpotBugs failures EI_EXPOSE_REP and EI_EXPOSE_REP2 when directly setting/retrieving class fields as Immutable objects.

For example:

```java
private Map<String, String> overrides;

public CLISetup(String commandLine, ImmutableMap<String, String> overrides) {
    this.commandLine = commandLine;
    this.overrides = overrides;
}
```
results in an EI_EXPOSE_REP2 failure from this.overrides = overrides.

This failure is incorrect because the object being passed in and set for the class field is shallowly immutable.

vladimirfx:
After upgrade to 4.3.0, I have a huge amount of EI_EXPOSE_REP and EI_EXPOSE_REP2 on previously clean code. (i.e. reports on getters returning "java.time.Instant")

I was forced to disable both rules. It is possible to restore previous logic in form of a separate rule OR introduce new logic as separate rules? It will be very useful for obvious cases when an array or something like "java.util.Date" is returned or stored.

xzel23:
I also ran into the issue with returning java.time.Instant. I don't know how this is supposed to work: how would SpotBugs know that java.time.Instant is immutable? Does it have a whitelist of classes, or is this done by analysing the bytecode? If it's the first, the immutable standard classes should be added to the list, if second, something's wrong with the byte code analyser.

xzel23:
OK, I just saw it. `java.time.Instant` should be added to `edu.umd.cs.findbugs.util.MutableClasses.KNOWN_IMMUTABLE_CLASSES`. 

Here is a list of the classes that are immutable in that package (stated directly in the classes' Javadoc):

java.time.Duration
java.time.Instant
java.time.LocalDate
java.time.LocalDateTime
java.time.LocalTime
java.time.MonthDay
java.time.OffsetDateTime
java.time.OffsetTime
java.time.Year
java.time.YearMonth
java.time.ZonedDateTime
java.time.ZoneOffset
java.time.ZoneRegion

[...]

PR #1603, opened 2021-07-05
Fix for the heuristics that recognizes immutable classes

baloghadamsoftware:
Class names containting the substring "immutable" are considered now immutable.
Fixing issue #1601.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
@@ -102,7 +102,13 @@ public class MutableClasses {
     }
 
     public static boolean looksLikeASetter(Method method, JavaClass cls) {
+        // Static methods are no setters.
         if (method.isStatic()) {
+            return false;
+        }
+
+        // Non-public setters cannot be used from outside the package so disregard them.
+        if (!method.isPublic()) {
             return false;
         }
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
@@ -5,6 +5,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import java.util.List;
 
 import org.apache.bcel.Repository;
@@ -45,12 +46,15 @@ public class MutableClasses {
             return false;
         }
 
-        if (IMMUTABLE_CLASS_NAME_MATCHER.reset(sig).matches()) {
+        if (IMMUTABLE_CLASS_NAME_MATCHER.reset(sig.substring(sig.lastIndexOf('.') + 1)).matches()) {
             return false;
         }
 
         try {
             JavaClass cls = Repository.lookupClass(dottedClassName);
+            if (Stream.of(cls.getAnnotationEntries()).anyMatch(s -> s.getAnnotationType().endsWith("Immutable"))) {
+                return false;
+            }
             return isMutable(cls);
         } catch (ClassNotFoundException e) {
             AnalysisContext.reportMissingClass(e);
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
@@ -45,7 +45,7 @@ public class MutableClasses {
 
     private static final List<String> SETTER_LIKE_NAMES = Arrays.asList(
             "set", "put", "add", "insert", "delete", "remove", "erase", "clear", "push", "pop",
-            "enqueue", "dequeue", "write", "append", "replace");
+            "enqueue", "dequeue", "append", "replace");
 
     public static boolean mutableSignature(String sig) {
         if (sig.charAt(0) == '[') {
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
@@ -72,7 +72,8 @@ public class MutableClasses {
 
         try {
             JavaClass cls = Repository.lookupClass(dottedClassName);
-            if (Stream.of(cls.getAnnotationEntries()).anyMatch(s -> s.getAnnotationType().endsWith(".Immutable"))) {
+            if (Stream.of(cls.getAnnotationEntries()).anyMatch(s -> (s.getAnnotationType().endsWith(".Immutable"))
+                    || s.getAnnotationType().equals("jdk.internal.ValueBased"))) {
                 return false;
             }
             return isMutable(cls);
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
@@ -3,8 +3,6 @@ package edu.umd.cs.findbugs.util;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


spotbugs/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -14,6 +14,10 @@ public void TestKnownImmutable() {
         Assert.assertFalse(MutableClasses.mutableSignature("Ljava/lang/String;"));
     }
 
+    public void TestNamedImmutable() {
```
KengoTODA, line 17: I guess we need an annotation here?

```suggestion
    @Test
    public void TestNamedImmutable() {
```

KengoTODA (changes requested):
Please resolve a conflict in the `CHANGELOG.md` file.

And I personally feel that it's not good to use class/package name to verify its immutability.
The target classes could be implemented by coder who has less knowledge about immutability, and they may use `.*immutable.*` as a magic word to remove "noisy" warning made by SpotBugs... :\

baloghadamsoftware:
> And I personally feel that it's not good to use class/package name to verify its immutability.
> The target classes could be implemented by coder who has less knowledge about immutability, and they may use `.*immutable.*` as a magic word to remove "noisy" warning made by SpotBugs... :\

I agree with you, coders may misuse this heuristic. I do not know how often that happens, actually I never thought that someone would intentionally silence a true positive just because of laziness. Now I changed it to only look for the word "immutable" in the class name but not in the package name. Is it OK? If not, then I can simply put the signatures of the Guava immutable classes into the list of know immutables. (However, this does not protect against false positives coming from other APIs.)

KengoTODA:
> Now I changed it to only look for the word "immutable" in the class name but not in the package name. Is it OK?

Not sure. I personally think it's not enough and better to use the signatures of Guava immutable classes, but let's wait opinions from other community members.

baloghadamsoftware:
> 
> 
> > Now I changed it to only look for the word "immutable" in the class name but not in the package name. Is it OK?
> 
> Not sure. I personally think it's not enough and better to use the signatures of Guava immutable classes, but let's wait opinions from other community members.

OK, I will put the Guava signatures into the list for now. However, it came to my mind that it is much easier for a developer to use a @Immutable annotation than change the name of the class or the package. And we must support this annotation to avoid false positives reported by the users. Thus I do not think that many would rename their classes or packages to forcefully silence false positives.

ThrawnCA:
I think there's too much scope for false positives. Just within the past few months, I overrode an `ImmutableMultiDict` class to enable a single limited operation for a specific reason, and named the result `MostlyImmutableMultiDict`. Using this heuristic, my subclass would be treated as immutable, when it shouldn't be.

rkraneis:
Should immutable classes from `java.time` package be added as well? Since 4.3.0 we get `EI_EXPOSE_REP` for those, too ...

baloghadamsoftware:
> 
> 
> Should immutable classes from `java.time` package be added as well? Since 4.3.0 we get `EI_EXPOSE_REP` for those, too ...
> (`java.time` package description reads “[...] All the classes are immutable and thread-safe. [...]”)

Sure. I will update this PR to include them.

callum-kilby:
Just checking in, what is the status of this request? There's other reporters with the same issue, with some additional suggestions for checking for immutability

baloghadamsoftware:
> 
> 
> Should immutable classes from `java.time` package be added as well? Since 4.3.0 we get `EI_EXPOSE_REP` for those, too ...
> (`java.time` package description reads “[...] All the classes are immutable and thread-safe. [...]”)

Sure. I will update this PR to include them.

> 
> 
> I think there's too much scope for false positives. Just within the past few months, I overrode an `ImmutableMultiDict` class to enable a single limited operation for a specific reason, and named the result `MostlyImmutableMultiDict`. Using this heuristic, my subclass would be treated as immutable, when it shouldn't be.

What about Detecting *.Immutable*? I think there is hardly anyone who calls a class ImmutableLikeXXX.

baloghadamsoftware:
@KengoTODA & @ThrawnCA For now I only added the Guava immutable classes, the java.date package and the \@Immutable annotation.

baloghadamsoftware:
> 
> 
> Just checking in, what is the status of this request? There's other reporters with the same issue, with some additional suggestions for checking for immutability

Where are they? I searched in the issues and discussions, but I could not find any. It would be good to read them and include all the classes they claims immutable into the list of known immutable classes.

callum-kilby:
> > Just checking in, what is the status of this request? There's other reporters with the same issue, with some additional suggestions for checking for immutability
> 
> Where are they? I searched in the issues and discussions, but I could not find any. It would be good to read them and include all the classes they claims immutable into the list of known immutable classes.

I'm looking at the other comments in my issue:
https://github.com/spotbugs/spotbugs/issues/1601

It's mostly `java.time` classes, but also `Java.lang.Class`. There's also some discussion on checking for a `jdk.internal.ValueBased` annotation, or providing the previous behaviour for these checks.

callum-kilby:
Is it possible to include an additional whitelist that is set as a plugin configuration option? It would let users specify immutable classes that they have defined in their project without asking them to be added to the spotbugs whitelist

baloghadamsoftware:
With the last modification it seems that we finally got rid of all false positives. Now all the findings are true positives. However, there are quite many of them so I would propose changing the priority of all the findings of the `FindReturnRef` detector to low.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -21,11 +23,29 @@
...
+            "com.google.common.collect.ImmutableSortedSet",
+            "com.google.common.collect.ImmutableTable"));
+
+    private static final Set<String> KNOWN_IMMUTABLE_PACKAGES = new HashSet<>(Arrays.asList(
+            "java.math", "java.time"));
 
     private static final List<String> SETTER_LIKE_NAMES = Arrays.asList(
             "set", "put", "add", "insert", "delete", "remove", "erase", "clear", "push", "pop",
-            "enqueue", "dequeue", "write", "append", "replace");
+            "enqueue", "dequeue", "append", "replace");
```
KengoTODA, line 48: This change is backward compatible and will affect many users.
So personally I don't want to merge this change to SpotBugs.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -68,12 +102,34 @@ private static boolean isMutable(JavaClass cls) {
     }
 
     public static boolean looksLikeASetter(Method method, JavaClass cls) {
+        // Static methods are no setters.
+        if (method.isStatic()) {
+            return false;
+        }
+
+        // Non-public setters cannot be used from outside the package so disregard them.
```
KengoTODA, line 110: I think this change is not so natural. The package affect nothing to the immutability of class architecture.

KengoTODA (changes requested):
This PR does not fix existing false positive, it introduces many new false negatives and it won't benefit most of Java developers.

baloghadamsoftware:
Thank you, for your review @KengoTODA!

> This PR does not fix existing false positive, it introduces many new false negatives and it won't benefit most of Java developers.

Which part, do you mean? Actually, this PR is quite complex and contains different fixes. Originally, I implemented a heuristics to replace the placeholder (which was just a list of 4 mutable classes and all the rest were considered as immutable) in PR #1533. However, my heuristics was incomplete and contained some bugs. It caused tons of findings, especially in the `FindReturnRef` detector, and some of them were false positives. Users reported them here: #1601.

In this patch I tried to fix these issues and also made a long and tiring manual research to find all the possible false positives and fix the heuristics to not to report them. The fixes in this PR:

1) Static methods are surely no setters, because they cannot write instance fields. I fixed this in this PR.

2) Setter-like methods which throw UnsupportedOperationException are no setters. I fixed this in this PR.

3) Private methods are no setters. If there are no public setters, then the only usage of private methods is in the constructor or block initializers. This means that the class is still immutable. What could be disputed here is the case of package-public methods. In this PR I decided to omit them, but if you think it is wrong then I can include them.

4) When checking the findigs I noticed that methods beginning with the word `write` are usually no setters, but stream output methods. For this reason I removed this tag from the list in this PR.

5) There are methods with name `set...()`, `add...()` etc which create a new object instead of modifying the current one. In the original heuristics I tried to detect them by comparing their result type with the class they are defined in, but for some reason it did not work as I expected. (I forgot to test them explicitly.) Furthermore, I found cases where they return a different class type but do not modify the `this` object. For this reason I changed this part of the algorithm in this PR to disregard methods which return any object type. This can be disputed, of course, if you do not agree.

6) For performance reasons (and also because the heuristics was too weak initially) I also had a fixed list of immutable classes. In this PR I extended this list with e.g. Guava immutable classes.

7) There are packages in the Java API that contain immutable classes only. Instead of putting these classes into the previous list one-by-one, I created a new list for these packages in this PR.

The result is that maybe we have some additional false negatives but all the false positives disappeared.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
baloghadamsoftware: Do you mean the omission of "write" from the list? Actually, I found that methods called `write...()` are usually no setters but they write to a stream.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
baloghadamsoftware: Yes, this can be disputed. If you whish, I can change this part to only omit methods which are private or protected.

candrews:
Here's a (simple) PR that fixes some incorrect class names in `MutableClasses` and adds a missing one (`java.net.Inet6Address`): https://github.com/spotbugs/spotbugs/pull/1654

candrews:
This PR fixes https://github.com/spotbugs/spotbugs/issues/1634

baloghadamsoftware:
This pull request is now split up to #1670, #1671, #1672, #1673, #1674, #1676, #1677 and #1678.

PR #1611, opened 2021-07-09
ci: bump up setup-java to v2

KengoTODA:
This change won't affect the build result. I'll use this PR to confirm the build performance with v1.


KengoTODA:
According to the performance comparison, v2 (13s) is slower than v1 (11s). We have less reason to apply latest ver, so I'll close this PR at this moment.
https://github.com/spotbugs/spotbugs/pull/1610#issue-686691785

PR #1619, opened 2021-07-21
Add option to user-defined list of immutable classes

baloghadamsoftware:
Although our heuristics considers a class as mutable only in the case it has a setter, some users may want to handle some of them as immutable in their code where security is less important than perofrmance. This patch allows them to define a comma-separated list of classes which are considered as immutable. Whole packages may also be whitelisted using an asterix for the class name.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Added
 * New detector `FindBadEndOfStreamCheck` for new bug type `EOS_BAD_END_OF_STREAM_CHECK`. This bug is reported whenever the return value of java.io.FileInputStream.read() or java.io.FileReader.read() is first converted to byte/int and only thereafter checked against -1. (See [SEI CERT rule FIO08-J](https://wiki.sei.cmu.edu/confluence/display/java/FIO08-J.+Distinguish+between+characters+or+bytes+read+from+a+stream+and+-1))
 ### Fixed
-- Class names containing the substring "immutable" (case insensitive) are considered immutable
+- Heuristics to detect mutable classes improved: non-public or static "setters" are no longer considered as setters. Also the return value must be of a primitive type to consider the function as a setter. The list of known immutable classes was also extended.
 
 ## 4.3.0 - 2021-07-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Added
 * New detector `FindBadEndOfStreamCheck` for new bug type `EOS_BAD_END_OF_STREAM_CHECK`. This bug is reported whenever the return value of java.io.FileInputStream.read() or java.io.FileReader.read() is first converted to byte/int and only thereafter checked against -1. (See [SEI CERT rule FIO08-J](https://wiki.sei.cmu.edu/confluence/display/java/FIO08-J.+Distinguish+between+characters+or+bytes+read+from+a+stream+and+-1))
+* New option through environmental variable (spotbugs.immutableclasses or spotbugs_immutableclasses) to specify a list of classes which are considered as immutable. If the class name is an asterix then all classes of the package are considered as immutable.
 ### Fixed
 - Heuristics to detect mutable classes improved: non-public or static "setters" are no longer considered as setters. Also the return value must be of a primitive type to consider the function as a setter. The list of known immutable classes was also extended.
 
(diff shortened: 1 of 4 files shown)


baloghadamsoftware:
This patch is a based on #1603. (GitHub cannot handle dependencies between patches.)

I made it as draft because I am not sure if this is the correct way to allow users customize the behavior of SpotBugs. Probably we should create a new SpotBugs option for this (and maybe also adjust the Maven and the Gradle plugin to let users use this option). SpotBugs plugins such as FindSecBugs use environmental variables and the SpotBugs Maven plugin allows to configure them since version 4.3.0.

The other reason I made it draft is because I am not sure how to write automatic tests for it. (I tested it manually and it works now without any problems.)

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -27,6 +29,8 @@
             "set", "put", "add", "insert", "delete", "remove", "erase", "clear", "push", "pop",
             "enqueue", "dequeue", "write", "append", "replace");
 
+    private static final Matcher IMMUTABLE_CLASS_NAME_MATCHER = Pattern.compile(".*immutable.*", Pattern.CASE_INSENSITIVE).matcher("");
```
pbludov: What about `unmodifiable` like 
https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/configuration/UnmodifiableConfiguration.java
?

spotbugs/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -14,6 +30,21 @@ public void TestKnownImmutable() {
         Assert.assertFalse(MutableClasses.mutableSignature("Ljava/lang/String;"));
     }
 
+    @Test
+    public void TestKnownImmutablePackage() {
+        Assert.assertFalse(MutableClasses.mutableSignature("Ljava/time/LocalTime;"));
+    }
+
+    @Test
+    public void TestNamedImmutable() {
```
ThrawnCA, line 39: ImmutableMap appears to be explicitly defined, not relying on name.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Added
 * New detector `FindBadEndOfStreamCheck` for new bug type `EOS_BAD_END_OF_STREAM_CHECK`. This bug is reported whenever the return value of java.io.FileInputStream.read() or java.io.FileReader.read() is first converted to byte/int and only thereafter checked against -1. (See [SEI CERT rule FIO08-J](https://wiki.sei.cmu.edu/confluence/display/java/FIO08-J.+Distinguish+between+characters+or+bytes+read+from+a+stream+and+-1))
+### Fixed
+- Class names containing the substring "immutable" (case insensitive) are considered immutable
```
ThrawnCA, line 15: I can't see any logic based on "class name contains 'immutable'"; is this still accurate?

ThrawnCA (changes requested):

baloghadamsoftware:
> What about `unmodifiable` like 
https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/configuration/UnmodifiableConfiguration.java
? 

This part is from #1603. Currently we do not check for the string "immutable" in the class name because it was disputed. Instead, we have a fixed list of known immutable classes and a heuristics which looks for a modifier. This particular patch adds an option for the user to extend this list with custom classes and packages.

baloghadamsoftware:
> I can't see any logic based on "class name contains 'immutable'"; is this still accurate? 

Thank you for noticing this. Currently there is no logic for that so I change this sentence.

KengoTODA:
I'm personally not so positive for this feature. Immutability is something able to judge automatically, no need to ask human to provide a list. Human makes mistake, so I prefer automation over this approach.

baloghadamsoftware:
> I'm personally not so positive for this feature. Immutability is something able to judge automatically, no need to ask human to provide a list. Human makes mistake, so I prefer automation over this approach.

Actually, I agree with you. If a code is sensitive, then using this option can make it vulnerable. If it is not, then the best option is to disable the "EI" bug reports (reported by `FindReturnRef`). Maybe we should change the priority of these bugs to "low".

This feature we requested by some of our users (I mean our, as the SpotBugs user community, not users in my company). See #1601.

ThrawnCA:
> I'm personally not so positive for this feature. Immutability is something able to judge automatically, no need to ask human to provide a list. Human makes mistake, so I prefer automation over this approach.

But the point of this detector is to spot cases where an automated check recognises that the class is not actually immutable, when the human's intent was that it _should_ be. Automation can only determine the real behaviour, not the intent.

CHANGELOG.md
JuditKnoll: This conversation can be resolved, since it got fixed in https://github.com/spotbugs/spotbugs/pull/1619/commits/65a43889a16ec6fc0f4c919e4cd4710a0e8d4966.

spotbugs/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
JuditKnoll: This conversation can be resolved, since this testcase got removed in https://github.com/spotbugs/spotbugs/pull/1619/commits/65a43889a16ec6fc0f4c919e4cd4710a0e8d4966, since the checker doesn't check for the string "immutable" in the name.

Issue #1630, opened 2021-08-02
Improving SARIF format for Github integration


h3xstream:
The SARIF format was a nice addition. I think many integrations will use it in the future.
The most common one currently will probably be Github.

Base on few tests, it appears that Github requires The field "text" to be present in **results > message > text**

From the REST webservice, it is giving an obscure error `"error": "expected a result message,expected a result message,expected a result message,[...]"`

From the code scanning section, the error "Analysis processing failed" is given.
![test](https://user-images.githubusercontent.com/593130/127538972-18d8c5d6-9b85-49ec-b65f-562a89b9c302.png)

In summary, Github does not take into account the `ruleId` field. The description need to be in every result apparently.  
It requires `message.text`:
Ref: https://docs.github.com/en/code-security/secure-coding/integrating-with-code-scanning/sarif-support-for-code-scanning#result-object

h3xstream:
I will open a PR with the proposition to fix the missing field.

KengoTODA:
closed by #1631

PR #1631, opened 2021-08-02
Add field "text" to the object message in result (SARIF)

h3xstream:
Change required for Github security events support. See  #1630

---- 
**TODO**


spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Message.java
```diff
@@ -13,17 +13,23 @@
...
  */
 final class Message {
+    @NonNull
+    final String text;
+
     @NonNull
     final List<String> arguments;
 
-    Message(@NonNull List<String> arguments) {
+    Message(@NonNull String text, @NonNull List<String> arguments) {
```
ThrawnCA, line 22: This will affect any other code that might use this class.

h3xstream: I removed the parameter from the constructor. Accessing the field is the only way to set the value.

Some null protection was added and the code calling it was refactored.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Message.java
```diff
@@ -13,22 +13,22 @@
...
+    Message(@NonNull List<String> arguments) {
         this.arguments = Collections.unmodifiableList(Objects.requireNonNull(arguments));
     }
 
     JsonObject toJsonObject() {
         JsonObject jsonObject = new JsonObject();
         jsonObject.addProperty("id", "default");
 
-        jsonObject.addProperty("text", text);
+        if(text != null) {
```
ThrawnCA, line 29: Should be a space between `if` and bracket.

ThrawnCA (dismissed):

KengoTODA (changes requested):
Please check the feedback from ThrawnCA, I think we can fix it automatic by `./gradlew spotlessApply`.

Then merge the latest `origin/master` to your topic branch, and resolve a conflict in the `CHANGELOG.md` file.

h3xstream:
Everything should be there.
Sorry for the delay

ThrawnCA (dismissed):

h3xstream:
@KengoTODA  I've merge the CHANGELOG file.

Let me know if something more is needed.

ThrawnCA (approved):

h3xstream:
@KengoTODA Let me know if anything is missing.

KengoTODA (approved):

PR #1639, opened 2021-08-05
Fix #1477 by removing invalid NO_BACKGROUND style

### LINKED ISSUE #1477: BugInfoView shouldn't use NO_BACKGROUND as style for Browser (regression in Eclipse 2021-03)
The `BugInfoView` in the Eclipse plug-in sets the `SWT.NO_BACKGROUND` style on its `Browser` instance, but the documentation for the Composite class says that the use of this style bit is undefined outside `Canvas` subclasses. This use has led to a regression when using the SpotBugs Eclipse plug-in in Eclipse 2021-03. The problem is that `SWT.NO_BACKGROUND` overlaps with the new constant `SWT.EDGE` for using the experimental WebView2 browser component on Windows (see [Bug 572208](https://bugs.eclipse.org/bugs/show_bug.cgi?id=572208) for SWT), and if that component is not available you get an exception:

```
org.eclipse.swt.SWTError: Not implemented [WebView2 runtime not found]
	at org.eclipse.swt.SWT.error(SWT.java:4915)
	at org.eclipse.swt.browser.Edge.createEnvironment(Edge.java:154)
	at org.eclipse.swt.browser.Edge.create(Edge.java:180)
	at org.eclipse.swt.browser.Browser.<init>(Browser.java:99)
	at de.tobject.findbugs.view.BugInfoView.createBrowser(BugInfoView.java:189)
```

It seems like the fix would be simple: use `SWT.NONE` when creating the `Browser` in the `BugInfoView`.

msternermxy:
Fixes #1477 by removing the invalid NO_BACKGROUND style for the Browser instance in the Bug Info view, to avoid inadvertently triggering EDGE browser mode.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fixed False positives for RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE ([#600](https://github.com/spotbugs/spotbugs/issues/600) and [#1338](https://github.com/spotbugs/spotbugs/issues/1338))
 - Inconsistent bug description on `EQ_COMPARING_CLASS_NAMES` ([#1523](https://github.com/spotbugs/spotbugs/issues/1523))
 - Add a declaration of charset encoding in generated reports ([#1623](https://github.com/spotbugs/spotbugs/pull/1623))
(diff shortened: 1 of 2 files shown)


iloveeclipse (approved):
Thanks.

ThrawnCA (approved):

Issue #1454, opened 2021-03-04
List of visitors that can be omitted


Meijuh:
I am looking for a list of visitors/detectors that I can omit. Specifically, I am look for the allowed values for https://spotbugs.github.io/spotbugs-maven-plugin/spotbugs-mojo.html#omitVisitors. 
Previously, in SpotBugs 3.1, this list could be found under "detectors" at e.g. https://spotbugs-in-kengo-toda.readthedocs.io/en/lqc-list-detectors/detectors.html#standard-detectors, but that page/menu does not exist anymore since SpotBugs 4 at https://spotbugs.readthedocs.io/en/stable/index.html.

KengoTODA:
@Meijuh The page is generated from https://github.com/lqc/spotbugs/tree/list-detectors that isn't merged to the SpotBugs master branch.
If you need, consider creating a PR based on that branch.

SumanohariR:
Could you specify the issue to be worked on, I am interested to work on this.

delanym:
This is an essential feature especially when extending. Spotbug analysis is expensive and I only want visitors from findsecbugs-plugin analyzed.

delanym:
The output of spotbugs:check is identifiers like EXS_EXCEPTION_SOFTENING_NO_CONSTRAINTS which don't help here at all.

sdavids:
Right now, one cannot easily use the [visitor CLI configuration options](https://spotbugs.readthedocs.io/en/latest/running.html#detector-visitor-configuration-options).

An example report:

```
AI	Method Example.example() that can return null, is missing a @Nullable annotation
[Bug type AI_ANNOTATION_ISSUES_NEEDS_NULLABLE (click for details)]
```

I want to easily get from `AI_ANNOTATION_ISSUES_NEEDS_NULLABLE` to the corresponding visitor, so I can add `-omitVisitors <whatever corresponds to AI_ANNOTATION_ISSUES_NEEDS_NULLABLE>`.

I am aware, that I could use `-exclude` with

```xml
<FindBugsFilter ...>
  <Match>
    <Class name="Example"/>
    <Bug pattern="SIC_INNER_SHOULD_BE_STATIC_ANON"/>
  </Match>
</FindBugsFilter>
```

instead.

JuditKnoll:
As far as I can see this issue was fixed by https://github.com/spotbugs/spotbugs/pull/1651.

PR #1651, opened 2021-08-19
Add detectors documentation

friscoMad:
Adds detector list (maybe it should be renamed to Visitors) in the docs.
Solves #1454 

Heavily based on https://github.com/lqc/spotbugs/commit/89fc59b7c1482ea3e1c6e690e0de391c12289655#

I am not sure how PO files are generated, but a lot of sphinx code is picked as text to translate I am not sure if that is something that can be fixed easily, I couldn't find any other link between different pages to use as an inspiration.

----


ThrawnCA (dismissed):

docs/extensions/generate_bug_description.py
```diff
@@ -49,5 +53,39 @@ def generate_bug_description(language):
...
+                reports = element.get("reports", "").split(",")
+
+                page = disabled_detector_page if disabled else enabled_detector_page
+
+                message = messages.find(".//Detector[@class='%s']" % klass)
+                short_name = klass.split(".")[-1]
+
+                page.write("%s\n%s\n\n" % (short_name, '^' * len(short_name)))
+
+                details = message.findtext('.//Details')
```
KengoTODA, line 74: This code throws an error like below if no translation is provided in `messages_ja.xml`:

```
Extension error (generate_bug_description):
Handler <function setup.<locals>.<lambda> at 0x7fc68a3d1550> for event 'builder-inited' threw an exception (exception: 'NoneType' object has no attribute 'findtext')
```

You can reproduce this error by following commands:

```sh
cd docs
docker build -t spotbugs-sphinx .
rm -rf .build
docker run -it -v $(pwd)/..:/docs spotbugs-sphinx make -e SPHINXOPTS="-D language='ja'" html
```

If you're using Windows, replace `$(pwd)/..` with fullpath of your SpotBugs root dir.

If possible, use the message in English (`messages.xml`) in case of no translation provided.
If it's difficult, it's OK to tell 'ja translation is not provided for foobar detector'.

docs/extensions/generate_bug_description.py
```diff
@@ -49,5 +53,39 @@ def generate_bug_description(language):
...
 
+    with codecs.open('generated/detectorListEnabled.inc', 'w', encoding='UTF-8') as enabled_detector_page:
+        with codecs.open('generated/detectorListDisabled.inc', 'w', encoding='UTF-8') as disabled_detector_page:
+            for element in findbugs.iterfind(".//Detector"):
+                hidden = parse_bool_attr(element, "hidden")
+                if hidden:
+                    continue
+                klass = element.get("class")
+                disabled = parse_bool_attr(element, "disabled")
+                speed = element.get("speed")
```
KengoTODA, line 64: The `speed` attribute is already deprecated, could you consider removing it from this document?

https://github.com/spotbugs/spotbugs/blob/51e586bed98393e53559a38c1f9bd15f54514efa/spotbugs/src/main/java/edu/umd/cs/findbugs/DetectorFactory.java#L125-L129

KengoTODA (changes requested):
Thanks for your PR!
There are two comments in my review, please check.

I also found many noisy comments in these `.po` files, could you revert or remove them?

docs/extensions/generate_bug_description.py
friscoMad: More or less fixed, I am using the English as a fallback but still, the process does raise a lot of warnings as sphinx complains about headings, several Japanese characters use 2 positions to print and the underlines are too short.
That problem is totally out of my capacity, I am a bit rusty in Python, and Unicode problems are a serious thing.

docs/extensions/generate_bug_description.py
friscoMad: Removed

friscoMad:
> Thanks for your PR!
> There are two comments in my review, please check.
> 
> I also found many noisy comments in these `.po` files, could you revert or remove them?

I think I have cleaned up all `#:~` in `bugDescription.po` file I hope you were referring to that.

KengoTODA (approved):

PR #1654, opened 2021-08-20
fix: MutableClasses: net classes are in java.net

### LINKED ISSUE #1653: Accessors of type java.net.URI fields being flagged as EI
Spotbugs 4.4.0 is flagging getters of type `java.net.URI` fields with EI_EXPOSE_REP and setters with EI_EXPOSE_REP2 indicating that these methods expose mutable application state. However [`java.net.URI` is immutable](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html), so I believe this finding is in error.

Here's a minimal example of the issue.
```java
import java.net.URI;

public class UriBugTest {
    public static final java.net.URI DEFAULT_URI= URI.create("https://github.com");
    private java.net.URI uri = DEFAULT_URI;

    public URI getUri() {
        // flagged EI_EXPOSE_REP
        return this.uri;
    }

    public void setUri(URI value) {
        // flagged EI_EXPOSE_REP2
        this.uri = value;
    }
}
```

This issue is related to https://github.com/spotbugs/spotbugs/issues/1634

candrews:
URI, Inet4Address, and InetSocketAddress are in the java.net package, not java.util

Fixes #1653

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2021-??-??
 
+### Fixed
+- Fixed immutable classes in java.net.* as being flagged as EI ([#1653](https://github.com/spotbugs/spotbugs/issues/1653)
+
 ## 4.4.0 - 2021-08-12
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ThrawnCA (approved):

KengoTODA (changes requested):
Please add an entry into `CHANGELOG.md`. We use https://keepachangelog.com/en/1.0.0/ so what you need is a `Fixed` type entry.

candrews:
> Please add an entry into `CHANGELOG.md`. We use https://keepachangelog.com/en/1.0.0/ so what you need is a `Fixed` type entry.

I've added a commit to this PR that updates `CHANGELOG.md`

Thank you for the review!

KengoTODA (approved):

PR #1671, opened 2021-08-30
Exclude private methods from setter-like methods when considering a class as immutable

baloghadamsoftware:
Private methods must not be considered as setters, even if their name is setter-like.

----


ThrawnCA:
Why would `protected` methods be excluded? If subclasses are allowed to alter an aspect of state, that doesn't sound immutable.

baloghadamsoftware:
> 
> 
> Why would `protected` methods be excluded? If subclasses are allowed to alter an aspect of state, that doesn't sound immutable.

Good point! There are two sides here: if a class has only `protected` setters then instances of this class are immutable. However, descendants of the class may be mutable. On the other hand, if we have a private attribute which is statically typed by such a class and return it in a public function then due to polymorphism we can actually store an instance of a mutable subclass in this attribute and thus return a reference to a private mutable.

ThrawnCA (dismissed):

KengoTODA (changes requested):
Even if the method is `private` it's possible to be a setter method. Here is example:

```java
public class Outer {
    static class Inner {
        private int field;
        private void setField(int field) {
            this.field = field;
        }
    }
    public void foo() {
        var inner = new Inner();
        inner.setField(0);
    }
}
```

So I believe current implementation is OK... while thinking that the responsibility of this method isn't so clear.

baloghadamsoftware:
> Even if the method is `private` it's possible to be a setter method. Here is example:

Thank you! You are right. In case of nested classes there can be a problem. Now I excluded them, private functions are only disregarded if the class is not a nested class.

ThrawnCA (changes requested):
Needs conflict resolved before this can be reviewed.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -126,7 +126,8 @@ private boolean computeMutable() {
             }
 
             for (Method method : cls.getMethods()) {
-                if (!method.isStatic() && looksLikeASetter(method)) {
+                if (!method.isStatic() && (!method.isPrivate() || cls.isNested())
```
rovarga, line 129: `isNested()` is not sufficient, as private methods are accessible also from the host, hence this check is not sufficient.

If I understand correctly, this heuristic is used to examine the class from the user's perspective.

If that is really the case, the user does not see intra-nest organization nor private methods and therefore we should not concern ourselves with private details.

rovarga (changes requested):

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
baloghadamsoftware: Sorry, I do not fully understand you. In this logical expression we are trying to determine whether the method is a setter. If it is static, then it is no setter. If it is private, it is no setter, except if the class is nested, because as you write, private methods are acessible from the host. Thus in case of nested classes private is not enough to say that this method is not a setter.

rovarga: My question is whether the check catches all everything you expect it to in this scenario:

```
public class Outer {
    private class Inner {
        private boolean fresh;
        
        private void clear() {
            fresh = true;
            cleared(this);
        }
    }

    private List<Inner> inners;
    private int counter;
    
    private void clear() {
        inners.forEach(Inner::clear);
    }
    
    private void cleared(Inner inner) {
        counter++;
    }
}
```

I have a separate argument that we should just ignore all private methods, because it is so easy to hide this mutability from us:
```
public class Outer {
    private class Inner {
        private boolean fresh;
    }

    private List<Inner> inners;
    private int counter;
    
    private void clear() {
        inners.forEach(inner -> {
            inner.fresh = true;
            counter++;
        });
    }
}
```

baloghadamsoftware:
@KengoTODA?

ThrawnCA (changes requested):
Conflict needs to be resolved.

ThrawnCA (approved):

PR #1672, opened 2021-08-31
Exclude methods throwing `UnsupportedOperationException` from setter-like methods when considering a class as immutable

baloghadamsoftware:
Methods throwing `UnsupportedOperationException` must not be considered as setters, even if their name is setter-like.

----


ThrawnCA (dismissed):

spotbugs/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -54,6 +54,10 @@ public int getN() {
         public Immutable setN(int n) {
             return new Immutable(n);
         }
+
+        public void setNUnsupported(int n) throws UnsupportedOperationException {
```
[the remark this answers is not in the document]
baloghadamsoftware: No, because we cannot check here the body of the function. IMO UnsupportedOperationException being thrown is better to be declared if the whole function call is not supported in this class.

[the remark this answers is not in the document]
pzygielo: Can method declare to throw `UnsupportedOperationException` and modify state anyway? And even not throw anything at all?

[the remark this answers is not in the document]
ThrawnCA: In the end, if code deliberately tries to mask bad behaviour, it will probably succeed.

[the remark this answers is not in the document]
baloghadamsoftware: Yes, this is just a heuristics to find immutable classes. If you declare a function as a thrower but modify the state then this is a setter and the class is not immutable. The heuristics cannot be 100% accurate.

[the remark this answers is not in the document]
pzygielo: Can it lead to false-negatives?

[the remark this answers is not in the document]
baloghadamsoftware: Of course. As I mentioned, this is a heuristics. Unfortunately, unlike C++'s const member functions, Java has no means that forbids a method to modify the attributes of the class instance. Instead, we try to find methods which look like setters. If we do not find any then we consider the class for immutable. However, if the methods are named in a foreign language, or the setters are ill named (like eg. foo()), then we consider the class as immutable which leads to false negatives.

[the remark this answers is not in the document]
pzygielo: I see, thank you  for your explanation.

ThrawnCA (changes requested):
Needs conflict resolved before this can be properly reviewed.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -137,6 +138,14 @@ private boolean computeMutable() {
 
         private boolean looksLikeASetter(Method method) {
             final String methodName = method.getName();
+            // If the method throws an UnsupportedOperationException then we ignore it.
+            ExceptionTable exceptions = method.getExceptionTable();
+            if (exceptions != null) {
```
ThrawnCA, line 143: These two `if` conditions could be combined.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -147,20 +156,55 @@ private boolean looksLikeASetter(Method method) {
             return false;
         }
 
-        // If the method throws an UnsupportedOperationException then we ignore it.
-        ExceptionTable exceptions = method.getExceptionTable();
-        if (exceptions != null) {
-            if (Arrays.asList(exceptions.getExceptionNames()).contains("java.lang.UnsupportedOperationException")) {
-                return false;
+        private String getSig() {
+            String local = sig;
```
ThrawnCA: Rather than the possible `null` assignment followed by double assignment on one line, how about doing a `null` check on `sig`, populating it, then assigning `local = sig` afterward?

Actually, why does `local` exist? We're populating `sig` anyway...

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -147,20 +156,55 @@ private boolean looksLikeASetter(Method method) {
...
+            if (local == null) {
+                sig = local = "L" + cls.getClassName().replace('.', '/') + ";";
+            }
+            return local;
+        }
+
+        private boolean isImmutableByContract() {
+            Boolean local = immutableByContract;
```
ThrawnCA: I don't see why the local variable is being used when we're populating the instance field anyway.

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
baloghadamsoftware: Hmm, something is wrong then in GitHub, because you commented an earlier change here. (Not even by me, but by @rovarga, already merged.)

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
baloghadamsoftware: Ditto.

ThrawnCA (dismissed):

KengoTODA (dismissed):
@ThrawnCA I think this change isn't so valuable and better to keep our implementation simple. No matter how methods are implemented, just check [fields and the class are `final` or not](https://docs.oracle.com/javase/tutorial/essential/concurrency/imstrat.html) then it should be enough. I know that the current SpotBugs implementation is not like what I mean, but it's better to stop making it far from the ideal state.

And even if this change works, `UnsupportedOperationException` is an unchecked exception so `ExceptionTable` could have no record for the exception. So I think this PR is not reasonable nor practical.

It's just my personal opinion. 🙇‍♂️  
How do you think? Is it still better to merge this PR? Thanks in advance!

baloghadamsoftware:
You are right, `UnsupportedOperationException` is an unchecked exception. This means that we should not look at the exception specification but the body of the function: if it consists of a single throw of this exception then it is definitely no setter. This is the way the Guava classes work, because they extend `java.util.AbstractCollection` which already has setters such as `add()` and `remove()`. The question is how can we check the body statically? We must somehow compare the body to the byte code of single line `throw new UnsuppoertedOperationException`.

Although Guava classes are considered as immutable explicitly, we cannot be sure that there are no similar classes in a user's code. Thus I think that this PR is useful to reduce the number of false positives but not in its current format but after rewriting it to check the body of the function instead of its exception specification.

ThrawnCA (approved):

jochenberger:
IIUC, this can be unblocked, i.e. the requested changes are no longer required.

baloghadamsoftware:
@KengoTODA?

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
JuditKnoll: This conversation can be resolved, since the if conditions got combined in https://github.com/spotbugs/spotbugs/pull/1672/commits/d30be689a17cced410cdc366aec91d5f706c943b.

ThrawnCA (approved):

JuditKnoll:
As far as I can see, Kengo Toda's requests (https://github.com/spotbugs/spotbugs/pull/1672#pullrequestreview-793825557) got addressed: when looking for the thrown exceptions, the code is checked, not the exception table / the throws declaration.
This PR only removes some fps and decreases the noisiness, so I think it could be merged. @gtoison @hazendaz What do you think?

gtoison:
I understand Kengo's sentiment that a pure immutability would be better, but that ship has sailed: we're using heuristics to detect immutability.
Maybe in the future we'll be able to do it formally, but that won't happen anytime soo.
This PR provides an improvement and looks good to me too.

gtoison (approved):

pzygielo:
To clarify changelog entry:
_Classes containing only methods throwing `UnsupportedOperationException` with setter-like names are no longer considered as mutable_

Is it necessary for such methods to **declare** that `UnsupportedOperationException` (which is `RuntimeException`) is thrown to no longer be considered as mutable? Will the actually mutating method be falsely considered as not mutable if it declares `UnsupportedOperationException` to be thrown?

gtoison:
My understanding of the code is that it looks for methods doing `throw new UnsupportedOperationException(...)`., the code does not look for methods _declaring_ that they throw `UnsupportedOperationException`

pzygielo:
> My understanding of the code is that it looks for methods doing `throw new UnsupportedOperationException(...)`., the code does not look for methods _declaring_ that they throw `UnsupportedOperationException`

Thanks for checking. Asked, because IIRC originally in 2021 it did :smile:

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -22,6 +33,22 @@ void set(int n) {
 }
 
 class MutableClassesTest {
+    @BeforeEach
+    void setUp() {
+        IAnalysisCache analysisCache = ClassFactory.instance().createAnalysisCache(new ClassPathImpl(), new PrintingBugReporter());
+        Global.setAnalysisCacheForCurrentThread(analysisCache);
+        FindBugs2.registerBuiltInAnalysisEngines(analysisCache);
```
JuditKnoll, line 48: Cherrypicked the changes from https://github.com/spotbugs/spotbugs/pull/3666 but added back the changes in the `build.gradle`, otherwise the build was failing. Thanks @gtoison

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -22,6 +33,22 @@ void set(int n) {
...
+
+        Project project = new Project();
+        AnalysisContext analysisContext = new AnalysisContext(project);
+        AnalysisContext.setCurrentAnalysisContext(analysisContext);
```
JuditKnoll, line 52: This is necessary, since - if the XMethod is resolved - the `XMethod.isUnsupported()` function is used to decide whether the method throws an UnsupportedOperationException unconditionally. For XMethod, XFactory and an AnalysisContext is needed, otherwise IllegalStateException is thrown [here](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/AnalysisContext.java#L123).

I think it can only happen when running `MutableClassesTest`, not when running a detector, that `XMethod` is unresolved - making `MutableClassesTest.testMutable` and `testKnownMutable` fail - , for this case case I added back @baloghadamsoftware 's implementation of `isSingleLineUnsupportedOperationThrower(Method)`.

The `XMethod.isUnsupported()` is preferred to `MutableClasses.isSingleLineUnsupportedOperationThrower(Method)`, because it's slightly broader and it's already calculated.

If anyone has a good idea how to make `XMethod` resolved with `MutableClassesTest`, then please share it.

JuditKnoll (commented):
@hazendaz Can you please dismiss Kengo Toda's stale review? So this change could be merged.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
gtoison: I'm not sure if that works, I thought the build was running fine on my computer but it failed on the github workers, and then it failed on my computer. Maybe it never worked or maybe it's a bit random, hence the "unstable" test name.

JuditKnoll: When I ran `MutableClassesTest` separately, it was successful on my side - both in IDE and cli - , but when I run all the tests with `MutableClassesTest` inside the tests and not in unstabletest, it failed. So, yes, it seemed unstable to me as well, this is why I reverted the changes to `build.gradle`.
The other changes to `MutableClassesTest` seems to be necessary when creating XMethod with XFactory, see my comments to line 42 to 44.

JuditKnoll: Setting the analysisCache is necessary when calling `edu.umd.cs.findbugs.ba.XFactory.createXMethod()`.
Without it `testImmutable`, `testKnownMutable` and `testsMutable` tests of `MutableClassesTest` throw the following exception:
```
Cannot invoke "edu.umd.cs.findbugs.classfile.IAnalysisCache.getAnalysisLocals()" because the return value of "edu.umd.cs.findbugs.classfile.Global.getAnalysisCache()" is null
java.lang.NullPointerException: Cannot invoke "edu.umd.cs.findbugs.classfile.IAnalysisCache.getAnalysisLocals()" because the return value of "edu.umd.cs.findbugs.classfile.Global.getAnalysisCache()" is null
	at edu.umd.cs.findbugs.AnalysisLocal.getMap(AnalysisLocal.java:39)
	at edu.umd.cs.findbugs.AnalysisLocal.get(AnalysisLocal.java:44)
	at edu.umd.cs.findbugs.ba.AnalysisContext.currentXFactory(AnalysisContext.java:208)
	at edu.umd.cs.findbugs.ba.XFactory.createXMethod(XFactory.java:338)
	at edu.umd.cs.findbugs.ba.XFactory.createXMethod(XFactory.java:334)
	at edu.umd.cs.findbugs.ba.XFactory.createXMethod(XFactory.java:277)
	at edu.umd.cs.findbugs.ba.XFactory.createXMethod(XFactory.java:270)
	at edu.umd.cs.findbugs.util.MutableClasses.looksLikeASetter(MutableClasses.java:170)
	at edu.umd.cs.findbugs.util.MutableClasses$ClassAnalysis.looksLikeASetter(MutableClasses.java:287)
	at edu.umd.cs.findbugs.util.MutableClasses$ClassAnalysis.computeMutable(MutableClasses.java:277)
	at edu.umd.cs.findbugs.util.MutableClasses$ClassAnalysis.isMutable(MutableClasses.java:264)
	at edu.umd.cs.findbugs.util.MutableClasses.mutableSignature(MutableClasses.java:118)
	at edu.umd.cs.findbugs.util.MutableClassesTest.testImmutable(MutableClassesTest.java:148)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

ThrawnCA (approved):

ThrawnCA (approved):

PR #1673, opened 2021-08-31
Disregard methods beginning with `write` when checking for mutability of a class

baloghadamsoftware:
Methods beginning with `write` are almost always output methods (e.g. to streams) and no setters. Disregard them when deciding about mutability of a class.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 - Fixed immutable classes in java.net.* as being flagged as EI ([#1653](https://github.com/spotbugs/spotbugs/issues/1653)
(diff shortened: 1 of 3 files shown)


ThrawnCA:
IMO writing to a stream qualifies as changing its state. Especially with classes like `ByteArrayOutputStream`.

baloghadamsoftware:
> 
> 
> IMO writing to a stream qualifies as changing its state. Especially with classes like `ByteArrayOutputStream`.

You are wright. However, if a non-stream class contains a method beginning with `write` it usually does not change the class but writes its contents into a stream.

rovarga:
I think the signal-to-noise ratio for `write*(...)` is way too high. In its current form it trips on anything java.io.Externalizable, as well as anything using writeReplace() to implement Serializable Proxy pattern -- suppressing that is quite a bit of work, too, see #1704 .

ThrawnCA:
This has a lot more in it than just ignoring 'write' methods...

ThrawnCA (approved):

PR #1703, opened 2021-09-14
Recognize OpenDaylight's Immutable interface contract

### LINKED ISSUE #1697: Recognize interface-defined immutability
During migration to spotbugs-4.4.1 in OpenDaylight we are noticing a number of new EI_EXPOSE_REP and MS_EXPOSE_REP warnings.
A large number of these are reported against types which are known to be immutable as part of their interface contract. This contract is not driven by annotations, but rather by implementing two dedicated interfaces,
[Mutable](https://www.javadoc.io/doc/org.opendaylight.yangtools/yangtools-docs/latest/org/opendaylight/yangtools/concepts/Mutable.html) and [Immutable](https://www.javadoc.io/static/org.opendaylight.yangtools/yangtools-docs/7.0.7/org/opendaylight/yangtools/concepts/Immutable.html). These cannot be implemented simultaneously and can therefore be used to specify API expectations, such as (made up)
```
  interface ImmutableMapBuilder<K extends Immutable, V extends Immutable> extends Mutable {
    ImmutableMapBuilder<K, V> put(K key, V value);
  }
```
as well as runtime decisions without going to `java.lang.reflect`:
```
  public void logAsync(Object obj) {
    final Object state;
    if (obj instanceof Immutable) {
      // Can safely defer toString()
      state = obj;
    } else {
      state = obj.toString();
    }
    enqueueToLog(state);
}
```

Since `findbugs.util.MutableClasses` also reacts to any annotation named Immutable being attached to a class, it would be beneficial if it would also recognize `/Immutable;` being extended or implemented in a type's hierarchy.

rovarga:
Aside from recognizing annotations, also treat any class that implements
an interface named Immutable as an immutable class. Fixes #1697.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Wrong description of the `SE_TRANSIENT_FIELD_OF_NONSERIALIZABLE_CLASS` ([#1664](https://github.com/spotbugs/spotbugs/pull/1664))
 - Treat types with `@com.google.errorprone.annotations.Immutable` as immutable ([#1705](https://github.com/spotbugs/spotbugs/pull/1705))
 - Fix annotation check for `jdk.internal.ValueBased` ([#1706](https://github.com/spotbugs/spotbugs/pull/1706))
+- Treat types which implement `org.opendaylight.yangtools.concepts.Immutable` as immutable ([#1697](https://github.com/spotbugs/spotbugs/issues/1697))
 
 ## 4.4.1 - 2021-09-07
 ### Changed
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -98,6 +98,7 @@ dependencies {
   testImplementation 'org.apache.ant:ant:1.10.11'
   testImplementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.14.0'
   testImplementation 'com.google.errorprone:error_prone_annotations:2.9.0'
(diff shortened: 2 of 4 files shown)


rovarga:
This will be reworked on top of #1704

rovarga:
This builds on top of #1705

ThrawnCA (dismissed):

ThrawnCA:
Please [avoid force-pushing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing#the-golden-rule-of-rebasing); it wrecks the history.

KengoTODA (commented):
The marker interface has no meaning in the bytecode level; annotated classes still be able to be mutable. Then I suggest not to merge this change but enhance the mutability check algorithm to reduce false positive and false negative.

rovarga:
> Please [avoid force-pushing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing#the-golden-rule-of-rebasing); it wrecks the history.

Ack

rovarga:
> The marker interface has no meaning in the bytecode level; annotated classes still be able to be mutable. Then I suggest not to merge this change but enhance the mutability check algorithm to reduce false positive and false negative.

That depends on what the expectations on MutableClasses are from three perspectives:
- perfomance, i.e. the depth of reasoning we make about the JavaClass
- accuracy of reported boolean
- what assumptions are made about the JavaClass: is it any reference we see? is it something we can load? is it a class that is part of the project being analyzed or is it on classpath? have we already performed bytecode/dataflow analysis on it to conclusively prove it is mutable (or immutable)?

At the end of the day, we are trusting `@Immutable` annotations here and this just uses a different aspect of what shape the class has.

PR #1704, opened 2021-09-14
Recognize Serializable's writeReplace()

rovarga:
java.io.Serializable defines semantics of writeReplace() method, it
should not be confused as a mutable method.

Introduce a helper class to encapsulate analysis logic, as there are a
few other cases which we'll need to cover.

----


rovarga:
#1673 is a better solution and if not, #1705 and #1703 provide the infra introduced here -- building on top of those is better.

PR #1727, opened 2021-09-27
More immutable classes.

victorwss:
I added some classes to the known immutable set.

- `java.math.Decimal` does not exists (should be `BigDecimal`). And the package `java.math` already covers it, so I just removed it along with `BigInteger`.
- Although the reflection objects can be altered with the `setAccessible(boolean)`, there is already other checks covering that corner-case, so for every other case, they are immutable. Also, the side-effect of using that method is likely the intended one, and not the accidental side-effect that immutability seeks to prevent.
- <strike>I am not sure if the `java.lang.Module` and `java.lang.ModuleLayer` should be considered immutable, but I left them out.</strike>
- <strike>`java.lang.Thread` is effectively immutable. Stopping or suspending a thread is deprecated and was even removed from the newest JDKs. Only `Thread.interrupt()` remains to interfere with a Thread, which is already covered by other checks and is a very unlikely to be a way to attack mutability.</strike> 
- The other classes that I added should be ok.

Edit:
- Sadly, `Optional<T>` is immutable if, and only if, `T` is immutable. But at this moment, it is too naive to know that.
- Although, `java.lang.Module` and `java.lang.ModuleLayer` aren't 100% immutable, it is extremely unlikely that mutating them causes undesirable side-effects (most likely, only the intended ones). So, I added them to the list.

Edit 2:
- Interfaces in `java.util.function` and annotations are immutable. However, sadly, for `Thread`, there are some stuff there that are too mutable to be safe.
- If a `Class` is fair to be considered immutable, so is the `ClassLoader`.


spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -86,7 +99,7 @@ public static boolean mutableSignature(String sig) {
     private static final class ClassAnalysis {
         private static final List<String> SETTER_LIKE_PREFIXES = Arrays.asList(
                 "set", "put", "add", "insert", "delete", "remove", "erase", "clear", "push", "pop",
-                "enqueue", "dequeue", "write", "append", "replace");
+                "enqueue", "dequeue", "write", "append", "replace", "prepend", "reset", "update");
```
KengoTODA, line 102: I'm personally not so positive to edit these prefixes. Name methods and mutability should have no relation.
Then it is better to keep the current implementation as is, to keep backward compatibility and [bring no surprise to users](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

KengoTODA (changes requested):

gtoison:
Superseded by #3140 
Thank you for the contribution!

PR #1732, opened 2021-10-01
fix resource leak due to Files.list

lujiefsi:


spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/JrtfsCodeBase.java
```diff
@@ -90,9 +91,10 @@ public JrtfsCodeBase(ICodeBaseLocator codeBaseLocator, @Nonnull String fileName)
     public Map<String, Object> createPackageToModuleMap(FileSystem fs) throws IOException {
         HashMap<String, Object> packageToModule = new LinkedHashMap<>();
         Path path = fs.getPath("packages");
-        Files.list(path).forEach(p -> {
-            try {
-                Iterator<Path> modIter = Files.list(p).iterator();
+        try(Stream<Path> packList = Files.list(path)) {
+         packList.forEach(p -> {
+            try(try(Stream<Path> pList = Files.list(p)) {
```
KengoTODA, line 96: Thanks for your PR, and nice catch! 👍 

For me `try try` is hard to understand. Here I simply mean that the code is too complex.  
The fix itself is very necessary, could you consider making the code simpler? Of course, it's welcome to refactor the code.

KengoTODA (changes requested):

PR #1740, opened 2021-10-05
Add detector for catching NullPointerExceptions

gamesh411:
----


gamesh411:
Hi,
I plan on making this detector handle the **ancestors** of `NullPointerException` as well (as per the SEICert guideline), but in my prototype, where I handle those as well, there are a lot of false positives thanks to `try-with-resources` language structures as those are compiled to exception handling in byte code catching `Throwable` instances.
In order to avoid those false positives I would need a minor refactoring regarding the `DominatorsAnalysis` API, which I will submit as a separate PR. Once I have that in place, I can extend this detector with the ancestor handling.

gamesh411:
Thanks for the quick response! I have applied the suggestions as best I could. I will ask it here as well, do you prefer the approach where I add every amendment in separate commits (and squash at the end), or should I just rebase every time when it is necessary?

PR #1746, opened 2021-10-08
Add check-project workflow

gamesh411:
Create a workflow for analyzing Java projects via PR versions of
Spotbugs. Baseline diff and time measurement features are added to
monitor the impact of PRs (result set changes or possible regressions).
Open-source projects spotbugs, matsim and jenkins are added as test
subjects.

----

### DIFF
--- a/.github/workflows/check-projects-comment.yml
+++ b/.github/workflows/check-projects-comment.yml
@@ -1,0 +1,69 @@
+name: check-projects-comment
+
+on:
+  workflow_run:
+    workflows: ['check-projects']
+    types:
+      - completed
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: 'Download artifact'
+        uses: actions/github-script@v3
+        with:
+          script: |
+            const fs = require('fs')
+            const artifactsQueryResult = await github.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: ${{ github.event.workflow_run.id }},
+            })
+            const artifacts = artifactsQueryResult.data.artifacts
+            for (const artifact of artifacts) {
+              const name = artifact.name
+              const download = await github.actions.downloadArtifact({
+                 owner: context.repo.owner,
+                 repo: context.repo.repo,
+                 artifact_id: artifact.id,
+                 archive_format: 'zip',
+              })
+              fs.writeFileSync(`${{github.workspace}}/${name}.zip`, Buffer.from(download.data))
+            }
+
+      - run: |
+          for archive in *.zip; do
+            unzip "${archive}" -d "${archive%.*}" && rm "${archive}"
+          done
+
+      - name: 'Comment on PR'
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const fs = require('fs')
+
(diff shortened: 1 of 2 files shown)


.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,329 @@
...
+        run: ./gradlew installDist --no-daemon
+
+      - name: Quick-build baseline Spotbugs
+        run: ./gradlew installDist --no-daemon
+        working-directory: spotbugs-baseline
+
+      - name: Cache analyzed Spotbugs
+        uses: actions/cache@v2
+        with:
+          path: spotbugs-4.4.1.jar
```
JuditKnoll, line 89: Could you move the version infos to variables, so it's easier to modify them in the future? (Not only for spotbugs, but the other projects as well.)

.github/workflows/check-projects-comment.yml
```diff
@@ -0,0 +1,69 @@
...
+
+            for (const projectDir of fs.readdirSync('${{github.workspace}}')) {
+              const projectName = projectDir
+              const issueNumber = Number(fs.readFileSync(`${projectDir}/prnum`))
+              const baselineTime = Number(fs.readFileSync(`${projectDir}/baseline_time`))
+              const newTime = Number(fs.readFileSync(`${projectDir}/new_time`))
+              const resultDiff = fs.readFileSync(`${projectDir}/result_diff`)
+
+              const body =
+              `${projectName}
```
JuditKnoll, line 58: IMO it would be easier to understand if the comments contained whole sentences, instead of only the bare minimal info. Also, it would be a bit more informative, if it contained not only the name of the tested project, but a link to the exact source and some version info (or commit hash).
This line could be e.g. `Testing/Running this PR on ${projectName} (${projectUrl}) version ${projectVersion}.`.

.github/workflows/check-projects-comment.yml
```diff
@@ -0,0 +1,69 @@
...
+                 baseline time: ${baselineTime}s
+                 new time: ${newTime}s
+                 results diff: ${resultDiff ? '\n```' + resultDiff + '```' : '-'}
```
JuditKnoll, line 61: If there are no differences, instead of an empty codeblock, there could a `No differences were found.` or `The results are identical.` or some other similar string.

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,329 @@
...
+    branches:
+      - master
+
+jobs:
+  check-spotbugs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v2
```
ThrawnCA, line 45: This version is old.

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,329 @@
...
+          path: spotbugs-baseline
+          persist-credentials: false
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v1
+        with:
+          java-version: '17.0.x'
+
+      - name: Cache Gradle
+        uses: actions/cache@v2
```
ThrawnCA, line 62: This version is old.

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,329 @@
...
+
+      - name: Checkout Baseline
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          path: spotbugs-baseline
+          persist-credentials: false
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v1
```
ThrawnCA, line 154: This version is old.

ThrawnCA (changes requested):

PR #1810, opened 2021-11-15
New detector for checking SEI CERT rule SEC02-J: Do not base security checks on untrusted sources

baloghadamsoftware:
A new detector called FindPotentialSecurityCheckBasedOnUntrustedSource approximately checks for
SEI CERT rule SEC02-J: Do not base security checks on untrusted sources. Specifically, it checks for the
scenario presented in the example code: the programmer calls a non-final method of a non-final class
instance received as a parameter of a public method of a public class to do a security check before
entering a doPrivileged block and then calls the same method inside the block. Since the method is
public it can be called by anyone and because the class of the parameter type is non-final, neither
the method is final which is called before and inside the doPrivileged it could have been overridden
by the caller to bypass the security check by returning different values on the first and subsequent
invocations.

----


baloghadamsoftware:
An alternative approach to [https://github.com/find-sec-bugs/find-sec-bugs/pull/657](#657) because I am more and more convinced that this rule has nothing to do with taint thus it has a better place here, in SpotBugs.

ThrawnCA (dismissed):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    private boolean isNestingMethodLocalVariable(OpcodeStack.Item object) {
+        XField field = object.getXField();
+        if (field == null) {
+            return false;
+        }
+
+        return field.getName().matches("val\\$.*");
```
KengoTODA, line 140: This [`String#matches(String)` method](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#matches-java.lang.String-) [compiles Regex `Pattern` every time](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/regex/Pattern.java#L1133).

For better performance, consider compile a `Pattern` instance in advance, keep it in an instance field, and reuse it in this method.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private void addToMethodsCalledInsidePrivilegedAction(XMethod calledMethod, OpcodeStack.Item object) {
+        Set<CallerInfo> objects = methodsCalledInsidePrivilegedAction.get(calledMethod);
+        if (objects == null) {
+            objects = new HashSet<>();
+            methodsCalledInsidePrivilegedAction.put(calledMethod, objects);
+        }
```
KengoTODA, line 148: The [`Map#computeIfAbsent(Function)`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) is better replacement to shorten the code:

```suggestion
        Set<CallerInfo> objects = methodsCalledInsidePrivilegedAction.computeIfAbsent(calledMethod, HashSet::new);
```

spotbugsTestCases/src/java/PotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,39 @@
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+public class PotentialSecurityCheckBasedOnUntrustedSource {
```
KengoTODA, line 7: If possible, could you add more test cases to test with lambda?
The `PrivilagedAction` is a functional interface, so coder may implement like: `AccessController.doPrivileged((Object object) -> { ... })`?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,242 @@
...
+    @Override
+    public void visit(JavaClass obj) {
+        isDoPrivileged = Subtypes2.instanceOf(getDottedClassName(), "java.security.PrivilegedAction")
+                || Subtypes2.instanceOf(getDottedClassName(), "java.security.PrivilegedExceptionAction");
+    }
+
+    @Override
+    public void visit(Method obj) {
+        nonFinalMethodsCalledOnParam.clear();
+        isDoPrivilegedRun = isDoPrivileged && "run".equals(getMethodName());
```
KengoTODA, line 87: better to check the method descriptor (arguments) too. user may add other `run` methods which isn't our target to verify.

KengoTODA (changes requested):
Add an entry into `CHANGELOG.md` to describe the change. I also feedback some, please consider to fix especially about the performance issue around `String#matches(String)`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
baloghadamsoftware: I had to use `k -> new HashSet<>()` because `HashSet::new` did not compile: javac could not deduce CallerInfo as the generic parameter of the HashSet. HashSet<>::new was syntactically invalid and there was also a problem using HashSet<CallerInfo>::new.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/BootstrapMethodsUtil.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.bcel.classfile.ConstantInterfaceMethodref;
+import org.apache.bcel.classfile.ConstantMethodHandle;
+import org.apache.bcel.classfile.ConstantMethodref;
+import org.apache.bcel.classfile.ConstantNameAndType;
+import org.apache.bcel.classfile.ConstantPool;
+import org.apache.bcel.classfile.JavaClass;
+import org.apache.bcel.classfile.Method;
+
+public class BootstrapMethodsUtil {
+    public static Method getMethodFromBootstrap(BootstrapMethods bms, int index, ConstantPool cp, JavaClass cls) {
```
KengoTODA, line 37: To tell null-ability of returned value, annotate this method with `@CheckForNull`, or return `Optional<Method>` instead of `Method`.

KengoTODA, line 37: And please consider to add a documentation comment, to tell what this method does.
Especially why this method returns only one method, even though `BootstrapMethods` is named as BootstrapMethod **"s"**.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,392 @@
...
+import edu.umd.cs.findbugs.ba.XField;
+import edu.umd.cs.findbugs.ba.XMethod;
+import edu.umd.cs.findbugs.ba.ch.Subtypes2;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import edu.umd.cs.findbugs.classfile.ClassDescriptor;
+import edu.umd.cs.findbugs.util.BootstrapMethodsUtil;
+
+public class FindPotentialSecurityCheckBasedOnUntrustedSource extends OpcodeStackDetector {
+
+    private static final Matcher NESTED_CLASS_VARIABLE_NAME_MATCHER = Pattern.compile("val\\$.*").matcher("");
```
KengoTODA, line 54: Better to keep `Pattern` instance which is immutable and thread-safe.
```suggestion
    private static final Pattern NESTED_CLASS_VARIABLE_NAME_PATTERN = Pattern.compile("val\\$.*");
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2021-??-??
 
+### Added
+* New rule `USC_POTENTIAL_SECURITY_CHECK_BASED_ON_UNTRUSED_SOURCE` to detect cases where a non-final method of a non-final class is called from public methods of public classes and then the same method is called on the same object inside a doPriviliged block. Since the called method may have been overridden to behave differently on the first and second invocations this is a possible security check based on an unreliable source. This rule is based on *SEC02-J. Do not base security checks on untrusted sources*. ([#SEC02-J](https://wiki.sei.cmu.edu/confluence/display/java/SEC02-J.+Do+not+base+security+checks+on+untrusted+sources))
```
KengoTODA, line 10: merge the latest upstream `master` branch, and move this entry under the `## Unreleased`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java
```diff
@@ -100,53 +92,32 @@ public void visit(JavaClass obj) {
...
-                    checkAndRecordCallFromClone(clone.method, method, clone.sourceLine);
-                }
-                if (callers != null) {
-                    for (XMethod caller : callers) {
-                        if (method.isPrivate() || method.isFinal()) {
-                            checkAndRecordCallBetweenNonOverridableMethods(caller, method);
-                        } else {
-                            checkAndRecordCallToOverridable(caller, method);
-                        }
+            Method method = BootstrapMethodsUtil.getMethodFromBootstrap(obj, i, getConstantPool(), getThisClass());
```
KengoTODA, line 101: Only this class depends on this new util method, so I suggest you to convert the util method into a method in this class. It is better to reduce published API as much as possible.

KengoTODA (changes requested):

CHANGELOG.md
ThrawnCA, line 10: Misspelled "doPrivileged"

spotbugs/etc/findbugs.xml
```diff
@@ -657,6 +657,8 @@
                     reports="REFLC_REFLECTION_MAY_INCREASE_ACCESSIBILITY_OF_CLASS,REFLF_REFLECTION_MAY_INCREASE_ACCESSIBILITY_OF_FIELD"/>
           <Detector class="edu.umd.cs.findbugs.detect.FindOverridableMethodCall"
                     reports="MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR,MC_OVERRIDABLE_METHOD_CALL_IN_CLONE"/>
+          <Detector class="edu.umd.cs.findbugs.detect.FindPotentialSecurityCheckBasedOnUntrustedSource" speed="fast"
```
ThrawnCA, line 660: Misspelled "UNTRUSTED"

spotbugs/etc/findbugs.xml
```diff
@@ -1266,4 +1268,6 @@
           <BugPattern abbrev="REFLF" type="REFLF_REFLECTION_MAY_INCREASE_ACCESSIBILITY_OF_FIELD" category="MALICIOUS_CODE" />
           <BugPattern abbrev="MC" type="MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR" category="MALICIOUS_CODE" />
           <BugPattern abbrev="MC" type="MC_OVERRIDABLE_METHOD_CALL_IN_CLONE" category="MALICIOUS_CODE" />
+          <BugPattern abbrev="USC" type="USC_POTENTIAL_SECURITY_CHECK_BASED_ON_UNTRUSED_SOURCE"
```
ThrawnCA, line 1271: Misspelled "UNTRUSTED"

spotbugs/etc/messages.xml
```diff
@@ -8612,6 +8620,26 @@ after the call to initLogging, the logger configuration is lost
...
+    <Details>
+      <![CDATA[<p>
+      A public method of a public class may be called from outside the package which means that untrusted
+      data may be passed to it. Calling a method before the doPrivileged to check its return value and then
+      calling the same method inside the class is dangerous if the method or its enclosing class is not
+      final. An attacker may pass an instance of a malicious descendant of the class instead of an instance
+      of the expected one where this method is overridden in a way that it returns different values upon
+      different invocations. For example, a method returning a file path may return a harmless path to check
+      before entering the doPrivileged block and then a sensitive file upon the call inside the doPrivileged
+      block. To avoid such scenario defensively copy the object reveiced in the parameter, e.g. by using
```
ThrawnCA, line 8635: Misspelled "received"

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindPotentialSecurityCheckBasedOnUntrustedSource.java
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    private void addToMethodsCalledInsidePrivilegedAction(XMethod calledMethod, OpcodeStack.Item object) {
+        Set<CallerInfo> objects = methodsCalledInsidePrivilegedAction.computeIfAbsent(calledMethod,
+                k -> new HashSet<>());
+        objects.add(new CallerInfo(object, getThisClass(),
+                SourceLineAnnotation.fromVisitedInstruction(this)));
+    }
+
+    private void addTononFinalMethodsCalledOnParam(ClassDescriptor calledClass, XMethod calledMethod,
```
ThrawnCA, line 265: "nonFinal" should be capitalised.

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java
baloghadamsoftware: Sorry, probably GitHub tricked you (as it also tricked me) because it hides the largest file by default. Actually, I created this utility class especially because both the current detector (`FindPotentialSecurityCheckBasedOnUntrustedSource`) and `FindOverridableMethodCall` uses this method. I also expect more (already existing) detectors in the future to cope with lambdas so this will probably also use it.

ThrawnCA (approved):

ThrawnCA:
Looks okay. I haven't gone into testing it myself.

baloghadamsoftware:
@KengoTODA Do I have to change anything else?

ThrawnCA (approved):

KengoTODA (changes requested):
Please merge the latest `origin/master` to your topic branch, and resolve the conflict in the `CHANGELOG.md`.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/BootstrapMethodsUtil.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.bcel.classfile.ConstantPool;
+import org.apache.bcel.classfile.JavaClass;
+import org.apache.bcel.classfile.Method;
+
+/**
+ * Utility methods for working with bootstrap methods
+ * 
+ * @author Ádám Balogh
+ */
+public class BootstrapMethodsUtil {
```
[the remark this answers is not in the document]
baloghadamsoftware: I intentionally moved this method to a new class from `FindOverridableMethodCall` to a utility class because it is used by both the `FindOverridableMethodCall` detector and this, new detector (`FindPotentialSecurityCheckBasedOnUntrustedSource`). Thus two different detectors use it.

KengoTODA (approved):
LGTM, thanks for your proposal!

ThrawnCA (approved):

PR #1833, opened 2021-11-23
optim: removed linear search in ValueNumberFrame.getLoad()

gtoison:
While I was waiting for a full analysis of a ~600k LoC project using the SpotBugs Eclipse plugin I fired up a profiler and noticed that about half (!) of the entire analysis time was spent on FindNullDeref.
More specifically ValueNumberFrame.getLoad() does a linear search and is called many times in FindNullDeref.

I've found a patch that was never applied in these old discussions:
https://sourceforge.net/p/findbugs/bugs/649/
https://sourceforge.net/p/findbugs/patches/188/

The purpose of this PR is to reopen the discussion, I think the original author was correct in his analysis and solution for this performance issue.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
@@ -54,6 +54,8 @@ public class ValueNumberFrame extends Frame<ValueNumber> implements ValueNumberA
 
     private Map<AvailableLoad, ValueNumber[]> availableLoadMap;
 
+    private Map<ValueNumber, AvailableLoad> availableLoadReverseMap;
+
     private Map<AvailableLoad, ValueNumber> mergedLoads;
 
     private Map<ValueNumber, AvailableLoad> previouslyKnownAs;
@@ -114,17 +116,8 @@ public class ValueNumberFrame extends Frame<ValueNumber> implements ValueNumberA
         if (!REDUNDANT_LOAD_ELIMINATION) {
             return null;
         }
-        for (Map.Entry<AvailableLoad, ValueNumber[]> e : getAvailableLoadMap().entrySet()) {
-            ValueNumber[] values = e.getValue();
-            if (values != null) {
-                for (ValueNumber v2 : values) {
-                    if (v.equals(v2)) {
-                        return e.getKey();
-                    }
-                }
-            }
-        }
-        return null;
+
+        return getAvailableLoadReverseMap().get(v);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gtoison:
Here's what I see when profiling the project analysis with version 4.5.0.r202111111015-5cc5fc9 of the eclipse plugin:

image

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
```diff
@@ -578,31 +590,76 @@ public boolean haveMatchingFlags(ValueNumber v1, ValueNumber v2) {
...
-            for (Map.Entry<AvailableLoad, ValueNumber[]> e : getAvailableLoadMap().entrySet()) {
-                if (e.getValue() != null) {
-                    Collections.addAll(result, e.getValue());
-                }
-            }
+            return Collections.unmodifiableCollection(getAvailableLoadReverseMap().keySet());
         }
 
-        return result;
+        return Collections.emptySet();
```
KengoTODA, line 597: Not sure how existing callers expect, but if there is a caller expecting mutability, it could be better to return a `new HashSet<>()`.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
```diff
@@ -54,6 +54,8 @@
 
     private Map<AvailableLoad, ValueNumber[]> availableLoadMap;
 
+    private Map<ValueNumber, AvailableLoad> availableLoadReverseMap;
```
KengoTODA, line 57: I think it's better to introduce a class like the [BiMap from Guava](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/BiMap.html)?
It will reduce the complexity of this class, and improve its readability.

We cannot use that `BiMap` because the value of `availableLoadReverseMap` is not an element but an array, but we can create a similar class on our own.

KengoTODA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
gtoison: Thanks a lot for looking into it, for this version I have applied the patch as-is.
I agree that the complexity of the class is high and using some kind of BiMap is a good idea.
I'll try implementing this differently

KengoTODA:
#1864 is merged so this PR is no longer needed.

Issue #1839, opened 2021-12-01
NP_NONNULL_RETURN_VIOLATION: false positive when using String.formatted()


xzel23:
Running Spotbugs 4.5.0 on this method:

    public @NotNull String version() {
        return "%d.%d.%d%s".formatted(major, minor, patchLevel, !suffix.isEmpty() ? "-"+suffix : "");
    }

yields:

> NP_NONNULL_RETURN_VIOLATION: Method may return null, but is declared @Nonnull
> 
> This method may return a null value, but the method (or a superclass method which it overrides) is declared to return @Nonnull.

gtoison:
Does it matter that the unit test uses `org.jetbrains.annotations.NotNull` while @xzel23 mentioned `javax.annotation.Nonnull` ?

hazendaz:
I revisited this issue and attempted to reproduce the reported behavior against the original SpotBugs 4.5.0 release.

The reproduction was run with Java 17, matching the environment reported in the original issue, and using the JetBrains `@NotNull` annotation shown in the original example.

The original reproduction passes on SpotBugs 4.5.0 and does not produce `NP_NONNULL_RETURN_VIOLATION`. The corresponding test also passes on current SpotBugs.

The branch has been rebased by me several times over the years, and @JuditKnoll has also made some cleanup changes to the test setup over time, so the current branch history should not be taken as evidence of when or how the reported behavior may have changed.

`String.formatted()` was introduced in Java 15.

After five years, and with the behavior not reproducible even against the reported SpotBugs version, there is no demonstrated defect remaining to fix. I'm therefore closing this issue as not reproducible and closing the associated reproduction PR as unnecessary. No regression test is being retained for behavior that cannot be reproduced.

PR #1863, opened 2021-12-11
WIP: Try to reproduce the issue 1839

KengoTODA:
Tried to reproduce the #1839 but failed.

----


ThrawnCA (approved):

hazendaz:
Thanks for the original reproduction attempt. I revisited this against SpotBugs 4.5.0 using Java 17 and the JetBrains `@NotNull` from the original report. The reproduction passes and the reported `NP_NONNULL_RETURN_VIOLATION` does not occur.

It also passes with the current test infrastructure/current SpotBugs. Since the issue cannot be reproduced even against the reported SpotBugs version, there is no demonstrated regression to preserve here. Closing this PR as unnecessary.

PR #1864, opened 2021-12-12
optim: removed linear search in ValueNumberFrame.getLoad()

gtoison:
This is a resubmission of https://github.com/spotbugs/spotbugs/pull/1833 hopefully implemented in a clearer way. I tried changing the existing code as little as possible and the new data structure is in an inner class.

I've tried benchmarking the change and it only seems to make a difference with the `-effort max` option. On a ~600k LoC sample project, running `FindBugs2` with the arguments `<PROJECT_PATH>\target\classes -effort max` I got a 19% improvement (176 seconds vs 217 seconds on the average of 3 runs).

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/vna/ValueNumberFrame.java
@@ -52,7 +52,7 @@ public class ValueNumberFrame extends Frame<ValueNumber> implements ValueNumberA
 
     private ArrayList<ValueNumber> mergedValueList;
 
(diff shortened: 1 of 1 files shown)


KengoTODA (approved):
LGTM. Thanks for realizing both performance and readability! 🆒

ThrawnCA (approved):
Isn't there anything from a standard library to do this? Either Apache Commons Collections or Google Guava?

KengoTODA:
> Isn't there anything from a standard library to do this? Either Apache Commons Collections or Google Guava?

Guava has [`BiMap` class](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/BiMap.html) but we cannot use it because the value of `availableLoadReverseMap` is not an element but an array.

Issue #1868, opened 2021-12-13
Release 4.5.2 to address CVE-2021-44228


iloveeclipse:
We should release 4.5.2 version to quickly address CVE-2021-44228.
The PR #1865 is prerequisite.

Note: there is no known exploit using spotbugs possible yet, but since we ship and use log4j in affected version, and there are web services that run spotbugs as a part of the build pipeline, an attacker could manage to craft a pull request that contains the Java code that would somehow manage the bad pattern be part of the message logged by spotbugs.

iloveeclipse:
@KengoTODA : the build fails with this error below: any idea?
```
> Task :closeSonatypeStagingRepository FAILED
FAILURE: Build failed with an exception.


Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
* What went wrong:

Execution failed for task ':closeSonatypeStagingRepository'.
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
> Staging repository is still transitioning after defined time. Consider its increment. StagingRepository(id=comgithubspotbugs-1253, state=open, transitioning=true)
```

KengoTODA:
> Staging repository is still transitioning after defined time. Consider its increment. StagingRepository(id=comgithubspotbugs-1253, state=open, transitioning=true)

It seems that the Sonatype Nexus (oss.sonatype.org) took a long time to handle our request. The deployment itself has been succeeded.
I will check the diff, and release pushed artifact if found no problem.

iloveeclipse:
Please approve and merge #1871 first

KengoTODA:
It seems that Apache Foundation released 2.16.0, I'll check its release note.

https://github.com/apache/logging-log4j2/releases/tag/rel%2F2.16.0

iloveeclipse:
see https://lists.apache.org/thread/6g0ls1cc9htn52hr0kkbo7dds5ts3rb5

iloveeclipse:
I assume we could pick this release too.

KengoTODA:
checked, thanks. Is it better to release our 4.5.2 with 2.16.0? It is possible for now.

iloveeclipse:
Please update.

iloveeclipse:
@KengoTODA : I think we should release maven and gradle plugins ASAP too, once core is released. WDYT? Would you push that?

KengoTODA:
[...]

PR #1875, opened 2021-12-15
Update log4j-slf4j18-impl for Eclipse plugin

KengoTODA:
Eclipse plugin unexpectedly contains both 2.15.0 and 2.16.0. This is because the `copyLibsForEclipse` task copies not only runtime dependencies but also test runtime dependencies. And when we release SpotBugs 4.5.2, test runtime dependencies have `log4j-slf4j18-impl` version 2.15.0.

By updating this dependency, we can solve this duplicated libraries in Eclipse plugin. I confirmed that generated `eclipsePlugin/META-INF/MANIFEST.MF` contains no vulnerable libraries as follows:

```
Bundle-ClassPath: .,spotbugs-plugin.jar,lib/asm-9.2.jar,lib/asm-analysis
 -9.2.jar,lib/asm-commons-9.2.jar,lib/asm-tree-9.2.jar,lib/asm-util-9.2.
 jar,lib/bcel-6.5.0.jar,lib/commons-lang3-3.12.0.jar,lib/commons-text-1.
 9.jar,lib/dom4j-2.1.3.jar,lib/error_prone_annotations-2.10.0.jar,lib/gs
 on-2.8.9.jar,lib/jaxen-1.2.0.jar,lib/jcip-annotations-1.0.jar,lib/jsr30
 5-3.0.2.jar,lib/junit-4.13.1.jar,lib/log4j-api-2.16.0.jar,lib/log4j-cor
 e-2.16.0.jar,lib/log4j-slf4j18-impl-2.16.0.jar,lib/slf4j-api-1.8.0-beta
 4.jar,lib/spotbugs-4.5.1-SNAPSHOT-javadoc.jar,lib/spotbugs-4.5.1-SNAPSH
 OT-sources.jar,lib/spotbugs-4.5.2-SNAPSHOT-javadoc.jar,lib/spotbugs-4.5
 .2-SNAPSHOT-sources.jar,lib/spotbugs-annotations-4.5.1-SNAPSHOT-javadoc
 .jar,lib/spotbugs-annotations-4.5.1-SNAPSHOT-sources.jar,lib/spotbugs-a
 nnotations-4.5.2-SNAPSHOT-javadoc.jar,lib/spotbugs-annotations-4.5.2-SN
 APSHOT-sources.jar,lib/spotbugs-annotations.jar,lib/spotbugs.jar
```

refs https://github.com/spotbugs/spotbugs/issues/1868#issuecomment-993710855

### DIFF
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -96,7 +96,7 @@ dependencies {
   api "com.google.code.gson:gson:2.8.9"
   testImplementation 'junit:junit:4.13.1'
   testImplementation 'org.apache.ant:ant:1.10.12'
-  testImplementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.15.0'
+  testImplementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.16.0'
   testImplementation 'com.google.errorprone:error_prone_annotations:2.10.0'
 
   guiImplementation sourceSets.main.runtimeClasspath
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -12,6 +12,7 @@ configurations {
 
 ext {
   asmVersion = '9.2'
+  log4jVersion = '2.16.0'
 }
 
 sourceSets {
@@ -85,7 +86,7 @@ dependencies {
   api 'org.apache.commons:commons-text:1.9'
   api 'org.slf4j:slf4j-api:1.8.0-beta4'
   implementation 'net.sf.saxon:Saxon-HE:10.6'
-  logBinding ('org.apache.logging.log4j:log4j-slf4j18-impl:2.16.0') {
+  logBinding ("org.apache.logging.log4j:log4j-slf4j18-impl:$log4jVersion") {
     exclude group: 'org.slf4j'
   }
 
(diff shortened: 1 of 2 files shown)


spotbugs/build.gradle
```diff
@@ -96,7 +96,7 @@ dependencies {
   api "com.google.code.gson:gson:2.8.9"
   testImplementation 'junit:junit:4.13.1'
   testImplementation 'org.apache.ant:ant:1.10.12'
-  testImplementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.15.0'
+  testImplementation 'org.apache.logging.log4j:log4j-slf4j18-impl:2.16.0'
```
iloveeclipse, line 99: We have at least two places where log4j dependency is hard coded, we should extract it to a variable like `asmVersion` above

iloveeclipse (changes requested):

spotbugs/build.gradle
KengoTODA: Nice catch, I'll do it :)

iloveeclipse (approved):
Thanks. I've overlooked that there was the second dependency in original PR. Not it shouldn't happen with 2.17.+ :-)

PR #1884, opened 2021-12-19
Corrected class name validation to no longer fail for Kotlin classes on class path containing special characters

### LINKED ISSUE #1883: Mixed Kotlin/Java projects can produce classes that crash the tool due to strange class names
When running the spotbugs gradle task in a mixed Kotlin/Java project, the task may fail due to the presence of classes containing special characters in their binary name. This is due to Kotlin allowing identifiers (including method names) to be backtick-quoted.

An example of such a class might be (in binary / slashed name format):
```
com/bla/TestClass$this test calls run() successfully$obj$1"
```
This class would be an anonymous class existing in the `TestClass` class in the `com.bla` package, which is assigned in the `` `this test calls run()` `` method, assigned to the `obj` variable. The rough code that would produce this in Kotlin would be:

```
package com.bla

// ...
class TestClass{

    // ...
    fun `this test calls run() successfully`() {
        // ...
        var obj = object {}
        // ...
    }
// ...
}
```

where the `var obj = object {}` produces the anonymous inner class.

[This Kotlin Playground example](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjEwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoicGFja2FnZSBjb20uYmxhXG5cbmZ1biBtYWluKCkge1xuICAgIFRlc3RDbGFzcygpLmB0aGlzIHRlc3QgY2FsbHMgcnVuKCkgc3VjY2Vzc2Z1bGx5YCgpXG59XG5cbi8vIC4uLlxuY2xhc3MgVGVzdENsYXNze1xuXG4gICAgLy8gLi4uXG4gICAgZnVuIGB0aGlzIHRlc3QgY2FsbHMgcnVuKCkgc3VjY2Vzc2Z1bGx5YCgpIHtcbiAgICAgICAgLy8gLi4uXG4gICAgICAgIHZhciBvYmogPSBvYmplY3Qge31cbiAgICAgICAgcHJpbnRsbihvYmouamF2YUNsYXNzLm5hbWUpXG4gICAgICAgIC8vIC4uLlxuICAgIH1cbi8vIC4uLlxufSJ9) shows the described class naming in action.

I've identified the root cause of this issue being the `edu.umd.cs.findbugs.util.ClassName.isValidClassName(String)` method not implementing proper name validation in accordance with [section 4.2 of the Java Virtual Machine specification](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.2). It appears that this method is being used in a number of different ways and must support all of the following inputs:

* dotted class names,
* binary (slashed) class names, and
* field descriptors.

I've made two commits in the PR that I will attach to this issue:
* the first adds test coverage showing the current behaviour and expected behaviour (with some tests failing).
* the second makes changes to the `isValidClassName()` method to make the tests pass.

studro:
Closes #1883.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
@@ -207,7 +207,8 @@ public abstract class ClassName {
      * @param className a class name to test for validity - must be non-{@code null} and non-empty.
      * @return {@code true} if {@code className} is a valid array field descriptor as
      *          per JVMS 4.3.2, otherwise {@code false}
-     * @throws IndexOutOfBoundsException if {@code className} is {@code null} or empty.
+     * @throws IndexOutOfBoundsException if {@code className} is empty.
+     * @throws NullPointerException if {@code className} is {@code null}.
      */
     private static boolean isValidArrayFieldDescriptor(String className) {
         String tail = className.substring(1);
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
@@ -175,9 +175,47 @@ public abstract class ClassName {
      * @return true if it's a valid class name, false otherwise
      */
     public static boolean isValidClassName(String className) {
-        // FIXME: should use a regex
+        return isValidBinaryClassName(className) ||
+                isValidDottedClassName(className) ||
+                isValidArrayFieldDescriptor(className) ||
+                isValidClassFieldDescriptor(className);
+    }
 
-        return className.indexOf('(') < 0;
+    private static boolean isValidBinaryClassName(String className) {
+        return className.indexOf('.') == -1 &&
+                className.indexOf('[') == -1 &&
+                className.indexOf(';') == -1;
+    }
+
+    private static boolean isValidDottedClassName(String className) {
+        return className.indexOf('/') == -1 &&
+                className.indexOf('[') == -1 &&
+                className.indexOf(';') == -1;
+    }
+
+    private static boolean isValidArrayFieldDescriptor(String className) {
+        return className.startsWith("[") &&
+                (isValidArrayFieldDescriptor(className.substring(1))) ||
+                isValidClassFieldDescriptor(className.substring(1)) ||
+                isValidBaseTypeFieldDescriptor(className.substring(1));
+
+    }
+
+    private static boolean isValidClassFieldDescriptor(String className) {
+        return className.startsWith("L") &&
+                className.endsWith(";") &&
+                isValidBinaryClassName(className.substring(1, className.length() - 1));
+    }
+
+    private static boolean isValidBaseTypeFieldDescriptor(String className) {
+        return "B".equals(className) ||
+                "C".equals(className) ||
+                "D".equals(className) ||
+                "F".equals(className) ||
+                "I".equals(className) ||
+                "J".equals(className) ||
+                "S".equals(className) ||
+                "Z".equals(className);
     }
 
     /**
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
@@ -175,12 +175,18 @@ public abstract class ClassName {
      * @return true if it's a valid class name, false otherwise
      */
     public static boolean isValidClassName(String className) {
-        return isValidBinaryClassName(className) ||
-                isValidDottedClassName(className) ||
-                isValidArrayFieldDescriptor(className) ||
-                isValidClassFieldDescriptor(className);
+        return !className.isEmpty() &&
+                (isValidBinaryClassName(className) ||
+                        isValidDottedClassName(className) ||
+                        isValidArrayFieldDescriptor(className) ||
+                        isValidClassFieldDescriptor(className));
     }
 
+    /**
+     * @see <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1">
+     *     JVMS (Java SE 8 Edition) 4.2.1. Binary Class and Interface Names
+     *     </a>
+     */
     private static boolean isValidBinaryClassName(String className) {
         return className.indexOf('.') == -1 &&
                 className.indexOf('[') == -1 &&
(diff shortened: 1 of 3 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -175,9 +175,47 @@ public static String extractSimpleName(@DottedClassName String className) {
...
+        return isValidBinaryClassName(className) ||
+                isValidDottedClassName(className) ||
+                isValidArrayFieldDescriptor(className) ||
+                isValidClassFieldDescriptor(className);
+    }
+
+    private static boolean isValidBinaryClassName(String className) {
+        return className.indexOf('.') == -1 &&
+                className.indexOf('[') == -1 &&
+                className.indexOf(';') == -1;
```
KengoTODA, line 187: According to the [JVMS 4.2.2](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.2):

> An unqualified name must contain at least one Unicode code point and must not contain any of the ASCII characters . ; [ / (that is, period or semicolon or left square bracket or forward slash).

So better to check existence of `/` too.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -175,9 +175,47 @@ public static String extractSimpleName(@DottedClassName String className) {
...
      */
     public static boolean isValidClassName(String className) {
-        // FIXME: should use a regex
+        return isValidBinaryClassName(className) ||
+                isValidDottedClassName(className) ||
+                isValidArrayFieldDescriptor(className) ||
+                isValidClassFieldDescriptor(className);
+    }
+
+    private static boolean isValidBinaryClassName(String className) {
```
KengoTODA, line 184: Better to have a hyperlink referring to the definition of 'valid binary class name':

```suggestion
    /**
     * @see <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1">JVMS (Java SE 8 Edition) 4.2.2. Unqualified Names</a>
     */
    private static boolean isValidBinaryClassName(String className) {
```

Note that we may need to run `./gradlew spotlessApply` to reformat my snippet.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -175,9 +175,47 @@ public static String extractSimpleName(@DottedClassName String className) {
...
+        return className.indexOf('/') == -1 &&
+                className.indexOf('[') == -1 &&
+                className.indexOf(';') == -1;
+    }
+
+    private static boolean isValidArrayFieldDescriptor(String className) {
+        return className.startsWith("[") &&
+                (isValidArrayFieldDescriptor(className.substring(1))) ||
+                isValidClassFieldDescriptor(className.substring(1)) ||
+                isValidBaseTypeFieldDescriptor(className.substring(1));
```
KengoTODA, line 200: It's just my personal and unreasonable motivation. JIT could be enough.

```suggestion
        String tail = className.substring(1);
        return className.startsWith("[") &&
                (isValidArrayFieldDescriptor(tail)) ||
                isValidClassFieldDescriptor(tail) ||
                isValidBaseTypeFieldDescriptor(tail);
```

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -175,9 +175,47 @@ public static String extractSimpleName(@DottedClassName String className) {
      * @return true if it's a valid class name, false otherwise
      */
     public static boolean isValidClassName(String className) {
-        // FIXME: should use a regex
+        return isValidBinaryClassName(className) ||
+                isValidDottedClassName(className) ||
+                isValidArrayFieldDescriptor(className) ||
+                isValidClassFieldDescriptor(className);
```
KengoTODA, line 181: :+1: Confirmed that [all descriptions listed in the JVMS](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2) are covered.

KengoTODA (changes requested):
Just one mandatory request exists in `isValidBinaryClassName()` methods, to follow the definition in the JVMS.
Other feedback are just my 2 cents. Thanks for your PR!

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
studro: In your suggestion, the link is to section 4.2.1, but the link text is to 4.2.2. I think 4.2.1 is most correct, as it refers to class names, while 4.2.2 refers to the elements of the class name separated by `/` or `.`.

I also tried to wrap the link, as it resulted in a very long line.

Feel free to throw this back to me if you feel that 4.2.1 is incorrect as we should stick with 4.2.2 instead.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
studro: I don't think this is entirely correct. An unqualified name may not contain `.`, `;`, `[`, or `/`, as you've quoted, but a binary class name consists of one or more unqualified names, which are separated by `/`s.

According to [JVMS 4.2.1](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1): 

>  In this internal form, the ASCII periods (.) that normally separate the identifiers which make up the binary name are replaced by ASCII forward slashes (/). The identifiers themselves must be unqualified names (§4.2.2).

If we were to exclude `/` here, names of the form `com/foo/Example` would be rejected.

I suggest that this remains as is, but I am happy to take any further suggestions.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
studro: I actually thinks this improves readability too, so this might be a good change. :+1:

However, I am concerned about the possibility of throwing an `IndexOutOfBoundsException` if this is called with an empty string (instead of returning false). I have added an `isEmpty()` check in the caller and documentation to mitigate this concern.

studro:
> Just one mandatory request exists in `isValidBinaryClassName()` methods, to follow the definition in the JVMS. Other feedback are just my 2 cents. Thanks for your PR!

Thanks for your quick and thorough review. I've attempted to address the comments you've raised and will be happy to make any further changes as necessary. 😄

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
KengoTODA: Thanks for your confirmation! The following sentence in the "4.2.1. Binary Class and Interface Names" is correct reference:

> The identifiers themselves must be unqualified names (§4.2.2).

KengoTODA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -175,9 +175,64 @@ public static String extractSimpleName(@DottedClassName String className) {
...
+
+    /**
+     * Determines whether a class name is a valid array field descriptor as per
+     * <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2">
+     * JVMS (Java SE 8 Edition) 4.3.2</a>
+     *
+     * @param className a class name to test for validity - must be non-{@code null} and non-empty.
+     * @return {@code true} if {@code className} is a valid array field descriptor as
+     *          per JVMS 4.3.2, otherwise {@code false}
+     * @throws IndexOutOfBoundsException if {@code className} is {@code null} or empty.
```
studro, line 210: I was thinking about the documentation of this later on in the day, and realised that I've made a mistake in documenting the exception behaviour. I think this minor change would be a good idea.

```suggestion
     * @throws IndexOutOfBoundsException if {@code className} is empty.
     * @throws NullPointerException if {@code className} is {@code null}.
```

studro:
I am very sorry to be difficult, but one final review is required, if possible.
I've updated the exception documentation as I noted in https://github.com/spotbugs/spotbugs/pull/1884#discussion_r772946651. While this was only a minor issue, what I had written was completely wrong.
I also noticed that I had the incorrect commit author email set for this project (my personal instead of no-reply github email), so I've updated history with a force push to correct that.
Thanks for your patience and hopefully this is now the end of this issue! 😊

KengoTODA (approved):

ThrawnCA (approved):

PR #1902, opened 2022-01-05
Stop using local Eclipse as a part of dependency

### LINKED ISSUE #1868: Release 4.5.2 to address CVE-2021-44228

### LINKED ISSUE #1882: Drop log4j from released distribution / Eclipse plugin
In the light of the recent never ending log4j security issues I propose to drop log4j binaries from all our build artifacts, including Eclipse bundle.

So instead of updating to 2.17.0 log4j (see https://logging.apache.org/log4j/2.x/security.html#CVE-2021-45105) we should remove it. 

I haven't tried that, but I assume we either can expect users to provide desired slf4j binding, or we could bundle either slf4j-simple or slf4j-jdk14 with core distribution. For Eclipse I assume we could rely on Eclipse product configuration - if there is nothing, nothing will be logged.

WDYT?

KengoTODA:
This change makes it possible to build Eclipse plugins without local Eclipse IDE.
It is also possible to use metadata of artifacts (pom.xml) to maintain license, vulnerability, and version of dependencies.

This change won't affect the manifest of the spotbugs distribution.  
This change will remove the following `Bundle-Classpath` from the manifest of Eclipse plugin:

* Compile time dependency
  * lib/error_prone_annotations-2.10.0.jar
* Test dependency
  * lib/junit-4.13.1.jar
* Logging libraries
  * lib/log4j-api-2.17.1.jar
  * lib/log4j-core-2.17.1.jar
  * lib/log4j-slf4j18-impl-2.17.1.jar

I've confirmed that we can remove the logging libraries from our Eclipse plugin, because Eclipse IDE itself provides SLF4J API and its binding in runtime:

image

I've tested with both of Eclipse 2021‑12 and [Neon.3 (which we use to build the plugin)](https://github.com/spotbugs/spotbugs/blob/6a2a4e8151ed9dcb01e7332df01d6d4f465a0b31/.github/workflows/release.yml#L22), so I believe that this change will not break compatibility.

close #1868
close #1882

### DIFF
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -316,6 +316,26 @@ apply from: "$rootDir/gradle/jigsaw.gradle"
 tasks.named('spotbugsTest').configure {
   ignoreFailures = true
 }
+
+dependencies {
+  constraints {
+    logBinding("org.apache.logging.log4j:log4j-core") {
+      version {
+        strictly("[2.17.1, 3[")
+        prefer("2.17.1")
+      }
+      because("CVE-2021-44228, CVE-2021-45046, CVE-2021-45105, CVE-2021-44832: Log4j vulnerable to remote code execution and other critical security vulnerabilities")
+    }
+    logBinding("ch.qos.logback:logback-core") {
+      version {
+        strictly("[1.2.9, 2[")
+        prefer("1.2.10")
+      }
+      because("CVE-2021-42550: Logback vulnerable to remote code execution vulnerabilities")
+    }
+  }
+}
+
 // TODO : generatemanual (we should decide what to do with the manual)
 // TODO : generatepdfmanual
 // TODO : bugdesc
--- a/spotbugs/build.gradle
+++ b/spotbugs/build.gradle
@@ -2,6 +2,19 @@ apply from: "$rootDir/gradle/maven.gradle"
 apply from: "$rootDir/gradle/checkstyle.gradle"
 apply from: "$rootDir/gradle/javadoc.gradle"
 
+tasks.named('compileJava', JavaCompile).configure {
+    options.release = 8
+}
+
+// Force Eclipse use Java 1.8, otherwise it will get Java 17 (!) from gradle
+eclipse {
+  jdt {
+    sourceCompatibility = 8
+    targetCompatibility = 8
+    javaRuntimeName = "JavaSE-1.8"
+  }
+}
+
 configurations {
   // used only in distribution. It is not listed in pom.xml, so users like maven plugin don't use this dependency.
   logBinding
@@ -107,13 +120,11 @@ tasks.withType(Jar).configureEach {
   //destinationDir = file("$projectDir/lib")
 }
 
-eclipse.classpath {
-  plusConfigurations += [ configurations.appleExtensions ]
-}
-
 eclipse.classpath.file {
     whenMerged {
       classpath ->
+         classpath.entries.removeAll { entry -> entry.kind == 'lib' && entry.path.contains('build/classes/')}
+         classpath.entries.removeAll { entry -> entry.kind == 'lib' && entry.path.contains('build/resources/')}
          classpath.entries.removeAll { entry -> entry.kind == 'lib' && entry.path.contains('xml-apis')}
          classpath.entries.forEach {
              entry ->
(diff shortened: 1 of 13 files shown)


spotbugs/build.gradle
```diff
@@ -361,6 +352,26 @@ tasks.named('jacocoTestReport', JacocoReport).configure {
...
+      because("CVE-2021-44228, CVE-2021-45046, CVE-2021-45105, CVE-2021-44832: Log4j vulnerable to remote code execution and other critical security vulnerabilities")
+    }
+    logBinding("ch.qos.logback:logback-core") {
+      version {
+        strictly("[1.2.9, 2[")
+        prefer("1.2.10")
+      }
+      because("CVE-2021-42550: Logback vulnerable to remote code execution vulnerabilities")
+    }
+  }
```
KengoTODA, line 372: These config will keep SpotBugs away from buggy logging libraries.
https://blog.gradle.org/log4j-vulnerability

iloveeclipse: I'm on it. I see few issues with Eclipse support also before this patch (wrong JLS settings, duplicated classpath elements, not compilable code, Java 11 dependencies introduced).

ThrawnCA (dismissed):

iloveeclipse:
I see following during the build:
```
> Task :buildSrc:compileKotlin
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```
I don't know why do we compile now kotlin, but that should probably also use Java 11.

Note: as I've mentioned before, Eclipse plugin code already depends on java 11 API but is compiled against Java 8. I will most likely want to change that to Java 11, and also increase minimal supported Eclipse SDK version to 4.15 (that works fine with Java 11). 

In general, for all spotbugs artifacts except annotations we should switch to Java 11, compiling with Java 8 JLS against Java 11 API and silently depending on Java 11 API's isn't a good idea anyway.

Another set of warnings related to Eclipse site generation:

```
> Task :eclipsePlugin:siteHtml
Execution optimizations have been disabled for task ':eclipsePlugin:siteHtml' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/data/git/spotbugs/eclipsePlugin/build/site/eclipse'. Reason: Task ':eclipsePlugin:pluginCandidateJar' uses this output of task ':eclipsePlugin:siteHtml' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.
  - Gradle detected a problem with the following location: '/data/git/spotbugs/eclipsePlugin/build/site/eclipse'. Reason: Task ':eclipsePlugin:testPluginJar' uses this output of task ':eclipsePlugin:siteHtml' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

This is shown for
- eclipsePlugin:siteXml
- eclipsePlugin:siteHtml
- eclipsePlugin:pluginDailyJar
- eclipsePlugin:pluginStableLatestJar

Similar warnings are reported for
```
> Task :spotbugs:spotbugsGui
Execution optimizations have been disabled for task ':spotbugs:spotbugsGui' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/data/git/spotbugs/spotbugs/build/libs/spotbugs.jar'. Reason: Task ':spotbugs:spotbugsGui' uses this output of task ':spotbugs:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :spotbugs:spotbugsGuiTest
Execution optimizations have been disabled for task ':spotbugs:spotbugsGuiTest' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/data/git/spotbugs/spotbugs/build/libs/spotbugs.jar'. Reason: Task ':spotbugs:spotbugsGuiTest' uses this output of task ':spotbugs:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :spotbugs:spotbugsMain
Execution optimizations have been disabled for task ':spotbugs:spotbugsMain' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/data/git/spotbugs/spotbugs/build/libs/spotbugs.jar'. Reason: Task ':spotbugs:spotbugsMain' uses this output of task ':spotbugs:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :spotbugs:spotbugsTest
Execution optimizations have been disabled for task ':spotbugs:spotbugsTest' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/data/git/spotbugs/spotbugs/build/libs/spotbugs.jar'. Reason: Task ':spotbugs:spotbugsTest' uses this output of task ':spotbugs:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.
SpotBugs ended with exit code 1
```

spotbugs/build.gradle
```diff
@@ -145,25 +145,16 @@ eclipse.classpath.file {
     }
 }
 
-def copyLibsForEclipse = tasks.register('copyLibsForEclipse', Copy) {
```
iloveeclipse, line 148: This makes Eclipse build unhappy. Since there are no libraries available for Eclipse, I can't compile / run SpotBugs from Eclipse anymore.

iloveeclipse (changes requested):
I'm unable to get this build running producing correct eclipse configuration files.
This is partly due already existing issues, but partly because of removed .libs below.
I will try to push a commit to this branch that first fixes Eclipse config file generation.

KengoTODA:
Really thank you for your review!

About the Gradle warning around the build optimization, I will handle them in another ticket #1911.
About the target Eclipse version to support, I agree to change the minimal supported version to 4.12.0. I will change the [Eclipse version in CI](https://github.com/spotbugs/spotbugs/blob/fa0ef8ebb0c15edcd75b71d6b4218bafec6c3816/.github/workflows/build.yml#L23) later.

About the target Java version to support, I think it's not so good to drop Java 8 support for now (I personally believe that everyone should migrate to Java 17 asap, tho). Would be better to do it as a part of next major release.

About the `copyLibsForEclipse` task, I prefer removing it to handle dependencies not as files but as dependency with metadata (pom.xml). I will try to find a better way.  
Can I assume that we use the `eclipse` task to generate Eclipse projects? Or it is also OK to use `Import existing Gradle project` provided by the Eclipse itself.

KengoTODA:
personal note: working on 06c018828f91ae910d4d4ed8a854cf669fceaa06 to fix generated `.classpath` file

iloveeclipse:
> Can I assume that we use the eclipse task to generate Eclipse projects? 

Sure.

> Or it is also OK to use Import existing Gradle project provided by the Eclipse itself

No, this functionality is not available in default SDK, so this will be not OK.

iloveeclipse:
I've just switched to my Windows notebook (no Java 17, only Java 11) and see that for some reason both eclipsePlugin-junit and eclipsePlugin-test projects have classpath errors in Eclipse - they have such (fully wrong because unneeded) ´pdebuild.jar` library added: 

```
<classpathentry kind="lib" path="C:/git/spotbugs/eclipsePlugin-junit/build/pdeTool/pdebuild.jar">
```
Not sure why I didn't saw this on my Linux workstation.
I have no idea why that should be on classpath at all, this is coming from the new eclipse-convention.gradle.kts fie.

ThrawnCA:
> I think it's not so good to drop Java 8 support for now

Yeah, OpenJDK 8 is still supported until 2026, so there could be a lot of organisations not in a hurry to migrate, and that's okay.

ThrawnCA (approved):

iloveeclipse:
So with #1914 fixed we are able to build spotbugs core on Java 8, so I've reverted Java 11 changes on production projects and pushed updated commits that make also spotbugs developing & debugging in Eclipse and on Windows happy. For me this state works fine & is a clear improvement over the old one.

iloveeclipse (approved):

Issue #1903, opened 2022-01-05
FindOverridableMethodCall.visitBootstrapMethods(BootstrapMethods) contains invisible \u200b character


iloveeclipse:
I'm unable to compile current spotbugs code in Eclipse because of `FindOverridableMethodCall.visitBootstrapMethods(BootstrapMethods) ` method.

The method ends with `\u200b` [unicode character](https://codepoints.net/U+200b?lang=en) and Eclipse compiler assumes that this method can't override the `visitBootstrapMethods` defined in the `org.apache.bcel.classfile.Visitor.visitBootstrapMethods(BootstrapMethods)`.

The method was added in #1716.

I will push a patch to remove the invisible character and will investigate which compiler is right.
I see that Eclipse 4.12 - 4.22 accepts `\u200b` in source file **and preserves** as part of the method name in the class file.
However, javac 1.8 - 17 accepts `\u200b` in the source file but **does not preserve** it as part of the method name in the class file.

Reported bug against ecj compiler: https://bugs.eclipse.org/bugs/show_bug.cgi?id=578063

iloveeclipse:
Patch that shows the difference (just run it on command line)
```
diff --git a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java
index b96fe8b..d1f87de 100644
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCall.java
@@ -41,2 +41,3 @@
 import org.apache.bcel.classfile.Method;
+import org.apache.bcel.classfile.Visitor;
 
@@ -99,3 +100,24 @@
 
-    @Override
+    private static void printName(String mName) {
+        char[] name = mName.toCharArray();
+        System.out.println("Name: '" + mName + "'");
+        for (char c : name) {
+            System.out.println("" + c + " : " + (int) c);
+        }
+    }
+
+    public static void main(String[] args) {
+        java.lang.reflect.Method method = Arrays.asList(Visitor.class.getMethods()).stream()
+                .filter(m -> m.getName().startsWith("visitBootstrapMethods")).findFirst().get();
+        String name = method.getName();
+        printName(name);
+
[...]

PR #1904, opened 2022-01-05
Remove \u200b from FindOverridableMethodCall.visitBootstrapMethods()

iloveeclipse:
The method ends with `\u200b` unicode character and Eclipse compiler
assumes that this method can't override the visitBootstrapMethods
defined in the org.apache.bcel.classfile.Visitor.visitBootstrapMethods(BootstrapMethods)

Since this character is invisible, it was added by mistake and should be
removed to allow SpotBugs compilation in Eclipse.

This fixes issue #1903.

----


iloveeclipse:
@KengoTODA : Sonar fails on code that I haven't introduced or modified.
So do we have now stricter checks?
Why doesn't this check fail on other PR's?

KengoTODA (approved):

ThrawnCA (approved):

Issue #1906, opened 2022-01-06
Spotbugs can't be compiled against JDK 1.8


iloveeclipse:
While reviewing #1902 I've found that SpotBugs that is supposed to be used on java 8 can't be compiled against Java 8, because it already uses Java 11 API. Specifying source 1.8 is not enough, if compiler uses Java 11 / 17 for Java class libraries used.

```
edu.umd.cs.findbugs.BugReporterDispatcherTest.dispatchingMethod():
The constructor PrintStream(ByteArrayOutputStream, boolean, Charset) is undefined
```

```
edu.umd.cs.findbugs.TextUICommandLineTest.htmlReportWithOption():
The method readString(Path, Charset) is undefined for the type Files
```

iloveeclipse:
My proposal would be to bump spotbugs lower JDK requirement to Java 11. Since we haven't received bug reports yet, it means, most of the users are already using new JDK anyway. However, I'm unable to make gradle build happy because of spotbugs test cases that need to be compiled with 1.8 / 11 / 17 and gradle says that if consumer is compiled 1.8, the spotbugs itself must be also compiled with 1.8. Stupid!

gtoison:
Currently the SonarQube LTS version (8.9) supports Java 8 and 11 while the latest version (9.2) only supports Java 11.
So bumping the lower JDK version required to 11 might be a problem for users of SonarQube 8.9 still on Java 8

iloveeclipse:
OK, another possibility would be to:
- fix tests to use Java 8 API
- or move tests to separated project that can use Java 11+

KengoTODA:
How do you think that we ask SpotBugs contributors to build the SpotBugs with JDK 11+?
Even though we use JDK 11+ to build, at least Gradle can generate class files that works on JVM 8, because we set `--release` option to realize it:

https://github.com/spotbugs/spotbugs/blob/fa0ef8ebb0c15edcd75b71d6b4218bafec6c3816/gradle/java.gradle#L10-L12

We may need additional work to configure Eclipse projects.

> fix tests to use Java 8 API

I think this is also nice option for us. It needs less changes than making another sub project.

ThrawnCA:
The build also uses the `--add-exports` parameter, which is not compatible with JDK 8.

KengoTODA:
> The build also uses the --add-exports parameter, which is not compatible with JDK 8.

Yes for Gradle. This parameter is not used when we build the project on Eclipse.

[...]

PR #1912, opened 2022-01-07
build: moved tests from spotbugs to spotbugs-tests project

iloveeclipse:
This is related to issue #1906 and fixes issue #1914

Tests are simply moved to existing tests project, with the only difference that two unstable tests are executed first now, so other tests don't break them.

----


iloveeclipse:
@KengoTODA : ready to review now. All tests moved, the core project is free from test code now & can be set to be compiled with java 1.8 release (to be done via #1906).

ThrawnCA (approved):

spotbugs-tests/build.gradle
```diff
@@ -14,21 +18,52 @@ dependencies {
...
 }
 
 tasks.named('jacocoTestReport', JacocoReport).configure {
+  dependsOn('unstableTest')
   additionalSourceDirs.setFrom files(project(':spotbugs').sourceSets.main.java.srcDirs)
   additionalClassDirs.setFrom files(project(':spotbugs').sourceSets.main.output.classesDirs)
+  executionData.setFrom files("$buildDir/jacoco/unstableTest.exec", "$buildDir/jacoco/test.exec")
 }
 
+// Tests below fail if executed with other tests
```
KengoTODA, line 46: It's strange, could you share how they fail?

KengoTODA (commented):
Let me confirm our objective: is it to build SpotBugs with Java 8?

In my understanding, even with this change we need to build the `spotbugs-tests` with Java 11+, so we cannot use Java 8 to build the whole projects.

spotbugs-tests/build.gradle
iloveeclipse: see https://github.com/spotbugs/spotbugs/issues/1906#issuecomment-1007739204

iloveeclipse:
> Let me confirm our objective: is it to build SpotBugs with Java 8?
> 
> In my understanding, even with this change we need to build the `spotbugs-tests` with Java 11+, so we cannot use Java 8 to build the whole projects.

The goal was not to hava Java 8 for **all** spotbugs projects, but to be able to use it for the core project.
So we can use any "modern" API in tests and keep core part safe.

Another goal is to cleanly separate core project from unnecessary dependencies / tests so we have one project with tests, not two and have freedom in the main one.

castro285 (approved):

iloveeclipse:
@KengoTODA :  any objections to merge?

KengoTODA (approved):
LGTM, thanks for your patience and sorry for my laziness :)

Issue #1415, opened 2021-01-31
Spotbugs as alternative to Lombok.


lifeinwild:
Lombok is a great attempt, but some people don't like it. I was looking at ideas around this and I thought static analysis could be alternative to Lombok. This proposal may attract new users.

There are six rules. But I thought this is minimal. If you divide it, it will be difficult to understand the meaning.

```
import java.util.Objects;

/**
 * Spotbugs as alternative to Lombok.
 *
 * Lombok is used for auto-generation of hashcode(), equals() and toString() to automatically support to subsequent class updates about members.
 * If Spotbugs detects missing updates about the methods Spotbugs will be an alternative to Lombok.
 *
 * my proposal has 6 rules that couldn't be split, for its meaning.
 * - MM_MEMBER_DOESNT_APPEAR_IN_HASHCODE: class defines member variable doesn't appear in its hashCode().
 * - MM_MEMBER_DOESNT_APPEAR_IN_EQUALS: class defines member variable doesn't appear in its equals().
 * - MM_MEMBER_DOESNT_APPEAR_IN_TOSTRING: class defines member variable doesn't appear in its toString().
 * - CL_NO_HASHCODE_UNLIKE_ABSTRACT_PARENT_CLASS: class doesn't define hashCode() unlike abstract parent class that defines hashCode().
 * - CL_NO_EQUALS_UNLIKE_ABSTRACT_PARENT_CLASS: class doesn't define equals() unlike abstract parent class that defines equals().
 * - CL_NO_TOSTRING_UNLIKE_ABSTRACT_PARENT_CLASS: class doesn't define toString() unlike abstract parent class that defines toString().
 */
public class ForgetToUpdateHashcodeEqualsToString {
    /**
     * i1 appear in the methods.
     * it means that programmer updated the methods when added i1 to this class.
     */
    private int i1 = 1;

    /**
     * i2 doesn't appear in the methods.
     * it means that programmer forgot to update the methods when added i2 to this class.
     *
     * MM_MEMBER_DOESNT_APPEAR_IN_HASHCODE
     * MM_MEMBER_DOESNT_APPEAR_IN_EQUALS
     * MM_MEMBER_DOESNT_APPEAR_IN_TOSTRING
     *
     * if class defines the methods, i2 should appear in the methods.
     * if class doesn't define the methods, i2 should be no problem in these rules.
     */
    private int i2 = 2;

    /**
     * transient member should be no problem in these rules.
     */
[...]

PR #1952, opened 2022-02-12
PR issue1415 ForgotToUpdateHashCodeEqualsToStringDetector

lifeinwild:
https://github.com/spotbugs/spotbugs/issues/1415

AnalysisRunner and AbstractIntegrationTest was updated because this detector has to lower the priority.
BugCounts in PackageStats.java was updated because it could not handle IGNORE_PRIORITY.

gradlew.bat check
BUILD SUCCESSFUL

gradlew.bat spotlessApply build smoketest
BUILD SUCCESSFUL

----


spotbugs/etc/messages.xml
```diff
@@ -1733,6 +1733,14 @@ factory pattern to create these objects.</p>
       ]]>
     </Details>
   </Detector>
+  <Detector class="edu.umd.cs.findbugs.detect.ForgotToUpdateHashCodeEqualsToStringDetector">
+    <Details>
+		<![CDATA[
+			<p> forgot to update hashCode() or equals() or toString().
```
ThrawnCA, line 1739: This is not enough detail. Please describe exactly what it is detecting, and how someone should look at addressing it.

spotbugs/etc/messages.xml
```diff
@@ -3283,6 +3291,66 @@ equal objects have equal hashCodes.
...
 fix this problem of highest importance.
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="HE_NO_EQUALS_UNLIKE_PARENT_OR_OUTER_CLASS_OR_INTERFACE">
+    <ShortDescription>There is no equals() unlike its parent or outer class or interface.</ShortDescription>
+    <LongDescription>There is no equals() unlike its parent or outer class or interface.</LongDescription>
+    <Details>
+<![CDATA[
+  <p>No equals() unlike its parent or outer class or interface.</p>
```
ThrawnCA, line 3302: Does this take into account the possibility of using Apache Commons classes like `EqualsBuilder.reflectionEquals`?

Also, this needs more detail, not just a single short sentence.

spotbugs/etc/messages.xml
```diff
@@ -3283,6 +3291,66 @@ equal objects have equal hashCodes.
...
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="HE_MEMBER_DOESNT_APPEAR_IN_EQUALS">
+    <ShortDescription>Instance member doesn't appear in equals().</ShortDescription>
+    <LongDescription>Instance member doesn't appear in equals().</LongDescription>
+    <Details>
+<![CDATA[
+  <p>
+  Instance member doesn't appear in equals().
```
ThrawnCA, line 3330: This is not a desirable rule. Detecting equality does not necessarily have to include every member of a class.

spotbugs/etc/messages.xml
```diff
@@ -3283,6 +3291,66 @@ equal objects have equal hashCodes.
...
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="HE_MEMBER_DOESNT_APPEAR_IN_HASHCODE">
+    <ShortDescription>Instance member doesn't appear in hashCode().</ShortDescription>
+    <LongDescription>Instance member doesn't appear in hashCode().</LongDescription>
+    <Details>
+<![CDATA[
+  <p>
+  Instance member doesn't appear in hashCode().
```
ThrawnCA, line 3341: This is not a desirable rule.

spotbugs/etc/messages.xml
```diff
@@ -3283,6 +3291,66 @@ equal objects have equal hashCodes.
...
+  </p>
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="USELESS_STRING_MEMBER_DOESNT_APPEAR_IN_TOSTRING">
+    <ShortDescription>Instance member doesn't appear in toString().</ShortDescription>
```
ThrawnCA, line 3347: This is not a desirable rule. Not every member of a class will necessarily be publicly exposed.

ThrawnCA (changes requested):
Several of these rules are not desirable, and others are insufficient documented.

I'm also unsure of their value when HE_EQUALS_NO_HASHCODE and HE_EQUALS_USE_HASHCODE and HE_HASHCODE_NO_EQUALS and HE_HASHCODE_USE_OBJECT_EQUALS already exist.

lifeinwild:
thank you for review. 
I will improve the document.

> HE_EQUALS_NO_HASHCODE and HE_EQUALS_USE_HASHCODE and HE_HASHCODE_NO_EQUALS and HE_HASHCODE_USE_OBJECT_EQUALS

these rules are for the contract of equals and hashcode.
So it only detects when one side is present.
If both equals and hashcode are missing, there is no contract violation.
And these rules cannot detect incomplete implementations, such as some member variables not being used.

The bug patterns of ForgotToUpdateHashCodeEqualsToStringDetector are not based on any contract, but 2 rules of thumb.

1. equals and hashcode often must use all member variables if it is overridden, except transient fields.

However, some classes don't require equals and hashcode. So don't require it to all classes. 
How do we distinguish between classes that require implementations of equals and hashcode and those that don't?
It can be distinguished by the No.2 rule.

If equals or hashcode is overridden in abstract class, it means that equivalence comparison is chosen, not identity comparison, as a design decision. 
It also means that equals and hashcode are expected to be used in the use cases of the abstract class.

2. if an abstract class overrides equals and hashcode, the concrete classes must also do so.

In the following two scenarios, the existing detectors don't detect, but ForgotToUpdateHashCodeEqualsToStringDetector should detects.

Scenario1: Suppose an abstract class that has some member variables and implements equals and hashcode. 
A programmer created a concrete class that has some member variables and inherit the abstract class. 
But he forgot to implement equals and hashcode in the concrete class. 

Scenario2: He implemented equals and hashcode properly in the concrete class. But one day he added a member variable to the concrete class and forgot to update equals and hashcode.
the equals and hashcode are now incomplete implementation only with a part of member variables.

Lombok: Some people don't like it.
EqualsBuilder: Slow. https://stackoverflow.com/questions/5038204/apache-commons-equals-hashcode-builder 
    And some code is not allowed to invoke reflection for security.
Objects: It requires this detector because each member variable is described.
IDE code generation: It requires this detector because each member variable is described.

ThrawnCA:
> _1. equals and hashcode often must use all member variables if it is overridden, except transient fields._

This is not a reliable enough rule of thumb for me to be comfortable enforcing it as a SpotBugs rule.

> _2. if an abstract class overrides equals and hashcode, the concrete classes must also do so._

Again, this is not always true. Subclasses will not necessarily change things in a way that alters equality.

lifeinwild:
This detector may produce many false positives. But bugs of equals and hashcode tend to be hard to find and serious. There is a trade-off between the cost of checking for false positives and the risk of bugs. So this detector uses EXP_PRIORITY. Users who want to see bugs of EXP_PRIORITY are prioritizing reducing bugs even if increasing false positives.

ThrawnCA:
> facts that make a prediction of the trend of the rule of thumb No.1:
> 
>     * @EqualsAndHashCode of Lombok

Actually, that Lombok feature specifically allows you to select fields for inclusion or exclusion. Does this rule take that into consideration?

>     * EqualsBuilder.reflectionEquals
> 
>     * equals and hashcode by IDE code generation of default setting

Those merely provide useful defaults. There is a very big difference between "this is a helpful pattern in many cases, so let's provide a shortcut" vs "this is enforced on all code."

lifeinwild:
**Patterns of equals**
1. No use of equals. So nothing is written.
2. Identity comparison. So it is not overridden.
3. Comparison only by id field or such a single field. So only a single member is used in equals.
4. **Dexterous implementation of equals depending the current use of the class**. So some members are used in equals. 
5. Full equivalence comparison. So all members are used in equals.

**For 1 and 2**, no problem.

**For 3**, use id.equals() instead of the equals of the class. 
```
obj1.getId().equals(obj2.getId()).
list.add(obj.getId());
list.contains(obj.getId());
map.put(obj.getId(), obj);
map.put(obj.getId(), anotherObj);
```

**For 4**, it is likely to be an anti-pattern. It will be difficult to maintain.
Because equals is used in a variety of situations, a dexterous implementation of equals only for a situation is anti-pattern in my opinion.
_Try to implement semantics that can withstand unknown uses._

**For 5**, it is the main case of the detector. 
Such equals are useful in test cases of Junit because assertEquals depends on equals().
The following is my thoughts about transient in this case.

_You can use transient to specify an exclude field._
Of course, that also affects serialization.
Serialization is used for communication and persistence.

_Q: Is a field that isn't used in equals() ignorable for communication and persistence?_
If it is ignorable, you will be comfortable specifying transient for excluding from equals.

Paradoxically, if equals() is used to determine the success or failure of communication or persistence, it should be ignorable.
1. When an object is communicated, if equals() can be performed on the object on sender and the object on receiver, the return value of equals() must mean that communication is successful or not.
2. When an object is read from DB, if equals() is performed on the objects before and after it is recorded in DB, the return value of equals() must mean that persistence is successful or not.

Such equals() is useful in test cases of DB or communication. So it is natural that equals() is used to determine the success or failure of communication or persistence.

If some of the member variables are not used in equals(), it is possible that the persistence fails even though the equals() returns true on the objects before and after it is recorded in DB.
Transient has a run-time effect only when DB and communication are involved. So if DB and communication are not involved, transient will be no problem. If DB and communication are involved, equals should means that.

Apparently, I implicitly expect the concept of equals as satisfying these requirements.
How many codes fit this idea? I don't know. But I think it is natural that code is written this way.

I looked into other similar ideas. There is a static analysis tool called **EqualsVerifier**.

This part is a little similar to my opinion.
https://jqno-nl.translate.goog/equalsverifier/manual/ignoring-fields/?_x_tr_sl=en&_x_tr_tl=ja&_x_tr_hl=ja&_x_tr_pto=op,sc
> By default, EqualsVerifier assumes that all (non-static) fields of your class participate in equals and hashCode.

EqualsVerifier does not allow equals to depend on mutable fields.
https://jqno-nl.translate.goog/equalsverifier/manual/immutability/?_x_tr_sl=en&_x_tr_tl=ja&_x_tr_hl=ja&_x_tr_pto=op,sc
> Mutability: equals depends on mutable field foo.

My detector allows mutable fields.

1. equals allows mutable fields.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)
> It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return true or consistently return false, provided no information used in equals comparisons on the objects is modified.

2. Almost all equals in the world depend on mutable fields. The implementation of IDE, Lombok, EqualsBuilder, etc.

KengoTODA:
I'll close this inactive PR. Please recreate a new one if necessary.

PR #1976, opened 2022-03-09
Added new detector for NUM09-J: Do not use floating-point variables as loop counters

adrianturtoczki:
Added new detector for NUM09-J: Do not use floating-point variables as loop counters.
Using floating-point variables should not be used as loop counters, as they are not precise, which may result in incorrect loops. A loop counter is a variable that is changed with each iteration and controls when the loop should terminate. It is decreased or increased by a fixed amount each iteration.
See rule <a href="https://wiki.sei.cmu.edu/confluence/display/java/NUM09-J.+Do+not+use+floating-point+variables+as+loop+counters">NUM09-J</a>.
----


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
```diff
@@ -0,0 +1,28 @@
...
+
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.StatelessDetector;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import java.util.stream.Stream;
+
+public class DontUseFloatsAsLoopCounters extends OpcodeStackDetector implements StatelessDetector {
+
+    BugReporter bugReporter;
```
KengoTODA, line 13: let's make this field `private final`.

KengoTODA (changes requested):
almost LGTM, thanks for your proposal!

KengoTODA (approved):

KengoTODA (dismissed):

KengoTODA (approved):

PR #1994, opened 2022-03-22
Created a new detector for rule MET04-J

sajtizsolt:
I've created a new detector for SEI CERT rule [MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods).

This detector should prevent the happening of increasing accessibility of methods from the super class. This can occur in three ways:
- increasing the accessibility of a `package private` method to `protected`
- increasing the accessibility of a `package private` method to `public` 
- increasing the accessibility of a `protected` method to `public`

A noteable exception is the `Cloneable` interface: when a class implements the `clone` method, it should increase it's accessiblity from `protected` to `public`.

My changes:
- [X] Updated `CHANGELOG.md`
- [X] Created a new detector called `FindIncreasedAccessibilityOfMethods` in `spotbugs` module
- [X] Created a test for the new detector called `FindIncreasedAccessibilityOfMethodsCheckTest` in `spotbugs-tests` module
- [X] Created test cases under `spotbugs/spotbugsTestCases/src/java/increasedAccessibilityOfMethods`
- [X] Added the required descriptions to `findbugs.xml` and `messages.xml`

My tests:
- I've ran my detector on the codebase of `Jenkins` to make sure it works properly.
- I've ran the `./gradlew spotlessApply build smoketest` command locally


ThrawnCA (dismissed):

sajtizsolt:
Thanks for the review @ThrawnCA ! Could you please give me a workflow approval as well? 🙏

KengoTODA (changes requested):
Added several comments, please check them.

Also, could you consider adding one more test, to verify that we can detect cases that publish methods provided not by direct superclass but by other ancestors, such as a parent of the superclass, or the default method defined in an interface?

sajtizsolt:
Hey @KengoTODA ! Thanks for you review, I've made the requested changes and also added some test cases (I've made small adjustments on the detector as well).

ThrawnCA:
Are malicious subclasses really a serious concern? There are reasons that the Security Manager is going away...

KengoTODA (approved):
LGTM, thanks for your contribution!

sajtizsolt:
I had to resolve conflicts to be able to merge. Please re-approve my PR and the workflow, thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,123 @@
...
+        List<Method> superClassMethods = new ArrayList<>();
+        for (JavaClass javaClass : superClasses) {
+            for (Method method : javaClass.getMethods()) {
+                if (isAccessibleFromSubClass(method)) {
+                    superClassMethods.add(method);
+                }
+            }
+        }
+
+        for (Method superClassMethod : superClassMethods) {
```
ThrawnCA, line 66: Isn't this double-handling? They could have been processed on line 61, instead of iterating through them again.

sajtizsolt:
@ThrawnCA @KengoTODA Please check and approve my PR and workflow once again (and hopefully for the last time :smile:). Thank you in advance!

ThrawnCA (approved):

sajtizsolt:
@ThrawnCA @KengoTODA Hi! I don't want to urge you, I'm just interested - is there any chance that my PR will be included in a future release?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Bump ObjectWeb ASM from 9.2 to 9.3 supporting JDK 19 ([#2004](https://github.com/spotbugs/spotbugs/pull/2004))
 
 ### Added
+* New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
```
KengoTODA, line 20: When you resolve the conflict, please move this CHANGELOG entry under the new `## Unreleased`.

KengoTODA: You've changed documents (messages.xml) in this PR, so please add one more entry for `### Changed` to tell such documentation updates.

KengoTODA (dismissed):
Please fetch the latest upstream branch, and resolve the conflict with it.

CHANGELOG.md
sajtizsolt: Done!

spotbugs/etc/messages.xml
```diff
@@ -124,7 +124,7 @@ This plugin contains all of the standard SpotBugs detectors.
   <BugCategory category="STYLE">
     <Description>Dodgy code</Description>
     <Abbreviation>D</Abbreviation>
-    <Details>Code that is confusing, anomalous, or
```
ThrawnCA, line 127: This seems better capitalised.

spotbugs/etc/messages.xml
```diff
@@ -245,7 +247,7 @@ applied to method parameters and uses of those method parameters. </p>
   <Detector class="edu.umd.cs.findbugs.detect.VarArgsProblems">
     <Details>
 <![CDATA[
-<p> Looks for problems with arising from Java 5 varargs.</p>
```
ThrawnCA, line 248: This is better as "Java 5".

spotbugs/etc/messages.xml
```diff
@@ -489,7 +491,7 @@ another package. </p>
   <Detector class="edu.umd.cs.findbugs.detect.EmptyZipFileEntry">
     <Details>
 <![CDATA[
-<p> This looks for creation of empty ZIP file entries. It is a moderately fast detector.</p>
```
ThrawnCA, line 492: This is better capitalised.

spotbugs/etc/messages.xml
```diff
@@ -1485,7 +1487,7 @@ factory pattern to create these objects.</p>
       <![CDATA[
       <p>
       This detector looks for final classes that declare protected members. As this
-      class cannot be derived from, the use of protected access for members is
```
ThrawnCA, line 1488: This is better as "cannot".

ThrawnCA (changes requested):
This seems to be reverting a bunch of typo fixes.

spotbugs/etc/messages.xml
sajtizsolt: I left the merging to my IDE which caused a bit of chaos. I looked at it manually, hopefully it's all good now!

spotbugs/etc/messages.xml
```diff
@@ -1670,7 +1668,7 @@ factory pattern to create these objects.</p>
     <Details>
 <![CDATA[
 <p> Checks for calls to methods which perform a byte to String (or String to byte) conversion using the user's default
- platform encoding. This can cause the application behaviour to vary between platforms.  </p>
```
ThrawnCA: This was already a valid (UK) spelling.

ThrawnCA (approved):

sajtizsolt:
@ThrawnCA 
Hey! I resolved the conflicts with the `master` branch and reviewed my changes. If I'm right, I resolved all previous merge issues and unnecessary changes, so I would be grateful if you would check my PR once more and approve it if you think it's right. Should I resolve the previous conversations or it is the task of the starter? Also, could you please give me a workflow approval as well?

ThrawnCA (approved):

sajtizsolt:
@KengoTODA @ThrawnCA I've rebased to the latest master, squashed my previous commits into one, amended my commit message to one which follows the conventional commit rules. Please check this PR one last time, give me a workflow approval and merge if you think that my changes are right too. Thanks in advance and thank you for your work!

spotbugs/etc/messages.xml
```diff
@@ -8785,11 +8785,29 @@ Using floating-point variables should not be used as loop counters, as they are
       </p>]]>
     </Details>
   </BugPattern>
+
+  <BugPattern type="IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY">
+    <ShortDescription>Accessibility of an overridden or hidden method was increased.</ShortDescription>
+    <LongDescription>Accessibility of an overridden or hidden method {1} was increased from {2} to {3}.</LongDescription>
+    <Details>
+      <![CDATA[<p>
+        Increasing the accessibility of overridden or hidden methods permits a malicious subclass to offer wider
```
ThrawnCA, line 8794: How many programs might actually be used in a context where malicious subclasses are a realistic threat? Applets are long out of fashion...

sajtizsolt: In my opinion, when developers are creating methods with an explicit visibility modifier, they do it for a reason. If a subclass overrides this behavior, then it ignores the intentions of the creator of the base class which, can lead to bugs and code which is hardly interpretable. Anyway, it is a very bad practice.

spotbugs/etc/messages.xml
```diff
@@ -8785,11 +8785,29 @@ Using floating-point variables should not be used as loop counters, as they are
       </p>]]>
     </Details>
   </BugPattern>
+
+  <BugPattern type="IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY">
+    <ShortDescription>Accessibility of an overridden or hidden method was increased.</ShortDescription>
+    <LongDescription>Accessibility of an overridden or hidden method {1} was increased from {2} to {3}.</LongDescription>
```
JuditKnoll, line 8791: I think the `an` is unnecessary here, since the method is explicitly named.

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
...
 
 ## Unreleased - 2023-??-??
+### Added
+- New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
+
 ### Changed
 - Bump up Apache Commons BCEL to the version 6.6.1 ([#2223](https://github.com/spotbugs/spotbugs/pull/2223))
 - Bump up slf4j-api to 2.0.3 ([#2220](https://github.com/spotbugs/spotbugs/pull/2220))
 - Bump up gson to 2.10 ([#2235](https://github.com/spotbugs/spotbugs/pull/2235))
+- Updated documentation (findbugs.xml, messages.xml) because of the introduction of the `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY` bug type.
```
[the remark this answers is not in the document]
sajtizsolt: I was requested earlier to add this line by one of the maintainers, but I will be happy to remove it.

spotbugs/etc/messages.xml
sajtizsolt: Nice catch! I appreciate precision like that!

CHANGELOG.md
```diff
@@ -86,6 +90,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 * New detector `DontUseFloatsAsLoopCounters` to detect usage of floating-point variables as loop counters (`FL_FLOATS_AS_LOOP_COUNTERS`), according to SEI CERT rules [NUM09-J. Do not use floating-point variables as loop counters](https://wiki.sei.cmu.edu/confluence/display/java/NUM09-J.+Do+not+use+floating-point+variables+as+loop+counters)
 * New test detector `ViewCFG` to visualize the control-flow graph for `SpotBugs` developers
 
+### Added
+* New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
+
```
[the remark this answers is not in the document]
sajtizsolt: Sheesh, that should have been some funny rebasing incident - thanks for noticing! 🙏

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.bcel.classfile.Method;
+
+import java.util.Arrays;
+
+public class FindIncreasedAccessibilityOfMethods implements Detector {
+
+    private final BugReporter bugReporter;
+
+    private JavaClass subClass;
+    private JavaClass directSuperClass;
```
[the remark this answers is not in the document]
sajtizsolt: Reasonable, I changed it!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsCheckTest.java
```diff
@@ -0,0 +1,84 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+public class FindIncreasedAccessibilityOfMethodsCheckTest extends AbstractIntegrationTest {
```
JuditKnoll, line 12: Can you add some tests with classes which has template parameters?

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,116 @@
...
+            for (Method superClassMethod : javaClass.getMethods()) {
+                if (isAccessibleFromSubClass(superClassMethod, subClass, directSuperClass)) {
+                    for (Method subClassMethod : subClass.getMethods()) {
+                        if (isOverridable(superClassMethod, subClassMethod) && !isCloneMethodFromCloneable(subClass, superClassMethod)) {
+                            if (superClassMethod.isProtected() && subClassMethod.isPublic()) {
+                                bugReporter.reportBug(new BugInstance(this, "IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY", NORMAL_PRIORITY)
+                                        .addClassAndMethod(subClass, subClassMethod)
+                                        .addString(subClassMethod.getName())
+                                        .addString("protected")
+                                        .addString("public"));
```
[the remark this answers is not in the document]
sajtizsolt: Honestly, I don't know if this would be useful information. The class and method already precisely define the method which is problematic. Could you please describe a use case when it would be useful, if you insist on it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,116 @@
...
+    public FindIncreasedAccessibilityOfMethods(BugReporter bugReporter) {
+        this.bugReporter = bugReporter;
+    }
+
+    @Override
+    public void visitClassContext(ClassContext classContext) {
+        JavaClass directSuperClass;
+        JavaClass subClass;
+        JavaClass[] superClasses;
+        subClass = classContext.getJavaClass();
```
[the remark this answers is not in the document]
sajtizsolt: True! ✔️

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsCheckTest.java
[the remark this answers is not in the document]
sajtizsolt: Of course! ✔️

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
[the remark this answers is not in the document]
JuditKnoll: By adding source line infos you can specify more precisely the location of the bugreport. Adding the method info already adds source line info for the entire method, which is a wider range of sourcelines. You can add the first sourceline, but it can be a bit problematic, when you only have the header and no body for the function. It can be okay like this, if you prefer it this way 😄

spotbugs/etc/messages.xml
```diff
@@ -8968,11 +8976,28 @@ Using floating-point variables should not be used as loop counters, as they are
         ]]>
     </Details>
   </BugPattern>
+  <BugPattern type="IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY">
+    <ShortDescription>Accessibility of the overridden or hidden method was increased.</ShortDescription>
+    <LongDescription>Accessibility of the overridden or hidden method {1} was increased from {2} to {3}.</LongDescription>
```
JuditKnoll, line 8981: Are you sure, that the message is correct? I think the indexes are one off, since the `addClassAndMethod()` uses the `addClass()` and the `addMethod()` functions separately. If you look at the annotationList of a bug instance, you will see, that it has five elements, not four.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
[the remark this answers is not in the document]
sajtizsolt: Thanks for the clarification! I would like to keep it this way then. 😄

spotbugs/etc/messages.xml
sajtizsolt: Thanks, you were right. I removed the unnecessary strings. If my PR seems to be okay now, can you please give me an approve, or maybe start a workflow? Also, thanks again for your hard work reviewing my changes!

sajtizsolt:
Good point, thanks for trying it! As the SEI CERT rule definition doesn't mention anything specific about constructors, I think you are right, I removed constructors from the analysis - also added some tests to validate if there are no false positive cases of them. If it is okay now, could you please give me a workflow approval?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,123 @@
...
+public class FindIncreasedAccessibilityOfMethods implements Detector {
+
+    private final BugReporter bugReporter;
+
+    public FindIncreasedAccessibilityOfMethods(BugReporter bugReporter) {
+        this.bugReporter = bugReporter;
+    }
+
+    @Override
+    public void visitClassContext(ClassContext classContext) {
```
JuditKnoll, line 43: In the current implementation there is a hit type, which may cause problems for the users, so we may need filter it out: we have three classes: SuperClass, Class, and SubClass, SuperClass is parent of Class, which is parent of SubClass. We have a method, which is declared in SuperClass as protected, overriden in both Class and SubClass as public. There will be a hit in Class, which is great, but there will be also a hit in SubClass, about which I'm not so sure. If the user have no write access to SuperClass and Class, then can't really do anything about the problem.

JuditKnoll (changes requested):
Rerun the detector on bigger projects, the links are the same as before, overwrote the earlier results.

sajtizsolt:
@JuditKnoll Can I run CodeChecker with the demo account? I really can't find any way to do it.

JuditKnoll:
> @JuditKnoll Can I run CodeChecker with the demo account? I really can't find any way to do it.

The demo account doesn't have rights to store the results. If you would like to run it yourself, you need to set up a separate codechecker server and configure the runs, which is not so straightforward.
I rerun the tests, the links are the same as earlier.

sajtizsolt:
> > @JuditKnoll Can I run CodeChecker with the demo account? I really can't find any way to do it.
> 
> The demo account doesn't have rights to store the results. If you would like to run it yourself, you need to set up a separate codechecker server and configure the runs, which is not so straightforward. I rerun the tests, the links are the same as earlier.

Thank you! I examined the results and they seem to be valid positive cases to me. What do you think? Can this detector be considered as complete or are there any more edge cases which comes into mind?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
JuditKnoll: If the user provided the necessary compiled files for the analysis, even if doesn't have write access to them, cause e.g. they are inside a jar file. Checked every hit of the tests on the bigger projects, but didn't find any test like the one, I mentioned. (Otherwise I would have referred it.) You can create a test by simply creating a class, which descends from one of the hits and overrides the appropriate function, or reusing your already existing test classes, using `SuperClassOfSuperClass` as SuperClass in my example, `SuperClass` as Class, and `SubClassFromSamePackage` as SubClass. If you simply add an override of `superPackagePrivateMethodToPublic()` to `SuperClass`, then you have the test:
```
@Override
public String superPackagePrivateMethodToPublic() {
    return "SuperClass.superPackagePrivateMethodToPublic";
}
```

IMO in this case, the error should be only reported in `SuperClass`, and not in `SubClassFromSamePackage`.

sajtizsolt: I will, I'm just a littly busy nowadays. I will tag you if it happens!

sajtizsolt: @JuditKnoll Now I thought about this problem again. To achieve the desired behaviour, we should be able to filter out potential bugs based on the bugs already found while we analyzed the super classes of the current class. I don't know, if there is some logic implemented in SpotBugs which enables us to query this list of bugs based on a class, but if there is not, then we should analyze every single super class again. I don't know, I feel that this would increase the complexity of the detector by a huge scale. Do you feel, that the added value of this solution complies with the amount of extra computational demand of it?

sajtizsolt: Anyway, I've added the requested changes in another commit - take a look at them. Not terribly slow if there are few super classes.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,121 @@
...
+        performAnalysis(
+                "increasedAccessibilityOfMethods/CloneableImplementation.class");
+
+        assertContainsCorrectNumberOfIAOMBugs(0);
+    }
+
+    @Test
+    public void findNoIAOMBugInClass_IncreasedAccessibilityOfMethods_InterfaceImplementation() {
+        performAnalysis(
+                "increasedAccessibilityOfMethods/InterfaceImplementation.class");
```
JuditKnoll, line 64: Isn't Interface missing from here?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            superClasses = subClass.getSuperClasses();
+        } catch (ClassNotFoundException e) {
+            AnalysisContext.reportMissingClass(e);
+            return iaomBugs;
+        }
+
+        List<IAOMBug> superClassBugs = new ArrayList<>();
+        for (JavaClass superClass : superClasses) {
+            superClassBugs.addAll(findIAOMBugs(superClass));
```
JuditKnoll, line 71: It does work, but it could be a bit more efficient, see my comment at the `visitClassContext()` method.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
JuditKnoll: Answering your question about filtering out bugs based on other bugs, there is a solution using `BugAccumulator`, but it's a bit different to what you need. As far as I know, that one only works inside one class, not across classes.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
sajtizsolt: It was missing, thanks!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+        assertContainsCorrectNumberOfIAOMBugs(0);
```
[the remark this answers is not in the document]
sajtizsolt: Yeah, I think it should not. The best practice with interfaces is to don't use accessibility modifiers - not because package private accessibility is important. but because it doesn't matter. I suspect, that you are not sharing my idea.

JuditKnoll:
Seems like the build is failing, can you please run `./gradlew :spotbugs:spotlessApply` command to solve the spotless problems?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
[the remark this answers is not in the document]
JuditKnoll: Thanks, yes, you are right.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,122 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.Test;
```
JuditKnoll, line 6: The project got migrated from junit to junit jupiter. Can you please change the import to `org.junit.jupiter.api.Test`? After merging to master, it shouldn't cause any issues, since all the other configs are taken care of.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.not;
+
+public class FindIncreasedAccessibilityOfMethodsTest extends AbstractIntegrationTest {
```
JuditKnoll, line 13: With using junit jupiter, the `public` modifier is no longer necessary for the Test classes. Can you please remove it to make the code more conform to the codebase?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Test
+    public void findIAOMBugInClass_IncreasedAccessibilityOfMethods_SubClassFromSamePackage() {
```
JuditKnoll, line 16: With using junit jupiter, the `public` modifier is no longer necessary for the Test methods. Can you please remove it from here and from the other tests to make the code more conform to the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public void visitClassContext(ClassContext classContext) {
+        JavaClass subClass = classContext.getJavaClass();
+        for (IAOMBug bug : findIAOMBugs(subClass)) {
+            bugReporter.reportBug(new BugInstance(this, BUG_TYPE, NORMAL_PRIORITY)
+                    .addClassAndMethod(bug.clazz, bug.method)
+                    .addString(bug.oldAccessibility)
+                    .addString(bug.newAccessibility));
+    }}
```
JuditKnoll, line 57: Can you please modify the formatting here to be conform to the formatting rules in the project? (`gradlew spotlessApply` should do the trick.)

JuditKnoll (changes requested):
Can you please rebase to the current master and resolve the merge conflicts?
Since your last rebase the project got migrated from Junit to Junit Jupiter (see https://github.com/spotbugs/spotbugs/pull/2605), which will cause build fail after the rebase, but I commented at the relevant lines, which needs to be modified to work properly.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+
+        return iaomBugs.stream().filter(bug -> !containsBug(superClassBugs, bug)).collect(Collectors.toList());
+    }
+
+    private List<Method> getMethodsExceptConstructors(JavaClass javaClass) {
+        ArrayList<Method> methods = new ArrayList<>();
+        for (Method method : javaClass.getMethods()) {
+            if (!Const.CONSTRUCTOR_NAME.equals(method.getName())) {
```
JuditKnoll, line 103: I think you can also filter out the static initializer.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethods.java
```diff
@@ -0,0 +1,156 @@
...
+
+        IAOMBug(JavaClass clazz, Method method, String oldAccessibility, String newAccessibility) {
+            this.clazz = clazz;
+            this.method = method;
+            this.oldAccessibility = oldAccessibility;
+            this.newAccessibility = newAccessibility;
+        }
+
+        boolean isSameMethodAndAccessibility(IAOMBug other) {
+            return Objects.equals(method.getSignature(), other.method.getSignature())
```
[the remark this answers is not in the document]
JuditKnoll: The `Method` class itself has an equals, which checks both the name of the method and the signature. Is there any reason, why you don't use that equals?

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        assertContainsCorrectNumberOfIAOMBugs(3);
+
+        assertContainsIAOMBug("SuperClassWithOverride", "superProtectedMethodToPublicOverriddenInSuperClass");
+        assertContainsIAOMBug("SuperClassWithOverride", "superPackagePrivateMethodToProtectedOverriddenInSuperClass");
+        assertContainsIAOMBug("SubClassWithRepeatedOverride", "superPackagePrivateMethodToProtectedOverriddenInSuperClass");
+        assertNotContainsIAOMBug("SubClassWithRepeatedOverride", "superProtectedMethodToPublicOverriddenInSuperClass");
+    }
+
+    private void assertContainsCorrectNumberOfIAOMBugs(int numberOfBugs) {
```
JuditKnoll, line 97: Some assert functions got introduced to `AbstractIntegrationTest` to minimize code repetition. Can you please use `assertBugTypeCount()` instead of this one and remove this function once it's not used?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,120 @@
...
+        BugInstanceMatcher bugTypeMatcher = new BugInstanceMatcherBuilder()
+                .bugType("IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY").build();
+        assertThat(getBugCollection(), containsExactly(numberOfBugs, bugTypeMatcher));
+    }
+
+    private void assertContainsIAOMBug(String className, String methodName) {
```
JuditKnoll, line 103: Some assert functions got introduced to `AbstractIntegrationTest` to minimize code repetition. Can you please use `assertBugInMethod()` instead of this one and remove this function once it's not used?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindIncreasedAccessibilityOfMethodsTest.java
```diff
@@ -0,0 +1,120 @@
...
+        BugInstanceMatcher bugInstanceMatcher = new BugInstanceMatcherBuilder()
+                .bugType("IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY")
+                .inClass(className)
+                .inMethod(methodName)
+                .build();
+        assertThat(getBugCollection(), hasItem(bugInstanceMatcher));
+    }
+
+    private void assertNotContainsIAOMBug(String className, String methodName) {
```
JuditKnoll, line 112: Some assert functions got introduced to `AbstractIntegrationTest` to minimize code repetition. Can you please use `assertNoBugInMethod()` instead of this one and remove this function once it's not used?

JuditKnoll (changes requested):

spotbugs/etc/messages.xml
```diff
@@ -9388,6 +9408,62 @@ Using floating-point variables should not be used as loop counters, as they are
       </p>]]>
     </Details>
   </BugPattern>
+  <BugPattern type="AT_NONATOMIC_64BIT_PRIMITIVE">
+    <ShortDescription>This write of this 64-bit primitive variable may not atomic</ShortDescription>
```
ThrawnCA, line 9412: "may not atomic" -> "may not be atomic"

spotbugs/etc/messages.xml
```diff
@@ -9388,6 +9408,62 @@ Using floating-point variables should not be used as loop counters, as they are
...
+    <LongDescription>This write of the 64-bit primitive variable "{2.name}" may not atomic</LongDescription>
+    <Details>
+      <![CDATA[
+        <p>
+          The long and the double are 64-bit primitive types, and depending on the Java Virtual Machine implementation assigning a value to them can be treated as two separate 32-bit writes, and as such it's not atomic.
```
ThrawnCA, line 9417: Should have a comma after "implementation". Also, the sentence should probably be split to avoid double "and". Perhaps change the first comma to a fulls top.

spotbugs/etc/messages.xml
```diff
@@ -9388,6 +9408,62 @@ Using floating-point variables should not be used as loop counters, as they are
...
+          See <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.7">JSL 17.7. Non-Atomic Treatment of double and long</a>.
+          See SEI CERT rule <a href="https://wiki.sei.cmu.edu/confluence/display/java/VNA05-J.+Ensure+atomicity+when+reading+and+writing+64-bit+values">VNA05-J. Ensure atomicity when reading and writing 64-bit values</a> for more info.
+
+          This bug can be ignored in platforms which guarantee that 64-bit long and double type read and write operations are atomic.
+        </p>
+        <p>
+          To fix it, declare the variable volatile, change the type of the field to the corresponding atomic type from <code>java.lang.concurrent.atomic<\code> or correctly synchronize the code.
```
ThrawnCA, line 9424: "volatile" should be monospaced.

spotbugs/etc/messages.xml
```diff
@@ -9388,6 +9408,62 @@ Using floating-point variables should not be used as loop counters, as they are
...
+          Declaring the variable volatile may not be enough in some cases: e.g. when the variable is assigned a value which depends on the current value or on the result of nonatomic compound operations.
```
ThrawnCA, line 9425: "e.g." does not need a colon, comma would be more appropriate.

ThrawnCA (changes requested):

hazendaz:
Closing as #3979 superseded this and has been applied for 4.10.0 release

PR #2009, opened 2022-04-10
Update the JKS file to sign eclipse plugin jar file

### LINKED ISSUE #2008: Certificate expired
The certificate used to sign plugin is expired, see screenshot:
image

KengoTODA:
close #2008
I also updated `GPG_SECRET_PASSPHRASE` and `KEYSTORE_PASS` in GitHub Secrets.

Note that the new certificate is valid for 90 days only. And I do not want to handle this operation so frequently.

Better to purchase a valid certificate that can reduce this frequency. It is also possible to keep a machine running `certbot` and automate the process with the [GitHub Secrets API](https://docs.github.com/en/rest/reference/actions#secrets).

Both need money, probably can we ask some org to sponsor SpotBugs? But [GitHub Sponsors for org accounts](https://docs.github.com/en/sponsors/receiving-sponsorships-through-github-sponsors/setting-up-github-sponsors-for-your-organization) needs a bank account as org, so it is hard for us SpotBugs team I believe.


iloveeclipse:
@KengoTODA : I personally don't think we should change old certificate if the new one will expire in 90 days. That makes no sense for me, just unnecessary effort.

Another possibility would be to sign that with pgp, but that would probably force use of "modern" Eclipse.
If that would be only needed for signing, would be OK to have "unsigned" plugin for older Eclipse versions (that don't understand pgp) and pgp signed for modern Eclipse.

@mickaelistria : could you provide some insights how that can be done & if that can be still done for spotbugs plugin that should also work (can be installed) with older Eclipse releases?

KengoTODA:
> That makes no sense for me, just unnecessary effort.

But the old passphrases (value of GitHub Secret) have been lost, we need to merge this PR to keep the build working in other branches...

iloveeclipse:
@KengoTODA : I honestly have no idea about all this. If that is something we **must** do, I'm OK, but I can't understand what is changed now and why is something lost/can't be build? The certificate is expired 2019, why is the problem appearing now?

KengoTODA:
> Another possibility would be to sign that with pgp, but that would probably force use of "modern" Eclipse.

If necessary, we can use the [next major release](https://github.com/spotbugs/spotbugs/milestone/15) to drop support for "legacy" Eclipse.
There is no plan to launch the next major release yet, but it would be good timing to consider, the [last major release v4 was two years ago already](https://github.com/spotbugs/spotbugs/releases/tag/4.0.0).

KengoTODA:
> The certificate is expired 2019, why is the problem appearing now?

I have no idea. Why/How you found the problem?

iloveeclipse:
@KengoTODA : I don't understand this:
> we need to merge this PR to keep the build working in other branches...

If we don't merge this PR, how it can affect building?

The problem itself in #2008 appears because the "Trust" dialog was improved to clearly explain issues with certification. I assume the problem was there longer, just not so prominently shown. This dialog appears if one imstalls spotbugs into Eclipse, not during build.

KengoTODA:
> If we don't merge this PR, how it can affect building?

The build cannot decrypt `spotbugs.jks.gpg` file, and the build job will be marked as failed.

iloveeclipse:
> The build cannot decrypt spotbugs.jks.gpg file, and the build job will be marked as failed.

And **why** the build is not able to decrypt that file? It was working so far? What is changed now, why is this PR  needed?

KengoTODA:
I suggest we merge this PR first and make a decision about how we support "legacy" Eclipse during the coming 90 days.

I haven't investigated the PGP-based sign yet, but it sounds nice and valuable to switch.

iloveeclipse (approved):
While I still can't understand why this PR is required to fix the build that was working fine beforehand, I surely won't break the build.
So feel free to merge if needed.

KengoTODA:
> > The build cannot decrypt spotbugs.jks.gpg file, and the build job will be marked as failed.
> 
> And **why** the build is not able to decrypt that file? It was working so far? What is changed now, why is this PR needed?

Because I've updated `GPG_SECRET_PASSPHRASE` and `KEYSTORE_PASS` that is necessary for the decrypt process. I did not memorize these values used in the previous key generation, so I always generate these passphrases randomly.

I've overwritten GitHub Secrets so previous values have been lost from GitHub.

iloveeclipse:
> it's possible to generate a self-signed x509 certificate and just use that too. I tried that with these instructions:
> 
> https://lists.gnupg.org/pipermail/gnupg-devel/2011-March/025989.html
> 
> The user will be prompted to approve such an "unrooted" certificate in the same way as is the case for a PGP key, and even in the same dialog. On the plus side, this works on all versions of Eclipse and you can keep doing the build how you are doing it now.

@KengoTODA : Ed mentioned that above - this seem to produce certificates that will be valid for years and would not require any changes in the build except the new certificate used. 
WDYT?

mickaelistria:
>  could you provide some insights how that can be done & if that can be still done for spotbugs plugin that should also work (can be installed) with older Eclipse releases?

If you want the "unsigned" popup to not show up in older Eclipse releases, your only option is to keep using jarsigner. Nothing can be done to improve old versions, they're to stay the same forever.
However, I would recommend any project to drop support (as in "stop caring") for older Eclipse versions if none of the contributors really needs them for their own goals. Dropping older release will allow your project to focus on creating more value to people who upgrade, and will also benefit Eclipse project as it will influence more users in getting rid of their old Eclipse IDE in favor of a newer one.

KengoTODA:
I'm OK to change the method to sign, but could you handle it on your side? I'm not so sure about the recent Eclipse, and I'm not a user of Eclipse either, so cannot be so sure about the fix. Thanks for your consideration!

iloveeclipse:
@KengoTODA : it turned out, pgp support is not mature yet / requires latest Eclipse platform, so I don't think it is appropriate to change the method now for our plugin used in older releases too.

Issue #2024, opened 2022-04-21
URLClassLoader leak in PluginLoader


gtoison:
PluginLoader creates some URLClassLoader and never closes them. When the URLClassLoader points to a .jar file, the JVM process keeps a reference to that file (a RandomAccessFile instance) preventing from deleting/renaming that .jar file.
This is an issue when the JVM process is long lived, for instance when it's running inside the Gradle daemon (see https://github.com/spotbugs/sonar-findbugs/issues/128)

The RandomAccessFile is created as soon as the classloader opens the .jar file, which for me was when looking for the non-existing resource `messages_en_GB.xml`. The URL does not exist because the .jar does not contain it so calling `JarURLConnection.setUseCaches()` wouldn't help here.

To sum up, this code reproduces the issue: omitting the `close()` on the last line creates a lock on the plugin jar file until the JVM process is stopped.

    URLClassLoader classLoader = new URLClassLoader(new URL[] {workingPluginFile.toURI().toURL()});
    classLoader.findResource("messages_en_GB.xml");
    classLoader.close();

gtoison:
This can be closed now that #2025 is merged

PR #2025, opened 2022-04-21
fix: let Plugin implement AutoCloseable

gtoison:
Proposed fix for https://github.com/spotbugs/spotbugs/issues/2024

When a custom plugin is loaded from a .jar file PluginLoader creates URLClassLoader and never closes it. This leaves a lock (a RandomAccessFile) on the .jar file and it cannot be deleted or renamed until the JVM process stops.

Keep track of the URLClassLoader created by the PluginLoader so we can close them when we are done with the Plugin

It is up to the code using these custom plugins to call `.close()` once the analysis is completed. The current code uses static initialization for the core plugin so I couldn't find a clean way (for instance using try-with-resources blocks) to close the plugins.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## Unreleased - 2022-??-??
 ### Changed
 - Updated documentation by adding parenthesis `()` to the negative odd check message ([#1995](https://github.com/spotbugs/spotbugs/issues/1995))
+- Let the Plugin class implement AutoCloseable so we can release the .jar file ([#2024](https://github.com/spotbugs/spotbugs/issues/2024))
 
 ### Fixed
 - Bumped Saxon-HE from 10.6 to 11.3 ([#1955](https://github.com/spotbugs/spotbugs/pull/1955), [#1999](https://github.com/spotbugs/spotbugs/pull/1999))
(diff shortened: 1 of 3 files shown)


ThrawnCA (approved):

KengoTODA:
Confirmed that `Plugin : PluginLoader` is `1 : 1` so closing parent (`PluginLoader`) from child (`Plugin`) would have less problem. 

I guess we have several places to call `Plugin.close()` in the current codebase, [like this](https://github.com/KengoTODA/spotbugs/blob/b98085b48bcc7aff68547af5f8b9ed75b101d0e9/eclipsePlugin/src/de/tobject/findbugs/FindbugsPlugin.java#L313) and `FindBugs2.dispose()`, I will recheck again later.

gtoison:
It might be better to "close" the plugin inside the static method `Plugin.removeCustomPlugin()` since all the clean-up code is already there.

It seems to me that the lifecycle of the plugin is not the same as the lifecycle of `FindBugs2` class: it seems reasonable to load some plugins, run a several analysis, then unload the plugins.
It would be great to close the Plugin loader as soon as the plugin is loaded but there's for instance a call to `Plugin.getClassLoader` in `SAXBugCollectionHandler` so I think that the `PluginLoader` needs to remain alive during the analysis.

In any case you guys know the codebase better than I do, so let me know if this PR needs to be amended

KengoTODA (approved):

PR #2055, opened 2022-05-16
LCK08-J. Ensure actively held locks are released on exceptional conditions

gonczmisi:
This PR is about SEI CERT rule: [LCK08](https://wiki.sei.cmu.edu/confluence/display/java/LCK08-J.+Ensure+actively+held+locks+are+released+on+exceptional+conditions)

First of all, the first part of the rule is unreleased locks on exceptional conditions. This part is already solved by a detector: `FindUnreleasedLock`. I wrote a unit test to demonstrate it.

The second part of the rule, is to try to catch exceptional conditions, which could end up unlocking a resource, without even locking it in the first place.
My PR is about this issue.
As mentioned, there is already a detector for unreleased locks, so I extended that one:
* Provided unit tests about each scenario
* Introduced new bug type: `COW_CLOSED_WITHOUT_OPENED`
* Provided informational descripton about the bug

I ran the new build on the [RxJava](https://github.com/ReactiveX/RxJava) source code to test if the new feature in `FindUnreleasedLock` would provide a lot of false positives or not.
No bug with type `CWO_CLOSED_WITHOUT_OPENED` was reported.


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueFrame.java
```diff
@@ -37,20 +37,31 @@ public class ResourceValueFrame extends Frame<ResourceValue> {
...
 
+    /**
+     * The resource is closed without being opened in the first place.
+     */
+    public static final int CLOSED_WITHOUT_OPENED = 3;
+
     /**
      * The resource is closed (or unlocked, etc).
      */
-    public static final int CLOSED = 3;
```
KengoTODA, line 43: better to keep public APIs compatible. Can we move new constants after the `NONEXISTENT`?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Fixed False positives for `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE` on try-with-resources with interface references ([#1931](https://github.com/spotbugs/spotbugs/issues/1931))
 
+### Added
+* New bug type `CWO_CLOSED_WITHOUT_OPENED` for locks that might be released without even being acquired. (See[SEI CERT rule LCK08-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK08-J.+Ensure+actively+held+locks+are+released+on+exceptional+conditions))
+
```
KengoTODA, line 13: Please merge the latest `master` branch in the upstream repo, then move this entry to under the `## Unreleased`.

It is also suggested to add a link to the related PR or issue, like `([#2055](https://github.com/spotbugs/spotbugs/pull/2055))`, at the end of this CHANGELOG entry.

KengoTODA (changes requested):
Thanks for your contribution, could you check my comments?

gonczmisi:
> Thanks for your contribution, could you check my comments?

Sure thing, I will look into them in the next week. Thanks for the feedback :+1:

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueFrame.java
gonczmisi: Yes, will do.

gonczmisi: Unfortunately no, because the mergeing of these states are based on a `Math.min()` call. This would mean, that a `CLOSED` would override a `CLOSED_WITHOUT_OPENED` state. Then the information that the lock was not even acquired in the first place would be lost.

gonczmisi: Hi @KengoTODA could you revisit this PR and check my comment above?

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueAnalysis.java
```diff
@@ -106,6 +106,9 @@ public void meetInto(ResourceValueFrame fact, Edge edge, ResourceValueFrame resu
                 // If status is OPEN, downgrade to OPEN_ON_EXCEPTION_PATH
                 tmpFact = modifyFrame(fact, null);
                 tmpFact.setStatus(ResourceValueFrame.OPEN_ON_EXCEPTION_PATH);
+            } else {
+                tmpFact = modifyFrame(fact, null);
```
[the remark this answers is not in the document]
gonczmisi: I wouldn't move the `tmpFact = modifyFrame(fact, null);` outside of the `if` block. That would mean the separation of two commands that belong together.

JuditKnoll (changes requested):
Can you please add more test cases? What happens if the release of the locks is after the try-catch block, outside of it?

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueFrame.java
JuditKnoll: Then it is a breaking change, and I think only should be merged for 5.0.0, together with #2069, which may cause some merge conflicts with this one. Also, it would be worth to mention the breaking change in the changelog as well.

ThrawnCA (approved):

ThrawnCA (approved):

JuditKnoll:
Added it to https://github.com/spotbugs/spotbugs/milestone/15, as this PR contains some breaking changes.

JuditKnoll:
Superseded by https://github.com/spotbugs/spotbugs/pull/3385

Issue #2042, opened 2022-05-06
MS_EXPOSE_REP cannot detect nested class


Belle-PL:
Hi, I found a false negative about the rule [MS_EXPOSE_REP](). Please read the following case, PMD should have reported a warning in line 5,but it doesn't. I think this is caused by the nest class. Thanks for your consideration.

```java
public class C1 {
  private static String[] s2;
  public class C2 {
    public String[] getS2() {
      return s2;  // should report a warning here
    }
  }
}
```
Version: Latest

ThrawnCA:
> PMD should have reported

I think you mean SpotBugs?

Belle-PL:
@ThrawnCA Hi, sorry for the mistake. I have fixed it 😸

baloghadamsoftware:
Thank you for reporting this. I am just checking what I can do here.

PR #2056, opened 2022-05-18
Extend detector `FindReturnRef` to detect exposition of references and buffers in nested and inner classes

baloghadamsoftware:
This patch fixes issue #2042

----


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -115,153 +123,206 @@ public void visit(Method obj) {
...
-        if (check) {
+        if (check || record) {
             super.visit(obj);
         }
     }
 
     @Override
     public void sawOpcode(int seen) {
 
-        if (!check) {
```
ThrawnCA, line 126: Why was this guard clause flipped? It's altered the indentation of the entire (very long) method, which makes it harder to identify the real changes.

JuditKnoll: This guard clause got flipped back, so it's not a problem anymore for the github diff to handle. IMO the code itself was a  bit more logical, but this way it's easy to see the changes.

ThrawnCA (approved):

ThrawnCA (approved):

JuditKnoll (approved):
This PR seems to solve the referenced issue.

ThrawnCA (approved):

ThrawnCA (approved):

ThrawnCA (approved):

ThrawnCA (approved):

ThrawnCA (approved):

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -80,13 +84,16 @@ public class FindReturnRef extends OpcodeStackDetector {
...
+    private final Map<XMethod, XField> accessMethodFields = new HashMap<>();
+
     private final BugAccumulator bugAccumulator;
 
     private static final Matcher BUFFER_CLASS_MATCHER = Pattern.compile("Ljava/nio/[A-Za-z]+Buffer;").matcher("");
     private static final Matcher DUPLICATE_METHODS_SIGNATURE_MATCHER =
             Pattern.compile("\\(\\)Ljava/nio/[A-Za-z]+Buffer;").matcher("");
     private static final Matcher WRAP_METHOD_SIGNATURE_MATCHER =
             Pattern.compile("\\(\\[.\\)Ljava/nio/[A-Za-z]+Buffer;").matcher("");
+    private static final Matcher ACCESS_METHOD_MATCHER = Pattern.compile("access\\$\\d{3}").matcher("");
```
ThrawnCA, line 96: `Matcher` instances are not thread-safe. This should store the `Pattern` and then create a `Matcher` for each string, not reset the same Matcher.

ThrawnCA (changes requested):

JuditKnoll:
Taken over by https://github.com/spotbugs/spotbugs/pull/3174.

Issue #2065, opened 2022-05-24
Spotbugs gradle build failing


HoelzelJon:
I'm trying to build spotbugs locally as described in the [contributing guideline](https://github.com/spotbugs/spotbugs/blob/master/.github/CONTRIBUTING.md), but I'm getting this error when running `./gradlew clean build`:

```
> Task :eclipsePlugin:compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':eclipsePlugin:compileJava'.
> Could not resolve all files for configuration ':eclipsePlugin:compileClasspath'.
   > Could not find org.eclipse.platform:org.eclipse.swt.cocoa.macosx.aarch64:3.111.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/eclipse/platform/org.eclipse.swt.cocoa.macosx.aarch64/3.111.0/org.eclipse.swt.cocoa.macosx.aarch64-3.111.0.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project :eclipsePlugin > org.eclipse.platform:org.eclipse.swt:3.111.0
```

As the error says, the `3.111.0` version of the dependency doesn't seem to exist in maven: https://repo.maven.apache.org/maven2/org/eclipse/platform/org.eclipse.swt.cocoa.macosx.aarch64/

I am on a M1 Max Macbook, running MacOS Monterey (Version 12.2.1). I suspect the fact I'm on an ARM mac may be the issue, since there is a x86 3.111.0 version available in maven: https://repo.maven.apache.org/maven2/org/eclipse/platform/org.eclipse.swt.cocoa.macosx.x86_64/

I'm working around this by excluding all eclipsePlugin tasks from the build, which works for me since I'm not touching the eclipsePlugin modules:
```
./gradlew clean build -x :eclipsePlugin:compileJava -x :eclipsePlugin-junit:compileTestJava -x :eclipsePlugin-test:compileJava -x :eclipsePlugin-test:checkstyleMain
```

I imagine this could be fixed by bumping some dependencies (so that `eclipsePlugin` depends on a version that _does_ exist in maven), but I haven't looked further since the workaround is fine for my purposes.

gtoison:
This should be fixed by #2488 and the continous integration build now runs on Mac thanks to #2507

PR #2066, opened 2022-05-24
Add support for Avro Nullable annotation

HoelzelJon:
A project I work on makes frequent use of the apache avro Nullable annotation (`org.apache.avro.reflect.Nullable`). Adding support for the avro nullable annotation would significantly reduce the number of false positive warnings generated by spotbugs.

----


HoelzelJon:
I am unable to run the full build as described in the contribution guidelines (see https://github.com/spotbugs/spotbugs/issues/2065), but everything seems to work fine when I skip the eclipsePlugin tasks.

KengoTODA (approved):

spotbugsTestCases/src/fakeAnnotations/org/apache/avro/reflect/Nullable.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
ThrawnCA, line 2: SpotBugs doesn't belong to Apache. Why is this copyright notice in use?

HoelzelJon: Sorry for the slow response

I copied the file from the Apache avro source, following what seems to have been done for the `android.support.annotation` classes (which use the same license).

I'm no expert on how this license works; if you think it's more reasonable to remove the license block here, I'd be happy to do so.

ThrawnCA: Ah. That does make sense.

JuditKnoll:
It seems like during the merge, this feature in the CHANGELOG.md got to the wrong place, not to the unreleased section, but into 4.7.2.

hazendaz:
Thank you will get that fixed.  We need a better way on change logs.  Anything aging is doomed to merge conflict or get placed incorrectly.

PR #2063, opened 2022-05-23
Disabled detector `ThrowingExceptions` by default (fixes #2040)

iloveeclipse:
This avoids many false positives reported via
https://github.com/spotbugs/spotbugs/issues/2040

PR #2067, opened 2022-05-24
Avoid unnecessary false positives generated by the recently introduce…

big-andy-coates:
Avoid unnecessary false positives generated by the recently introduced `THROWS_METHOD_THROWS_CLAUSE_BASIC_EXCEPTION` and `THROWS_METHOD_THROWS_CLAUSE_THROWABLE` checks.

These were throwing even for methods on synthetic classes, i.e. for methods generated by the compiler. For example:

```java
interface SomeInterface() {
  void doIt() throws Exception();  // Method throwing exception
}

private void accept(SomeInterface thing) {
  ...
}

private void test() {
  accept(() -> {});     // Previous red herring due to synthetic class creation
}
```

Previously, not only did the above code create an error for `SomeInterface.doIt()`, as it throws `Exception`, it also created an error on the `accept(() -> {})` line, due to the synthetic class
the compiler is generating to handle the lambda. This synthetic class will also have a method with signature `void doIt() throws Exception()`, causing the issue.

Methods on synthetic classes are now excluded.

The following false positives still exist (which is why I still think this check should be disabled https://github.com/spotbugs/spotbugs/pull/2063):

1. Generic exceptions in the signatures of overriding methods.

If an overriding method follows the signature of the throwing declaration in the superclass then a bug report is still raised. Really, the issue should be raised on the superclass declaration of the method (if part of the code being analysed).

2. Generic exceptions in the signature of methods that make calls to methods that throw generic exceptions.

```java
public void myOtherMethod throws Exception {
  doTheThing();  // this method throws Exception
}
```

To compile the method must throw the generic exception, so this should not be a bug.

3. Finally, given this check would report on bug on JDK types like `Callable`, as it `throws Exception`, I'm struggling to understand how this check can be considered valid at all.  When defining a library interface like `Callable` it seems reasonable to be able to mark the method as being allowed to throw any exception it likes, checked or otherwise.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ThrowingExceptionsTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ThrowingExceptionsTest.java
@@ -10,7 +10,7 @@ import edu.umd.cs.findbugs.AbstractIntegrationTest;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
 
-public class CheckThrowingExceptions extends AbstractIntegrationTest {
+public class ThrowingExceptionsTest extends AbstractIntegrationTest {
     @Test
     public void throwingExceptionsTests() {
         performAnalysis("MethodsThrowingExceptions.class");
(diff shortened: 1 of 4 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ThrowingExceptionsTest.java
```diff
@@ -10,7 +10,7 @@
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
 
-public class CheckThrowingExceptions extends AbstractIntegrationTest {
+public class ThrowingExceptionsTest extends AbstractIntegrationTest {
```
big-andy-coates, line 13: I've renamed this to be consistent with other test classes...

KengoTODA (approved):
LGTM, thanks for your PR!

ThrawnCA (approved):

ThrawnCA (approved):

iloveeclipse:
> @KengoTODA is this PR something Spotbugs is interested in? 

Sure.

> If not, then please feel free to close. I don't want to have to manage merge issues going forward.

But we all work in our spare time on this project. So sometimes it's not easy to find a free slot if there are other more important items on the table.

Issue #1927, opened 2022-01-18
False negatives on IO_INEFFICIENT_INDEX_OF


gloNelson:
Hi,
I thought it should be false negatives on IO_INEFFICIENT_INDEX_OF when using SpotBugs to analyze my project.
> IO_INEFFICIENT_INDEX_OF: This code passes a constant string of length 1 to String.indexOf(). It is more efficient to use the integer implementations of String.indexOf(). f. e. call myString.indexOf('.') instead of myString.indexOf(".")
```
void testIIO_INEFFICIENT_INDEX_OF(int index){
    StringBuilder sb = new StringBuilder();
    int lastIndex = sb.indexOf(">", index);
}
```

It seems that SpotBugs does not do the same analysis on the StringBuilder class. Is it more efficient to use StringBuilder indexOf('.') than StringBuilder indexOf(".")？
Similar to #1928

tomyy:
StringBuilder has no `indexOf(int)` or `indexOf(int, int)` method and therefore `sb.indexOf('>', index)` doesn't even compile.

SpotBugs must not recommend code that doesn't compile

gloNelson:
StringBuilder has `indexOf(String str, int fromIndex)` method and therefore `sb.indexOf('>', index)` can be compile.
Here is the [StringBuilder Class Doc](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html).
You can copy and paste the test case and compile it to execute SpotBugs.

iloveeclipse:
As said, there is no plans to change anything, because that's not an issue.

PR #2074, opened 2022-05-28
Issue1927 fixed

hzy20001211:
----

Issue1927 fixed

https://github.com/spotbugs/spotbugs/issues/1927


iloveeclipse:
Issue #1927 does not need a fix, it is invalid, so this PR doesn't make sense.

Issue #1767, opened 2021-10-22
New False Negative for DM_STRING_CTOR


Belle-PL:
I found a false negative for this rule. I think we should consider this API because its internal implementation is to invoke new String(java.lang.String). This program can lead to lower performance.

The rule link: [link](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#dm-method-invokes-inefficient-new-string-string-constructor-dm-string-ctor). Please refer to the following code sample:
```java
public void test(StringBuffer buf) {
    String str = new String(buf);  // Need report warning
}
```
My SpotBugs version is 4.4.2 and I think it needs report a warning for the line 2, but no expected outcome. So, this is a false negative. Thanks for your consideration.

iloveeclipse:
As said, there is no plans to change anything, because that's not an issue.

PR #2075, opened 2022-05-28
issue 1767 fixed

hzy20001211:
----

Issue1767 fixed

https://github.com/spotbugs/spotbugs/issues/1767


iloveeclipse:
Issue #1767 does not need a fix, it is invalid, so this PR doesn't make sense.

PR #2105, opened 2022-06-27
Run SQ analysis even in the default branch

KengoTODA:
Currently SQ is working only for PR. This change will
enable SQ even in the default branch.

refs https://github.com/spotbugs/spotbugs/discussions/2076#discussioncomment-2871205


ThrawnCA (approved):

baloghadamsoftware (approved):
Good idea!

PR #2112, opened 2022-07-06
Warn for catching RuntimeException, Exception and Throwable

baloghadamsoftware:
According to [SEI CERT rule ERR08-J](https://wiki.sei.cmu.edu/confluence/display/java/ERR08-J.+Do+not+catch+NullPointerException+or+any+of+its+ancestors) catching `RuntimeException`, `Exception` and `Throwable` is a bad practice because it may hide exceptions which must be handled or avoided. This PR extends detector `RuntimeExceptionCapture` to warn for these cases.

----


spotbugs/etc/messages.xml
```diff
@@ -6991,6 +6991,74 @@ does not need to be created, just access the static methods directly using the c
...
+  <p>
+    This method uses a try-catch block that catches Throwable objects, but Throwable is not
+    thrown explicitly within the try block. This is risky, because future changes to signatures
+    of method invoked in the try block could add more exceptions to the list of potential
+    exceptions the caller must handle, or adding method calls into the try block may also throw
+    additional exceptions which must be handled. Additionally, adding code into the try block in
+    the future may introduce the throwing of new runtime exceptions either because of programming
+    errors (e.g. improper checks such as a null pointer check) or intentionally but some of them may
+    require special handling. However, catching Throwable itself hides the problem of not
+    handling these exceptions and not fixing the errors.However, catching Throwable itself hides
```
ThrawnCA, line 7052: Missing space after full stop.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/RuntimeExceptionCapture.java
```diff
@@ -200,8 +216,20 @@ public void visit(CodeException obj) {
                             System.out.println("Dead exception store at " + first);
                         }
                         caughtException.dead = true;
-                        break;
                     }
+
+                    // See if the excpetion is rethrown at the end of the handler.
```
ThrawnCA, line 221: Misspelled "exception"

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/RuntimeExceptionCapture.java
```diff
@@ -200,8 +216,20 @@ public void visit(CodeException obj) {
...
+                    while (block != null) {
+                        InstructionHandle last = block.getLastInstruction();
+                        if (last != null) {
+                            if (last.getInstruction().getOpcode() == Const.ATHROW) {
```
ThrawnCA, line 225: This could be combined with the previous line.

spotbugsTestCases/src/java/RECTest.java
```diff
@@ -171,4 +186,32 @@ public void testPass5() {
             dontTriggerEmptyExceptionHandler();
         }
     }
+
+    // should pass -- catches E, but rethrows it
+    public void testPass6() {
+        try {
+            for (int i = 0; i < 1000; i++)
```
ThrawnCA, line 193: `for` should always use braces.

iloveeclipse (changes requested):
This PR will produce again a lot of false positives. The idea of catching exact the exception types that some code might throw and don’t catch others is purely academic. In most cases exception handlers that catch Exception or Throwable are written because they want catch **all** potential exceptions from the try block, not only some specific. I know lot of absolutely valid code that does that.

So this new checks should be disabled by default.

baloghadamsoftware:
Thank you for your comment @iloveeclipse!

I agree that this PR produces several new bug reports. However, based on my experience most of these reports are true positives. Unfortunately, many programmers are lazy to properly handle exceptions. Beside the true positives there are of course some false positives because you are right, there are some cases where you must catch all the exceptions. The SEI CERT rule itself describes such cases.

Case ERR08-J-EX0 where the exception is rethrown (possibly after logging) is handled. Case ERR08-J-EX2 only affects some special software. Where we can reduce the false positives are the cases described in ERR08-J-EX1: we must check for methods called in the try block whether they are methods of a class received as a generic parameter of the current method or the class. It is a valid case to catch `RuntimeException` (and maybe the other two) when calling such methods. Could you please mention some more cases?

I think that setting a check to disabled by default is the last resort. First we should try fix it by recognizing corner cases and reducing the number of false positives. Only if that effort fails should we disable the checker by default.

ThrawnCA (approved):

baloghadamsoftware:
@iloveeclipse Any comments after seeing the statistics?

spotbugs/etc/messages.xml
```diff
@@ -6991,6 +6991,74 @@ does not need to be created, just access the static methods directly using the c
...
+  <p>
+    This method uses a try-catch block that catches Exception objects, but Exception is not
+    thrown within the try block. This is risky, because future changes to signatures of method
+    invoked in the try block could add more exceptions to the list of potential exceptions the
+    caller must handle, or adding method calls into the try block may also throw additional
+    exceptions which must be handled. Additionally, adding code into the try block in the future
+    may introduce the throwing of new runtime exceptions either because of programming errors
+    (e.g. improper checks such as a null pointer check) or intentionally but some of them may
+    require special handling. However, catching RuntimeException itself hides the problem of not
+    handling these exceptions and not fixing the errors.However, catching Exception itself hides
```
JuditKnoll, line 7008: The space is missing from here to after the full stop, before "However".

spotbugsTestCases/src/java/RECTest.java
```diff
@@ -110,6 +110,21 @@ public void testFail6() {
         }
     }
 
+    // should fail with LOW priority -- catches E, but E not thrown, however e
+    //// is used in the handler
```
JuditKnoll, line 114: There is an unnecessary '//' at the beginning of the line of comment.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/RuntimeExceptionCapture.java
JuditKnoll: This conversation can be resolved, since got fixed in https://github.com/spotbugs/spotbugs/pull/2112/commits/7b0d8382230ded583db874c4f0f0900a5f8505c8.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/RuntimeExceptionCapture.java
JuditKnoll: These two lines got combined in https://github.com/spotbugs/spotbugs/pull/2112/commits/7b0d8382230ded583db874c4f0f0900a5f8505c8, so the conversation can be resolved.

ThrawnCA (approved):

hazendaz:
> e ERR08-J-EX0 where the exception is rethrown (possibly after logging) is handled. Case ERR08-J-EX2 only affects some special software. Where we can reduce the false positives are the cases described in ERR08-J-EX1: we must check for methods called in the try block whether they are methods of a class received as a generic parameter of the current method or the class. It is a valid case to catch `RuntimeException` (and maybe the other two) when calling such methods. Could you please mention some more cases?

I don't think the issue should be active by default either because sometimes its other libraries that abuse usage and that would be a false positive on wrong code bases.  For example, Selenium is really bad about using Exception.  Even worse with Selenium, is that they have unchecked exceptions that should actually be checked exceptions.  Anyways point is that its not always in direct developer control.  I agree we mark them but not as a default setting.  Note: I didn't review this PR so I didn't check to see if that was modified, instead what I did was resolve most of the coding notes others left.

ThrawnCA (approved):

spotbugsTestCases/src/java/REC2Test.java
```diff
@@ -0,0 +1,142 @@
...
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    // should pass -- rethrows Throwable
+    public void testPass2() {
+        try {
+            for (int i = 0; i < 1000; i++)
```
ThrawnCA, line 105: Loops should always use braces.

hazendaz:
@baloghadamsoftware If you have this set as disabled by default, it can come in.  I personally like this but I know many libraries abuse what they catch and then rethrow which would cause the false positive but I personally would enable this to get my staff to be smarter about exception handling.  Its not like sonar already doesn't say same thing ;)  But let users opt into this one.

PR #2119, opened 2022-07-21
Test #1338: false positive try-with-resource on JDK11 and later might fail with RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE

lars-sh:
see #1338

----

- [x] ~~Added an entry into `CHANGELOG.md` if you have changed SpotBugs code~~
(no SpotBugs code changes, just added one more test case)


ThrawnCA (approved):

KengoTODA:
🤔 this test case can pass even we built SpotBugs with Java 17 on GitHub Actions. Is this test case enough to reproduce the problem?

lars-sh:
@KengoTODA, I cannot tell you if there might be similar cases, but at least this is my test case, which I initially used for #1338 and while it failed before it's been fixed now.

In case you need additional help from my side, feel free to ask.

KengoTODA (approved):

KengoTODA:
Thanks for your PR! :)

PR #2123, opened 2022-07-27
Do not disable the security manager on Java 17 VMs and newer as it is deprecated for removal.

### LINKED ISSUE #1579: System::setSecurityManager is deprecated and will be removed in a future JDK release
We have several usage of `System::setSecurityManager` in SpotBugs implementation.

https://github.com/spotbugs/spotbugs/search?q=setSecurityManager&type=

```
spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java:                System.setSecurityManager(null);
spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierValue.java:                        System.setSecurityManager(ValidationSecurityManager.INSTANCE);
spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierValue.java:                    System.setSecurityManager(ValidationSecurityManager.INSTANCE);
```

raphw:
This fixes #1579. If needed, disabling the security manager can still be requested on Java 17 and newer by setting the *edu.umd.cs.findbugs.securityManagerDisabled* property to `false`. Using this property also allows to disable this functionality on older VMs.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
@@ -1,0 +1,65 @@
+package edu.umd.cs.findbugs.util;
+
+import org.apache.commons.lang3.JavaVersion;
+import org.apache.commons.lang3.SystemUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Since Java 17, the security manager is deprecated for removal and invoking related methods
+ * causes a warning to be printed to the console. This intermediate disables the use of
+ * security manager-related APIs on Java 17 or later, unless using the security manager is
+ * explicitly configured by setting the <i>edu.umd.cs.findbugs.securityManagerDisabled</i>
+ * property.
+ */
+public class SecurityManagerHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SecurityManagerHandler.class);
+
+    /**
+     * Determines if the security manager is used by SpotBugs.
+     */
+    public static boolean SECURITY_MANAGER_DISABLED;
+
+    static {
+        boolean securityManagerDisabled;
+        try {
+            String property = System.getProperty("edu.umd.cs.findbugs.securityManagerDisabled");
+            if (property != null) {
+                securityManagerDisabled = Boolean.parseBoolean(property);
+            } else {
+                securityManagerDisabled = SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_17);
+            }
+        } catch (Throwable t) {
+            securityManagerDisabled = false;
+            LOGGER.debug("failed to detect the ability of security manager feature, so treat it as available", t);
+        }
+        SECURITY_MANAGER_DISABLED = securityManagerDisabled;
+    }
+
+    /**
+     * Disables the security manager by setting {@link System#setSecurityManager(SecurityManager)}
+     * to {@code null}.
+     */
+    public static void disableSecurityManager() {
+        if (SECURITY_MANAGER_DISABLED) {
+            return;
+        }
+        doDisableSecurityManager();
+    }
+
+    /**
+     * This method is a safeguard for running this library on a JVM that might no longer include
+     * the security manager API after removal. As the JVM verifies methods lazily, and since this
+     * method will never be invoked, validation of this method with a missing type can never fail.
+     *
+     * As some environments do not support setting the security manager but always return null
+     * when getting it, we check if a security manager is set before disabling it.
+     */
+    private static void doDisableSecurityManager() {
+        SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            System.setSecurityManager(null);
+        }
+    }
+}
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
@@ -19,7 +19,7 @@ public class SecurityManagerHandler {
     /**
      * Determines if the security manager is used by SpotBugs.
      */
-    public static boolean SECURITY_MANAGER_DISABLED;
+    private static final boolean SECURITY_MANAGER_DISABLED;
 
     static {
         boolean securityManagerDisabled;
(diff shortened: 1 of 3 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
```diff
@@ -0,0 +1,59 @@
+package edu.umd.cs.findbugs.util;
+
+/**
+ * Since Java 17, the security manager is deprecated for removal and invoking related methods
+ * causes a warning to be printed to the console. This intermediate disables use security
```
ThrawnCA, line 5: "use security" -> "use of security"

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
raphw: Changed via force push.

ThrawnCA (dismissed):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+                    version = version.substring(2, 3);
+                } else {
+                    int index = version.indexOf(".");
+                    if (index != -1) {
+                        version = version.substring(0, index);
+                    }
+                }
+                securityManagerDisabled = Integer.parseInt(version) > 16;
+            }
+        } catch (Throwable ignored) {
```
KengoTODA, line 35: We're using Java 8 so let's consider to catch multiple concrete types instead of a single abstract `Throwable`:

```suggestion
        } catch (SecurityException | NumberFormatException ignored) {
```

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+            securityManagerDisabled = false;
```
KengoTODA, line 36: It's also nice to have a debug log here for debugging purpose:

```suggestion
            LoggerFactory.getLogger(SecurityManagerHandler.class)
                .debug("failed to detect the ability of security manager feature, so treat it as available", e);
            securityManagerDisabled = false;
```

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
```diff
@@ -0,0 +1,60 @@
...
+                String version = System.getProperty("java.version");
+                if (version.startsWith("1.")) {
+                    version = version.substring(2, 3);
+                } else {
+                    int index = version.indexOf(".");
+                    if (index != -1) {
+                        version = version.substring(0, index);
+                    }
+                }
+                securityManagerDisabled = Integer.parseInt(version) > 16;
```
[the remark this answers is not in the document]
KengoTODA: nitpick: [GraalVM native image does not support SecurityManager](https://www.graalvm.org/22.0/security-guide/#security-manager-and-untrusted-code) even in v11 release, so it would be better to check `java.vm.name` or something to judge.

KengoTODA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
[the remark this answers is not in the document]
raphw: I changed to use the utility. As for Graal, I added a check to only set the security manager to null if one is currently set. This works for Graal and some other platforms like Android, as they still provide compatible API where the security manager is always `null`.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
raphw: I don't quite agree as the check is supposed to handle any type of error to avoid backwards compatibility issues. With the Apache utilities, for example, a `NullPointerException` might be thrown without this being documented. If updating the code, one would need to adjust the catch block without this being obvious or compile-checked. A catch-all seems like the more appropriate option.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
raphw: I added some logging.

KengoTODA:
Please merge the latest `origin/master` a to resolve a conflict in the `CHANGELOG.md` then I'll review your code. Thanks in advance!

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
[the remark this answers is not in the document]
KengoTODA: Please push the change to GitHub 😀

raphw:
Indeed, pushed the changes, including the rebase.

KengoTODA (approved):
LGTM, thank you!

KengoTODA:
Not sure why the build is failing. 🤔

It seems that there are some problem like #2146?

ThrawnCA (dismissed):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/SecurityManagerHandler.java
```diff
@@ -0,0 +1,65 @@
...
+            if (property != null) {
+                securityManagerDisabled = Boolean.parseBoolean(property);
+            } else {
+                securityManagerDisabled = SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_17);
+            }
+        } catch (Throwable t) {
+            securityManagerDisabled = false;
+            LOGGER.debug("failed to detect the ability of security manager feature, so treat it as available", t);
+        }
+        SECURITY_MANAGER_DISABLED = securityManagerDisabled;
```
KengoTODA, line 37: SpotBugs reports a potential bug here, please consider to make this static field `final`.

```
edu.umd.cs.findbugs.util.SecurityManagerHandler.SECURITY_MANAGER_DISABLED isn't final but should be
This static field public but not final, and could be changed by malicious code or by accident from another package. The field could be made final to avoid this vulnerability.
```

KengoTODA (changes requested):

KengoTODA (approved):

Issue #2145, opened 2022-08-19
CFGBuilderException when assigning double or long in groovy constructor


tmoschou:
This minimal Groovy class

`src/main/groovy/Class1.groovy`
```groovy
class Class1 {
    final double field1
    Class1(double field1) {
        this.field1 = field1
    }
}
```

with Java class main method entry point and maven `pom.xml`

`src/main/java/Main.java`
```java
public class Main {
    public static void main(String[] args) {
        new Class1(0);
    }
}
```

<details>
  <summary><code>pom.xml</code></summary>

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.github.tmoschou</groupId>
    <artifactId>bug-repro-spotbugs</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>11</maven.compiler.source>
        <maven.compiler.target>11</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <groovy.version>4.0.4</groovy.version>
        <spotbugs.version>4.7.1</spotbugs.version>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.apache.groovy</groupId>
            <artifactId>groovy-all</artifactId>
            <version>${groovy.version}</version>
            <type>pom</type>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>com.github.spotbugs</groupId>
                <artifactId>spotbugs-maven-plugin</artifactId>
                <version>4.7.1.1</version>
                <configuration>
                    <maxRank>1</maxRank>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>check</goal>
                        </goals>
                    </execution>
                </executions>
                <dependencies>
                    <dependency>
                        <groupId>com.github.spotbugs</groupId>
                        <artifactId>spotbugs</artifactId>
                        <version>${spotbugs.version}</version>
                    </dependency>
                </dependencies>
            </plugin>

            <plugin>
                <groupId>org.codehaus.gmavenplus</groupId>
[...]

Issue #1386, opened 2020-12-09
CFGBuilderException when assigning a long value in a constructor and running groovy 2.5.13


andrewhall-premier:
When running groovy 2.5.13, I receive the following exception when running `spotbugsMain` on a Class which assigns a long value in its constructor:

```
  Detector edu.umd.cs.findbugs.detect.FindUnrelatedTypesInGenericContainer caught exception while analyzing TestClass.<init> : (J)V
    edu.umd.cs.findbugs.ba.CFGBuilderException: Invalid stack at   27: lload[22](2) 5 when checking   32: putfield[181](3) 32
      At edu.umd.cs.findbugs.ba.BetterCFGBuilder2.isPEI(BetterCFGBuilder2.java:1032)
      At edu.umd.cs.findbugs.ba.BetterCFGBuilder2.build(BetterCFGBuilder2.java:914)
      At edu.umd.cs.findbugs.ba.BetterCFGBuilder2.build(BetterCFGBuilder2.java:777)
      At edu.umd.cs.findbugs.classfile.engine.bcel.CFGFactory.analyze(CFGFactory.java:94)
      At edu.umd.cs.findbugs.classfile.engine.bcel.CFGFactory.analyze(CFGFactory.java:65)
      At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.analyzeMethod(AnalysisCache.java:368)
      At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.getMethodAnalysis(AnalysisCache.java:321)
      At edu.umd.cs.findbugs.ba.ClassContext.getMethodAnalysis(ClassContext.java:995)
      At edu.umd.cs.findbugs.ba.ClassContext.getMethodAnalysisNoDataflowAnalysisException(ClassContext.java:980)
      At edu.umd.cs.findbugs.ba.ClassContext.getCFG(ClassContext.java:299)
      At edu.umd.cs.findbugs.detect.FindUseOfNonSerializableValue.analyzeMethod(FindUseOfNonSerializableValue.java:143)
      At edu.umd.cs.findbugs.detect.FindUseOfNonSerializableValue.visitClassContext(FindUseOfNonSerializableValue.java:95)
      At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
      At edu.umd.cs.findbugs.FindBugs2.lambda$null$1(FindBugs2.java:1108)
[...]

PR #2146, opened 2022-08-19
Add failing test case for class generated by Groovy compiler

tmoschou:
This PR adds a failing test case for Groovy class that is causing Spotbugs to throw `edu.umd.cs.findbugs.ba.CFGBuilderException`. It does not fix the issue. Refer to #2145

* Groovy compilation support is added to `spotbugsTestCases/build.gradle`
* `AbstractIntegrationTest` has been cleaned up / refactored to additionally search for `.class` files generated by the Groovy compiler. At the same time simplifies some fall-back search path logic (e.g. presumably left over code for nonexistent `java9/` directory) and switches use of `java.io.File` over to the `java.nio.file` package instead.

----

  _**NOTE: No fix in this PR**_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Bumped gson from 2.9.0 to 2.9.1 ([#2136](https://github.com/spotbugs/spotbugs/pull/2136))
 - Fixed InvalidInputException in Eclipse while bug reporting ([#2134](https://github.com/spotbugs/spotbugs/issues/2134))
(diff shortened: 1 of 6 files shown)


ThrawnCA (approved):

KengoTODA (approved):
Thanks for your PR! 
<del>I confirmed that the bug comes from 771f26facba9effee104834562803a5361880494 which assumes the size of field is always 1. I will wait for review from other teammates to merge this fix and your test case.</del> I had misunderstanding about the intention of existing code, I will reconsider my fix.

KengoTODA (approved):
Added a new fix that treats `dup_x1` and `dup_x2` in a better way.

KengoTODA:
I believe this PR can close https://github.com/spotbugs/spotbugs/issues/2145. This PR also close https://github.com/spotbugs/spotbugs/issues/1386.

ThrawnCA (approved):

PR #2156, opened 2022-08-29
Use stable SLF4J API instead of unstable 1.8

KengoTODA:
SLF4J released a stable API v2.0.0, so migrate our dependency on unstable 1.8 with it.
It's a kind of drop-in replacement so we do not have to update our code, in theory.

----


ThrawnCA (approved):

gtoison:
Apparently there are some breaking changes in slf4j 2.0 so it's not always a drop in replacement (see #2176)

Issue #2126, opened 2022-07-30
False positive for FL_FLOATS_AS_LOOP_COUNTERS


DanySK:
Relevant code:
```java
double y = Double.NaN
while (Double.isNaN(y) && someOtherCondition) {
    ...
    y = Double.parseDouble(...);
    ...
}
```

Reported:

#### FL_FLOATS_AS_LOOP_COUNTERS: Do not use floating-point variables as loop counters

Using floating-point variables should not be used as loop counters, as they are not precise, which may result in incorrect loops. A loop counter is a variable that is changed with each iteration and controls when the loop should terminate. *It is decreased or increased by a fixed amount each iteration*.

See rule [NUM09-J](https://wiki.sei.cmu.edu/confluence/display/java/NUM09-J.+Do+not+use+floating-point+variables+as+loop+counters).

---

However, there is no fixed increment, so I think it does not apply here.

Bytecode:
[FollowTarget.zip](https://github.com/spotbugs/spotbugs/files/9226167/FollowTarget.zip)

DanySK:
Another false positive here:
```java
    protected static WeibullDistribution weibullFromMean(
            final double mean,
            final double deviation,
            final RandomGenerator random
    ) {
        final double t = FastMath.log((deviation * deviation) / (mean * mean) + 1);
        double kmin = 0, kmax = 1;
        while (Gamma.logGamma(1 + 2 * kmax) - 2 * Gamma.logGamma(1 + kmax) < t) {
            kmin = kmax;
            kmax *= 2;
        }
        double k = (kmin + kmax) / 2;
        while (kmin < k && k < kmax) { // WARNING
            if (Gamma.logGamma(1 + 2 * k) - 2 * Gamma.logGamma(1 + k) < t) {
                kmin = k;
            } else {
                kmax = k;
            }
            k = (kmin + kmax) / 2;
        }
        final double shapeParameter = 1 / k;
        final double scaleParameter = mean / FastMath.exp(Gamma.logGamma(1 + k));
        return new WeibullDistribution(
                random,
                shapeParameter,
                scaleParameter,
                WeibullDistribution.DEFAULT_INVERSE_ABSOLUTE_ACCURACY
        );
    }
```

atnak:
Here's another vastly simplified false positive:
```java
double total = 0;

for (int attempt = 1;; attempt++) {
	total += RandomUtils.nextDouble(); // FL_FLOATS_AS_LOOP_COUNTERS

[...]

PR #2167, opened 2022-09-06
Fix of detector `DontUseFloatsAsLoopCounters` to prevent false positives

baloghadamsoftware:
Bug `FL_FLOATS_AS_LOOP_COUNTERS` was reported for several cases where floating-point types appear in the loops but not as loop counters. See issue ([#2126](https://github.com/spotbugs/spotbugs/issues/2126)).

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
@@ -8,6 +8,7 @@ import edu.umd.cs.findbugs.StatelessDetector;
 import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


baloghadamsoftware:
The build failure is not caused by this PR. Also `master` fails on the same test (`Issue2145`). Yesterday it was OK, and this test was added 3 weeks ago so the bug should be somewhere between yesterday's and today's commits.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
```diff
@@ -6,23 +6,104 @@
...
-                        Const.DLOAD_2, Const.DLOAD_3).anyMatch(x -> x == getCodeByte(getBranchTarget())) && (getBranchTarget() < getPC())) {
+                getBranchTarget() < getPC() &&
+                checkLoopEnd() && checkLoopStart(getBranchTarget())) {
             bugReporter.reportBug(new BugInstance(this, "FL_FLOATS_AS_LOOP_COUNTERS", NORMAL_PRIORITY)
                     .addClassAndMethod(this).addSourceLine(this, getBranchTarget()));
         }
     }
+
+    private boolean checkLoopEnd() {
+        if (!FLOAT_STORERS.contains(getPrevOpcode(1))) {
```
ThrawnCA, line 83: "if not...else" is a double negative, perhaps refactor.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
```diff
@@ -6,23 +6,104 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 
 public class DontUseFloatsAsLoopCounters extends OpcodeStackDetector implements StatelessDetector {
 
     private final BugReporter bugReporter;
 
+    private static final Map<Integer, Integer> FLOAT_LOADERS = new HashMap<>();
```
ThrawnCA, line 20: Perhaps initialise this inside the `static` block, with a `Collections.unmodifiableMap`?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontUseFloatsAsLoopCounters.java
```diff
@@ -6,23 +6,104 @@
...
+        FLOAT_LOADERS.put((int) Const.FLOAD_2, 1);
+        FLOAT_LOADERS.put((int) Const.FLOAD_3, 1);
+        FLOAT_LOADERS.put((int) Const.DLOAD, 2);
+        FLOAT_LOADERS.put((int) Const.DLOAD_0, 1);
+        FLOAT_LOADERS.put((int) Const.DLOAD_1, 1);
+        FLOAT_LOADERS.put((int) Const.DLOAD_2, 1);
+        FLOAT_LOADERS.put((int) Const.DLOAD_3, 1);
+    }
+
+    private static final Map<Integer, Integer> FLOAT_CONSTANT_PUSHERS = new HashMap<>();
```
ThrawnCA, line 34: Perhaps wrap in `Collections.unmodifiableMap`?

ThrawnCA (changes requested):

ThrawnCA (approved):

Han-Cui:
Hi, this fix indeed covered more cases which should be detected by the detector. While if I change the test case like this, Spotbugs will give no report, which may be a _false negative_.
```
public class Test2167 {
    public static void test9(boolean someOtherCondition) {
        double y = Double.NaN;
        double k = 10.0;
        while (Double.isNaN(y) && someOtherCondition && k > 0) {
            System.out.println("y before" + y);
            y = Double.parseDouble(System.console().readLine());
            k--;
            System.out.println("y after" + y);
        }
    }
}
```

Issue #2188, opened 2022-09-24
Incompatibility with later version of jdk (>= 13) in SourceFinder.java:361


Bluesbreaker45:
A method with signature **newFileSystem(Path path, Map<String,?> env)** was introduced in version 13, which results in the conflict in SourceFinder.java:361.
According to FileSystems.java in version 8 (which is recommended to compile this project), **newFileSystem(Path path, ClassLoader loader)** should be used in this context.
So an explicit cast should be used here.

PR #2189, opened 2022-09-24
Improve compatibility with later version of jdk (>= 13)

Bluesbreaker45:
A method with signature **newFileSystem(Path path, Map<String,?> env)** was introduced in version 13, which results in the conflict in SourceFinder.java:361. According to FileSystems.java in version 8 (which is recommended to compile this project), **newFileSystem(Path path, ClassLoader loader)** should be used in this context. So an explicit cast should be used here.

([issue #2188](https://github.com/spotbugs/spotbugs/issues/2188))

----


KengoTODA (approved):
LGTM, thank you!

ThrawnCA (approved):

Bluesbreaker45:
> LGTM, thank you!

My pleasure!

PR #2077, opened 2022-05-29
Introduced detector for rule VNA-00 and VNA-02 with test cases

gonczmisi:
Introducing new detector for SEI CERT rules:
* [VNA00-J. Ensure visibility when accessing shared primitive variables](https://wiki.sei.cmu.edu/confluence/display/java/VNA00-J.+Ensure+visibility+when+accessing+shared+primitive+variables)
* [VNA02-J. Ensure that compound operations on shared variables are atomic](https://wiki.sei.cmu.edu/confluence/display/java/VNA02-J.+Ensure+that+compound+operations+on+shared+variables+are+atomic)

The detectors looks for classes that could be part of multithreaded code. This means that the detector only analyzes classes, if **either of the following conditions** are true for the context of the class:
* the class implements the `java.lang.Runnable` interface (this includes extending the `java.lang.Thread` class as well, due to transitive cheking for the `Runnable` interface)
* the class contains any `synchronized` method or block
* the class contains any `volatile` or `atomic` type variables (the latter means the `java.util.concurrent.atomic` package)
* there is a `java.util.concurrent.locks.Lock` interface call anywhere in it (`lock()`, `trylock()`, `unlock()`)

If any of the above mentioned conditions are present, then the **VNA00** detector looks for primitive variables that **match all of the following conditions**:
* the variable gets compared in a not multithreaded method (contains loop conditions as well)
* the variable is witten in any method, **other then the one it got compared in**
* the variable is not declared as `volatile`, nor as an atomic type (`java.util.concurrent.atomic` package)
* there are no manual lock calls in the method where the access is (`java.util.concurrent.locks.Lock` interface call: `lock()`, `trylock()`, `unlock()`)

If any of the above mentioned conditions are present, then the **VNA02** detector looks for primitive variables that **match all of the following conditions**:
* the variable gets red in a not secured method
* the variable is witten by a compound operation in a not secured method, **other then the one it got compared in**
* the variable is not declared as an atomic type (`java.util.concurrent.atomic` package), **important that volatile is not a guarantee in this rule**
* there are no manual lock calls in the method where the access is (`java.util.concurrent.locks.Lock` interface call: `lock()`, `trylock()`, `unlock()`)

> Note: a method is considered secured, if the following either of the following conditions are true:
> * the method is snychronized
> * the method contains any local variables that has an atomic type
> * the method contains a synchronized block
> * the method contains any manual lock calls

Both detectors were tested against the code of the following repositories:
[...]

PR #2208, opened 2022-10-11
SEI_CERT_VNA00_J_VNA02_J

baloghadamsoftware:
Introduced detector for rule VNA-00 and VNA-02 with test cases

Introducing new detector for SEI CERT rules:

VNA00-J. Ensure visibility when accessing shared primitive variables
VNA02-J. Ensure that compound operations on shared variables are atomic
The detectors looks for classes that could be part of multithreaded code. This means that the detector only analyzes classes, if either of the following conditions are true for the context of the class:

the class implements the java.lang.Runnable interface (this includes extending the java.lang.Thread class as well, due to transitive cheking for the Runnable interface)
the class contains any synchronized method or block
the class contains any volatile or atomic type variables (the latter means the java.util.concurrent.atomic package)
there is a java.util.concurrent.locks.Lock interface call anywhere in it (lock(), trylock(), unlock())
If any of the above mentioned conditions are present, then the VNA00 detector looks for primitive variables that match all of the following conditions:

the variable gets compared in a not multithreaded method (contains loop conditions as well)
the variable is witten in any method, other then the one it got compared in
the variable is not declared as volatile, nor as an atomic type (java.util.concurrent.atomic package)
there are no manual lock calls in the method where the access is (java.util.concurrent.locks.Lock interface call: lock(), trylock(), unlock())
If any of the above mentioned conditions are present, then the VNA02 detector looks for primitive variables that match all of the following conditions:

the variable gets red in a not secured method
the variable is witten by a compound operation in a not secured method, other then the one it got compared in
the variable is not declared as an atomic type (java.util.concurrent.atomic package), important that volatile is not a guarantee in this rule
there are no manual lock calls in the method where the access is (java.util.concurrent.locks.Lock interface call: lock(), trylock(), unlock())
Note: a method is considered secured, if the following either of the following conditions are true:

the method is snychronized
the method contains any local variables that has an atomic type
the method contains a synchronized block
the method contains any manual lock calls
Both detectors were tested against the code of the following repositories:

RxJava
Jenkins
Spring Framework
I provieded unit tests for each detectors, that covers all possible test cases (multiple types of multithreaded classes, every kind of compund operations for VNA02 etc.).

Although the numbers of the found bugs were high (specially in the case of Jenkins), all bugs seemed to be a possible vulnerability described by the VNA00 or VNA02 rules.

Co-authered-by: @gonczmisi

This PR is a continuation of #2077 (taken over by me) which can be closed.

----


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CompoundOperationsOnSharedVariables.java
```diff
@@ -0,0 +1,215 @@
...
+                stepsMadeInCompoundOperationProcess = 1;
+                return;
+            }
+
+            // Possible invoke virtual first if it is a wrapper type
+            if (stepsMadeInCompoundOperationProcess == 1 && seen == Const.INVOKEVIRTUAL) {
+                return;
+            }
+
+            if (stepsMadeInCompoundOperationProcess == 1 && (isReadOpCode(seen) || isPushConstant(seen))) {
```
ThrawnCA, line 126: Some of this is shared with the previous condition; they could be combined.

ThrawnCA (approved):

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+    branches:
+      - master
+
+jobs:
+  check-matsim-libs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v2
```
KengoTODA, line 13:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          persist-credentials: false
+          
+      - name: Checkout Baseline
+        uses: actions/checkout@v2
```
KengoTODA, line 18:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          path: spotbugs-baseline
+          persist-credentials: false
+
+      - name: Checkout Matsim-libs
+        uses: actions/checkout@v2
```
KengoTODA, line 25:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Cache Gradle
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
```
KengoTODA, line 45: Java 15 is not supported, better to use Java 17 instead.

```suggestion
      - name: Set up JDK 17
        uses: actions/setup-java@v3
        with:
          java-version: '17.0.x'
          cache: 'gradle'
          distribution: 'temurin'
```

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Package artifacts
+        run: |
+          mkdir -p ./pr
+          echo ${{ github.event.number }} > ./pr/NR
+          mv diff_matsim pr/
+          mv baseline_time_matsim pr/
+          mv new_time_matsim pr/
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v2
```
KengoTODA, line 96:
```suggestion
        uses: actions/upload-artifact@v3
```

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          name: pr
```
KengoTODA, line 98: Please give a human-friendly name, or contributors cannot get its meaning.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/BadVisibilityOnSharedPrimitiveVariablesTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+public class BadVisibilityOnSharedPrimitiveVariablesTest extends AbstractIntegrationTest {
+
+    @Before
+    public void setup() {
+        SystemProperties.setProperty("ful.debug", "true");
+    }
```
KengoTODA, line 37: Please reset the property after each test case by `@After`, to keep each test case independent.

.github/workflows/check-projects-comment.yml
```diff
@@ -0,0 +1,63 @@
+name: check-projects-comment
+
+on:
+  workflow_run:
+    workflows: ['check-projects']
+    types:
+      - completed
+
```
KengoTODA, line 8: Add `permissions` to reduce the privilege of GITHUB_TOKEN.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

.github/workflows/check-projects.yml
```diff
@@ -0,0 +1,99 @@
+name: check-projects
+on:
+  pull_request:
+    branches:
+      - master
+
```
KengoTODA, line 6: Add `permissions` to reduce the privilege of GITHUB_TOKEN.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

.github/workflows/check-projects-comment.yml
```diff
@@ -0,0 +1,63 @@
...
+              return artifact.name == 'pr'
+            })[0]
+            const download = await github.actions.downloadArtifact({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            })
+            const fs = require('fs')
+            fs.writeFileSync('${{github.workspace}}/pr.zip', Buffer.from(download.data))
```
KengoTODA, line 35: Better to use GitHub official CLI to shorten your code:
https://cli.github.com/manual/gh_run_download

KengoTODA (changes requested):
Please merge the latest upstream branch, to resolve conflicts in the CHANGELOG.

baloghadamsoftware:
@gonczmisi decided to continue the original PR, #2077.

PR #2217, opened 2022-10-15
fix: Disable access to external entities in XML parsing

gtoison:
Fixes security issues detected by sonarcloud such as: https://sonarcloud.io/project/issues?issues=AXDRMcroPCexr-BXKaE_&open=AXDRMcroPCexr-BXKaE_&id=com.github.spotbugs.spotbugs

> XML standard allows the use of entities, declared in the DOCTYPE of the document, which can be [internal](https://www.w3.org/TR/xml/#sec-internal-ent) or [external](https://www.w3.org/TR/xml/#sec-external-ent).
> 
> When parsing the XML file, the content of the external entities is retrieved from an external storage such as the file system or network, which may lead, if no restrictions are put in place, to arbitrary file disclosures or [server-side request forgery (SSRF)](https://owasp.org/www-community/attacks/Server_Side_Request_Forgery) vulnerabilities.

Added methods in `XMLUtil` to setup XML objects with more secure options

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
@@ -27,7 +27,6 @@ import javax.xml.transform.TransformerFactory;
 import org.dom4j.Node;
 import org.dom4j.io.SAXReader;
 
-import edu.umd.cs.findbugs.FindBugs;
 
 /**
  * @author pugh
@@ -50,7 +49,7 @@ public class XMLUtil {
 
         return reader;
     }
-    
+
     public static TransformerFactory buildTransformerFactory() {
         TransformerFactory factory = TransformerFactory.newInstance("net.sf.saxon.TransformerFactoryImpl", null);
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
@@ -21,7 +21,13 @@ package edu.umd.cs.findbugs.xml;
 
 import java.util.List;
 
+import javax.xml.XMLConstants;
+import javax.xml.transform.TransformerFactory;
+
 import org.dom4j.Node;
+import org.dom4j.io.SAXReader;
+
+import edu.umd.cs.findbugs.FindBugs;
 
 /**
  * @author pugh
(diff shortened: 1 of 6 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
```diff
@@ -33,4 +38,28 @@ public static <T> List<T> selectNodes(Node node, String arg0) {
         return (List<T>) node.selectNodes(arg0);
     }
 
+    public static SAXReader buildSAXReader() {
+        SAXReader reader = new SAXReader();
+
+        try {
+            reader.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e);
```
KengoTODA, line 47: It seems that it's not argument issue (this `buildSAXReader()` method has no argument).
It is probably better to use another exception type?

spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
```diff
@@ -33,4 +38,28 @@ public static <T> List<T> selectNodes(Node node, String arg0) {
...
+    }
+
+    public static TransformerFactory buildTransformerFactory() {
+        TransformerFactory factory = TransformerFactory.newInstance("net.sf.saxon.TransformerFactoryImpl", null);
+
+        try {
+            factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+            factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, "");
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e);
```
KengoTODA, line 60: Same issue here. 🙇‍♂️

KengoTODA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/xml/XMLUtil.java
gtoison: I  have modified the exception type and added an error message, please let me know if you had something else in mind or if this requires further adjustments

ThrawnCA (approved):

KengoTODA (commented):
LGTM, please add a changelog entry into the `CHANGELOG.md` file. We probably need a new `### Security` under the `## Unreleased`.

gtoison:
Thanks for the review, I have rebased and added a security category in the changelog with a new entry for this PR

KengoTODA (approved):
LGTM, thank you! 🤝

PR #2239, opened 2022-10-29
Removed String constructors and replaced with empty String

imkulwant:
Replaced String constructors with empty String

----

### DIFF
--- a/eclipsePlugin-test/defaultScenario/B.java
+++ b/eclipsePlugin-test/defaultScenario/B.java
@@ -1,4 +1,4 @@
 public class B {
     // should report URF_UNREAD_FIELD (UrF) and DM_STRING_CTOR (Dm)
-    String s = new String("");
+    String s = "";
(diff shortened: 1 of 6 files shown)


ThrawnCA (approved):

eclipsePlugin-test/defaultScenario/B.java
```diff
@@ -1,4 +1,4 @@
 public class B {
     // should report URF_UNREAD_FIELD (UrF) and DM_STRING_CTOR (Dm)
-    String s = new String("");
+    String s = "";
```
gtoison, line 3: Isn't it a sample bug for unit tests, with a bug purposedely placed here for SpotBugs to find

KengoTODA:
I believe that such "bad" codes in `*-test` project are intentional, to test behavior of SpotBugs. We do not have to fix these code to keep our product tested.

Issue #543, opened 2018-01-22
Support for filtering by annotation


PhilippWendler:
Please add support for filtering by annotations to the SpotBugs filter file. Such a filter could for example look like
```
<Match>
  <AnnotatedWith name="annotation.class.name"/>
</Match>
```
Of course it should be possible to combine this filter with other existing filters. The annotation filter should work on packages, classes, methods, fields, and local variables, basically every Java element for which already a filter exists.

More specific filters (like annotated with a specific annotation and a specific value of this annotation) would be possible, but also more complicated, thus I suggest to leave them out for now.

Of course this annotation filter would match only annotations that have a retention level of at least `class`, but there are still enough use cases for such a filter. For example, we have annotations that mark elements that are accessed via reflection, and because SpotBugs does not know about this reflective accesses there are false alarms for some checks. Now we have to disable these checks globally, with such a filter we could disable the checks only where they need to be disabled.

For completeness: related to #464 and [feature request 256 of FindBugs](https://sourceforge.net/p/findbugs/feature-requests/256/), but I was asked to create a new issue.

oujesky:
Just ran to an use case for this - JavaFX views supports private event handler methods to be annotated with `@FXML` annotation, SpotBugs hovewer marks such methods with a `UPM_UNCALLED_PRIVATE_METHOD` rule violation.

mannodermaus:
I've recently been thinking about what the syntax for annotation support in Spotbugs might look like, and I'd like to get a discussion going with the current findings.

Assume the following annotation:

```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.CLASS)
public @interface MyAnnotation {
    String value();
    int counter() default 0;
}
```

Applied on these classes:

```java
@MyAnnotation("Test")
public void ExampleFragment1 {
}

@MyAnnotation(value = "Test", counter = 42)
public void ExampleFragment2 {
}
```

The two approaches I was able to come up with are detailed in the filter file description below. Each scenario is supposed to match the class named inside the `<Class>` element (please disregard the matched `pattern`, that's just for demonstration).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<FindBugsFilter>
    <Match>
        <Class name="~.*\.*Fragment" />
        <Bug pattern="UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR" />
    </Match>

    <!-- 
        Approach 1: New attributes in Class, Method, Field, Local
    -->

[...]

PR #2252, opened 2022-11-09
Issue 543

phisad:
Hello all,

I added a simple AnnotationMatcher that matches classes annotated with that annotation name.
For example, use in an `excludeFilter.xml` as follows
```
<Match>
<Annotation name="org.immutables.value.Generated" />
</Match>
```
 to ignore classes generated by the Immutable framework.

**Request For Comments**
The main issue was to add the java class annotation names of the analysed classes to the PackageMemberAnnotation for later use during exclude file filtering. For this I simply changed the `BugInstance.add()` method as this seemed to me a central point to get this information and avoiding to re-write all the detectors. For better performance I access the `IAnalysisCache`.

Please suggest a better place for this, if necessary, because I do not know the codebase too well.

**The PR includes:**
- a new AnnotationMatcher for excludeFilter.xml files; added an AnnotationMatcherTest
- add AnnotationType to findbugsfilter.xsd
- add a `javaAnnotationNames` attribute to PackageMemberAnnotation
- added `addJavaAnnotationNames()` method to `BugInstance` as a central point to get the java annotations that a class is annotated with
- a comma-separated "classjas" attribute is now written to XML when `writeXml()` is called on ClassAnnotation, MethodAnnotation or FieldAnnotation; a test case is added to BugInstanceTest 
- adjust SAXBugCollectionHandler to handle the new "Annotation" matcher and "classjas" attribute; a test case added to SAXBugCollectionHandlerTest

----


ThrawnCA:
Build is failing due to a formatting error.

pzygielo:
https://github.com/spotbugs/spotbugs/issues/543

phisad:
> Build is failing due to a formatting error. 

I ran `spotlessApply` on spotbugs-spotbugs and spotbugs-test and hope it is fixed now.

phisad:
Fixed the test and run `spotlessApply` and `test`

phisad:
btw. any comments on the changes to `BugInstance.add()` ?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/AnnotationMatcherTest.java
```diff
@@ -0,0 +1,157 @@
+/*
+ * FindBugs - Find Bugs in Java programs Copyright (C) 2003-2008 University of Maryland Author: Andrey Loskutov This
```
ThrawnCA, line 2: New files should reference SpotBugs not FindBugs.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -100,6 +104,8 @@
  */
 public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Cloneable {
 
+    private static final Logger LOG = LoggerFactory.getLogger(BugInstance.class);
```
ThrawnCA, line 107: This could be made copy-paste compatible using `MethodHandles`, eg:

    private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());

ThrawnCA (changes requested):

phisad:
Thanks for the valuable comments. I applied the requested changes.

spotbugs/src/main/java/edu/umd/cs/findbugs/ClassAnnotation.java
```diff
@@ -198,6 +199,11 @@ public void writeXML(XMLOutput xmlOutput, boolean addMessages, boolean isPrimary
             attributeList.addAttribute("primary", "true");
         }
 
+        if (!getJavaAnnotationNames().isEmpty()) {
+            attributeList.addAttribute("classjas", // similar to classname, short for classJavaAnnotations
```
hazendaz, line 203: @phisad is 'classjas' just made up naming?  Seems odd even from change log.  Why not just 'classAnnotations'?  We already know its java, that is redundant.  At least then it follows the idea you have there and is more clear on name without new jargon.  Then you probably don't need to describe why its named that way...

phisad: Thanks for the comment! I was also thinking about "classAnnotations" but felt that it might be confused with the ClassAnnotation from spotbugs. But I think you are right, that a user of the XML file will not know that the term "classAnnotation" is also otherwise used. If you say that this is clearer, then I will certainly adjust it.

hazendaz: Good point, thanks for calling that out, that certainly would cause unnecessary confusion. how about 'classAnnotationNames' which is more in line with what its loading there?

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -171,6 +178,16 @@
             </annotation></attribute>
     </complexType>
 
+    <complexType name="AnnotationType">
+        <annotation>
+            <documentation></documentation>
+        </annotation>
+        <attribute name="name" type="string">
+            <annotation>
+                <documentation>The exact or regex match pattern for a java class annotation name. If the name starts with the ~ character the rest of attribute content is interpreted as a Java regular expression.</documentation>
```
ThrawnCA, line 187: "the rest of attribute content" -> "the rest of the attribute content

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -2092,9 +2099,43 @@ public BugInstance addOptionalAnnotation(@CheckForNull BugAnnotation annotation,
...
+            PackageMemberAnnotation pma = (PackageMemberAnnotation) annotation;
+            ClassDescriptor classDescriptor = pma.getClassDescriptor();
+            ClassContext classContext = analysisCache.getClassAnalysis(ClassContext.class, classDescriptor);
+            JavaClass javaClass = classContext.getJavaClass();
+            AnnotationEntry[] annotationEntries = javaClass.getAnnotationEntries();
+            List<String> javaAnnotationNames = Arrays.asList(annotationEntries).stream().map((AnnotationEntry ae) -> {
+                // map annotation entry type to dotted class name, for example
+                // Lorg/immutables/value/Generated; -> org.immutables.value.Generated
+                String annotationType = ae.getAnnotationType().substring(1); // remove starting L
+                annotationType = annotationType.replace("/", ".");
```
ThrawnCA, line 2117: These two lines could be combined.

ThrawnCA (changes requested):

spotbugs/etc/findbugsfilter.xsd
phisad: Adjusted. Note that this is a copy paste error (that still exists in the other documentation tags as well).

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
phisad: Inlined the replace operations.

spotbugs/src/main/java/edu/umd/cs/findbugs/ClassAnnotation.java
phisad: OK, I changed the attribute in the XMLs from _classjas_ to _classAnnotationNames_

ThrawnCA (approved):

ThrawnCA:
Changelog needs to be deconflicted and merged.

phisad:
Rebased on upstream/master. Adjusted CHANGELOG.

hazendaz:
@phisad Fix fix formatting issues GH-actions raised.

The following files had format violations:
      src/main/java/edu/umd/cs/findbugs/BugInstance.java
          @@ -2113,7 +2113,7 @@
           ············List<String>·javaAnnotationNames·=·Arrays.asList(annotationEntries).stream().map((AnnotationEntry·ae)·->·{
           ················//·map·annotation·entry·type·to·dotted·class·name,·for·example
           ················//·Lorg/immutables/value/Generated;·->·org.immutables.value.Generated
          -················String·annotationType·=·ae.getAnnotationType().substring(1).replace("/",·".").replace(";",·"");·
          +················String·annotationType·=·ae.getAnnotationType().substring(1).replace("/",·".").replace(";",·"");
           ················return·annotationType;
           ············}).collect(Collectors.toList());
           ············pma.setJavaAnnotationNames(javaAnnotationNames);
  Run './gradlew :spotbugs:spotlessApply' to fix these violations.

phisad:
Did run './gradlew :spotbugs:spotlessApply' to fix the violations.

phisad:
@hazendaz Does it look fine now?

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -58,6 +58,13 @@
             <documentation>This element matches warnings associated with a particular class.</documentation>
         </annotation>
     </element>
+    
+    <element name="Annotation" type="fb:AnnotationType">
+        <annotation>
+            <documentation>This element matches warnings associated with a particular class that is annoted with a
```
hazendaz, line 64: @phisad See a spelling mistake here, should 'annoted' be 'annotated'?

hazendaz:
@phisad Overall good, just one spelling mistake I found, can you do a double check across this?

@ThrawnCA Look good to you now as well?

cpfeiffer:
Supserseded by #2395

PR #2256, opened 2022-11-14
Stop exposing junit-bom to consumers (#2255)

### LINKED ISSUE #2255: spotbugs-annotations artifact depends on and exports junit-bom
When a project depends on `com.github.spotbugs:spotbugs-annotations`, it will add the latest JUnit Jupiter BOM to the project classpath. This causes issues for consumers, as adding or bumping `spotbugs-annotations` will lead to a bump of JUnit, as the BOM ends up in the [POM](https://repo1.maven.org/maven2/com/github/spotbugs/spotbugs-annotations/4.7.3/spotbugs-annotations-4.7.3.pom) and Gradle [module](https://repo1.maven.org/maven2/com/github/spotbugs/spotbugs-annotations/4.7.3/spotbugs-annotations-4.7.3.module) files.

This happens because the BOM is configured to be an `implementation` dependency (https://github.com/spotbugs/spotbugs/blob/a9ae3aaab8917de470a0301d7c51819246b58007/build.gradle#L40). It would likely be better to add the BOM exclusively to the `testImplementation` configurations for all projects, as it really shouldn't be leaked to any production code consumers.

davidburstrom:
Fixes #2255 by limiting the scope where the JUnit-bom is visible.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 - Fixed missing classes not in report if using IErrorLogger.reportMissingClass(ClassDescriptor) ([#219](https://github.com/spotbugs/spotbugs/issues/219))
+- Stop exposing junit-bom to consumers ([#2255](https://github.com/spotbugs/spotbugs/pull/2255))
 
 #### Security
 - Disable access to external entities when processing XML ([#2217](https://github.com/spotbugs/spotbugs/pull/2217))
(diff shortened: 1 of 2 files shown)


davidburstrom:
When/if this change is shipped, there is a risk that some consuming projects will break since they no longer get a suggested JUnit version transitively, but ultimately, they should be in control of that themselves.

ThrawnCA (approved):

hazendaz:
@davidburstrom Thanks, no worries IMHO regarding consumers on this one.  They should not re relying on us forcing such resolution on them.

PR #2266, opened 2022-11-20
SEI_CERT_LCK01 updated

katheg:
New function added to detector `SynchronizationOnSharedBuiltinConstant`to detect `DL_SYNCHRONIZATION_ON_INTERNED_STRING`

https://wiki.sei.cmu.edu/confluence/display/java/LCK01-J.+Do+not+synchronize+on+objects+that+may+be+reused

----


ThrawnCA (approved):

ThrawnCA:
Is this something people actually do? It seems bizarre.

hazendaz:
Seems odd but also suggestions are new'ing objects unnecessary instead of valueOf which would themselves be further violations.  Might want to start cleaning that up so its properly using valueOf where appropriate.  At least so I'm not distracted by telling users to write bad code.

hazendaz:
Sorry totally read this wrong first time through, approved and will merge.

hazendaz (approved):

Issue #184, opened 2017-05-29
AbstractBugReporter emits wrong non-sensical debug output during filtering


sewe:
Either the filtering logic or the debug output in `AbstractBugReporter` is wrong. With `-Dabreporter.debug=true` set, I just got the following non-sensical output:

> AbstractBugReporter: Filtering due to rankThreshold 20 > 20
> AbstractBugReporter: Filtering due to rankThreshold 19 > 20

Neither inequality is true.

JuditKnoll:
This issue was fixed by https://github.com/spotbugs/spotbugs/pull/2295.

PR #2295, opened 2023-01-02
Fix Issue #184

Alex-Vol-Amz:
The logic conditionals were entangled in an invalid way that produced nonsense debug log messages. Simplified the logic and reduced the logic branches to make the operation direct and thus fix the problem.

----


spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
```diff
@@ -190,15 +190,13 @@ public final void reportBug(@Nonnull BugInstance bugInstance) {
         }
         int priority = bugInstance.getPriority();
         int bugRank = bugInstance.getBugRank();
-        if (priority <= priorityThreshold && bugRank <= rankThreshold) {
-            doReportBug(bugInstance);
+        if (priority > priorityThreshold) {
```
ThrawnCA, line 193: This logic seems wrong. If the priority is higher than the threshold, it should be reported, not filtered.

ThrawnCA (dismissed):

spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
Alex-Vol-Amz: Threshold for both priority and rank operates with a "lower value is more important".

Look at the original code

```
        if (priority <= priorityThreshold && bugRank <= rankThreshold) {
            doReportBug(bugInstance);
        }
```

Alex-Vol-Amz: I could have effected the same fix just by reversing the logic in the `else` clause from `<=` to `>` but after reading this code the current form is easier to follow and reduced the number of comparisons to exactly 2 with 3 logical outcomes.

Alex-Vol-Amz: You may need to look at the issue reported. The problem is not with the reporting but with the invalid debug logging. What was happening before is that when the reporting was skipped due to priority the debug log would instead report skipping due to rankThreshold with nonsense conditions. The opposite would also do the same, a skip due to rank would debug log that skip due to priority with invalid conditions.

Obviously if the original logic was flawed this would have been fixed a long time ago, it was only the debug logging that was confusing and this was never addressed.

In my opinion this makes the intended logic clear and unambiguous.

Alex-Vol-Amz: Perhaps a unit test would help restore confidence in this logic and eliminate any questions. There were no unit tests I could find that validated logged messages but it may be useful to at least make a unit test that proves `doReportBug` is called based on the rules of this logic.

Alex-Vol-Amz:
Performed the execution of the new unit test with and without the main code change. The result was the same passing test as expected since the intention of the code change was to fix the debug log, not change the reporting behavior.

spotbugs/src/main/java/edu/umd/cs/findbugs/AbstractBugReporter.java
Alex-Vol-Amz: > This logic seems wrong. If the priority is higher than the threshold, it should be reported, not filtered, right?
> 
> Seems to me that the correct fix is simply to swap the first condition so that it reports the bug if either priority or bugRank is _greater_ than the threshold, and leave the rest basically the same.

I am not going to make the changes due to the reasons stated above. The code is correct as it stands and I expect this would be approved now. I added the unit test to prove that the intended filtering was working the same way before and after my changes.

hazendaz: @Alex-Vol-Amz Think I'm following the logic error.  The problem was that both were considered for the original first check.

```
priority <= priorityThreshold && bugRank <= rankThreshold
```

Then it was checking this again

```
priority <= priorityThreshold
```

If it didn't hit that then it assumed rank threshold to log.  Problem as I see original code then just to clarify my understanding is that priority or bug ran could be ```>``` and such the only path was for rank threshold to write the debug line.

I agree it was overly complex as I had to think too hard to even respond here.  Its far easier in new state as follows.

Write debug...previously this was a bug, if it was > then it was never logged, it fell through to rankThreshold incorrectly.
```
priority > priorityThreshold
```

Write debug...
```
bugRank > rankThreshold
```

Otherwise....reportBug

Alex-Vol-Amz: Exactly, which is now also covered by unit test so we know for sure it will stay fixed in this new form.

I wish the code was amenable to actually mock the logger so one could even verify that the debug is actually logged and correct, but given the direct approach followed here it becomes simple binary logic.

hazendaz: Thanks @Alex-Vol-Amz I'm good with the changes.  Going to merge it in now.

hazendaz (approved):

PR #2322, opened 2023-01-21
Add synchronized block around System.in usages

blacelle:
This proposes a change to fix concurrent issues when using Spotbugs (e.g. through the mvn plugin (https://github.com/spotbugs/spotbugs-maven-plugin/issues/488)(.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
@@ -405,7 +405,6 @@ public class TextUICommandLine extends FindBugsCommandLine {
         } else if ("-exitcode".equals(option)) {
             setExitCode = true;
         } else if ("-auxclasspathFromInput".equals(option)) {
-            // 'synchronized' to handle concurrent executions (e.g. by the mvn plugin)
             synchronized (System.in) {
                 try {
                     BufferedReader in = UTF8.bufferedReader(System.in);
@@ -765,7 +764,6 @@ public class TextUICommandLine extends FindBugsCommandLine {
      */
     public void handleXArgs() throws IOException {
         if (getXargs()) {
-            // 'synchronized' to handle concurrent executions (e.g. by the mvn plugin)
             synchronized (System.in) {
                 try (BufferedReader in = UTF8.bufferedReader(System.in)) {
                     while (true) {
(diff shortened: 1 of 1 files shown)


ThrawnCA (approved):

blacelle:
@ThrawnCA This PR needs an approval for CI due to first-time contribution. THanks

hazendaz:
@blacelle 

I'm somewhat confused how this relates in general to maven plugin, I think that comment should be removed.  I agree synchronization here is a good thing otherwise.  I'm not that deeply versed on the paths taken but UI part would indicate GUI part used which can be ran from maven.  But to indicate the issue is due to maven plugin is a bit off.  It would happen with gradle too or anything that would execute in same way.  I would like a bit more clarification on what is being run though that causes this code to activate and hence concerns a number of maven users.

This will be merged but can you remove comment that its due to maven plugin?  I don't think we want comments in the core code indicating perceived issues from addon tools.

blacelle:
I'm happy to remove the comments if they seem irrelevant.

I can't say if spotbugs-maven is legitimate in relying in `TextUICommandLine`. Here is a typical stack demonstrating how we get from `mvn`/`ant` up to `TextUICommandLine` through `FindBugs2` and `FindBugs` classes.
```
	at java.io.BufferedReader.readLine(java.base@17.0.2/BufferedReader.java:396)
	at edu.umd.cs.findbugs.TextUICommandLine.handleOption(TextUICommandLine.java:411)
	at edu.umd.cs.findbugs.config.CommandLine.parse(CommandLine.java:338)
	at edu.umd.cs.findbugs.config.CommandLine.parse(CommandLine.java:300)
	at edu.umd.cs.findbugs.FindBugs.processCommandLine(FindBugs.java:345)
	at edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1221)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(java.base@17.0.2/Native Method)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(java.base@17.0.2/NativeMethodAccessorImpl.java:77)
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@17.0.2/DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(java.base@17.0.2/Method.java:568)
	at org.apache.tools.ant.taskdefs.ExecuteJava.run(ExecuteJava.java:218)
	at java.lang.Thread.run(java.base@17.0.2/Thread.java:833)
```

gtoison:
Wouldn't it be better to fork a new JVM when the maven plugin (or something else really) uses SpotBugs in a multithreaded way?
Adding synchronization here might potentially lead to deadlocks when one thread has the the lock on `System.in` and also blocks on reading `System.in`

hazendaz:
closing as issue fixed in the maven plugin, it was calling this incorrectly since inception.  There were actually multiple tickets dating back years that described the issue but did not find it as well as this effort.  No need to add this here now.

Issue #2325, opened 2023-01-23
Unused field not reported due to assumed reflection


trancexpress:
Snippet to reproduce:

```
package test;
public class SomeClass {

	public static Class<?> cl = SomeClass.class;
	private String unusedField;
}
```

For `unusedField`, nothing is reported. After commenting out `public static Class<?> cl = SomeClass.class;`, a bug is reported as expected.

Seems to be coming from: https://github.com/spotbugs/spotbugs/commit/9ae2687e771dbd927e11b6a19ad23de3419fe7c8

In particular as soon as `LDC Ltest/SomeClass;.class` is seen, `SomeClass` is assumed to be accessed with reflection and the change above suppresses the unused field bug report.

PR #2326, opened 2023-01-23
Unused field not reported due to assumed reflection

trancexpress:
This change ensures an unused field is reported despite assumptions on reflection use for the defining class.

If reflection is assumed, the priority of the created spotbugs problem is lowered.

Fixes: #2325

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Stop exposing junit-bom to consumers ([#2255](https://github.com/spotbugs/spotbugs/pull/2255))
 - Fixed AbstractBugReporter emits wrong non-sensical debug output during filtering ([#184](https://github.com/spotbugs/spotbugs/issues/184))
 - Added support for jakarta namespace ([#2289](https://github.com/spotbugs/spotbugs/pull/2289))
+- Report a low priority bug for an unread field in reflective classes ([#2325](https://github.com/spotbugs/spotbugs/issues/2325))
 
 ### Security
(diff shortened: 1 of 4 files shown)


trancexpress:
@iloveeclipse I don't see how to lower the confidence of the bug. Just the priority (++ is lowering it).

iloveeclipse (changes requested):
Please do two things:
- add a trivial test to spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect
- add an entry to the CHANGELOG.md file

Beside this, I think we should be fine.

Dirty code should be properly reported, and by filtering on prio one can define what one will see what not. 
Not reporting such obvious thing makes people assume that SpotBugs is stupid, which isn't the case in general :-)

iloveeclipse (approved):
New with "reflective" class usage we report
```
Rank: Of Concern (20), confidence: Low
Pattern: UUF_UNUSED_FIELD
```

Same but **without** "reflective" class usage
```
Rank: Of Concern (18), confidence: Normal
Pattern: UUF_UNUSED_FIELD
```

So one should be able to sort that out.
Thanks, LGTM

vectro:
With this change, we now generate `SIC_INNER_SHOULD_BE_STATIC_ANON` for these common patterns for getting a `Method` for the current method:
```
        class L {
        };
        Method currentlyExecutingMethod = L.class.getEnclosingMethod();
```
```
        Method currentlyExecutingMethod = new Object(){}.getClass().getEnclosingMethod();
```
The latter also generates `DM_NEW_FOR_GETCLASS`.

Is it worth opening a new issue for that?

PR #2375, opened 2023-03-09
Don't report for stream methods called on Random

### LINKED ISSUE #2370: False positive for DMI_RANDOM_USED_ONLY_ONCE with Stream methods
I'm getting a false positive on `DMI_RANDOM_USED_ONLY_ONCE` for code like the following:

```java
import java.util.Random;
import java.util.stream.IntStream;

public class RandomTest {
  public IntStream randomPercents(int count) {
    return new Random().ints(count, 0, 100);
  }
}
```

Returns this error:

```
H B DMI: Random object created and used only once in RandomTest.randomPercents(int)  At RandomTest.java:[line 6]
```

Since `ints()` is returning a stream, using that method or any of the analogous methods like `longs()` should act as if the `Random` used more than once. I suppose there's an edge case of `count` is `0` or `1`.

md5:
If one of the stream-returning methods (`doubles`, `ints`, or `longs`) is called on `Random` or `SecureRandom`, don't report that the `Random` was only used once

Fixes #2370

----

### DIFF
--- a/spotbugsTestCases/src/java/ghIssues/issue2370/Issue2370Doubles.java
+++ b/spotbugsTestCases/src/java/ghIssues/issue2370/Issue2370Doubles.java
@@ -1,0 +1,22 @@
+package ghIssues.issue2370;
+
+import java.util.Random;
+import java.util.stream.DoubleStream;
+
+public class Issue2370Doubles {
+  public DoubleStream shouldNotReport0() {
+    return new Random().doubles();
+  }
+
+  public DoubleStream shouldNotReport1() {
+    return new Random().doubles(10);
+  }
+
+  public DoubleStream shouldNotReport2() {
+    return new Random().doubles(0.0, 100.0);
+  }
+
+  public DoubleStream shouldNotReport3() {
+    return new Random().doubles(10, 0.0, 100.0);
+  }
+}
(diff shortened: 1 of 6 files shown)


spotbugsTestCases/src/java/ghIssues/issue2370/Issue2370Doubles.java
```diff
@@ -0,0 +1,22 @@
+package ghIssues.issue2370;
+
+import java.util.Random;
+import java.util.stream.DoubleStream;
+
+public class Issue2370Doubles {
+  public DoubleStream shouldNotReport0() {
+    return new Random().doubles();
```
md5, line 8: Let me know if I should also add analogous tests for `SecureRandom` or if there is a preferred pattern for making these tests more DRY

iloveeclipse (approved):
LGTM

spotbugsTestCases/src/java/ghIssues/issue2370/Issue2370Doubles.java
iloveeclipse: Should be OK as is.

iloveeclipse:
Thanks!

PR #2395, opened 2023-03-31
Issue 543

cpfeiffer:
This new PR just fixes [the last comment](https://github.com/spotbugs/spotbugs/pull/2252#issuecomment-1489469776) of #2252. I don't have write permission to the fork of @phisad, so here's his work, rebased on top of master with the suggested spelling fix.

#2252 can be closed in favor of this branch, since my colleague @phisad does not have time for it anymore.


hazendaz:
@cpfeiffer Will take this once build completes and will double check rebase against @phisad original PR.  Thanks for picking up.

hazendaz (approved):

hazendaz:
@cpfeiffer @phisad Merged now, thanks!  No ETA on actually release yet.  I don't know the full process on the core here (I do the maven part).  I think we need to get a number of vulnerable libraries updated still but think we have to be close to releasing in general otherwise.  Thanks for sticking around on this one.

PR #2407, opened 2023-04-14
minor: make private class final with default constructor

Kevin222004:
minor: make private class final with default constructor

This is part of https://github.com/checkstyle/checkstyle/pull/12737

According to the https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.8.9 The default constructor has the same access modifier as the class.
and according to the check https://checkstyle.org/config_design.html#FinalClass a class that has only private constructors and has no descendant classes is declared as final.

### DIFF
--- a/eclipsePlugin-junit/src/test/java/de/tobject/findbugs/reporter/ThrottledProgressMonitorTest.java
+++ b/eclipsePlugin-junit/src/test/java/de/tobject/findbugs/reporter/ThrottledProgressMonitorTest.java
@@ -42,7 +42,7 @@ public class ThrottledProgressMonitorTest {
         verify(delegate).worked(2 + 4);
     }
 
-    private static class Clock {
+    private static final class Clock {
         private long currentTime;
 
         protected long getCurrentTime() {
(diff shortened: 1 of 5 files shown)


Kevin222004:
I have given information on this changes in description. 

can we have your valuable feedback on this changes.

hazendaz:
@Kevin222004 I'm taking a look, in future though, there is no reason to open a PR then immediately ask for an update on it.  We get emailed on these and it just sends needless emails doing that.  Its better to just open it, describe it well and leave it at that.  Thanks.

hazendaz (approved):

Kevin222004:
@hazendaz we need your feedback on these changes.

Kevin222004:
What do you think should we care about this thing or not

hazendaz:
I merged this an hour ago.  Please make sure to follow github here and what is happening.  My first comment was to assist in not unnecessarily sending spam comments.  You have followed that by two more needless spam emails.

To answer your question yes this is valid.  Code should be written properly.  Many fail to do so on most if not all repos thinking java is too wordy.  I personally don't agree with ignoring such code requirements in java and it typically seems silly that even a library such as this doesn't fully follow all coding conventions.  Tools like open rewrite would have likely applied the same here, same with Eclipse / Intellij, etc.  You changes were otherwise accurate and have been merged.

Issue #2321, opened 2023-01-20
Add CWE Taxonomy to SARIF Report


Jeeppler:
## Description

The SARIF reports generated by FindSecurityBugs do not contain the [Common Weakness Enumeration (CWE)](https://cwe.mitre.org/) taxonomy from MITRE.

Adding the CWE Taxonomies into SARIF reports would make the SARIF report easier to understand for security practitioners already familiar with CWE.

SARIF does support [taxonomies](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317536) in reports.
For example, SARIF reports from the [GoSec](https://github.com/securego/gosec) tool contain the MITRE CWE taxonomy.

Jeeppler:
I already opened the same issue in FindSecurityBugs: find-sec-bugs/find-sec-bugs/issues/688. However, the maintainers of FindSecurityBugs told me, SpotBugs is responsible for generating the SARIF report.

PR #2410, opened 2023-04-22
Add CWE taxonomy to SARIF report

Jeeppler:
Add CWE taxonomy to SARIF report.

- added class to calculate GUID based on UUID5
- added classes for CWE information and data
- added slimmed down JSON file with information about all CWEs (based on official CWE XML file). The script to create the slimmed down version is not included.
- added new tests
- changed, added and modified classes to create the CWE taxonomy

closes: #2321

----


spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
```diff
@@ -0,0 +1,123 @@
...
+
+            JsonElement weaknessElements = rootElement.getAsJsonObject().get("weaknesses");
+
+            for (JsonElement weaknessElement : weaknessElements.getAsJsonArray()) {
+                Weakness weakness = gson.fromJson(weaknessElement, Weakness.class);
+
+                INSTANCE.weaknesses.put(Integer.valueOf(weakness.getCweId()), weakness);
+            }
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
```
hazendaz, line 77: @Jeeppler Please use logger here rather than print stack trace.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
Jeeppler: @hazendaz did it.

Jeeppler:
@hazendaz and @ThrawnCA I wrote a small Python CLI app/script to convert the MITRE XML into a reduced JSON file.

Do you want me to add the Python script to this pull request?
- If so, in which folder shall I put it?

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
```diff
@@ -16,13 +19,17 @@
...
+import edu.umd.cs.findbugs.AbstractBugReporter;
+
 /**
  * The weakness catalog contains a number of weaknesses
  *
  * @author Jeremias Eppler
  * @see Weakness
  */
 public class WeaknessCatalog {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractBugReporter.class);
```
ThrawnCA, line 31: Wrong class. Perhaps use `LoggerFactory.getLogger(MethodHandles.lookup().lookupClass())` to avoid this issue?

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
hazendaz: seems expensive doing that nevermind fact slf4j will let you provide nothing now (at least in version 2).  Personally just fix it to correct class name WeaknessCatalog.class as many scanning tools expect that.

I don't know about other code in spotbugs, but upper case LOG is long incorrect.  It should simply be 'logger' in lower case as the accepted practice.

hazendaz:
> * If so, in which folder shall I put it?

Sure on including but not sure where to place it.  Make your best guess :)

Jeeppler:
I used the [REUSE specification](https://reuse.software/spec/) header comments for the CWE conversion script to put it under the proper license.

1. Is that okay?
2. Is there a reason why SpotBugs is not using the REUSE spec (SPDX format) for all files?

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
ThrawnCA: It's a static final field. The [Google Java style guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) still says to use UPPER_SNAKE_CASE for those, and so does [Carnegie Mellon](https://www.cs.cmu.edu/~rdriley/121/resources/styleguide/#s5.2.4-constant-names), and also [Oracle](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html). What Java style guide says otherwise?

hazendaz:
Yes but not on logger.  Long disputed special case.  Has to do with high frequency usage of loggers.

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
hazendaz: To my response, see https://itecnote.com/tecnote/java-should-a-static-final-logger-be-declared-in-upper-case/

Main point as many state similar and in fact slf4j states to use it that way is that its a reference not a constant.

hazendaz:
> 2\. Is there a reason why SpotBugs is not using the REUSE spec (SPDX format) for all files?

I don't know why, this project is really old so just maybe not something that was generally thought of.  I'd have all files marked with license headers if this were maven but its gradle and my skills there are very low.  If you understand gradle enough and know how to get that to be turned on and automated so files auto do it across the board, I think that would be a separate welcomed PR.  I'm sure there are plugins for gradle to do same as there are for maven so its not something we have to think about :)

hazendaz:
@ThrawnCA I think I'm good here.  WDYT?

hazendaz (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
ThrawnCA: The linked page doesn't seem to give any actual basis for treating static final references differently from constants.

hazendaz:
2009 same statement.  https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

Look at follow up 2020.  Look at fact it's defined in jave style docs.

Then look again at Google.  They in fact have it lower cased.

https://google.github.io/styleguide/javaguide.html#s5.2-specific-identifier-names

Look specifically at what they call constants.  Logger is not.


hazendaz:
Every reference you provided says the same thing about logger.  Make sure to look at what a constant really is in the examples when determining to use upper case.  That is the important part because it's not a constant and thus remains lower cased.  Can be easy to be overlooked...

spotbugs/src/main/java/edu/umd/cs/findbugs/cwe/WeaknessCatalog.java
hazendaz: @ThrawnCA Please mark as ok here as it was fixed.  Noted few times here on why 'logger', please look closer at the documentation you referenced even.  Feel like this review went south here but hopefully it was worth learning for all :)

I'm ok with the PR at this point, wanting your final opinion on PR itself now so we can finish this off.

Jeeppler:
@hazendaz and @ThrawnCA is there anything outstanding which I have to change from your perspective? Is there any outstanding issue to get this merged?

hazendaz:
I'm good with this.

hazendaz:
@Jeeppler Merged, sorry for delays

Jeeppler:
@hazendaz thanks for merging it.

PR #2426, opened 2023-05-10
SER11-J. Prevent overwriting of externalizable objects

Spatrik95:
I extended the `SerializableIdiom` detector to check `readExternal()` function for the SER11-J bug.  
I created several test classes to make sure that every assigned class variable are need to be in an if condition. Also I checked that the variable in the if condition will get a new value. (so the initialized variable in the example changes value)  
I ran the extended detector for some larger repositories like ElasticSearch, but I didn't get any false-positive finding.

----


spotbugs/etc/messages.xml
```diff
@@ -4544,6 +4544,16 @@ it may indicate a misunderstanding of how serialization works.
   <p> This class contains a field that is updated at multiple places in the class, thus it seems to be part of the state of the class. However, since the field is marked as transient and not set in readObject or readResolve, it will contain the default value in any
 deserialized instance of the class.
 </p>
+]]>
+    </Details>
+  </BugPattern>
+  <BugPattern type="SE_PREVENT_EXT_OBJ_OVERWRITE">
+    <ShortDescription>Prevent overwriting of externalizable objects</ShortDescription>
+    <LongDescription>Any caller can reset the value of the object by using the readExternal() method.</LongDescription>
+    <Details>
```
ThrawnCA, line 4553: This doesn't say anything about how the issue might be resolved.

Spatrik95: I add a solution suggestion section to the description.

hazendaz: > This doesn't say anything about how the issue might be resolved.

Are you good with the proposed solution?  I think it reads ok generally but what are your thoughts?

@Spatrik95 I fixed the change log merge conflict for you.  If @ThrawnCA is good I Think we can merge in.

hazendaz (approved):

spotbugs/etc/messages.xml
hazendaz: I went ahead and merged.  Thanks.

PR #2431, opened 2023-05-16
Fix TestASM doesn't note that it reports bugs

JuditKnoll:
If we created a test, in which the `TestASM` detector found an error, running the test throws the following (or a really similar) error:

`The following errors occurred during analysis: 
TestASM doesn't note that it reports BugPattern[NM_METHOD_NAMING_CONVENTION] in category BAD_PRACTICE.`

This is because the bugtypes aren't registered in the detectors `reports` list in the `findbugs.xml`. In this PR I fix this problem.
Some tests are provided as well, and the message for the `NM_FIELD_NAMING_CONVENTION` bugtype is modified, so it is correct even if the [TestASM](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/TestASM.java) detector finds the problem.

The `ICAST_INT_CAST_TO_DOUBLE_PASSED_TO_CEIL` is detected both by [TestASM](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/TestASM.java) and (a bit more precisely) by [IDivResultCastToDouble](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IDivResultCastToDouble.java), so the same problem is reported twice, but I didn't want to remove a working solution.

----


hazendaz:
looks like the build fails with this one.

JuditKnoll:
> looks like the build fails with this one.

Yes, it was forked from the failing master, and this hid a test error which I left in the commit. Merged the master to my branch and fixed the problem, so should be okay now.

ThrawnCA (approved):

hazendaz:
thank you!

PR #2446, opened 2023-06-09
Partial fix org.apache.bcel:bcel dependency update

JuditKnoll:
This is a partial fix to the org.apache.bcel:bcel dependency update (https://github.com/spotbugs/spotbugs/pull/2278), since to solve it completely we will also need a new, currently unreleased version of bcel.

There are several issues with updating to the newest bcel version currently as @nbauma109 already mentioned (https://github.com/spotbugs/spotbugs/pull/2278#issuecomment-1532038444):
- the Constant Pool doesn not allow 0, or anything right after an 8 byte constant*,
- a problem introduced to bcel in https://github.com/apache/commons-bcel/pull/171.
    - @nbauma109 already created a PR for solving this problem: https://github.com/apache/commons-bcel/pull/221
    - I tested it on my machine, it does solve the problem.

This PR in itself only solves the first problem, since the second one is in bcel itself, and there is already a working solution for it.
Using a snapshot bcel version with the mentioned PR (https://github.com/apache/commons-bcel/pull/221) and the changes in this PR, spotbugs builds successfully on my machine.
So for a successful bcel dependency version update, we will need a new version of bcel containing the mentioned PR, and the changes in this PR.

*Quote from the JVM specification: "All eight byte constants take up two spots in the constant pool. If this is the n'th byte in the constant pool, then the next item will be numbered n+2"

----

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Bug3107916Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Bug3107916Test.java
@@ -1,0 +1,56 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.Test;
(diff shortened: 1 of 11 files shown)


hazendaz (approved):

hazendaz:
Thanks @JuditKnoll  Appreciate all the work you are putting in here.

PR #2319, opened 2023-01-19
Large command line args

simonjghammer:
----

Make sure these boxes are checked before submitting your PR -- thank you!

- [X] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code
- [X] Created test for introduced code

Let me know if I need something else. I'll try and see if I can check github once in a while.

PR #2466, opened 2023-07-07
Remove generated changing files

JuditKnoll:
Adding to gitignore and removing files which are regenerated at every build with changing data.
The https://github.com/spotbugs/spotbugs/pull/2319 PR added two files which are generated during running the [UnionResultsTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/workflow/UnionResultsTest.java#L46) test: [spotbugs-test/src/test/resources/firstFile.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/resources/firstFile.xml) and [spotbugs-test/src/test/resources/secondFile.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/resources/secondFile.xml).
The [spotbugs-test/src/test/resources](https://github.com/spotbugs/spotbugs/tree/master/spotbugs-tests/src/test/resources) directory only contains these two files and this directory was added during the mentioned PR.
Since the spotbugs build runs the tests, these files are regenerated at every build, and timestamp data is changing every time.
With having these generated files versioned, contributors need to be extra careful, not to commit these changes unnecessarily.
Tested it, deleting the directory doesn't cause any problems during build.

----


hazendaz:
@JuditKnoll Would it be better to add that to generated resources rather than into the source path?  That would remove need for the gitignore.

JuditKnoll:
> @JuditKnoll Would it be better to add that to generated resources rather than into the source path? That would remove need for the gitignore.

I would happy to move the files to generated resources, but couldn't find any directory like this. There are some generated resources (e.g. manifest files, doc, bin, deps, and build directories), but IMO these files doesn't belong in any of those directories. Is there a directory for these kind of files? Where would you put these?

hazendaz:
> > @JuditKnoll Would it be better to add that to generated resources rather than into the source path? That would remove need for the gitignore.
> 
> I would happy to move the files to generated resources, but couldn't find any directory like this. There are some generated resources (e.g. manifest files, doc, bin, deps, and build directories), but IMO these files doesn't belong in any of those directories. Is there a directory for these kind of files? Where would you put these?

Not sure how this is done in gradle land, in maven typically auto generated files go to the target folder under generated sources so they never land in the tracked source code but are properly bundled up and used with the project.

JuditKnoll:
> > > @JuditKnoll Would it be better to add that to generated resources rather than into the source path? That would remove need for the gitignore.
> > 
> > 
> > I would happy to move the files to generated resources, but couldn't find any directory like this. There are some generated resources (e.g. manifest files, doc, bin, deps, and build directories), but IMO these files doesn't belong in any of those directories. Is there a directory for these kind of files? Where would you put these?
> 
> Not sure how this is done in gradle land, in maven typically auto generated files go to the target folder under generated sources so they never land in the tracked source code but are properly bundled up and used with the project.

The files are generated during running the tests (and the tests run during build).
What do you think of moving these files to (the currently non-existent) spotbugs-test/resources folder?

hazendaz:
Hi @JuditKnoll After looking at this again, given these are the only files as  you have indicated, trying to figure out the right path may not really matter right now.  So I'll merge as-is.  If for some reason someone needs to use resources for other items in the future we will have to review again at that time but for now think we are ok.

hazendaz (approved):

PR #2484, opened 2023-07-22
[eclipse] Set eclipse to release 4.14.0 in eclipse-convention.gradle.kts

hazendaz:
missed from original change as versions did not align correctly

Already in change log for this one, so did not update further.  Everything was moved to 4.14, but discovered two parts that were not properly on the older version listed elsewhere.  So now all will be 4.14.0

### DIFF
--- a/buildSrc/src/main/kotlin/eclipse-convention.gradle.kts
+++ b/buildSrc/src/main/kotlin/eclipse-convention.gradle.kts
@@ -7,7 +7,7 @@ val pdeTool by configurations.creating {
 }
 
 eclipseMavenCentral {
-  release("4.12.0") {
+  release("4.14.0") {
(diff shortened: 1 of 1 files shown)


hazendaz:
failure seems unrelated, rebuild was same so not worrying for moment, worked locally, will wait until other items I have catch up then will try again with rebased copy.

JuditKnoll:
I could recreate this build fail using Linux locally (in Windows the build is successful), the change in the `spotless.gradle` file triggers it. The changes in the #2351 does seem to solve it.

hazendaz:
build was still failing on spotless piece so I split this to try the kotline eclipse convention part.

ThrawnCA (approved):

gtoison (approved):

PR #2507, opened 2023-08-11
Multiplatform GHA build

JuditKnoll:
This PR extends the build github action running on Linux with Windows and MacOS platforms, complete with the eclipsePlugin build.
There is one downside to this approach: the builds on different platforms are tied together, so if any of them fails, the others are cancelled. If it would be separated to different jobs I think the builds weren't tied together like this, but that way the code would contain too much code repetition. Please let me know, if it would be better the other way (or if you know any other way to separate the jobs)!
Didn't add anything to the `CHANGELOG.md`, since this doesn't change SpotBugs code, it's only a github action change. Let me know, if you think I should.

----


gtoison:
I think you can use `continue-on-error` if you don't want that all jobs are cancelled when one of them fails: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

JuditKnoll:
> I think you can use `continue-on-error` if you don't want that all jobs are cancelled when one of them fails: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

Thank you, @gtoison ! This solves the problem. See the [checks](https://github.com/spotbugs/spotbugs/actions/runs/5852554882/job/15864977579) for the https://github.com/spotbugs/spotbugs/pull/2507/commits/001332e6ab156435c336d6e1d1aa029f36744c16 commit, where the Windows fails before the MacOS finishes, but it doesn't get cancelled.

Now, the build is successful on all three platforms, but the Read the docs fails. I don't think I made any changes which should cause the Read the docs to fail. Does anyone have any idea why it fails? Is it a problem?

gtoison:
Looks like readthedocs requires a new "v2" configuration: https://blog.readthedocs.com/migrate-configuration-v2/

gtoison (approved):
LGTM, thank you!

PR #2509, opened 2023-08-14
Not able to detect SE_NO_SERIALVERSIONID bug - added unit test

yuliyasokh:
----

- [v] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
@@ -1,0 +1,20 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class Issue2436Test extends AbstractIntegrationTest {
+
+    @Test
+    public void test() {
+        performAnalysis("lambdas/Issue2436.class");
+        BugInstanceMatcher bugMatcher = new BugInstanceMatcherBuilder().build();
+        assertThat(getBugCollection(), containsExactly(1, bugMatcher));
+    }
+
+    }
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
@@ -1,6 +1,7 @@
 package edu.umd.cs.findbugs.detect;
 
 import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.annotations.Confidence;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
```diff
@@ -0,0 +1,20 @@
...
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class Issue2436Test extends AbstractIntegrationTest {
+
+    @Test
+    public void test() {
+        performAnalysis("lambdas/Issue2436.class");
+        BugInstanceMatcher bugMatcher = new BugInstanceMatcherBuilder().build();
```
JuditKnoll, line 16: Can you please add at least a bugtype to the BugInstanceMatcher?

spotbugsTestCases/src/java/lambdas/Issue2436.java
```diff
@@ -0,0 +1,10 @@
+package lambdas;
```
JuditKnoll, line 1: This issue is not about lambas, can you please move this file to the appropriate package? If you can't find better, then `ghIssues` would be okay. (Don't forget to modify the path in the test file as well.)

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
yuliyasokh: thanks, added bugtype and confidence

spotbugsTestCases/src/java/ghIssues/Issue2436.java
```diff
@@ -0,0 +1,10 @@
+package ghIssues;
+
+import java.io.Serializable;
+
+public class Issue2436 implements Serializable {
+    public void doNothing(){
```
JuditKnoll, line 6: Is this function really needed for a minimal example?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class Issue2436Test extends AbstractIntegrationTest {
+
+    @Test
+    public void test() {
+        performAnalysis("ghIssues/Issue2436.class");
+        BugInstanceMatcher bugMatcher = new BugInstanceMatcherBuilder()
+                .bugType("SE_NO_SERIALVERSIONID")
+                .withConfidence(Confidence.LOW)
+                .build();
```
JuditKnoll, line 20: There are usually two ways when we use BugInstanceMatcher:
1. negative tests: the BugInstanceMatcher should contain minimal info (only the bugtype, maybe the containing class), used when we want to assert the number of hits of a certain bugtype, or we want to check that a bug is not reported at all.
2. positive tests: the BugInstanceMatcher should contain maximal info, used when we want to assert that a bug is reported in a certain way exactly.

IMO adding the confidence makes it closer to the positive tests, than to the negative, general ones.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
```diff
@@ -0,0 +1,24 @@
...
+        assertThat(getBugCollection(), containsExactly(1, bugMatcher));
+    }
+
+    }
```
JuditKnoll, line 24: Can you please modify the indenting to be conform to the project? Or you can simply run `./gradlew spotlessApply` or more specifically the `./gradlew :spotbugs-tests:spotlessApply` command as you can see in the build fail.

JuditKnoll (changes requested):
Please make sure that the build is successful when creating your PR.

spotbugsTestCases/src/java/ghIssues/Issue2436.java
yuliyasokh: true, removed it

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2436Test.java
yuliyasokh: ok, got it

JuditKnoll (approved):

yuliyasokh:
@JuditKnoll could you please merge this request as I do not have write access?

PR #2511, opened 2023-08-14
 DCL01-J. Do not reuse public identifiers from the Java Standard Library

norbo03:
I created _DontReusePublicIdentifiers_ detector for rule [ DCL01-J](https://wiki.sei.cmu.edu/confluence/display/java/DCL01-J.+Do+not+reuse+public+identifiers+from+the+Java+Standard+Library) to check clashes between user-defined identifiers and publicly available identifiers from the Java Standard Library.

I created several test cases considering the possible locations where user-defined names would appear. I checked class and interface names, field and variable declarations, and method names. In the case of classes, I checked multiply nested classes too. The test classes are placed into a separate package because of the number of source files.

I have slightly adjusted [AnnotationMatcherTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/AnnotationMatcherTest.java) because my detector found bugs in the test classes and the test was failing. Instead of checking that found bug matches the annotation, it inspects the number of bugs.

----


ThrawnCA (changes requested):
Build is failing.

CHANGELOG.md
```diff
@@ -26,6 +27,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fixed annotation on generic or array incorrectly considered for the nullability of a method parameter or return type ([#2502](https://github.com/spotbugs/spotbugs/issues/2502))
 
 ### Added
+* New simple name-based AnnotationMatcher for exclude files (now bug annotations store the class java annotations in an attribute called `classAnnotationNames`). For example, use like <Match><Annotation name="org.immutables.value.Generated" /></Match> in an excludeFilter.xml to ignore classes generated by the Immutable framework. This ignores all class, method or field bugs in classes with that annotation.
+* Added the Common Weakness Enumeration (CWE) taxonomy to the Static Analysis Results Interchange Format (SARIF) report. The short and long description for the CWEs are retrived from a JSON file which is a slimmed down version of the official comprehensive CWE XML from MITRE. The JSON contains information about all CWEs. ([#2410](https://github.com/spotbugs/spotbugs/pull/2410)).
```
JuditKnoll, line 31: These lines are already contained in this file with `-` instead of `*`. Can you please resolve this merge problem?

spotbugs-tests/src/test/resources/firstFile.xml
JuditKnoll, line 1: Can you please remove this file?

spotbugs-tests/src/test/resources/secondFile.xml
JuditKnoll, line 1: Can you please remove this file?

JuditKnoll (changes requested):
It looks like there are some merge problems. Can you please resolve them?

JuditKnoll:
> Build is failing.

I think earlier "only" the Read the docs build failed, not the actual spotbugs build. Approved the workflow, the build succeeded. @ThrawnCA Do you know how to prevent the Read the docs build from failing?

JuditKnoll:
> > Build is failing.
> 
> I think earlier "only" the Read the docs build failed, not the actual spotbugs build. Approved the workflow, the build succeeded. @ThrawnCA Do you know how to prevent the Read the docs build from failing?

The Read the docs build problem got solved.
@norbo03 Can you please merge master to your branch?

spotbugs-tests/src/test/resources/firstFile.xml
norbo03: firstFile.xml was removed

spotbugs-tests/src/test/resources/secondFile.xml
norbo03: secondFile.xml was removed

CHANGELOG.md
norbo03: @JuditKnoll Yeah, sure. Thank you for pointing out this problem.

gtoison:
Thanks for the PR @norbo03 
I looked at the list of public identifiers and I think it should only include the _public_ parts of the JDK. For instance the `Axis` class name is twice (!) in the JDK, in `com.sun.org.apache.xerces.internal.impl.xpath.XPath.Axis` and `com.sun.org.apache.xml.internal.dtm.XPath`. Neither of these classes in "exported" packages according to the `module-info` of these modules, so basically they shouldn't be used at all, and I'd argue that naming your own class "Axis" would be fine.

I suppose that this identifiers list was built with some kind of script, would it be possible to share the script? 
Maybe this might be part of the project build, so new classes added to the JDK get added automatically as we upgrade versions?
There are also some `.db` in the project for this kind of long list, for instance https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/longInstant.db

spotbugs/etc/messages.xml
```diff
@@ -8991,7 +8999,144 @@ Using floating-point variables should not be used as loop counters, as they are
...
+      <p>See SEI CERT rule <a href="https://wiki.sei.cmu.edu/confluence/display/java/DCL01-J.+Do+not+reuse+public+identifiers+from+the+Java+Standard+Library">DCL01-J. Do not reuse public identifiers from the Java Standard Library</a>.</p>
+      ]]>
+        </Details>
+    </BugPattern>
+    <BugPattern type="PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_FIELD_NAMES">
+        <ShortDescription>Do not reuse public identifiers from JSL as field name</ShortDescription>
+        <LongDescription>The used identifier {2} as field name in class {1} in source file {3} shadows the publicly available identifier from the Java Standard Library.</LongDescription>
+        <Details>
+            <![CDATA[<p>
+      It is crucial to avoid reusing public identifiers from the Java Standard Library as field names in your code.
```
ThrawnCA, line 9045: "Crucial" is probably putting it too strongly. It's poor practice, but it can work if you're careful, especially if your application doesn't make use of the relevant part of the standard library.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/PublicIdentifiers.java
```diff
@@ -0,0 +1,11451 @@
...
+        addPUBLIC_IDENTIFIERSSS();
+        addPUBLIC_IDENTIFIERSTU();
+        addPUBLIC_IDENTIFIERSVZ();
+    }
+
+    public static boolean isPartOfStandardLibrary(String containingPackage) {
+        return STANDARD_LIBRARY_PACKAGES.stream().anyMatch(containingPackage::startsWith);
+    }
+
+    private static void addSTANDARD_PACKAGES() {
```
ThrawnCA, line 54: What is going on with the capitalisation of this name?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/PublicIdentifiers.java
ThrawnCA, line 1: Is it possible to handle this via reflection instead of an exhaustive list?

ThrawnCA (changes requested):

spotbugs/etc/messages.xml
norbo03: @ThrawnCA  I have reconsidered the phrasing and I agree with your point. Thanks for mentioning it!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/PublicIdentifiers.java
norbo03: @ThrawnCA This has been updated. It was some strange idea of mine and a mistake.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/PublicIdentifiers.java
norbo03: I think using reflection during runtime for collecting all the publicly available identifiers would require too much computational effort. Therefore I don't think it is a viable solution for this issue.

norbo03:
> Thanks for the PR @norbo03 I looked at the list of public identifiers and I think it should only include the _public_ parts of the JDK. For instance the `Axis` class name is twice (!) in the JDK, in `com.sun.org.apache.xerces.internal.impl.xpath.XPath.Axis` and `com.sun.org.apache.xml.internal.dtm.XPath`. Neither of these classes in "exported" packages according to the `module-info` of these modules, so basically they shouldn't be used at all, and I'd argue that naming your own class "Axis" would be fine.
> 
> I suppose that this identifiers list was built with some kind of script, would it be possible to share the script? Maybe this might be part of the project build, so new classes added to the JDK get added automatically as we upgrade versions? There are also some `.db` in the project for this kind of long list, for instance https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/longInstant.db

I believe integrating an automatic update feature into the project is a fantastic idea! In the past, I utilized scripts to extract public identifiers. Regrettably, these scripts are now outdated, and the results they produced necessitated extensive manual adjustments. While these scripts could serve as a foundation, they do require updates. I've gathered them [here](https://github.com/norbo03/java-identifier-collector) for reference.

gtoison:
Thanks, I can't seem to accept the repository you have linked (I get a 404 error page), maybe it is private?

norbo03:
> Thanks, I can't seem to access the repository you have linked (I get a 404 error page), maybe it is private?

I have already changed it, can you please check it now?

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/PublicIdentifiers.java
hazendaz: I don't feel this is very maintainable like this.  This had to have taken quite some time to get together.  Since others approved it, will merge but just don't see how its maintainable given its already crazy size.  How would anyone know something is missing?

PR #2513, opened 2023-08-15
Migrate readthedocs to v2

gtoison:
Readthedocs now requires a configuration file, see: https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+    # You can also specify other tool versions:
+    # nodejs: "20"
+    # rust: "1.70"
+    # golang: "1.20"
+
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
```
JuditKnoll, line 9: However, this comes from the default template, the other `*.yaml` use `ubuntu-latest`. It may be worth considering using that to be a bit more uniform. It's okay like this as well, so I'll approve this PR.

JuditKnoll (approved):
It was a bit strange for me, that much earlier there were already failing builds (see https://github.com/spotbugs/spotbugs/pull/2351), but most of the builds worked, now most builds are failing, but there are still some which succeeds.
But it can be explained by this blogpost about the migration: https://blog.readthedocs.com/migrate-configuration-v2/

Thank you @gtoison !

.readthedocs.yaml
gtoison: It might be better to use a fixed version so we don't have PR randomly failing, because there was an unrelated change somewhere

Issue #2369, opened 2023-03-03
False positive MS_EXPOSE_REP when returning static record field


xzel23:
This triggers a false positive (code simplified):

```
record A {
    private static final A DEFAULT_A = new A();
    public static A defaultA() {
        return DEFAULT_A;
    }
}
```

With the following message about returning a reference to an array (sic!): 
> A public static method returns a reference to an array that is part of the static state of the class. Any code that calls this method can freely modify the underlying array. One fix is to return a copy of the array.

gtoison:
Sorry for the belated answer, I've tried reproducing the issue in the above PR but that does not seem to work (the bug is not detected)

gtoison:
So, the issue that records could be considered mutable was fixed by #2514 and the misleading message (only mentionning arrays) was fixed in #2535
There's also the open issue that a class is considered mutable if it has a "setter like" method, see issue #1797

PR #2515, opened 2023-08-15
test: attempt to reproduce issue #2369

gtoison:
This sample apparently does not reproduce issue #2369


ThrawnCA (approved):

gtoison:
The issue was fixed by #2514

PR #2522, opened 2023-08-19
Issue #2465 reproducer and fix

gtoison:
The AvailableLoad -> ValueNumber[] was modified into a bidirectional map in https://github.com/spotbugs/spotbugs/pull/1864 to remove a costly linear search in the getLoad() method.
Fixed a direct update to the underlying map in mergeAvailableLoadSets() to also update the reverse map

----


hazendaz (approved):

hazendaz:
Some day we should rewrite a lot of this code so its a lot more clear on what its doing.  Some of these field names without any comments makes it hard to review and tricky to fix things like this :(  LGTM.

gtoison:
I must admit that I have little to no idea how all of this works...
I feel like some of the code could be simplified because it tries hard to optimize away the allocations by caching/reusing objects. Maybe these optimisations are now counter productive with modern JVM.
But that's the easy part... the algorithms are not trivial

hazendaz:
Same 😀

Issue #2071, opened 2022-05-27
FL_FLOATS_AS_LOOP_COUNTERS: false positive in SpotBugs 4.7.0


xzel23:
SpotBugs 4.7.0 reports FL_FLOATS_AS_LOOP_COUNTERS twice for this code:

```
    public static int ilog10(double x) {
        if (x >= 1.0) {
            int i = 0;
            while ((x /= 10.0) >= 1.0) { // <-- HERE
                i++;
            }
            return i;
        } else {
            int i = -1;
            while ((x *= 10.0) < 1.0) { // <-- HERE
                i--;
            }
            return i;
        }
    }
```

The explanation refers to [NUM09-J](https://wiki.sei.cmu.edu/confluence/display/java/NUM09-J.+Do+not+use+floating-point+variables+as+loop+counters), note the part in bold:

> Floating-point variables must not be used as loop counters. Limited-precision IEEE 754 floating-point types cannot represent
> 
> All simple fractions exactly.
> All decimals precisely, even when the decimals can be represented in a small number of digits.
> All digits of large values, meaning that incrementing a large floating-point value might not change that value within the available precision.
> For the purpose of this rule, a loop counter is an induction variable that is used as an operand of a comparison expression that is used as the controlling expression of a do, while or for loop. An induction variable is **a variable that gets increased or decreased by a fixed amount** on every iteration of a loop [Aho 1986]. Furthermore, the change to the variable must occur directly in the loop body (rather than inside a function executed within the loop.)
> 

In both cases `x` is not increased or decreased by a fixed amount, so this rule should not apply.

As a side note I think SpotBugs should release less often and test more thoroughly since for some time now each new version of SpotBugs introduces a bunch of false positives that I already think about removing SpotBugs altogether.

KengoTODA:
Thanks for your comment,

> As a side note I think SpotBugs should release less often and test more thoroughly since for some time now each new version of SpotBugs introduces a bunch of false positives

This is an important note, let me hold a discussion at https://github.com/spotbugs/spotbugs/discussions/2076

gtoison:
This was fixed by #2167

PR #2539, opened 2023-08-29
test: sample checking that FP in issue #2071 was fixed by PR #2167

gtoison:
Added the sample provided in #2071 so we can check that it was fixed in #2167


JuditKnoll (approved):
Maybe it's worth to mention the issue number in a comment in the code, but otherwise good, so I'll approve.

ThrawnCA (approved):

PR #2581, opened 2023-09-30
Update templates

hazendaz:
note: Appears direct updates didn't quite make it and some where messed up.  There were branches for these that github created, so clearing those out and bringing in on one.  While the intellij item could be seen as redundant these are all just following same info indicating to everyone that sees that we don't support intellij plugin as well as guiding to where to update.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,10 +2,10 @@
 
 Before you submit issue or pull request, please check the following points.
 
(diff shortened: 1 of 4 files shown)


hazendaz:
sonar cloud is unimportant on this.  originally wrote these manually on github but not sure what happened as there were 4 branches for same.  Just syncing it up. If we need to change again we can do so afterwards.

PR #2589, opened 2023-10-01
[GHA] Update eclipse to 4.29 and allow to run CI on any branch

hazendaz:


hazendaz:
Eclipse version used for this was incredibly out of date.  Eclipse is used for the build.  Like most Eclipse plugins they should be built against the release trains.  No old ones are supported.  Thus this brings it up to date.

PR #2604, opened 2023-10-05
Add build support for jdk 11 - fixes #2598

### LINKED ISSUE #2598: GHA - Add jdk 11 along with temp drop back on Eclipse to java 11 build
In trying to isolate personal issues on my windows platform, I learned that gradle was using oracles hack to the jvm after one of the installs and ignored my path setup.  After fixing that, I was able to get same results on local as on GHA.  While working on this I got GHA working with jdk 11 and 20 in addition to 17.  There is one test case in this that simply does not work with java 11.  Going to ignore on that for java 11 only, drop eclipse slightly back as well.  Originally I wanted jdk 21 to show case it working due to various user inquires but fact of the matter is gradle doesn't support it yet until 8.4 and still won't with kotlin which means we cannot actually show that.  So that is where jdk 20 comes into place.

Will have this over in next few days to be in 4.8.0

hazendaz:
fixes #2598
- Moved to 'zulu' for builds
- Run in parallel 4 max
- Drop Eclipse back to 4.25 to allow jdk 11 on build
- Drop toolchains back to prior usage to allow jdk 11 (github doesn't have toolchains on jdk 11 installed so they fail with toolchains)
- Drop spring back to 5.3.30 as project is not on jakartaEE yet

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
         include:
          - os: flyci-macos-large-latest-m1
            java: 21
-        java: [17, 21]
+        java: [11, 17, 21]
       max-parallel: 6
     runs-on: ${{ matrix.os }}
     continue-on-error: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,22 +31,22 @@ jobs:
       - name: Download Eclipse on Ubuntu
         if: matrix.os == 'ubuntu-latest'
         run: |
-          curl -L 'https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-4.29-202309031000/eclipse-SDK-4.29-linux-gtk-x86_64.tar.gz&r=1' --output eclipse-SDK-4.29-linux-gtk-x86_64.tar.gz
-          tar xzf eclipse-SDK-4.29-linux-gtk-x86_64.tar.gz eclipse
+          curl -L 'https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-4.24-202206070700/eclipse-SDK-4.24-linux-gtk-x86_64.tar.gz&r=1' --output eclipse-SDK-4.24-linux-gtk-x86_64.tar.gz
+          tar xzf eclipse-SDK-4.24-linux-gtk-x86_64.tar.gz eclipse
           echo "eclipseRoot.dir=$(pwd)/eclipse" | tee eclipsePlugin/local.properties
       - name: Download Eclipse on Windows
         if: matrix.os == 'windows-latest'
         run: |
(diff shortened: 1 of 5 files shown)


hazendaz:
Requires junit 5 changes before this will work.  More specifically it requires a single code change to tests as a bad test exists here on java 11 so its disabled specifically on java 11 with code updates for junit 5.

hazendaz:
Toolchains will be added back at later date in same way once we move to requiring java 17.  For now this shows a bit more of what we support with obvious caveat that java 8 is not shown here.

.github/workflows/build.yml
```diff
@@ -9,39 +9,41 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        java: [11, 17, 20]
```
soloturn, line 12: why not yet 21 ?

JuditKnoll: It's stated in https://github.com/spotbugs/spotbugs/issues/2598: 
> Originally I wanted jdk 21 to show case it working due to various user inquires but fact of the matter is gradle doesn't support it yet until 8.4 and still won't with kotlin which means we cannot actually show that. So that is where jdk 20 comes into place.

hazendaz:
I'll post the specific error on that shortly for clarity

hazendaz:
Why jdk 21 not here...

```
* What went wrong:
Could not determine the dependencies of task ':buildSrc:compileKotlin'.
> Unknown Kotlin JVM target: 21
```

hazendaz:
going to partially use this and leave rest to look at later.  Not sure why but it wants to download latest components from eclipse.  Its been about 2 month since I last worked on this but seem to recall jdk 11 working then, now it does not.  Looked over changes not seeing it but don't want to hold up release tonight.  Will however peal off parts of this (ie jdk 21 parts).

JuditKnoll:
> going to partially use this and leave rest to look at later. Not sure why but it wants to download latest components from eclipse. Its been about 2 month since I last worked on this but seem to recall jdk 11 working then, now it does not. Looked over changes not seeing it but don't want to hold up release tonight. Will however peel off parts of this (ie jdk 21 parts).

I think the java version of the downloaded eclipse's code is too new for java 11 (see https://github.com/spotbugs/spotbugs/pull/2589). We would need eclipse 4.24 or older to support java 11 (see https://wiki.eclipse.org/Eclipse/Installation).

hazendaz:
guessing the issue is osgi related, need to look more into it.  Its pulling almost the latest binaries which is the problem.

ThrawnCA (approved):

ThrawnCA (changes requested):
Build is broken.

hazendaz:
found issue, will be pushing up shortly...

ThrawnCA (approved):

Issue #2280, opened 2022-12-15
Code failed to build on Mac M1


testn:
It seems to refer to the older version of Eclipse which does not support Mac M1

gtoison:
This should be fixed according to #2507

hazendaz:
closing as addressed.  Further we expect after this next release to start raising our jdk requirements so we can then bring eclipse up much further.  it is true its stuck in time but we are aware of the issues and working on it.  As noted this was fixed already on master with changes I did earlier.  The PR noted will show multi platform usage.  If still an issue after this next release, please reopen but as noted we will fast track moving up to newer releases.

Spatrik95:
Hello! I still have this kind of problem. I'm on an Apple Mac M2 with Sonoma 14.0.  
On a `./gradlew clean bulld` I get the following message:
```
Execution failed for task ':eclipsePlugin:compileJava'.
> Could not resolve all files for configuration ':eclipsePlugin:compileClasspath'.
   > Could not find org.eclipse.platform:org.eclipse.swt.cocoa.macosx.aarch64:3.113.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/eclipse/platform/org.eclipse.swt.cocoa.macosx.aarch64/3.113.0/org.eclipse.swt.cocoa.macosx.aarch64-3.113.0.pom
     Required by:
         project :eclipsePlugin > org.eclipse.platform:org.eclipse.swt:3.113.0
```

Currently the only way, is a workaround: I exclude these modules from the build process:  
```
./gradlew clean build -x :eclipsePlugin:compileJava -x :eclipsePlugin-junit:compileTestJava -x :eclipsePlugin-test:compileJava -x :eclipsePlugin-test:checkstyleMain
```

hazendaz:
Are you building from latest master?  The version you listed is older than our current master and latest eclipse supports the arm module at least per eclipse github.  The 3.113 version is from 2019.  I think if you rebase it may work.  Clearly 2019 eclipse would not support that arch type 😉

Now even if that solves it for you, we are going to slightly move back to eclipse 4.25 release so you may want to check existing recent pull request I raised a few days ago to check it as well.

Thanks for work around in any case as it may help other contributors.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Spatrik95:
[...]

PR #2611, opened 2023-10-07
#2280 Update Eclipse plugin release number to 4.29.0

Spatrik95:
Update to a newer Eclipse plugin release version to make the eclipseModul buildable on ARM based Apple Mac systems.

----


hazendaz:
taking but expect to downgrade to 4.25.  Will follow up on the issue to have that rechecked and will now double check online as I'm on real PC now and can confirm what eclipse is or isn't in support level for m1.

PR #2609, opened 2023-10-06
Fix first batch of Gradle buildDir deprecations

beatbrot:
Gradle recently deprecated [Project.getBuildDir()](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getBuildDir--). This PR removes many of the simpler usages.

Of course, this is not all. There are more usages of buildDir and even other deprecations in the build script. But this is a first step :)

PR #2616, opened 2023-10-09
Fix `createReleaseBody` Gradle Task

beatbrot:
Fixes the `createReleaseBody` task that I broke in my earlier PR #2609

@hazendaz FYI


JuditKnoll (approved):
Tested `./gradlew assemble createReleaseBody` locally on this, it runs successfully.

Issue #1416, opened 2021-02-01
Schema Target Namespace


kev22257:
If you look at https://raw.githubusercontent.com/findbugsproject/findbugs/master/findbugs/etc/findbugsfilter.xsd you'll see the schema target namespace as http://findbugs.sourceforge.net/filter/3.0.0 which is a dead URL.  Any way to update this to a valid value?  This is causing XML validation errors.  I poked around and didn't see any 4.0.0 version either.  Also, this is documented in the filter file documentation so that probably needs to be updated too (https://spotbugs.readthedocs.io/en/latest/filter.html)?

KengoTODA:
It's already replaced with `https://github.com/spotbugs/filter/3.0.0` in this repository:

https://github.com/spotbugs/spotbugs/blob/2367aeea72ca654d2818c6a37578aa1f58db131e/spotbugs/etc/findbugsfilter.xsd#L2-L3

It's still invalid value responding 400.

NeetigyaPod:
Hi, would like to work on this, is this still up for grab?

hazendaz:
@NeetigyaPod Sure take a look into this one.  I presume its still valid and not being worked on.

NeetigyaPod:
How should I go about to find the correct URL and the correct version? I am a bit new to Open-source, will look around in the docs till then.

hazendaz:
Hi, you would want to add this element ```xsi:schemaLocation=https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/master/spotbugs/etc/findbugsfilter.xsd ```

No need to call out the version...

NeetigyaPod:
Hi, have created a PR with the solution.

PR #2625, opened 2023-10-12
fix: Schema Location for findbugsfilter

NeetigyaPod:

----

Closes: #1416 . Added xsi:schemaLocation to solve issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2022-??-??
 
+### Fixed
+- Fixed schema location for findbugsfilter.xsd ([[#1416](https://github.com/spotbugs/spotbugs/issues/1416)])
+
 ## 4.8.0 - 2023-10-11
 
 ### Changed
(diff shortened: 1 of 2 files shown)


ThrawnCA (approved):

CHANGELOG.md
```diff
@@ -40,7 +40,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report SIC_INNER_SHOULD_BE_STATIC for classes annotated with JUnit Nested ([#560](https://github.com/spotbugs/spotbugs/issues/560))
 - Detect created, but not-thrown exceptions, which are created by not the constructor ([#2547](https://github.com/spotbugs/spotbugs/issues/2547))
 - Fixed eclipse plugin Effort.values pass to effortViewer as required cast to varargs ([#2579](https://github.com/spotbugs/spotbugs/pull/2579))
-
+- Fixed schema location for findbugsfilter.xsd ([[#1416](https://github.com/spotbugs/spotbugs/issues/1416)])
```
hazendaz, line 43: @NeetigyaPod This needs moved into new section at top as release just occurred, make sure you are rebased and you an create similar section as you used here in the unreleased area at top.

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schema targetNamespace="https://github.com/spotbugs/filter/3.0.0" elementFormDefault="unqualified"
```
hazendaz, line 2: @NeetigyaPod YOu need the other side of this that I has put on that ticket.  This itself is broken.  The schema location is a list of property=value pairs where the left side is the namespace (what you have now) and the right is the location of the actual xsd.  Note that the namespace is not actually an endpoint.  So the purpose is so you can link to the source right from your IDE.

NeetigyaPod: Oh makes sense. Committing the change and pushing, thank you.

CHANGELOG.md
NeetigyaPod: Rectified.

PR #2637, opened 2023-10-16
Disabled DontReusePublicIdentifiers due to false positives rate

### LINKED ISSUE #2627: 4.8.0 PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES False positives
Upgrading to 4.8.0 I get a lot of what I believe are false positives like:

```
PI | Class name ?>?1/1??? in source file ?>?2/1??? shadows the publicly available identifier from the Java Standard Library.
Bug type PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES
In class com.my.package.SomeException
At SomeException.java:[lines 9-19]
```

Where SomeException is a custom exception subclass.

And the class name and source in the message are weird characters instead of the actual class name and source (which is shown in the details)

gtoison:
The DontReusePublicIdentifiers  detector searches for class names in the entire JDK, including classes that are in the internal packages such as `com.sun.*`. Class names such as `Filter` appear in the packages and using them is reported as an issue.

Fixes #2627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Fixed schema location for findbugsfilter.xsd ([[#1416](https://github.com/spotbugs/spotbugs/issues/1416)])
 - Fixed missing null checks ([[#2629](https://github.com/spotbugs/spotbugs/issues/2629)])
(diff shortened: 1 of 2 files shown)


iloveeclipse (approved):

iloveeclipse:
Thanks!

Issue #2647, opened 2023-10-16
DMI_DOH reported on org.junit.Assert.fail


bwmeier:
The following error is reported on invocation of `org.junit.Assert.fail(String)`:

> DMI	D'oh! A nonsensical invocation of fail(String) in functionXYZ(String)
> [Bug type DMI_DOH (click for details)](https://github.com/spotbugs/spotbugs/issues/new#DMI_DOH)
> In class Example
> In method functionXYZ(String)
> Called method org.junit.Assert.fail(String)
> Passing String constant as value that should be null checked
> String constant "Element should exist"
> At Example.java:[line 361]

JuditKnoll:
Can you please provide an example code, where you got the error reported?

gtoison:
This seems to be fixed by #2643

PR #2643, opened 2023-10-16
Fix escaping exception from DumbMethods

JuditKnoll:
The brackets in [DumbMethods](https://github.com/spotbugs/spotbugs/pull/2605/files#diff-bb14d9be707bded935e80462e027bcb6aa99a83af5f5fa4cd39faf3cc1352a8b) in https://github.com/spotbugs/spotbugs/pull/2605 got mixed up, which caused https://github.com/spotbugs/spotbugs/issues/2640. This PR fixes this problem.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [X] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #2650, opened 2023-10-17
test: sample reproducing issue #2647

gtoison:
This is fixed by #2643, the test should pass once it is merged


gtoison:
Looks like #2643 fixed that issue as well

JuditKnoll (approved):

PR #2671, opened 2023-11-01
Simplify PDE classpath resolving

iloveeclipse:
I've noticed that starting SpotBugs analysis in the IDE where workspace contains lot of plugin project takes sometimes longer as the analysis itself.

Turned out, we do lot of duplicated work resolving plugins classpath, while Eclipse already resolved almost everything for us via JavaRuntime.computeDefaultRuntimeClassPath().

Just removing lot of code here allows reduce startup times for spotbugs analysis, especially if running analysis on ~100 plugins in parallel.

### DIFF
--- a/eclipsePlugin/src/de/tobject/findbugs/builder/PDEClassPathGenerator.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/builder/PDEClassPathGenerator.java
@@ -18,20 +18,11 @@
  */
 package de.tobject.findbugs.builder;
 
-import java.io.File;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


hazendaz (approved):

hazendaz:
LGTM and makes sense.  Thanks!

hazendaz:
@iloveeclipse Can you add an entry to the change log and send another PR so we don't lose track of this change?

iloveeclipse:
> I will push a new PR soon and along with that also updated README.

Here is it: https://github.com/spotbugs/spotbugs/pull/2673

Issue #2658, opened 2023-10-20
Error while parsing `message_fr.xml`


CyrilSicard:
Problem
--
An error occured while reading file `message_fr.xml`.

Version
--
SpotBugs 4.8.0

Error Log
--
```java
edu.umd.cs.findbugs.PluginException: Couldn't parse "jar:file:/home/gradle/.gradle/caches/modules-2/files-2.1/com.github.spotbugs/spotbugs/4.8.0/b4f9cd4fc6352be6226efa0c5f59f0dc10219732/spotbugs-4.8.0.jar!/messages_fr.xml"
	at edu.umd.cs.findbugs.PluginLoader.addCollection(PluginLoader.java:1336)
	at edu.umd.cs.findbugs.PluginLoader.getMessageDocuments(PluginLoader.java:1235)
	at edu.umd.cs.findbugs.PluginLoader.loadPluginComponents(PluginLoader.java:750)
	at edu.umd.cs.findbugs.PluginLoader.<init>(PluginLoader.java:229)
	at edu.umd.cs.findbugs.PluginLoader.getPluginLoader(PluginLoader.java:1378)
	at edu.umd.cs.findbugs.Plugin.addCustomPlugin(Plugin.java:671)
	at edu.umd.cs.findbugs.Plugin.addCustomPlugin(Plugin.java:663)
	at edu.umd.cs.findbugs.Plugin.loadCustomPlugin(Plugin.java:655)
	at edu.umd.cs.findbugs.Plugin.loadCustomPlugin(Plugin.java:648)
	at edu.umd.cs.findbugs.FindBugsCommandLine.handleOptionWithArgument(FindBugsCommandLine.java:147)
	at edu.umd.cs.findbugs.TextUICommandLine.handleOptionWithArgument(TextUICommandLine.java:619)
	at edu.umd.cs.findbugs.config.CommandLine.parse(CommandLine.java:333)
	at edu.umd.cs.findbugs.config.CommandLine.parse(CommandLine.java:300)
	at edu.umd.cs.findbugs.FindBugs.processCommandLine(FindBugs.java:345)
	at edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1221)
Caused by: org.dom4j.DocumentException: Error on line 1723 of document  : Le type d'élément "MessageCollection" doit se terminer par la balise de fin correspondante "</MessageCollection>".
	at org.dom4j.io.SAXReader.read(SAXReader.java:511)
[...]

PR #2676, opened 2023-11-05
test: added a smoke test checking that we can read the XML message files

gtoison:
Check that the XML files are well formed by reading them

This is a follow-up to issue #2658 where I had made a mistake when updating the `messages_fr.xml` file


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/LoadMessagesTest.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * FindBugs - Find bugs in Java programs
```
ThrawnCA, line 2: New files should reference SpotBugs not FindBugs, and the copyright notice should not mention the University of Maryland.

ThrawnCA (changes requested):

JuditKnoll (approved):

ThrawnCA (approved):

PR #2679, opened 2023-11-05
Recognize JUnit's `@MethodSource` when looking for uncalled private methods

gtoison:
`@MethodSource` allows for private methods as parameters providers, recognize them when looking for uncalled private methods


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindUncalledPrivateMethods.java
```diff
@@ -56,13 +47,30 @@ public class FindUncalledPrivateMethods extends BytecodeScanningDetector impleme
     private HashSet<MethodAnnotation> definedPrivateMethods, calledMethods;
 
     private HashSet<String> calledMethodNames;
+    private HashSet<String> jUnitSourceMethodNames;
```
ThrawnCA, line 50: Fields should probably use the interface type, `Set`.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindUncalledPrivateMethods.java
```diff
@@ -24,16 +24,7 @@
...
-import org.apache.bcel.classfile.Constant;
-import org.apache.bcel.classfile.ConstantCP;
-import org.apache.bcel.classfile.ConstantMethodHandle;
-import org.apache.bcel.classfile.ConstantMethodref;
-import org.apache.bcel.classfile.ConstantNameAndType;
-import org.apache.bcel.classfile.ConstantPool;
-import org.apache.bcel.classfile.ConstantUtf8;
-import org.apache.bcel.classfile.JavaClass;
-import org.apache.bcel.classfile.Method;
+import org.apache.bcel.classfile.*;
```
hazendaz, line 27: @gtoison Please do not use 'star' imports.

JuditKnoll (changes requested):
The `@MethodSource` can have no arguments declared (source [the API description](https://junit.org/junit5/docs/5.8.0/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/MethodSource.html)):

> If no factory method names are declared, a method within the test class that has the same name as the test method will be used as the factory method by default.

Can you please add a testcase for this?

gtoison:
@JuditKnoll good point, will do!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2379Test.java
```diff
@@ -0,0 +1,29 @@
...
+        BugInstanceMatcher bugTypeMatcher = new BugInstanceMatcherBuilder()
+                .bugType("UPM_UNCALLED_PRIVATE_METHOD")
+                .inMethod("unusedValues")
+                .build();
+
+        // Check that we've found the uncalled private method
+        assertThat(getBugCollection(), containsExactly(1, bugTypeMatcher));
+
+        // Check that it was the only bug we've found
+        assertThat(getBugCollection(), containsExactly(1, new BugInstanceMatcherBuilder().build()));
```
JuditKnoll, line 27: Can you please add the bugtype to the `BugInstanceMatcher`? If later a new detector is added, which finds a bug in the test class, this test would fail, which is not intended.

JuditKnoll (changes requested):
I have one small change request, otherwise it looks good. Thanks :)

JuditKnoll (approved):

PR #2681, opened 2023-11-06
New detector for Env Var usages instead of Java Properties

JuditKnoll:
New detector which finds environment variable usages (`System.getenv()` calls), where Java Properties (`System.getProperty()`) could have been used. The latter is more portable and secure. Since there is no Java Property for every environment variable, a black list of the corresponding pairs are used. 
I used the Java Properties found at the [API](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getProperties--) of the java.lang.System.getProperties(), the [System Utilities / System Properties](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html) and the [Rich Internet Application / System Properties](https://docs.oracle.com/javase/tutorial/deployment/doingMoreWithRIA/properties.html) Oracle sites.

See the SEI CERT rule [ENV02-J. Do not trust the values of environment variables](https://wiki.sei.cmu.edu/confluence/display/java/ENV02-J.+Do+not+trust+the+values+of+environment+variables).

Co-authored-by: @katheg

----


iloveeclipse:
> The latter is more portable and secure

Could you please explain how command line is more secure, compared to the environment variables?

JuditKnoll:
> > The latter is more portable and secure
> 
> Could you please explain how command line is more secure, compared to the environment variables?

The default values of system properties are set by the JVM upon startup, and these can be trusted. However, these can be overwritten when starting the program either by command line or by property file. These can't be modified after starting the program, but the environment variables are globally accessible and can be modified outside of the program.

Even the javadoc at the `System.getenv()` function mentions that environment variables may have unintended side effects:
> Environment variables have a more global effect, because they are visible to all descendants of the process which defines them, not just the immediate Java subprocess. They can have subtly different semantics, such as case insensitivity, on different operating systems. For these reasons, environment variables are more likely to have unintended side effects. It is best to use system properties where possible. Environment variables should be used when a global effect is desired, or when an external system interface requires an environment variable (such as PATH).

I think this part is summed up quite well at the [SEI CERT rule site](https://wiki.sei.cmu.edu/confluence/display/java/ENV02-J.+Do+not+trust+the+values+of+environment+variables).

iloveeclipse:
The rule, in regards of **security** is just plain wrong. If attacker can change environment variables supplied to ProcessBuilder, he can in same way change system properties. Also, similar to program arguments, environment variables can't be changed after process startup. The only issue I can see is the portability - but again, I don't see how that would affect security.

So with that, if sou would like to keep the detector, I would propose to set this detector disabled by default, to avoid confusion / spreading false information.

JuditKnoll:
> The rule, in regards of **security** is just plain wrong. If attacker can change environment variables supplied to ProcessBuilder, he can in same way change system properties. Also, similar to program arguments, environment variables can't be changed after process startup. The only issue I can see is the portability - but again, I don't see how that would affect security.

There may be some way to change system properties, but the language tries to make it a bit more difficult. E.g. they are not visible to all descendants, just the immediate subprocess. (Of course, with reflection or other sneaky ways you could access it.)

> So with that, if sou would like to keep the detector, I would propose to set this detector disabled by default, to avoid confusion / spreading false information.

IMO just because the some part of the message may cause confusion, we should not disable the detector by default. It only means that confusing part of the message should be adjusted. If we concat to every detectors' message something irrelevant, not every detector should be disabled, just the newly added irrelevant part removed.
I will modify the message and the bug category tomorrow.

spotbugs/etc/messages.xml
```diff
@@ -9128,6 +9137,51 @@ Using floating-point variables should not be used as loop counters, as they are
       ]]>
         </Details>
   </BugPattern>
+
+  <BugPattern type="ENV_USE_PROPERTY_INSTEAD_OF_ENV">
+    <ShortDescription>It is preferable to use portable Java property instead of environment variable.</ShortDescription>
+    <LongDescription>It is preferable to use portable Java property '{3}' instead of environment variable '{2}' in method {1}.</LongDescription>
+    <Details>
+      <![CDATA[<p>
+        Environment variables are not portable, the variable itself (not only the value) may be different depending on the running OS.
```
gtoison, line 9146: Minor nitpicking that "the variable **name** iftself" might be better here, but I'm French so please feel to disregard if you disagree with my broken English :)

JuditKnoll: Thanks, this way it's more clear and unambiguous. English is also not my mother tongue 😄

gtoison (approved):

spotbugs/etc/messages.xml
```diff
@@ -9128,6 +9137,51 @@ Using floating-point variables should not be used as loop counters, as they are
...
+
+  <BugPattern type="ENV_USE_PROPERTY_INSTEAD_OF_ENV">
+    <ShortDescription>It is preferable to use portable Java property instead of environment variable.</ShortDescription>
+    <LongDescription>It is preferable to use portable Java property '{3}' instead of environment variable '{2}' in method {1}.</LongDescription>
+    <Details>
+      <![CDATA[<p>
+        Environment variables are not portable, the variable name itself (not only the value) may be different depending on the running OS.
+        Not only the names of the specific environment variables can differ (e.g. `USERNAME` in Windows and `USER` in Unix systems),
+        but even the semantics differ, e.g. the case sensitivity (Windows being case-insensitive and Unix case-sensitive).
+        Thus, the environment variables may not obey the general contract of the <code>Object.equals(java.lang.Object)</code> and <code>Object.hashCode()</code> methods.
```
iloveeclipse, line 9149: >  Thus, the environment variables may not obey the general contract of the <code>Object.equals(java.lang.Object)</code> and <code>Object.hashCode()</code> methods.

This sound wrong. The contract of mentioned methods is for sure working for String values read from environment variables.

spotbugs/etc/messages.xml
```diff
@@ -9128,6 +9137,51 @@ Using floating-point variables should not be used as loop counters, as they are
...
+        Consequently, environment variables may have unintended side effects.
+        Also, The visibility of environment variables are also less restricted compared to Java Properties: they are visible to all descendants
+        of the defining process, not just the immediate Java subprocess. Environment variables may be accidentally or maliciously altered during
```
iloveeclipse, line 9152: > Environment variables may be accidentally or maliciously altered during
        process execution using the <code>java.lang.ProcessBuilder</code> class.

Another piece of wrong claims. As I've mentioned before, if attacker or user can control command line creation via ProcessBuilder, he can change **everything**, not only variables. Please remove this false sentence.

iloveeclipse (changes requested):

spotbugs/etc/messages.xml
JuditKnoll: However, it sounds counterintuitive, I think it is right according to the following sources:

This part from `java.lang.System.getenv(String)`: 
> On UNIX systems the alphabetic case of name is typically significant, while on Microsoft Windows systems it is typically not. For example, the expression System.getenv("FOO").equals(System.getenv("foo")) is likely to be true on Microsoft Windows.

This part from `java.lang.System.getenv()`:
> The returned map and its collection views may not obey the general contract of the [Object.equals(java.lang.Object)](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#equals-java.lang.Object-) and [Object.hashCode()](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--) methods.

Do you think I misunderstood these parts from the Java API? Or was my wording misunderstandable?

spotbugs/etc/messages.xml
JuditKnoll: Thank you for clarifying this part. You are right, if someone has access to the ProcessBuilder or some similar construction, then everything can be changed.
The sentence itself is not entirely wrong, since environment variables can be changed, but it is misleading, indicating that the java properties can't be, so I removed it.

spotbugs/etc/messages.xml
gtoison: I seems to me that the potential issue about equals/hashcode is with the map itself, not with the keys and values.
So you might run into issues doing something like `someMap.put("env", System.getEnv())`.

The current message implies that the values (or keys) inside the map do not implement equals/hashcode according to the general contract. My understanding is that `someMap.put("env", System.getEnv("whatever"))` is fine

JuditKnoll: Thank you both for the feedback. I modified the message. Could you please look at it one more time?

iloveeclipse (approved):
I haven't checked what the code does, but the messages seem to be acceptable now.
Thanks for reworking that.

gtoison:
Was there an attempt to detect calls like `System.getenv().get(...)`?
Unless I misunderstood these wouldn't be detected, but I suppose that this is more complicated to catch.

JuditKnoll:
> Was there an attempt to detect calls like `System.getenv().get(...)`? Unless I misunderstood these wouldn't be detected, but I suppose that this is more complicated to catch.

Thanks, good catch. Earlier, it wouldn't be detected, but the newest commit solves this. Added some new tests as well.

gtoison:
Sorry to come back again with a question but I was wondering if it would detect this:

```
private static final String OS = "OS";

public String getOs() {
  return System.getenv(OS);
}
```

Since it is generally recommended to extract constants we can probably expect these keys to be static fields.

JuditKnoll:
> Sorry to come back again with a question but I was wondering if it would detect this:
> 
> ```
> private static final String OS = "OS";
> 
> public String getOs() {
>   return System.getenv(OS);
> }
> ```
> 
> Since it is generally recommended to extract constants we can probably expect these keys to be static fields.

Thank you for raising your concerns. It's not a problem, it's much better if the problems come to light during the review process, than later, after a release as an issue.
It works, added some tests to verify it.

spotbugs/etc/messages.xml
```diff
@@ -9128,6 +9137,51 @@ Using floating-point variables should not be used as loop counters, as they are
...
+    <LongDescription>It is preferable to use portable Java property '{3}' instead of environment variable '{2}' in method {1}.</LongDescription>
+    <Details>
+      <![CDATA[<p>
+        Environment variables are not portable, the variable name itself (not only the value) may be different depending on the running OS.
+        Not only the names of the specific environment variables can differ (e.g. `USERNAME` in Windows and `USER` in Unix systems),
+        but even the semantics differ, e.g. the case sensitivity (Windows being case-insensitive and Unix case-sensitive).
+        Moreover, the Map of the environment variables returned by <code>java.lang.System.getenv()</code> and its collection views may not obey
+        the general contract of the <code>Object.equals(java.lang.Object)</code> and <code>Object.hashCode()</code> methods.
+        Consequently, using environment variables may have unintended side effects.
+        Also, The visibility of environment variables are also less restricted compared to Java Properties: they are visible to all descendants
```
ThrawnCA, line 9152: "Also, The" -> "Also, the"

"are also less restricted" -> "is less restricted"

ThrawnCA (changes requested):

spotbugs/etc/messages.xml
JuditKnoll: Thanks. Corrected them.

ThrawnCA (approved):

JuditKnoll:
If there is no more change requests to this PR, and no objection against it, I will merge this PR on next Monday.

PR #2684, opened 2023-11-07
Use java.nio to load filter files

gtoison:
`java.io.FileInputStream` seems to be running into issues when trying to load files with non-ascii charatecter in the file name.
Other similar issues point to this happening on a Mac, so it might be dependent on the combination of OS/JDK version/user settings.
Although the test did not reproduce the problem I think that using `java.nio` should help.
Note that I'm using `FileSystems.getDefault().getPath(...)` instead of `Path.of(...)` because it is only available since JDK 11.

Originally reported here:
https://github.com/JetBrains/spotbugs-intellij-plugin/issues/1492

Other links discussing similar problems:
https://stackoverflow.com/questions/22775758/java-io-file-accessing-files-with-invalid-filename-encodings
https://ogris.de/howtos/java-utf8-filenames.html
https://stackoverflow.com/questions/14171565/java-read-write-unicode-utf-8-filenames-not-contents

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 - Fixed false positive UPM_UNCALLED_PRIVATE_METHOD for method used in JUnit's MethodSource ([[#2379](https://github.com/spotbugs/spotbugs/issues/2379)])
+- Use java.nio to load filter files ([[#2684](https://github.com/spotbugs/spotbugs/pull/2684)])
 
 - tbd
 
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/Utf8FilterFileNameTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/Utf8FilterFileNameTest.java
@@ -1,0 +1,52 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2023, the SpotBugs authors
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Fixed
 - Fixed false positive UPM_UNCALLED_PRIVATE_METHOD for method used in JUnit's MethodSource ([[#2379](https://github.com/spotbugs/spotbugs/issues/2379)])
+- Use java.nio to load filter files ([[#2684](https://github.com/spotbugs/spotbugs/pull/2684)])
```
JuditKnoll, line 12: Are you sure, you wanted to refer the PR here and not the fixed issue?

JuditKnoll (commented):
Tried to reproduce the problem with the provided test without the solution both on a Windows and on a Linux machine (using OpenJDK 17 on both), but I couldn't, the test was successful before using `java.nio`, and after as well.

If using `java.nio` solves the problem, it may be worth to look at and consider using `java.nio` in other places operating with Files as well to prevent similar issues.

CHANGELOG.md
gtoison: The issue in the IntelliJ plugin, since Jetbrains does not seem to be actively maintaining it I think it will take a while until the plugin gets the updated version

JuditKnoll: In this exact instance, I get it. However, I couldn't find the general logic yet, when exactly is referenced the issue, and when the PR, when there are no links? The links seems the same, when looking at the preview of the file (not the raw version), which may be a bit confusing for the reader. Until this point, I only refered the fixed issues. Could you please clarify this to me? (It's a bit off topic here, sorry about it.)

gtoison: I don't know either to be honest, but as a user I would expect a link to something giving some context as to what was changed and why

ThrawnCA (approved):

JuditKnoll (approved):

lastarc:
I could reproduce this issue by building spotbugs from current master branch (da3610b) in fresh ubuntu:jammy-20240911.1 docker image:

```sh
docker run -it --rm ubuntu:jammy-20240911.1
apt update
apt install -y git curl openjdk-17-jdk
git clone --depth 1 https://github.com/spotbugs/spotbugs.git
cd spotbugs
./gradlew build
```

I narrowed the cause down to the `LANG` environment variable, absence of which causes this test case (Utf8FilterFileNameTest.java) to fail. 

Reproduction with github actions:

yaml file: https://github.com/lastarc/spotbugs-utf8-issue-reproduction/blob/main/.github/workflows/build.yml

one of the failing runs on ubuntu-24.04 with jdk 17.0.15-tem: https://github.com/lastarc/spotbugs-utf8-issue-reproduction/actions/runs/15413291985/job/43370101296

PR #2689, opened 2023-11-08
Fix the weird messages in DontReusePublicIdentifiers

### LINKED ISSUE #2646: PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES  message missing substitutions
I'm seeing messages like the following from Spotbugs 4.8.0:

```M B PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES PI: Class name ?>?1/1??? in source file ?>?2/1??? shadows the publicly available identifier from the Java Standard Library.  In Clock.java```

Presumably this is meant to print an actual class name and file path rather than `?>?1/1???` and `?>?2/1???`

JuditKnoll:
This PR fixes #2646, the problem with the messages in the PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS bugs.
With this change, the annotations of the BugInstances in the [DontReusePublicIdentifiers](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DontReusePublicIdentifiers.java) detector match the messages.
Tried it out locally by printing out the messages, and all of them looked okay. I haven't committed this, since IMO it wouldn't be nice to check the hardcoded exact messages in the tests. If anyone has a good idea, how this could be properly tested, I'm open to any idea.
Also, I found `PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_INNER_CLASS_NAMES` bugpattern in messages.xml, which seems to be only existing in a stage of the development of https://github.com/spotbugs/spotbugs/pull/2511. (The other references to it are already deleted, both in the code and in findbugs.xml.)

Co-authored-by: @norbo03 

----


ThrawnCA (approved):

hazendaz:
@JuditKnoll Looks ok to me, are you saying you want to hold off on this for now?  Maybe place in draft or was this one ok to merge?  Seems we are going to be releasing again very shortly :(

gtoison:
I think the missing piece is finding a better way to catch the case when some annotations are missing. But detectors create them programmatically in non always trivial ways, so it is not easily checked. 
In any case detecting such errors would be better suited for another PR, that one fixes an outstanding error

gtoison (approved):

norbo03:
@JuditKnoll The modifications look great to me.

JuditKnoll:
I think it's okay to merge, since it solves the underlying problem. I was just wondering how it could be properly tested and showed that it works. If you think, there is no need to add test to this PR, then it can be merged.

JuditKnoll:
> I think the missing piece is finding a better way to catch the case when some annotations are missing. But detectors create them programmatically in non always trivial ways, so it is not easily checked. In any case detecting such errors would be better suited for another PR, that one fixes an outstanding error

Created an issue from this comment, so it's doesn't get forgotten so easily.

@hazendaz You assigned this issue to yourself. Does it mean that you would like to merge it? I think it can be merged, if nobody finds any problem with it. I think it would be great, if this PR could be merged, before releasing 4.8.2.

hazendaz:
@JuditKnoll Sorry about that, I was probably thinking it would be merged around time I added myself there and simply didn't remove myself after it appeared it might sit a bit longer.  If you see that elsewhere, don't think that means I mean to merge.  Feel free to do so where needed.

JuditKnoll:
> @JuditKnoll Sorry about that, I was probably thinking it would be merged around time I added myself there and simply didn't remove myself after it appeared it might sit a bit longer. If you see that elsewhere, don't think that means I mean to merge. Feel free to do so where needed.

No problem. Thanks. Next time I will do so 😄

Issue #2701, opened 2023-11-15
FindOverridableMethodCall detector is not thread safe


iloveeclipse:
Spotbugs analysis in Eclipse can be executed on multiple projects in parallel.
In order to achieve that, detectors should not assume they run "in isolation" with a single analysis in the entire JVM.
Very basic point is not to use static storage for analysis specific data.
If they don't do that, we have errors like below:

```
eclipse.buildId=4.30.0.I20231114-1800
java.version=21-internal
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data /data/4x_platform_workspace -os linux -ws gtk -arch x86_64

com.github.spotbugs.plugin.eclipse
Error
Wed Nov 15 09:08:35 CET 2023
Exception analyzing org.eclipse.jdt.core.tests.model.IndexManagerTests using detector edu.umd.cs.findbugs.detect.FindOverridableMethodCall

java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1095)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1049)
	at edu.umd.cs.findbugs.detect.FindOverridableMethodCall.visitBootstrapMethods(FindOverridableMethodCall.java:121)
	at org.apache.bcel.classfile.BootstrapMethods.accept(BootstrapMethods.java:82)
	at edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitJavaClass(PreorderVisitor.java:408)
	at org.apache.bcel.classfile.JavaClass.accept(JavaClass.java:244)
	at edu.umd.cs.findbugs.BytecodeScanningDetector.visitClassContext(BytecodeScanningDetector.java:38)
	at edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
	at edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$1(FindBugs2.java:1108)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at edu.umd.cs.findbugs.CurrentThreadExecutorService.execute(CurrentThreadExecutorService.java:86)
	at java.base/java.util.concurrent.AbstractExecutorService.invokeAll(AbstractExecutorService.java:247)
[...]

PR #2702, opened 2023-11-15
FindOverridableMethodCall is not thread safe

iloveeclipse:
Don't use static maps for analysis specific data, it can't work with multiple analysis running in same JVM, like it might happen in Eclipse.

Fixes https://github.com/spotbugs/spotbugs/issues/2701


gtoison (approved):
Good find, too bad there wasn't a detector to find that issue :)

Issue #651, opened 2018-06-05
findbugs:NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE does not consider null validation


alixwar:
**Problem**: A violation is triggered at the last line, saying that args may be null. This is clearly validated the line before (using org.apache.commons.lang3.Validate)

**Affected version**: 3.1.2 (SpotBugs)
**Seen in**: SonarQube (FindBugs plugin 3.7)

Reproducible code example (from an Android app):

```java
@Override
public void onCreate(@Nullable Bundle savedInstanceState) {
   super.onCreate(savedInstanceState);
   final Bundle args = getArguments();
   Validate.notNull(args);
   email = args.getString(ARG_EMAIL);
}
```

npfistner:
**Problem:** Also if validated using java assert spotbugs still complains NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE

**Affected version:** 3.1.5 (SpotBugs)

```java
        assert entity.getId() != null;
        final long id = entity.getId().longValue();
```

ThrawnCA:
I'm not sure that the `assert` keyword should be relied on, since assertions can be disabled. But the Commons Lang one is likely worth including; thanks for pointing it out.

alixwar:
Same issue with `java.util.Objects.requireNonNull()`

nmatt:
+1 for supporting `requireNonNull()`.

gmbett:
Same issue with custom method that is annotated with @Contract("null -> false")

alixwar:
I'm trying to add a reproducer in the SpotBugs project but I keep getting: 

```
The following classes needed for analysis were missing:
  org.apache.commons.lang.Validate
```

even after adding the dependency to the spotbugs-tests module 😕

gmbett:
Any updates on this issue?

victornoel:
@ThrawnCA any news on that?

C-Otto:
There's a database of hardcoded method parameter annotations in https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/DefaultNullnessAnnotations.java#L406

At the given line, the first argument of `Preconditions.checkNotNull` is assumed to be "non-null". As such, the analysis throws an exception of `null` is passed to `checkNotNull`.

The same holds true for other implementations, for example `Objects.requireNonNull`.

marcopelegrini:
[...]

Issue #456, opened 2017-10-17
Possible Nullpointer dereference - False positive


phoenix384:
`File[] content = new File(".").listFiles();`
`Objects.requireNonNull(content);`

results in "Possible Nullpointer dereference due to return value of called method" for the second line.

iloveeclipse:
Why do you need `Objects.requireNonNull()`? If the array is null, next access to it will throw NPE, same as Objects thing - but probably you simply want to do something with the array only if it is non null?

But yea, we could consider that method in the NPE analysis similar to an "assert".

phoenix384:
Because with Object.requireNonNull I can add a specific error message like
`Objects.requireNonNull(content, "Unable to read directory");`
That is better or more helpful than just the normal NPE.

Just like
`Validate.notNull(content, "...");`
from Apache Commons Lang3 which is NOT marked by SpotBugs, although it's doing the same thing.

Maaartinus:
TL;DR Using `Objects.requireNonNull` means that the NPE is the wanted outcome and should produce no warning.

> But yea, we could consider that method in the NPE analysis similar to an "assert".

AFAICT it's not similar.

When I do `Objects.requireNonNull` or `Preconditions.checkNotNull`, then I'm well aware of the possibility of NPE and definitely want no warning for this line. *Later, the object is known not to be null.*

After an `assert`, *the object is *not* known not to be null.* When I do

	assert started != null;
	if (started == null) ...

then I hope that `started` isn't `null`, but I'm not perfectly sure and the conditional is needed to deal with it (and should produce no warning like in findbugsproject/findbugs/issues/166).

xenoterracide:
> TL;DR Using Objects.requireNonNull means that the NPE is an expected outcome and should produce no warning.

plus one, we have this problem in findbugs, and it's obnoxious. If I explicitly throw an NPE, whether it's by `if ( i == null ) { throw new NullPointerException() }` or use `Objects.requireNonNull` it should be the same thing. Often I use requireNonNull so it doesn't fail farther down in a method, and usually it's to catch programming errors, vs actual runtime errors.

YeisonMelo:
you can use `if (method == null) throw new AssertionError("error message)` also

xpyct707:
There is a quote from Bloch's 'Effective Java' (3rd Edition) - Item 49: Check parameters for validity:
[...]

Issue #438, opened 2017-10-07
BuildCheckReturnAnnotationDatabase should visit classes in AUX_CLASSPATH


KengoTODA:
This issue was found by #429.

Currently `BuildCheckReturnAnnotationDatabase` does not visit classes in AUX_CLASSPATH. It means that features like `CheckReturnAnnotationDatabase` cannot analyze annotations on class/package in AUX_CLASSPATH.

PR #2709, opened 2023-11-24
Fix NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE does not consider null val…

mdouaihy:
This fixes the issue reported in 

- https://github.com/spotbugs/spotbugs/issues/651
- https://github.com/spotbugs/spotbugs/issues/456

NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE is raised when using Objects.requireNonNull whereas it's not raised when using Guava Preconditions.checkNotNull.

----


hazendaz (approved):

ThrawnCA (approved):

gtoison:
Thanks a lot for the contribution @mdouaihy 
#651 mentionned other methods than `Objects.requireNonNull`, did you consider adding things like:
```
com.google.common.base.Verify.verifyNotNull
com.google.common.base.Preconditions.checkNotNull
org.apache.commons.lang.Validate.notNull
```

I think it would be good adding unit tests for these too

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/IssueRequireNotNullTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unit test to reproduce <a href="https://github.com/spotbugs/spotbugs/issues/456">#259</a>.
+ */
+@ExtendWith(SpotBugsExtension.class)
+class IssueRequireNotNullTest {
+
+    @Test
+    void testIssue(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/nullnessAnnotations/TestNonNull7.class"));
+        assertThat(bugCollection, emptyIterable());
```
JuditKnoll, line 26: It's better to add a `BugInstanceMatcher` matching to the bugtype when creating negative tests, so later if a new bug will be added, which finds a hit in the testcase, then this test won't fail. See https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2147Test.java as an example.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -1386,7 +1386,7 @@ boolean callToAssertionMethod(Location loc) {
             if (i instanceof InvokeInstruction) {
                 InvokeInstruction ii = (InvokeInstruction) i;
                 String name = ii.getMethodName(classContext.getConstantPoolGen());
-                if (name.startsWith("check") || name.startsWith("assert")) {
+                if (name.equals("requireNonNull") || name.startsWith("check") || name.startsWith("assert")) {
```
JuditKnoll, line 1389: Can you please check class of the called function as well?

spotbugsTestCases/src/java/nullnessAnnotations/TestNonNull7.java
```diff
@@ -0,0 +1,16 @@
+package nullnessAnnotations;
+
+import java.io.File;
+import java.util.Objects;
+
+
```
JuditKnoll, line 6: It's a good practice to refer the issues at the testcases, if there are any.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Eclipse: Do not export javax.annotation packages ([[#2699](https://github.com/spotbugs/spotbugs/pull/2699)])
 - Fixed not thread safe FindOverridableMethodCall detector ([[#2701](https://github.com/spotbugs/spotbugs/issues/2701)])
 - Fix the weird messages of PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS bugs. ([#2646]((https://github.com/spotbugs/spotbugs/issues/2646)]))
+- Fix false positive NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE even if Objects.requireNonNull is used. ([[#651](https://github.com/spotbugs/spotbugs/issues/651)], [[#456](https://github.com/spotbugs/spotbugs/issues/456)])
```
JuditKnoll, line 16: Can you please remove the unnecessary brackets from the links. Now, it looks like this: ([[#651](https://github.com/spotbugs/spotbugs/issues/651)], [[#456](https://github.com/spotbugs/spotbugs/issues/456)])

JuditKnoll (changes requested):

CHANGELOG.md
hazendaz: oh I see, was a problem but @iloveeclipse fixed it already.  We just need this one fixed still.

hazendaz:
Leaving this one out of 4.8.2 release as a few other items need looked at here still.

robinjhector:
Hi, this would be great to get merged and released! Any chance you can take a look at the requested changes @mdouaihy ?

mdouaihy:
hi @robinjhector, 

I am looking into applying the requested changes. I am just having a ClassNotFound when adding the tests to cover the changes and I am trying to figure out why.

JuditKnoll:
> hi @robinjhector,
> 
> I am looking into applying the requested changes. I am just having a ClassNotFound when adding the tests to cover the changes and I am trying to figure out why.

In this case, it can be a good idea to convert the PR to draft.
Is the class file for not found class added to the `performAnalysis()` function inside the testclass? If yes and you are stuck, feel free to commit your changes, and I can look into it.

hazendaz:
@mdouaihy Can you look at last concern here?   Would like to include this in next release in 2 weeks.

mdouaihy:
@hazendaz, due to some personal problems, I was not able to fix the issues before. I'll fix that during the coming couple of days.

hazendaz:
@JuditKnoll Any chance you could pick this up and do the necessary items you were wanting adjusted?  Seems @mdouaihy Has some personal issues to deal with so not available to make this go.  If its not too difficult, we could get this included in upcoming release hopefully this weekend :)

JuditKnoll:
> Thanks a lot for the contribution @mdouaihy #651 mentionned other methods than `Objects.requireNonNull`, did you consider adding things like:
> 
> ```
> com.google.common.base.Verify.verifyNotNull
> com.google.common.base.Preconditions.checkNotNull
> org.apache.commons.lang.Validate.notNull
> ```
> 
> I think it would be good adding unit tests for these too



> Thanks a lot for the contribution @mdouaihy #651 mentionned other methods than `Objects.requireNonNull`, did you consider adding things like:
> 
> ```
> com.google.common.base.Verify.verifyNotNull
> com.google.common.base.Preconditions.checkNotNull
> org.apache.commons.lang.Validate.notNull
> ```
> 
> I think it would be good adding unit tests for these too

There is already a unit test for `com.google.common.base.Preconditions.checkNotNull` with [RV_RETURN_VALUE_IGNORED_Guava_Preconditions](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugPatterns/RV_RETURN_VALUE_IGNORED_Guava_Preconditions.java) as sample class tested in [PreconditionsCheckNotNullCanIgnoreReturnValueTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/PreconditionsCheckNotNullCanIgnoreReturnValueTest.java), but when running it, the `com.google.common.base.Preconditions` class needed for analysis was missing, and the test does not fail because of the missing class.

gtoison:
@JuditKnoll would you know if the missing class error is because of the way we are running unit tests?
I was wondering if it could be a problem similar to https://github.com/spotbugs/spotbugs/issues/438

JuditKnoll:
> @JuditKnoll would you know if the missing class error is because of the way we are running unit tests? I was wondering if it could be a problem similar to #438

Yes, it seems really similar.
The `AbstractIntegrationTest.performAnalysis()` adds the auxClassPathEntries from the `spotbugsTestCases/lib` directory, adding the necessary jars here, removes these messages. I think we should add the aux classpathes listed in the `spotbugsTestCases/build/spotbugs/auxclasspath/spotbugsMain` file, or if you have a better idea, I would love to hear it, but this should be in a separate PR.

P.S. I'm still looking into whether `com.google.common.base.Verify.verifyNotNull()`, `org.apache.commons.lang3.Validate.notNull()` are already handled, or just not found because of some connecting problem in the tests.

JuditKnoll:
I addressed the issues mentioned in the reviews. However, when running the test separately from the IDE, the following is written to the std err*:
```
  com.google.common.base.Verify
  org.apache.commons.lang3.Validate
  com.google.common.base.Preconditions
```
Locally I added these to the `spotbugsTestcases/lib`, which does solve this issue, and the tests in `IssueRequireNotNullTest` are still successful, but some other tests fail, which also fail on master if the necessary jars are added.
It looks like the functions are handled by the `IsNullValueDataflow`, and not by `callToAssertionMethod()` function.


*When running from terminal using gradle, these missing classes messages does not appear in the output, the build is successful. IMO if there are missing classes, the tests should fail.

JuditKnoll (approved):

gtoison (approved):
This LGTM, thank you!

ThrawnCA (approved):

hazendaz:
fixes #651 and fixes #456

PR #2721, opened 2023-11-29
Fix doc on gradle plugin version and run on gradle 8.5 rc4

hazendaz:


hazendaz:
Expecting to finish up the jdk 11, 17, 21, 22-ea builds attempt and will add change log there to note it but not really worth nothing our build wrapper differs in general.  History of change log doesn't show that mentioned very often much like library updates.

hazendaz:
> Why run on an RC version?

jdk 21 kotlin support.

hazendaz:
Will be following it up in a few with separate set to start building with jdk 17 and 21.  Longer plan is to show it building with more supported versions but having issues with eclipse.  So going to split the effort there for what works.

hazendaz:
note on gradle: gradle 8.4 supports jdk 21 but kotlin did not.  Kotlin only recently released support and that is a big part of what 8.5 is.

PR #2722, opened 2023-11-29
Run build on both jdk 17 and 21

hazendaz:


hazendaz:
Part of this comes from #2604.  This abbreviated cut attempts to only use jdk 17 and 21 to do the build without using toolchain which would just drop back to java 17.  It further shows that gradle now supports jdk 21 fully.

PR #2747, opened 2023-12-06
Fix FNs in CT_CONSTRUCTOR_THROW

JuditKnoll:
Run into false negatives in CT_CONSTRUCTOR_THROW, when trying to solve another issue. The method signature was inconsistently stored, in one place as dotted, at another place as slashed. This caused that if the exception was thrown from or through a method, the return type of which is not void or primitive type, then the problem is not found.
This PR solves this.

----


JuditKnoll:
In https://github.com/spotbugs/spotbugs/pull/2732 the [spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest18.java](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest18.java) got added, but the test running on this got forgotten, that's why it's added in the ConstructorThrowTest file.

hazendaz (approved):

Issue #2695, opened 2023-11-10
CT_CONSTRUCTOR_THROW: false positives constructing lambdas


pburka:
In spotbugs 4.8.1 and Java 17.0.6, the `CT_CONSTRUCTOR_THROW` bug is incorrectly detected if the constructor creates a lambda which could throw an exception if used, even if the lambda is not used in the constructor.

Test case:

```
// CT_CONSTRUCTOR_THROW test case

import java.util.function.*;

public class Foo {
    Supplier<String> s;

    public Foo() {
        s = Foo::supplier;
    }

    private static String supplier() {
        //return "Foo";
        throw new IllegalStateException();
    }
}
```

This reports:

```
M B CT_CONSTRUCTOR_THROW CT: Exception thrown in class Foo at new Foo() will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks.  At Foo.java:[line 9]
M P URF_UNREAD_FIELD UrF: Unread field: Foo.s  At Foo.java:[line 9]
```

The unread field detection is correct; the constructor throws detection is not.

If we change the test case by commenting out the `throw` and uncommenting the `return`, the `CT_CONSTRUCTOR_THROW` bug is no longer reported.

big-andy-coates:
Given that finalizer methods are pretty much a dead pattern in Java, is there not an argument that `CT_CONSTRUCTOR_THROW` should be off by default?

JuditKnoll:
@pburka Thank you for raising this FP. I will look into this issue and propose a solution soon.

@big-andy-coates This detector draws attention to a vulnerability to finalizer attacks, which can be solved by several ways, not just by defining a finalizer method, but also declaring the class final, defining the constructor as private, removing the exception throw from the constructor (or from the called method), or wrapping the exception throw and logging the exception properly inside the constructor. The vulnerability still exists and can be exploited. So IMO just because one of the ways it can be solved is a dead pattern, it doesn't means that the problem doesn't exist.

big-andy-coates:
It seems adding a final, empty finalise method avoids the fault.

I wonder if a better implementation would be to find classes with non-empty finaliser and only warn of any of its constructors throw?

This would avoid a lot of noise and avoid missing issues due to people disabling a noisy check.

JuditKnoll:
> It seems adding a final, empty finalise method avoids the fault.
> 
> I wonder if a better implementation would be to find classes with non-empty finaliser and only warn of any of its constructors throw?
[...]

PR #2748, opened 2023-12-06
Fix FP in CT_CONSTRUCTOR_THROW with lambdas

JuditKnoll:
This is a dirty fix for https://github.com/spotbugs/spotbugs/issues/2695.
This PR fixes https://github.com/spotbugs/spotbugs/issues/2695 by deleting the lambda handling parts from the ConstructorThrow detector, so it causes false negatives, since from now, if an exception is thrown through an invoke dynamic (e.g. a lamdba), it won't be able to find it. I think the false positives in the original version can be more annoying, than the false negatives this PR introduces.

I plan to propose a proper solution later, but I didn't want to wait longer for proposing a solution to this issue, so the users won't disable the rule because of its noisiness.

----


ThrawnCA (approved):

hazendaz:
I'm uneasy with concept of a dirty fix but given context it seems better than current scenario.

PR #2755, opened 2023-12-07
Use String.isEmpty() with null guards

rovarga:
There are a few places which check for null-or-empty String, where
the empty part is performed with "".equals().

Rather than loading an explicit literal, use isEmpty() on the string
itself, as we know it is not null.

----


ThrawnCA (approved):
This is fine, but I'm pretty sure an `equals` check is fairly efficient when the object it's called on is an empty string.

We could also consider using Apache Commons StringUtils.isBlank for this.

hazendaz:
@ThrawnCA Actually this is faster than using string utils.  And this is faster than the check it had.  See internals of isEmpty() below.  Its just checking the length.  Other tools are already flagging the same code (believe sonar at least).

```
public boolean isEmpty() {    
        return value.length == 0;    
    }   
```

PR #2602, opened 2023-10-05
[draft] Renovate bcel 6.7.0 + custom changes

hazendaz:
Before opening a 'pull request'

* Search existing issues and pull requests to see if the issue was already discussed.
* Check our discussions to see if the issue was already discussed.
* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin) *

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [ ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #2757, opened 2023-12-08
Fixes for the bcel update

JuditKnoll:
This PR applies the custom changes from https://github.com/spotbugs/spotbugs/pull/2602, and migrates the tests to be compatible with the current master with the junit migration.
The base for this PR is not the master branch, but renovate/org.apache.bcel-bcel-6.x, which will be merged to master by https://github.com/spotbugs/spotbugs/pull/2756.

----

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Bug3107916Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Bug3107916Test.java
@@ -1,0 +1,56 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+public class Bug3107916Test extends AbstractIntegrationTest {
+    @Test
+    public void test() {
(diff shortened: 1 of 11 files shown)


hazendaz (approved):

hazendaz:
@JuditKnoll Going to go ahead and merge so I can get bcel update to run again and apply on top.  We should have a change log noted.  Can you take care of getting change log over for this after?

hazendaz:
@JuditKnoll Lets not do this to add onto renovate.  It will catch us every time.  If fixing renovate item, just new PR to master with the renovate changes is good enough.  Otherwise its a deviation of normal pattern and not typical Renovate usage IMO so like before I clicked the button on Renovate before this showed up so it wiped the change again.  I'll fix the issue but had noted then to just stay away from applying on Renovate for that reason :)  So my fault but think its best we don't tack on items to renovate since that button click is far too easy and there is no recourse after clicked.

hazendaz:
argghhh.  to fix taking same path, just easier.  I added your remote and just repulled your branch then merged in upstream renovate, will quickly merge it.  I'll add change log to that as well now so no worries there.

JuditKnoll:
> @JuditKnoll Going to go ahead and merge so I can get bcel update to run again and apply on top. We should have a change log noted. Can you take care of getting change log over for this after?

I see that, meanwhile you updated the changelog as well, so I won't add another entry. Thanks 😄 

> @JuditKnoll Lets not do this to add onto renovate. It will catch us every time. If fixing renovate item, just new PR to master with the renovate changes is good enough. Otherwise its a deviation of normal pattern and not typical Renovate usage IMO so like before I clicked the button on Renovate before this showed up so it wiped the change again. I'll fix the issue but had noted then to just stay away from applying on Renovate for that reason :) So my fault but think its best we don't tack on items to renovate since that button click is far too easy and there is no recourse after clicked.

I'm sorry for the problems. Next time I will create a new PR directly to master with the renovate changes as well.

PR #2777, opened 2023-12-18
Update the changelog automatically when releasing

gtoison:
Here's an untested proposal to automate the update of the changelog when releasing
This should hopefully put an end to this endless source of merge conflicts

The process is using https://github.com/stefanzweifel/git-auto-commit-action and would entail that commits follow [Conventionnal commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)


hazendaz (approved):

hazendaz:
Lets see if it does the trick!

JuditKnoll:
It looks like this one will need a bit more attention. Looking at the release run on the current master (https://github.com/spotbugs/spotbugs/actions/runs/7492405140/job/20395818143), the relevant steps doesn't run properly, the `previousTag` seems to be empty. (The `Commit CHANGELOG.md` is skipped intentionally because of https://github.com/spotbugs/spotbugs/pull/2806.

Issue #2793, opened 2024-01-03
SE_NO_SERIALVERSIONID with Serializable records


rovarga:
Having a Serializable record without a serialVersionUID constant triggers SE_NO_SERIALVERSIONID.

This is essentially a false positive, as records [are special to serialization](https://docs.oracle.com/en/java/javase/16/serializable-records/index.html) -- SpotBugs should not report this warning for records.

hazendaz:
I guess in theory true its not an error, but it could be a proper warning.  It clearly states in there that sometimes you do need to do so in certain situations or that is how I read it.  Maybe it needs its own classification though since it sets to 0L if not listed.

gtoison:
My understanding is that `serialVersionUID` is not considered when loading a record:
_the requirement for matching the serialVersionUID value is waived for record classes_

There's already a special case to ignore enums, I think it should be the same records
https://github.com/spotbugs/spotbugs/blob/3da060edfcc43017623c9922bdb2563a98591b51/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SerializableIdiom.java#L172-L175

It also seems to report `SE_NO_SUITABLE_CONSTRUCTOR_FOR_EXTERNALIZATION` which is not great because records are not externalizable anyway (so there might be another bug to report: a record should not implement `Externalizable`)

hazendaz:
Fixed via #2795

PR #2795, opened 2024-01-05
Issue #2793

gtoison:
`SerializableIdiom` reports that a `serialVersionUid` is needed for a record class.
It also reports that `SE_NO_SUITABLE_CONSTRUCTOR_FOR_EXTERNALIZATION` for an externalizable record even if records serialization does not work with externalizable


hazendaz (approved):

hazendaz:
@gtoison Can you add change log?

gtoison:
Done here #2799

PR #2796, opened 2024-01-05
ci: tell sonar where the jacoco root report is

### LINKED ISSUE #2619: Build: Fix jacoco reporting to sonar
Originally discussed over at #2605

Currently sonar does not show any code coverage, we should fix that situation.  This may be related to fact that tests are separate from the core and jacoco may need some help in passing that data along.

gtoison:
The aggregated jacocoRootReport is not in one of the standard locations searched by sonar, add a property with the report file path so sonar can reflect the coverage

This should fix #2619


hazendaz (approved):

hazendaz:
@gtoison Confirmed fixed!  https://sonarcloud.io/component_measures?metric=coverage&id=com.github.spotbugs.spotbugs.  Shows us at 64% now rather than 0%.

PR #2798, opened 2024-01-06
[checkstyle] Bump to 10.12.7, add spring dependency management to fix gradle exclusion defects

hazendaz:
As documented with spring here https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/211, there is a long standing defect in exclusion handling within gradle.  With more recent checkstyle, we cannot apply due to exclusion that is not being adhered to.  

See https://github.com/checkstyle/checkstyle/issues/14211 for checkstyle info on situation.

See https://github.com/spotbugs/spotbugs/pull/2752 for our failures trying to pick up latest with the exclusion.

### DIFF
--- a/gradle/checkstyle.gradle
+++ b/gradle/checkstyle.gradle
@@ -1,8 +1,15 @@
 // Setup checkstyle
 apply plugin: 'checkstyle'
+apply plugin: 'io.spring.dependency-management'
+
+dependencyManagement {
+    dependencies {
+        dependency 'com.puppycrawl.tools:checkstyle:10.12.7'
+    }
+}
 
 checkstyle {
-  toolVersion '10.12.5'
+  toolVersion '10.12.7'
   ignoreFailures false
   configFile file("$rootDir/spotbugs/etc/checkstyle.xml") // TODO : This config file is lame and should be moved out...
 }
--- a/gradle/checkstyle.gradle
+++ b/gradle/checkstyle.gradle
@@ -1,11 +1,10 @@
 // Setup checkstyle
 apply plugin: 'checkstyle'
-apply plugin: 'io.spring.dependency-management'
 
-dependencyManagement {
-    dependencies {
-        dependency 'com.puppycrawl.tools:checkstyle:10.12.7'
-    }
+configurations.checkstyle {
+  resolutionStrategy.capabilitiesResolution.withCapability("com.google.collections:google-collections") {
+    select("com.google.guava:guava:0")
+  }
 }
 
 checkstyle {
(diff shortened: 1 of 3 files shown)


hazendaz:
note: To not hammer system it redoing build, I've fixed the changelog PR in separate effort with some other change log entries that I'll push separately.

gradle/checkstyle.gradle
```diff
@@ -1,8 +1,15 @@
...
+
+dependencyManagement {
+    dependencies {
+        dependency 'com.puppycrawl.tools:checkstyle:10.12.7'
+    }
+}
 
 checkstyle {
-  toolVersion '10.12.5'
+  toolVersion '10.12.7'
```
gtoison, line 12: I tried removing that line and it seems to work, so it seems to be redundant now

hazendaz: Gradle would be using older version then tied to gradle since its documented in their docs.  Checkstyle change at 10.12.6 caused failures for any using gradle with latest checkstyle.  Checkstyle 10.12.7 was supposed to fix which does for typical maven usage but gradle failed to support properly so in theory I suspect you would see this again with gradle 8.6 as they do not handle exclusions properly as well documented with spring.

hazendaz: or are you saying you remove it from current master - will fail and using this PR with it removed its then ok?  Wondering if it just pulls latest but not clear on your comment.  I looked around and cannot find any documentation that states one way or the other.  But if that is the case here, I think we still keep it just in case given it did cause a failure just so we are mindful of that.  If for example, we removed the version on master and it were to fail the repo would be broken whereas having version its not.

gtoison: I meant to say that it seemed redundant to have the checkstyle version twice in the same file.
But you're right: it doesn't use the 10.12.7 version if I remove that line, it uses 9.3

At the same time I wonder if we really need to force the version. Using the built-in version of checkstyle wouldn't have caused the issue in the first place

hazendaz: I think its better we use the latest given checkstyle is a small team and a drop back to 9.3 would put us back to January of 2022.

There may be some follow up need here that we need to confirm.  I think spring plugin here writes the pom file with dep management and there is a flag to turn that off.  I have not confirmed that aspect yet.  I think we want this to only impact us and not influence spotbugs itself needlessly which could then impact downstream usage with gradle.  I'll look into that before we move forwards here and will just put this into DRAFT for time being.

hazendaz:
Once build finishes will merge, per previous discussions then help from checkstyle comments on better way to do this.  This way ensures we don't mess up our dependency management and cause users other unexpected side effects.  Fixed up change log as well as two build areas in this release.

Issue #2759, opened 2023-12-09
NullPointerException in FieldItemSummary 


romani:
detected during version bump spotbugs-maven-plugin from 4.7.3.5 to 4.8.2.0
https://github.com/checkstyle/checkstyle/pull/14117

https://app.circleci.com/pipelines/github/checkstyle/checkstyle/22221/workflows/e397e60b-04ca-4618-b94e-5ca0be2a20d8/jobs/446175

OS:
```
 Kernel Version: 5.15.0-1039-aws
 Operating System: Ubuntu 20.04.6 LTS
 OSType: linux
 Architecture: x86_64
``` 

Special docker [container](https://github.com/checkstyle/contribution/blob/master/devops/docker/no-exception-build-image/jdk-11-groovy-git-mvn/Dockerfile):
```
Starting container checkstyle/jdk-11-groovy-git-mvn:11.0.20.1__2.4.21__2.42.0__3.9.5
this means
jdk11.0.20.1
mvn 3.9.5
```

Execution log:
```
.....
[[1;34mINFO[m] [1m--- [0;32mspotbugs:4.8.2.0:spotbugs[m [1m(spotbugs)[m @ [36mcheckstyle[0;1m ---[m
[[1;34mINFO[m] Fork Value is true
     [java] The following errors occurred during analysis:
     [java]   Exception analyzing com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParser$ClassBlockContext using detector edu.umd.cs.findbugs.detect.FieldItemSummary
     [java]     java.lang.NullPointerException
     [java]       At edu.umd.cs.findbugs.visitclass.DismantleBytecode.visit(DismantleBytecode.java:719)
     [java]       At edu.umd.cs.findbugs.detect.FieldItemSummary.visit(FieldItemSummary.java:131)
     [java]       At edu.umd.cs.findbugs.visitclass.BetterVisitor.visitCode(BetterVisitor.java:218)
     [java]       At edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitCode(PreorderVisitor.java:250)
     [java]       At edu.umd.cs.findbugs.bcel.OpcodeStackDetector.visitCode(OpcodeStackDetector.java:65)
     [java]       At org.apache.bcel.classfile.Code.accept(Code.java:121)
     [java]       At edu.umd.cs.findbugs.visitclass.PreorderVisitor.doVisitMethod(PreorderVisitor.java:322)
[...]

PR #2807, opened 2024-01-08
Add support for CONSTANT_Dynamic

gtoison:
Class files modified by jacoco cause the spotbugs analysis to crash due to #2759


gtoison:
```
Issue2759Test > testIssue(SpotBugsRunner) FAILED
    java.lang.AssertionError: Analysis failed with exception. Check stderr for detail.
        at edu.umd.cs.findbugs.test.AnalysisRunner.run(AnalysisRunner.java:121)
        at edu.umd.cs.findbugs.test.SpotBugsRunner.performAnalysis(SpotBugsRunner.java:61)
        at edu.umd.cs.findbugs.test.SpotBugsRunner.performAnalysis(SpotBugsRunner.java:75)
        at edu.umd.cs.findbugs.detect.Issue2759Test.testIssue(Issue2759Test.java:23)
```

It would  be great if the build could upload the test results in case there are failures

JuditKnoll:
There was a PR about `CONSTANT_Dynamic` partial support, which may be connected, but it looks like the author decided to delete the spotbugs fork. Fortunately, I took a look at it locally earlier, so I could push it under my fork: https://github.com/JuditKnoll/spotbugs/tree/support-constant-dynamic

gtoison:
Thanks @JuditKnoll that's very helpful!

ThrawnCA (approved):

build.gradle
```diff
@@ -28,6 +28,9 @@ subprojects {
 allprojects {
   repositories {
     mavenCentral()
+    maven {
```
gtoison, line 31: This seems to be working, the only missing piece is that we're using a BCEL release candidate.
Hopefully they will release the new version soon, then I'll remove this bit

hazendaz: @gtoison Its ready to be removed now. I've merged in bcel 6.8.1

JuditKnoll (approved):

Issue #1219, opened 2020-07-14
False Positive: BC_UNCONFIRMED_CAST_OF_RETURN_VALUE when using generic bounded type


pcimcioch:
Following code is safe in terms of casting, but spotbugs reports `BC_UNCONFIRMED_CAST_OF_RETURN_VALUE` 
```
package test;

class Parent {}
class Child extends Parent {}

class Factory<T extends Parent> {
    private final Class<T> type;

    Factory(Class<T> type) {
        this.type = type;
    }

    public T create() {
        try {
            return type.newInstance();
        } catch (InstantiationException | IllegalAccessException e) {
            return null;
        }
    }
}

public class Test {
    public static void main(String[] args) {
        Factory<Child> cs = new Factory<>(Child.class);
        Child child = cs.create(); // BC_UNCONFIRMED_CAST_OF_RETURN_VALUE
        System.out.println(child);
    }
}
```

- If `Parent` is interface instead of class, this bug is **not reported**
```
interface Parent {}
class Child implements Parent {}
```
- If `Factory` doesn't bound type T to `Parent`, this bug is **not reported**
```
class Factory<T> {
```
- If type is bounded on method-level, this bug is **not reported**
```
class Factory {
    public <T extends Parent> T create(Class<T> type) {
        try {
            return type.newInstance();
        } catch (InstantiationException | IllegalAccessException e) {
            return null;
        }
    }
}
```

Spotbugs Version: 4.0.6
Report Level: Low

alzimmermsft:
I'm also seeing a similar issue as this with a method such as:

```java
public <T extends Throwable> T logThrowableAsError(T throwable) {
    // Logic to log the throwable.
    return throwable;
}

// Other code later
public void attemptSomething() throws IOException {
    try {
        trySomething();
    } catch (IOException ex) {
        throw logThrowableAsError(ex);
    }
}
```

uhafner:
No, I am running latest 4.4.2 and the bug still is there:
https://ci.jenkins.io/job/Plugins/job/warnings-ng-plugin/job/PR-1075/15/spotbugs/source.011b3afe-630a-44df-8856-41442c138a42/#177

danielnorberg:
I'm also seeing this issue.

zman0900:
Also seeing this with fairly basic use of Jackson.  For example:

```Java
package crap

import com.fasterxml.jackson.databind.ObjectReader;
[...]

PR #2811, opened 2024-01-13
Ignore variable types so FindBadCast2 does not report BC_UNCONFIRMED_CAST_OF_RETURN_VALUE

gtoison:
Variable types are currently not handled. This should not be reported as bug:
```
Factory<Child> cs = new Factory<>(Child.class);
Child child = cs.create(1);
```


gtoison:
This should fix #1219

hazendaz (approved):

Issue #2782, opened 2023-12-21
False Positive: BC_UNCONFIRMED_CAST for pattern matching switch


robinjhector:
In JDK 21 we can now use pattern matching switches. So instanceof checking can be done in the switch label.

Here's a failing unit test branch that returns `BC_UNCONFIRMED_CAST` (a false positive): https://github.com/spotbugs/spotbugs/compare/master...robinjhector:spotbugs:issue2782

I would be happy to contribute with a fix for this

In-line example

```java
public class Issue2782 {

    public interface MyInterface<T> {
    }

    static class Impl1 implements MyInterface<String> {
        public MyInterface<String> helloFromImpl1() {
            return this;
        }
    }

    static class Impl2 implements MyInterface<Boolean> {
        public MyInterface<Boolean> helloFromImpl2() {
            return this;
        }
    }

    public static <T> void method(MyInterface<T> instance) {
        Object bla = switch (instance) {
            case Impl1 impl1 -> impl1.helloFromImpl1();
            case Impl2 impl2 -> impl2.helloFromImpl2();
            default -> throw new IllegalArgumentException("Blah");
        };

        System.out.println(bla);
    }
}
```

robinjhector:
I can't get the development environment to work locally with gradle for some reason, so I can't pinpoint where the false-positive occurs. Hopefully someone else can take a look at this!

robinjhector:
This also applies to `BC_UNCONFIRMED_CAST_OF_RETURN_VALUE`, even though the return value is never cast to anything

gtoison:
> I can't get the development environment to work locally with gradle for some reason

What's the error you're getting? We're looking for ways to help external contributors onboarding into the project

robinjhector:
> > I can't get the development environment to work locally with gradle for some reason
> 
> What's the error you're getting? We're looking for ways to help external contributors onboarding into the project

It compiled fine iirc, but I couldn't get the debugger to attach correctly in IntelliJ. No breakpoints were respected. I don't know how to trouble shoot it since I'm more of a Maven guy myself, and has rarely worked with gradle

maxl2287:
Hi,
is it already resolved?
Because with 4.8.5.0 (maven-spotbugs) I still got this spotbug for:

```java
boolean isValid = switch (geographicArea) {
      case Polygon polygon -> ...
      case PointUncertaintyCircle circle -> ...
      case Point point -> ...
      default -> {
        yield true;
      }
    };
```

[...]

PR #2813, opened 2024-01-14
Ignore checkcast in Java 21 type switch

gtoison:
Repurposed `SwitchHandler` to keep track of the `InvokeDynamic` corresponding to a Java 21 type switch.

We detect that the switch uses the object type by looking for a `typeSwitch` invocation.
Then we ignore the `checkcast` for each of the switch's target PC + 1

```
    11  invokedynamic 0 typeSwitch(java.lang.Object, int) : int [13]
    16  lookupswitch default: 72
          case 0: 44
          case 1: 58
    44  aload_2
    45  checkcast Issue2782$Impl1 [17]
    48  astore 4 [impl1]
    50  aload 4 [impl1]
    52  invokevirtual Issue2782$Impl1.helloFromImpl1() : Issue2782$MyInterface [19]
    55  goto 82
    58  aload_2
    59  checkcast Issue2782$Impl2 [23]
    62  astore 5 [impl2]
    64  aload 5 [impl2]
    66  invokevirtual Issue2782$Impl2.helloFromImpl2() : Issue2782$MyInterface [25]
    69  goto 82
    72  new java.lang.IllegalArgumentException [28]
```

~~I have removed the sample provided by @robinjhector in #2782 so we don't have to compile Java 21 code because it doesn't seem to work for now. Instead I'm using the compiled .class files directly.~~

~~Update: this PR effectively upgrades the build to JDK 21 (from 17 AND 21) since we now use Java 21 language features in bug samples~~

This PR enables the compilation of bug samples using Java 21 language features when running on JDK 21 or above


spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
```diff
@@ -120,6 +126,45 @@ public SourceLineAnnotation getCurrentSwitchStatement(BytecodeScanningDetector d
...
+    public boolean isTypeSwitchCaseCheckCast(int opCode, int pc) {
+        if (opCode != Const.CHECKCAST) {
+            return false;
+        }
+
+        // For a type switch each case starts with an aload_2 followed by a checkcast
+        // Since we're on a checkcast look for a switch with a target on the previous PC
+        SwitchDetails switchDetails = findSwitchDetailsByPc(pc - 1);
+
+        if (switchDetails != null) {
```
ThrawnCA, line 149: This is essentially "if not...else" which is a double negative. Perhaps refactor to "if null then return false".

spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
```diff
@@ -120,6 +126,45 @@ public SourceLineAnnotation getCurrentSwitchStatement(BytecodeScanningDetector d
...
+            return typeSwitchPcs.contains(switchDetails.switchPC);
+        }
+
+        return false;
+    }
+
+    private SwitchDetails findSwitchDetailsByPc(int pc) {
+        for (SwitchDetails switchDetails : switchOffsetStack) {
+            for (int i = 0; i < switchDetails.swOffsets.length; i++) {
```
ThrawnCA, line 158: Couldn't this be a second foreach loop?

spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
gtoison: Yes, I have made the change

hazendaz:
Am I missing where jdk 21 only is required?  I don't see any new imports and tests seem to all indicate only run with jdk 21 so I think jdk 17 is still allowed?

Also this is still in Draft so I presume you will remove it from draft once its ready to be merged.

gtoison:
I have reverted the change to continue building on JDK 17 and 21, the bug samples using java 21 langage features are not built when running on JDK 17 (and the correspoding tests are skipped)

.github/workflows/release.yml
```diff
@@ -42,10 +42,10 @@ jobs:
           commit_message: 'docs: update CHANGELOG.md for ${{ github.ref_name }}'
           file_pattern: CHANGELOG.md
 
-      - name: Set up JDK 21
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          java-version: '21'
+          java-version: '17'
```
JuditKnoll, line 48: If Java 21 is needed for all the test to run, I think it's better if we stuck to this version for the release.

ThrawnCA (approved):

hazendaz (approved):

Issue #2820, opened 2024-01-18
Nightly builds do not deploy Eclipse plugin anymore


iloveeclipse:
I haven't checked which PR broke that, but since a week or so the latest builds on master do not deploy Eclipse plugin anymore to https://spotbugs.github.io/eclipse-latest/.

This makes it impossible to use latest nightly builds in Eclipse.

JuditKnoll:
Yes, it's my bad, I'm sorry and thank you for bringing it to my attention. I created a PR fixing this: https://github.com/spotbugs/spotbugs/pull/2821

iloveeclipse:
Still nothing deployed ... Reopening.

hazendaz:
Agree seeing the same issue.  I run off snapshots only.  Always expect an update everytime code is merged.  I'll take a look as well.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

iloveeclipse:
See commits pushed to "eclipse-latest" gh-pages repo: https://github.com/spotbugs/eclipse-latest/commits/gh-pages/
Last one was pushed for https://github.com/spotbugs/spotbugs/commit/572b53fade25474faca1f61c95a12bef82e0ea48.

So next commit https://github.com/spotbugs/spotbugs/commit/e05a747ede654f9e41be58c7a851a2179c34be1a on spotbugs didn't produced commit on "eclipse-latest" gh-pages repo...

Here is the link to github actions at this commit: https://github.com/spotbugs/spotbugs/actions/runs/7455816001/job/20285476104
but I don't see any issue reported there.

Looking at the magic number of **800 commits** on the traget repo, and thinking about the repo size (approx. 1.8 GB), I wonder if we have hit some limit on the github repo:

![image](https://github.com/spotbugs/spotbugs/assets/964108/8820bf2f-15f3-45eb-bf80-3b4295d7fa4e)

iloveeclipse:
See https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages

But I also see a change in *how* the commits to that repo were created.
Till 172adddd2de954d135dfad5b6977ca2a43de686f the commits never deleted site.xml, but updated it.
After the commit above commits started to alternate and add/delete contents on each commit, and then it stopped to commit anything...

iloveeclipse:
I believe the 
[...]

PR #2821, opened 2024-01-18
Fix release gha build

JuditKnoll:
It looks like I made a mistake with the condition in the `Build otherwise` step in https://github.com/spotbugs/spotbugs/pull/2806.
This PR should fix https://github.com/spotbugs/spotbugs/issues/2820.


iloveeclipse (approved):

iloveeclipse:
Let see...

PR #2831, opened 2024-01-21
Small update to daily release condition

iloveeclipse:
Change condition to run every time but not on tags.

See https://github.com/spotbugs/spotbugs/issues/2820

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,7 @@ jobs:
           SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
           KEYSTORE_PASS: ${{ secrets.KEYSTORE_PASS }}
       - name: Build otherwise
-        if: ${{ !((github.ref_type == 'tag') && (github.event.base_ref == 'refs/heads/master')) }}
+        if: ${{ (github.ref_type != 'tag') && (github.event.base_ref == 'refs/heads/master') }}
         run: |
           gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg
           echo sonatypeUsername=eller86 >> gradle.properties


ThrawnCA (approved):

.github/workflows/release.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
           KEYSTORE_PASS: ${{ secrets.KEYSTORE_PASS }}
       - name: Build otherwise
-        if: ${{ !((github.ref_type == 'tag') && (github.event.base_ref == 'refs/heads/master')) }}
+        if: ${{ (github.ref_type != 'tag') && (github.event.base_ref == 'refs/heads/master') }}
```
JuditKnoll, line 72: If we want it to run every time but on tags, the condition should only be `(github.ref_type != 'tag')` (or `(github.ref_type == 'branch')` since it is only run on tags and branches). The second part of the condition is unnecessary in this case. (Originally I wanted this part to run both on simple commits and on tags not on master, but you are right it's better to run it on only commits.)
See that the value of `github.event.base_ref` is empty, when there is only a commit pushed from https://github.com/spotbugs/spotbugs/pull/2806#issue-2070658001:

> I tested this solution on my repository with a simple [test.yml](https://github.com/JuditKnoll/spotbugs/blob/master/.github/workflows/test.yml), which only echoes some values, the `Echo on tag if condition` step has the same condition as this PR.
> 
> * when the tag is pushed on to separate branch, which is later merged to master, [the job only runs when pushing the tag, so the relevant step does not run](https://github.com/JuditKnoll/spotbugs/actions/runs/7448441800/job/20263873941),
> * when there is no tag, just a commit pushed to master, [the relevant step does not run](https://github.com/JuditKnoll/spotbugs/actions/runs/7448533288/job/20263124283),
> * when the tag is only pushed to master, [the relevant step runs](https://github.com/JuditKnoll/spotbugs/actions/runs/7448557603/job/20263201209).

A working link to the workflow file: https://github.com/JuditKnoll/spotbugs/actions/runs/7448156139/workflow

iloveeclipse: Actually I wonder now why do we want to have it conditionally anyway? I would like to get my nightly build updated to latest master **also** if the master commit is a new version.

So should we remove the condition here completely?

JuditKnoll: If there is a tag pushed to master, there is already a build run in step `Build on tag` (from line 57). The difference between the two builds is only the ˙closeAndReleaseSonatypeStagingRepository˙. 
I'm not sure whether running both builds could cause any issues for the eclipse-latest repository (e.g. some minor differences between the builds), but I don't think it really should.

hazendaz:
Do we still want this?

iloveeclipse:
> Do we still want this?

I was just wanted to see daily builds appearing, but that seem to work **without** this PR.
So I would delegate the question to @JuditKnoll if we need this PR, some other change or let the current state "as is".

hazendaz:
@JuditKnoll Whatever you feel is best there, please go ahead and made necessary changes.  Thanks.

hazendaz:
Going to leave it as is.  There is a reason it is checking master because tags can be on any branch and we want to prevent that.  Changing then doesn't make much sense.  I agree we probably could have better naming but that should probably be a separate PR that attempts to cleanup the actions entirely as the naming even on this isn't what is suggested these days based on google searches I performed.  Since we are currently working going to just close.  Thanks.

PR #2694, opened 2023-11-10
Some `NonNull` annotations not recognized

gtoison:
Added support for:
- `edu.umd.cs.findbugs.annotations.NonNull`
- `edu.umd.cs.findbugs.annotations.Nullable`
- `org.netbeans.api.annotations.common.NonNull`
- `org.netbeans.api.annotations.common.NullAllowed`
- `lombok.NonNull`

PR #2840, opened 2024-01-26
Fix missing classes needed for analyses for Tests

JuditKnoll:
Several tests were emiting the `The following classes needed for analyses were missing:` message to the standard error, but it doesn't appear in the log when running the tests from gradle, and this does not make the tests fail.

The missing classes can be either coming from
- outer dependency (e.g. google guava), adding aux classpathes to `AbstractIntegrationTest` solves this for its descendants,
- inner dependency (e.g. an inner class or other class from `spotbugsTestCases`), which needs to be added each test.

Converted some test which missed some classes and was using `@ExtendWith(SpotBugsExtension)` to extend `AbstractIntegrationTest`, since the former is for SpotBugs plugins, and as such it can't be expected to look for the aux classes inside `spotbugsTestCases`.

In some of the affected classes, done some minor cleanups.

The PR is still a draft, since it disables `RegressionIdeas20110722Test`, because that test fails, and I think the `FindNullDeref` detector is the faulty one, but that's quite complex.

The following tests are still missing some classes:
- `Issue1771Test` misses `java.util.Collections$EmptyNavigableMap`, `java.util.Collections$EmptyNavigableSet`;
    - It may be because this tests needs Java 11.
- `Issue2759Test` misses `com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParser$ClassBodyDeclarationContext`, `com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParser`, `org.antlr.v4.runtime.tree.RuleNode`, `org.antlr.v4.runtime.tree.ParseTreeVisitor`, `com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParserVisitor`, `org.jacoco.agent.rt.internal_4742761.Offline`, `com.puppycrawl.tools.checkstyle.grammar.java.JavaLanguageParser$BlockContext`.
   - It's read from a class file, with several outside dependencies. IMO we shouldn't add all of those dependencies, and it's okay like this.
   - When I tried to perform the analyses on some of the other files which were added in https://github.com/spotbugs/spotbugs/pull/2807, the there was an exception thrown - not connected to the original issue.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
@@ -3,6 +3,7 @@ package edu.umd.cs.findbugs.detect;
 import edu.umd.cs.findbugs.AbstractIntegrationTest;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
 import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
@@ -11,6 +12,8 @@ import static org.hamcrest.Matchers.hasItem;
 
 class RegressionIdeas20110722Test extends AbstractIntegrationTest {
 
+    // This test, or rather the detector needs to be fixed
+    @Disabled
     @Test
     void testArgumentAssertions() {
         performAnalysis("bugIdeas/Ideas_2011_07_22.class");
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
@@ -12,7 +12,7 @@ import static org.hamcrest.Matchers.hasItem;
 
 class RegressionIdeas20110722Test extends AbstractIntegrationTest {
 
-    // This test, or rather the detector needs to be fixed
+    // Problems in null checks, see https://github.com/spotbugs/spotbugs/issues/2890
     @Disabled
     @Test
     void testArgumentAssertions() {
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
@@ -123,6 +123,19 @@ public abstract class AbstractIntegrationTest {
             throw new UncheckedIOException(e);
         }
 
+        final Path dependencies = getFindbugsTestCases().resolve("build/spotbugs/auxclasspath/spotbugsMain");
+        try {
+            final List<String> lines = Files.readAllLines(dependencies);
+            for (String line : lines) {
+                Path path = Paths.get(line);
+                if (Files.isReadable(path)) {
+                    runner.addAuxClasspathEntry(path);
+                }
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+
         // TODO : Unwire this once we move bug samples to a proper sourceset
         Path[] paths = Arrays.stream(analyzeMe)
                 .map(AbstractIntegrationTest::getFindbugsTestCasesFile)
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue429Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue429Test.java
@@ -3,29 +3,22 @@ package edu.umd.cs.findbugs.ba;
 import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
 import static org.hamcrest.MatcherAssert.assertThat;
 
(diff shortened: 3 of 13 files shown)


gtoison:
Hum, I think this might fix the issues I was wondering about in #2694

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
```diff
@@ -11,6 +12,8 @@
 
 class RegressionIdeas20110722Test extends AbstractIntegrationTest {
 
+    // This test, or rather the detector needs to be fixed
```
ThrawnCA, line 15: Can you describe in more detail what is broken?

JuditKnoll:
> Hum, I think this might fix the issues I was wondering about in #2694

Maybe the solution is the same to the problems in this PR and #2694, since both are connected to bugs reported by `FindNullDeref`, but unfortunately, this PR in its current form doesn't seem to help with #2694 - tried it locally.

gtoison:
Ah too bad, that means I'll have to find out how that thing works :)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20110722Test.java
JuditKnoll: I'm still trying to fix this problem, this is the reason the PR is still a draft, and why I haven't explained it in the comment. If I give up, I definitely will provide some details in the code, or rather open an issue and refer that here.

Basically, the found buginstances should be:
- 4 RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE 
   - at getHashCode3() method line 34
   - at getHashCode4() method line 41
   - at getHashCode4() method line 48
   - at getHashCode4() method line 55
- 2 NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE
  - at getHashCode6() method parameter x
  - at getHashCode7() method parameter x

Instead it is the following:
- 4 NP_NULL_ON_SOME_PATH
   - at getHashCode3() method line 37
   - at getHashCode4() method line 44
   - at getHashCode5() method line 51
   - at getHashCode6() method line 58

I think the root of the problem might be in `IsNullValueAnalysis`, which I'm not so familiar with and need to understand how it works to be able to solve it.

hazendaz:
@JuditKnoll We still draft here?

JuditKnoll:
> @JuditKnoll We still draft here?

Unfortunately, I haven't got time lately to properly look into the underlying problem, it still exists. I didn't want to convert it to ready for review without proper investigation first. However, the PR doesn't introduce new problems, but solves one, and meanwhile sheds light on an already existing, hidden problem (tested it locally with a reproducer with SpotBugs 4.8.3.). So I decided to create the issue https://github.com/spotbugs/spotbugs/pull/2840 for this problem and make this PR ready for review.

hazendaz (approved):

gtoison:
Hello @JuditKnoll 
I get the error below when running `gradlew clean test` and I think this PR might have something to do with it.
On a second `gradle test` run, the error does not show up; so I'm guessing that the missing file was created by the first run.
Would you know what's creating the file?

```
SpotBugsExtensionTest > test(SpotBugsRunner) FAILED
    java.io.UncheckedIOException: java.nio.file.NoSuchFileException: build\spotbugs\auxclasspath\spotbugsMain
        at edu.umd.cs.findbugs.test.SpotBugsExtensionTest.addAuxClassPathEntry(SpotBugsExtensionTest.java:56)
        at edu.umd.cs.findbugs.test.SpotBugsExtensionTest.test(SpotBugsExtensionTest.java:39)

        Caused by:
        java.nio.file.NoSuchFileException: build\spotbugs\auxclasspath\spotbugsMain
            at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
            at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
            at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
            at java.base/sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:234)
            at java.base/java.nio.file.Files.newByteChannel(Files.java:379)
            at java.base/java.nio.file.Files.newByteChannel(Files.java:431)
            at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:420)
            at java.base/java.nio.file.Files.newInputStream(Files.java:159)
            at java.base/java.nio.file.Files.newBufferedReader(Files.java:2897)
            at java.base/java.nio.file.Files.readAllLines(Files.java:3392)
            at java.base/java.nio.file.Files.readAllLines(Files.java:3433)
            at edu.umd.cs.findbugs.test.SpotBugsExtensionTest.addAuxClassPathEntry(SpotBugsExtensionTest.java:47)
            ... 1 more

1 test completed, 1 failed
```

JuditKnoll:
> Hello @JuditKnoll I get the error below when running `gradlew clean test` and I think this PR might have something to do with it. On a second `gradle test` run, the error does not show up; so I'm guessing that the missing file was created by the first run. Would you know what's creating the file?

I'm not sure, but I will look into it later this week. At first glance, it looks like maybe this change in `SpotBugsExtensionTest` is not needed, since it's in `test-harness-jupiter`, which even after a build does not contain the referenced folder, so it's strange, that it only causes issue the first time around.

Issue #1097, opened 2020-03-02
Spotbugs-4.0.0 doesn't detect FLOATING_POINT_EQUALITY


ChristopherNeufeld:
I have seen a change in behaviour, using the test cases in spotbugs.
Downloaded the binary packages for spotbugs-3.1.12 and spotbugs-4.0.0.
In the source package, there is a file, spotbugs-4.0.0/spotbugsTestCases/src/java/FloatingPointEquality.java.
Compiled this file: javac FloatingPointEquality.java
My javac is version 1.8.0_232 on Linux.
I then ran both spotbugs versions against the resulting .class file.  With spotbugs-3.1.12:

```
H D FE: Test for floating point equality in FloatingPointEquality.sum(double)  At FloatingPointEquality.java:[line 5]
Warnings generated: 1
```
With spotbugs-4.0.0, no output is generated.

PR #2843, opened 2024-01-27
Fix for FindFloatEquality same field test

gtoison:
The `isSameField()` method was changed in https://github.com/spotbugs/spotbugs/commit/84859d131c1c25346ab8f8ec83d1d674a7eced0f and the case when both arguments aren't fields was changed.
Reverted the change to fix `FE_FLOATING_POINT_EQUALITY`


hazendaz (approved):

hazendaz:
fixes #1097

Issue #2857, opened 2024-02-05
Sonar run from master fails with java 11


hazendaz:
After adding in build support for java 11 build, master started failing with sonar.  On a quick look not seeing where its happening.  The PR bringing in support did not display the condition.  See https://github.com/spotbugs/spotbugs/actions/runs/7777531716/job/21205933927 with all working.  While I can understand master does the real push to sonar and thus issue, its confusing to me that PRs now get the same issue after the fact.

hazendaz:
ok sonar runs against any and all branches when on spotbugs repo.  We probably need to limit that down somewhat as we don't need it scanned for java 11 / 17 for example if we have 21 working.  Reason my PR didn't show the issue was that it was coming from a fork.

hazendaz:
resolved.

PR #2859, opened 2024-02-07
[GHA] Split up build using sonar so it runs against java 21 ubuntu only

hazendaz:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Set up JDK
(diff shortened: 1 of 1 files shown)


hazendaz:
Fixes #2857

hazendaz:
merging as this won't see anything on PR but rather after it lands on master.

JuditKnoll:
It seems to fix the problem 😄 
According to the title it looks like you only wanted to run sonar on ubuntu java 21, but there is no check for ubuntu, only for the java version.

hazendaz:
Good catch.  I intended that but couldn't get it to work so I dropped back to jdk check only.

hazendaz:
Fixed title.

PR #2865, opened 2024-02-10
Hibernate's bytecode enhancement DLS_DEAD_LOCAL_STORE false positives

### LINKED ISSUE #2864: `DLS_DEAD_LOCAL_STORE` false positives for Hibernate enhanced entities
Hibenate's [bytecode enhancement](https://docs.jboss.org/hibernate/orm/5.4/topical/html_single/bytecode/BytecodeEnhancement.html) adds generated methods that trigger `DLS_DEAD_LOCAL_STORE` false positives.

Since the bytecode is generated it is not very useful to report these issues.

gtoison:
The methods added by Hibernate's bytecode enhancement trigger DLS_DEAD_LOCAL_STORE false positives

Fixes #2864


hazendaz (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2864Test.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.nio.file.Paths;
+
+/**
+ * @author gtoison
+ *
+ */
+@ExtendWith(SpotBugsExtension.class)
+class Issue2864Test extends AbstractIntegrationTest {
```
JuditKnoll, line 24: You don't need to use both `@ExtendWith(SpotBugsExtension.class)` and extend `AbstractIntegrationTest`. Only one of them is enough. I prefer the `AbstractIntegrationTest`, since that's more simple, and https://github.com/spotbugs/spotbugs/pull/2840 solves the missing classes problem for that.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2864Test.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    void testIssue(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/src/classSamples/classEnhanceByHibernate/HibernateEnhancedEntity.class"));
```
JuditKnoll, line 29: When running this test, several classes are missing:
```
  org.hibernate.engine.spi.ManagedEntity
  org.hibernate.engine.spi.ExtendedSelfDirtinessTracker
  org.hibernate.bytecode.enhance.internal.tracker.SimpleFieldTracker
  org.hibernate.bytecode.enhance.internal.tracker.SimpleCollectionTracker
  org.hibernate.engine.spi.EntityEntry
  org.hibernate.bytecode.enhance.internal.tracker.DirtyTracker
  org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor
  org.hibernate.bytecode.enhance.spi.CollectionTracker
  org.hibernate.collection.spi.PersistentCollection
```

I'm not sure, how much these classes are really needed, and what exactly causes when there are missing classes. The PR solves the referenced issue, so maybe it's not a big deal in this case.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2864Test.java
gtoison: Thanks for reviewing, I've made the change

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2864Test.java
gtoison: The classes references are from the bytecode added by Hibernate's enhancer.
I looked into adding Hibernate as a dependency and also adding the enhancer plugin to the build but that seems to be overkill, so I just compiled a class and copied it into the samples folder.

JuditKnoll (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindDeadLocalStores.java
```diff
@@ -209,6 +209,10 @@ private void analyzeMethod(ClassContext classContext, Method method) throws Data
             return;
         }
 
+        if (method.getName().startsWith("$$_hibernate_")) {
```
ThrawnCA, line 212: Might be worth ignoring other `$$` methods?

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindDeadLocalStores.java
gtoison: I thought about it by I think the underlying issue might very well be a bug in Hibernate's bytecode enhancer, to at least it seems to be producing suboptimal bytecode here.
In case other false positives are reported it will be easy to ignore more methods

ThrawnCA (approved):

gtoison:
Is it OK if I merge this PR?
There are a few outstanding PRs and its getting time consuming to merge the conflicts in the changelog
Let me know if you think I could/should/shouldn't/musn't/can't merge my own PR once approved :)

JuditKnoll:
> Is it OK if I merge this PR? There are a few outstanding PRs and its getting time consuming to merge the conflicts in the changelog Let me know if you think I could/should/shouldn't/musn't/can't merge my own PR once approved :)

I don't see any reason not to merge it. 
Usually @hazendaz or @iloveeclipse merge the PRs, so if they could express their opinion it would be really useful.

hazendaz:
Go for it

Issue #2870, opened 2024-02-21
Treat `@PostConstruct` methods as constructors for `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`


famod:
Very similar to #453.

As of SB 4.8.3, `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` is reported for the following class:
```java
// CDI bean
@ApplicationScoped
puplic class Foo {
    private Object bar;

    @PostConstruct
    void init() {
        bar = new Object();
    }

    public String doSomethingWithBar() {
        return bar.toString();
    }
}
```
This will never be a problem during runtime (only maybe in simple tests that don't use CDI, but you'll notice soon enough).

Issue #453, opened 2017-10-16
Treat unit test setup methods as constructors for UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR


ThrawnCA:
JUnit and TestNG test classes typically initialise their fields within a `setup` method, rather than the constructor, which often results in UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR. Can appropriately-annotated setup methods (`@Before`, `@BeforeMethod`, `@BeforeClass`) be treated as constructors for this purpose?

rscarvalho:
any updates on this?

jansohn:
We also start seeing this with JUnit 5 and its annotation `@BeforeEach`

hazendaz:
Would need @BeforeAll as well.

PR #2872, opened 2024-02-22
Consider initializer annotations when looking for `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`

gtoison:
See #453 and #2870

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
@@ -819,7 +819,7 @@ public class UnreadFields extends OpcodeStackDetector {
 
     /**
      *
-     * @return true if the method is considered to be an initializer method. Fields migth be initialized outside of a constructor,
+     * @return true if the method is considered to be an initializer method. Fields might be initialized outside of a constructor,
      * for instance through JUnit's BeforeEach
      */
     private boolean isInitializerMethod() {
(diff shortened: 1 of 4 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
```diff
@@ -816,6 +817,32 @@ public void sawOpcode(int seen) {
         previousOpcode = seen;
     }
 
+    /**
+     *
+     * @return true if the method is considered to be an initializer method. Fields migth be initialized outside of a constructor,
```
JuditKnoll, line 822:
```suggestion
     * @return true if the method is considered to be an initializer method. Fields might be initialized outside of a constructor,
```

gtoison:
Thanks for reviewing @JuditKnoll, I have made the changes

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
```diff
@@ -816,6 +817,32 @@ public void sawOpcode(int seen) {
...
+     */
+    private boolean isInitializerMethod() {
+        if ("init".equals(getMethodName()) || "initialize".equals(getMethodName())) {
+            return true;
+        }
+
+        for (AnnotationEntry a : getMethod().getAnnotationEntries()) {
+            String typeName = a.getAnnotationType();
+
+            if ("Ljakarta/annotation/PostConstruct;".equals(typeName)
```
ThrawnCA, line 833: Could put these in an ordered array and use `Arrays.binarySearch`, or an `ArrayList` and use `contains`.

ThrawnCA (approved):

PR #1861, opened 2021-12-10
Issue 574

ramraval:
Attempt at addressing #574 to ignore URF_UNREAD_FIELD warning for fields with certain annotations (e.g. @SerializedName, @XmlElement, and @RegisterExtension). New contributor / completed as part of university project.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #2873, opened 2024-02-23
Do not report `URF_UNREAD_FIELD` for fields with certain annotations

### LINKED ISSUE #574: URF_UNREAD_FIELD: should be ignored on test class fields annotated with JUnit5's @RegisterExtension
JUnit 5.1 allows you to [register an extension programmatically](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic) by adding a field to your class as shown below. Similiar to JUnit4s Rule/ClassRule.

```
@RegisterExtension
JerseyExtension jerseyExtension = new JerseyExtension(this::configureJersey);
```
(apologies I haven't yet had a chance to add a failing test for this)

gtoison:
Since there was no answer in #1861 I've merged the proposed changes and made my own changes on top:
- Upgraded the test to JUnit 5
- Replaced the annotation classe references by class names so there's no runtime dependency
- Added jakarta annotations

I think the comment to move `UnreadFieldsAnnotationChecker` into `UnreadFields` has merits (since the class is only used there) but I haven't made the change to preserve the Git history. Let me know if you think the change should be made.

This should fix #574 and supersede #1861


spotbugsTestCases/src/java/ghIssues/issue574/Issue574RegisterExtension.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+/**
+ * Tests that "URF_UNREAD_FIELD" warning is suppressed for fields with @RegisterExtension annotation.
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/574">GitHub issue</a>
+ */
+
+
+public final class Issue574RegisterExtension {
```
JuditKnoll, line 11: Couldn't find any tests, which calls the `performAnalysis()` function for this class.

JuditKnoll (changes requested):
Thanks a lot for picking up that dangling PR.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/UnreadFieldsAnnotationChecker.java
```diff
@@ -0,0 +1,66 @@
...
+import edu.umd.cs.findbugs.classfile.analysis.AnnotationValue;
+import edu.umd.cs.findbugs.classfile.ClassDescriptor;
+import edu.umd.cs.findbugs.classfile.DescriptorFactory;
+
+/**
+ * Utility class checking for presence of certain annotations. If one of the listed annotations is associated with
+ * a given field, then the "URF_UNREAD_FIELD" detector should be skipped for the field.
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/574">GitHub issue</a>
+ */
+public final class UnreadFieldsAnnotationChecker {
```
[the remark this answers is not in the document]
gtoison: I have moved the code to `UnreadFields`

spotbugs/src/main/java/edu/umd/cs/findbugs/util/UnreadFieldsAnnotationChecker.java
```diff
@@ -0,0 +1,66 @@
...
+        "javax/xml/bind/annotation/XmlElement",
+        "javax/xml/bind/annotation/XmlAttribute",
+        "javax/xml/bind/annotation/XmlValue",
+        "jakarta/xml/bind/annotation/XmlElement",
+        "jakarta/xml/bind/annotation/XmlAttribute",
+        "jakarta/xml/bind/annotation/XmlValue",
+        "org/junit/jupiter/api/extension/RegisterExtension"
+    };
+
+    private static final Set<ClassDescriptor> IGNORE_RULE_FOR_THESE_CLASS_DESCRIPTORS =
```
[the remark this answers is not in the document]
gtoison: I have moved the code to `UnreadFields` and am using one collection only

spotbugsTestCases/src/java/ghIssues/issue574/Issue574RegisterExtension.java
gtoison: Isn't it analyzed in `Issue574Test.testRegisterExtensionAnnotation`?

ThrawnCA (approved):

spotbugsTestCases/src/java/ghIssues/issue574/Issue574RegisterExtension.java
JuditKnoll: It's the dependency of the `Issue574RegisterExtension` class, which is analyzed there, but it itself is not analyzed, the dependency is missing, it is even listed as a missing class on the std err (or to be more precise the `BypassFileNotFoundExceptionExtension` is listed, but if you add that, the `Issue574RegisterExtension` will be missed. The dependencies outside the `spotbugsTestCases` directory will be handled by https://github.com/spotbugs/spotbugs/pull/2840, but the dependencies inside it (such as parents, nested classes), needs to be added.

JuditKnoll: Yes, you are right, left the comment at the wrong place. I'm sorry. Can you please add `BypassFileNotFoundExceptionExtension` to the relevant test?

JuditKnoll (approved):

PR #2889, opened 2024-03-04
Fix deprecated renovate setup / Make sure files follow GIT convention and end with empty end of line

hazendaz:


hazendaz:
Hi All, I don't expect everyone to notice the EOL markers that often as pulls come in with then and the process 'git bash script' to fix them takes forever.  So I just plan to do this myself typically around time we go for a release.  The point is that when git performs diffs, if it does not have end of line marker, it has to add one so technically its slightly faster to do it appropriately but its often easily missed.

hazendaz:
For the renovate item, they put a notice out and are supposed to eventually fix those on their own PRs but still haven't done so.  So I mass did some 160 repos I have and just pushing them up when I have time.  It makes no real difference as it was just a naming change.

That all said I did not bother with change log because technically nothing changes here as completely no-op.

JuditKnoll (approved):

Issue #2047, opened 2022-05-12
Specifying the same report multiple times results in output issues


Vogel612:
For reasons I don't want to get into here, I accidentally ran spotbugs with the following abridged command: 

```
> java -jar /spotbugs/spotbugs.jar -noClassOk -xml:withMessages=spotbugs.xml [...] -xml:withMessages=spotbugs.xml -onlyAnalyze com.example.SomeClass dist/*.ear
```

Specifying the **same** XML report output multiple times seems to consistently generate a non-parseable XML file on the `FileStats` section, given sufficiently many files analyzed.
It seems that there is some output flushing issue between two instances of the XML reporter. I assume that also applies for other reporters.

I'd expect a reporter to always generate valid data of its output format. Spotbugs should at least warn me about reporters fighting over a file (or even throw an error), especially when generating multiple reports using different options (i.e. `-xml:withMessages=spotbugs.xml -xml:minimal=spotbugs.xml`

PR #2894, opened 2024-03-09
Keep track of constructed BugReporters for TextUiCommandLine (#2047)

Vogel612:
Updates TextUiCommandLine to avoid writing to the same output file multiple times.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -32,6 +32,23 @@ class TextUICommandLineTest {
     }
 
     @Test
+    void differentFilepathsOnSameReporter_createsTwoReporters() throws IOException {
+        Path file = Files.createTempFile("spotbugs", ".xml");
+        Path file2 = Files.createTempFile("spotbugs", ".xml");
+        TextUICommandLine commandLine = new TextUICommandLine();
+
+        TextUIBugReporter call1 = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
+                () -> new SortingBugReporter(), (reporter, reporterOptions) -> {
+                });
+        TextUIBugReporter call2 = commandLine.initializeReporter("withMessages=" + file2.toFile().getAbsolutePath(),
+                () -> new SortingBugReporter(), (reporter, reporterOptions) -> {
+                });
+
+        assertNotNull(call1);
+        assertNotNull(call2);
+    }
+
+    @Test
     void handleOutputFilePathUsesGzip() throws IOException {
         Path file = Files.createTempFile("spotbugs", ".xml.gz");
         TextUICommandLine commandLine = new TextUICommandLine();
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -21,10 +21,10 @@ class TextUICommandLineTest {
         TextUICommandLine commandLine = new TextUICommandLine();
 
         TextUIBugReporter call1 = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
-                () -> new SortingBugReporter(), (r, rOpts) -> {
+                () -> new SortingBugReporter(), (reporter, reporterOptions) -> {
                 });
         TextUIBugReporter call2 = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
-                () -> new SortingBugReporter(), (r, rOpts) -> {
+                () -> new SortingBugReporter(), (reporter, reporterOptions) -> {
                 });
 
         assertNotNull(call1);
@@ -37,7 +37,7 @@ class TextUICommandLineTest {
         TextUICommandLine commandLine = new TextUICommandLine();
         TextUIBugReporter reporter = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
                 () -> new SortingBugReporter(),
-                (r, rOpt) -> {
+                (constructed, reporterOptions) -> {
                 });
 
         reporter.finish();
@@ -53,7 +53,7 @@ class TextUICommandLineTest {
         TextUICommandLine commandLine = new TextUICommandLine();
         TextUIBugReporter reporter = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
                 () -> new SortingBugReporter(),
-                (r, rOpt) -> {
+                (constructed, reporterOptions) -> {
                 });
 
         reporter.finish();
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -11,24 +11,34 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 class TextUICommandLineTest {
     @Test
-    void handleOutputFileReturnsRemainingPart() throws IOException {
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -266,30 +270,40 @@ public boolean justPrintVersion() {
     private List<TextUIBugReporter> reporters = new ArrayList<>();
 
     /**
-     * Parse {@code optionExtraPart} and configure {@Link TextUIBugReporter} if it contains the
+     * Parse {@code optionExtraPart} create a path-associated {@Link TextUIBugReporter} if it contains the
```
ThrawnCA, line 273: "create" -> "and create"

ThrawnCA (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -266,30 +270,40 @@ public boolean justPrintVersion() {
...
+     * Finally configure the created BugReporter with the {@code optionExtraPart}'s configuration information.
      *
-     * @param reporter the reporter to set a {@link PrintStream} based on the given file path
      * @param optionExtraPart extra part of the specified commandline option
-     * @return Remaining part of {@code optionExtraPart}
+     * @param ctor            A supplier for an unconfigured {@Link TextUIBugReporter}.
+     * @param handleOptions   A function that can configure the created {@code BugReporter} instance.
+     * @param <T>             The implementation type of the {@Link TextUIBugReporter} to propagate type
+     *                        information between {@link ctor} and {@link handleOptions}.
+     * @return The fully configured reporter
```
gtoison, line 282: Can you please add a mention that this methods returns null in case the report was already produced?

gtoison (changes requested):
Thanks for the contribution! Just a small comment on my part
This fixes #2047

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
Vogel612: updated

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
Vogel612: good shout, updated

gtoison (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
gtoison: Thank you, LGTM!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
```diff
@@ -11,24 +11,34 @@
...
-    void handleOutputFileReturnsRemainingPart() throws IOException {
+    void duplicateFilepathIsIgnored() throws IOException {
         Path file = Files.createTempFile("spotbugs", ".xml");
         TextUICommandLine commandLine = new TextUICommandLine();
-        SortingBugReporter reporter = new SortingBugReporter();
-        String result = commandLine.handleOutputFilePath(reporter, "withMessages=" + file.toFile().getAbsolutePath());
 
-        assertThat(result, is("withMessages"));
+        TextUIBugReporter call1 = commandLine.initializeReporter("withMessages=" + file.toFile().getAbsolutePath(),
+                () -> new SortingBugReporter(), (r, rOpts) -> {
```
hazendaz, line 24: Can we not abbreviate here and use something closer to what the method says it is?  That helps in reading it in the future without having to attempt to figure it out.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -343,35 +357,41 @@ protected void handleOption(String option, String optionExtraPart) {
...
+                                r.setAddMessages(true);
+                                xmlWithAbridgedMessages = true;
+                            } else if ("minimal".equals(arg)) {
+                                r.setMinimalXML(true);
+                            } else {
+                                throw new IllegalArgumentException("Unknown option: -xml:" + arg);
+                            }
+                        }
+                    });
+            if (reporter != null)
```
hazendaz, line 385: Anytime we have an if statement, just because there is only one item, please always add {} as expected by general google styling rules.   Like my first comment this makes it easier to read in the future.

hazendaz (changes requested):
See my notes.  I think clarity trumps anything else especially when this bloated the code instead of reducing it.  I think its fine otherwise but nice clarity will help.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
Vogel612: Updated

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
Vogel612: considering that these are NOOPs I didn't put much work into these names, but happy to update

gtoison (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -266,30 +270,40 @@ public boolean justPrintVersion() {
...
-     * @return Remaining part of {@code optionExtraPart}
+     * @param ctor            A supplier for an unconfigured {@Link TextUIBugReporter}.
+     * @param handleOptions   A function that can configure the created {@code BugReporter} instance.
+     * @param <T>             The implementation type of the {@Link TextUIBugReporter} to propagate type
+     *                        information between {@link ctor} and {@link handleOptions}.
+     * @return The fully configured reporter, or {@code null}, if the reporter would output to a file that is already used as a reporter output file.
      */
-    /* visible for testing */ String handleOutputFilePath(TextUIBugReporter reporter, String optionExtraPart) {
+    /* visible for testing */ <T extends TextUIBugReporter> TextUIBugReporter initializeReporter(String optionExtraPart,
+            Supplier<T> ctor, BiConsumer<T, String> handleOptions) {
```
JuditKnoll, line 285: It's a bit strange for me to have `optionExtraPart` as the first parameter, where it was the last one earlier and at the called functions it's the last one. However, since it has different type, it can't cause parameter mixup issues.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -380,22 +403,33 @@ protected void handleOption(String option, String optionExtraPart) {
...
             bugReporterType = HTML_REPORTER;
-            HTMLBugReporter htmlBugReporter = new HTMLBugReporter(project, "default.xsl");
-            optionExtraPart = handleOutputFilePath(htmlBugReporter, optionExtraPart);
-            if (!"".equals(optionExtraPart)) {
-                htmlBugReporter.setStylesheet(optionExtraPart);
+            TextUIBugReporter reporter = initializeReporter(optionExtraPart,
+                    () -> new HTMLBugReporter(project, "default.xsl"),
+                    (r, rOpt) -> {
+                        if (!"".equals(rOpt))
+                            r.setStylesheet(rOpt);
```
JuditKnoll, line 410: Can you please add the brackets to the if to be conform to the Google style guide?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
```diff
@@ -11,24 +11,34 @@
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 class TextUICommandLineTest {
     @Test
-    void handleOutputFileReturnsRemainingPart() throws IOException {
+    void duplicateFilepathIsIgnored() throws IOException {
```
JuditKnoll, line 19: I would love to see one more test, where there are two different files are added with the same commandline reporter.

JuditKnoll (approved):

hazendaz (approved):

hazendaz:
@Vogel612 Thanks!

PR #2915, opened 2024-03-27
Update the filter XSD namespace and location for the upcoming 4.8.4

gtoison:
The XSD was modified but the namespace and location remained on 3.0.0
Update them to the upcoming 4.8.4 version (instead of using the ever moving master branch), the location will be available once the 4.8.4 Git tag is created


JuditKnoll (approved):

hazendaz:
We may want to hold on this until day we are cutting release so there isn't a gap of availability.

ThrawnCA (approved):

PR #2919, opened 2024-04-03
Add 'VNA03-J Sequence of calls on a synchronized abstraction may not be atomic' detector with test cases

PatrikScully:
## Summary

New detector `AtomicOperationsCombinedDetector` targeting [SEI CERT VNA03-J](https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/visibility-and-atomicity-vna/vna03-j/) violations: sequences of calls to individually atomic methods on synchronized collections or `java.util.concurrent.atomic` types that are not synchronized as a group.

A thread-safe class only guarantees atomicity of its individual methods. A grouping of calls to such methods requires additional synchronization. For example, calling `synchronizedList.add()` followed by `synchronizedList.toArray()` without holding a common lock allows another thread to modify the list between the two calls.

## Bug types

- `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC` -- reported when multiple operations on a synchronized/atomic field within a method are not protected by a common lock.
- `AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION` -- reported when a single operation on a field appears in a method, but that field has combined operations in another method, indicating the field requires synchronization everywhere.

## Detection approach

1. **Collect fields**: scan all methods for PUTFIELD/PUTSTATIC instructions where the assigned value comes from `Collections.synchronized*()` or a `java.util.concurrent.atomic` constructor.
2. **Analyze violations**: for each non-constructor, non-synchronized method, use `LockDataflow` to identify unsynchronized locations and a single-pass `MethodStackScanner` to capture operand stack state, then record unsynchronized accesses to tracked fields.
3. **Filter**: remove false positives from private helper methods that are only called from synchronized contexts.
4. **Report**: fields with multiple unsynchronized accesses in a method get `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC`; fields with single access but combined access elsewhere get `AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION`.

## Changes

- Add `AtomicOperationsCombinedDetector` with `MethodStackScanner` for single-pass bytecode analysis.
- Extend `CallListDataflow` / `CallListAnalysis` to attach `XField` attributes to calls.
- Register two new bug patterns with CWE-362.
- Add integration test suite with safe and unsafe scenarios covering synchronized collections, atomic types, nested sync blocks, static fields, finally blocks, lambdas, and private helper methods.


PatrikScully:
We run the checkers on some big projects, here are the results:

* [Bt](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20Bt%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29&is-unique=off&diff-type=New&checker-name=AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION&checker-name=AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC) (3 combined, 7 simple),
* [Jenkins](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20Jenkins%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29&is-unique=off&diff-type=New) (0 hits),
* [MatSim-Libs](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20MATSim-Libs%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29&is-unique=off&diff-type=New&checker-name=AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION&checker-name=AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC) (24 combined, 9 simple),
* [NanoHTTPD](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20NanoHTTPD%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29) (0 hits),
* [OpenGrok](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20OpenGrok%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29) (0 hits),
* [Ttorrent](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=SpotBugs%20on%20Ttorrent%20%28user%3A%20PatrikScully,%20branch%3A%20feature%2FVNA03-J-combine-atomic-operations-detector%29) (0 hits).

hazendaz:
Due to size of this effort and fact we have had some missteps with false positives on releases, deferring this to 5.0.0.  We need to get a proper stable 4.8.x before we move forwards with new features.

JuditKnoll:
> Due to size of this effort and fact we have had some missteps with false positives on releases, deferring this to 5.0.0. We need to get a proper stable 4.8.x before we move forwards with new features.

I have taken a look at it, and I don't think this necessarily needs to be in 5.0.0, since it does not have anything breaking. It can be even in 4.9.0 before 5.0.0, or 5.1.0.

@PatrikScully At https://github.com/spotbugs/spotbugs/pull/2986, you mentioned that you reused some methods from here which you would like to move to a utility class. Can you please do that both here and in the other PR, so these don't depend on each other?

PatrikScully:
> @PatrikScully At #2986, you mentioned that you reused some methods from here which you would like to move to a utility class. Can you please do that both here and in the other PR, so these don't depend on each other?

@JuditKnoll Yes, I have moved the relevant methods into utility classes in both PR.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    private final BugReporter bugReporter;
+    private final BugAccumulator bugAccumulator;
+
+    private final Set<XField> interestingFields = new HashSet<>();
+    private final Set<XField> combinedAtomicFields = new HashSet<>();
+
+    private final Map<Method, Map<XField, List<BugPrototype>>> interestingFieldCalls = new HashMap<>();
+    private final Map<Method, List<BugPrototype>> interestingLocalVariableCalls = new HashMap<>();
```
JuditKnoll, line 108: What makes a local variable or field interesting? Can you please rename these fields and functions to describe better what does these contain or what the function does?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,323 @@
...
+                OpcodeStack stack = OpcodeStackScanner.getStackAt(classContext.getJavaClass(), method, handle.getPosition());
+                OpcodeStack.Item stackItem = stack.getStackItem(0);
+                if (isInterestingField(stackItem.getReturnValueOf())) {
+                    interestingFields.add(XFactory.createXField((FieldInstruction) instruction, cpg));
+                }
+            }
+        }
+    }
+
+    private static boolean isInterestingField(ClassMember classMember) {
```
JuditKnoll, line 155: As far as I can see this method is only called once with an `XMethod`. Is there any reason why the parameter is not an `XMethod`?

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
PatrikScully: I removed the "interesting" word from variables and methods.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
PatrikScully: I removed the "interesting" word from variables and methods.

PatrikScully: Fixed the type, thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/util/CollectionAnalysis.java
```diff
@@ -0,0 +1,48 @@
...
+
+import edu.umd.cs.findbugs.ba.ClassMember;
+
+/**
+ * Utility class for analyzing collections.
+ */
+public final class CollectionAnalysis {
+
+    private CollectionAnalysis() {
+    }
```
JuditKnoll, line 33: Can you please add a comment that it's empty because it's a utility class, so the Sonar analysis won't mark it as problematic?

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MethodAnalysis.java
```diff
@@ -0,0 +1,45 @@
...
+import edu.umd.cs.findbugs.classfile.MethodDescriptor;
+import edu.umd.cs.findbugs.classfile.engine.bcel.FinallyDuplicatesInfoFactory;
+
+/**
+ * Utility class for method analysis.
+ */
+public final class MethodAnalysis {
+
+    private MethodAnalysis() {
+    }
```
JuditKnoll, line 32: Can you please add a comment that it's empty because it's a utility class, so the Sonar analysis won't mark it as problematic?

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MethodAnalysis.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Check if the location is duplicated in the method.
+     *
+     * @return {@code true} if the location is duplicated in the method, {@code false} otherwise
+     * @throws CheckedAnalysisException if an error occurs during the analysis
+     */
+    public static boolean isDuplicatedLocation(MethodDescriptor methodDescriptor, int pc) throws CheckedAnalysisException {
```
JuditKnoll, line 40: Can you please mention the finally block in the javadoc as the source of the location duplication?

JuditKnoll (changes requested):
Can you please update it to the latest master, where in the meantime some assert functions got introduced to `AbstractIntegrationTest` and use those? I mentioned them at your tests util functions.

JuditKnoll (changes requested):

JuditKnoll (approved):

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Added
 - New detector `ResourceInMultipleThreadsDetector` and introduced new bug type:
   - `AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD` is reported in case of unsafe resource access in multiple threads.
+- New detector `AtomicOperationsCombinedDetector` and introduced new bug types:
+  - `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC` is reported in case of combined atomic operations are not synchronized.
+  - `AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION` is reported when an atomic operation is not synchronized, but should be because of thread safety.
```
JuditKnoll, line 42: It looks like this Added block got here by mistake when merging, as there is already an Added section in this Unreleased part. Can you please move this entry there after updating master?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,322 @@
...
+        CFG cfg = classContext.getCFG(method);
+        ConstantPoolGen cpg = classContext.getConstantPoolGen();
+
+        for (Location location : cfg.orderedLocations()) {
+            InstructionHandle handle = location.getHandle();
+            Instruction instruction = handle.getInstruction();
+
+            if (instruction instanceof PUTFIELD) {
+                OpcodeStack stack = OpcodeStackScanner.getStackAt(classContext.getJavaClass(), method, handle.getPosition());
+                OpcodeStack.Item stackItem = stack.getStackItem(0);
```
JuditKnoll, line 146: Can you please check that there are enough item in the stack before getting an item?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,322 @@
...
+                    } else if (isAtomicOperationsCombined(stack)) {
+                        combinedAtomicFields.addAll(findFieldsInvolvedInAtomicOperations(location, callListDataflow));
+                    }
+                }
+            }
+        }
+    }
+
+    private Optional<XField> findFieldRequiringAtomicityCheck(OpcodeStack stack) {
+        if (stack.getStackDepth() > 1 && stack.getStackItem(0).getReturnValueOf() != null) {
```
JuditKnoll, line 226: Are you sure that the stack depth check should be `stack.getStackDepth() > 1`, when the requested item is the one on index 0?

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Added
 - New detector `SharedVariableAtomicityDetector` for new bug types `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE`, `AT_NONATOMIC_64BIT_PRIMITIVE` and `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` (See SEI CERT rules [VNA00-J](https://wiki.sei.cmu.edu/confluence/display/java/VNA00-J.+Ensure+visibility+when+accessing+shared+primitive+variables), [VNA02-J](https://wiki.sei.cmu.edu/confluence/display/java/VNA02-J.+Ensure+that+compound+operations+on+shared+variables+are+atomic) and [VNA05-J](https://wiki.sei.cmu.edu/confluence/display/java/VNA05-J.+Ensure+atomicity+when+reading+and+writing+64-bit+values)).
 - New detector `FindHiddenMethod` for bug type `HSM_HIDING_METHOD`. This bug is reported whenever a subclass method hides the static method of super class. (See [SEI CERT MET07-J] (https://wiki.sei.cmu.edu/confluence/display/java/MET07-J.+Never+declare+a+class+method+that+hides+a+method+declared+in+a+superclass+or+superinterface)).
+- New detector `ResourceInMultipleThreadsDetector` and introduced new bug type:
+  - `AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD` is reported in case of unsafe resource access in multiple threads.
```
JuditKnoll, line 13: These two lines got here when merging by accident. Can you please delete them?

JuditKnoll (changes requested):

CHANGELOG.md
PatrikScully: I removed them.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
PatrikScully: No, I think it's enough to check if depth > 0, fixed.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
PatrikScully: Thank you, I fixed this.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
...
 - New detector `FindHiddenMethod` for bug type `HSM_HIDING_METHOD`. This bug is reported whenever a subclass method hides the static method of super class. (See [SEI CERT MET07-J](https://wiki.sei.cmu.edu/confluence/display/java/MET07-J.+Never+declare+a+class+method+that+hides+a+method+declared+in+a+superclass+or+superinterface)).
+- New detector `AtomicOperationsCombinedDetector` and introduced new bug types:
+  - `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC` is reported in case of combined atomic operations are not synchronized.
+  - `AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION` is reported when an atomic operation is not synchronized, but should be because of thread safety.
```
JuditKnoll, line 16: Can you please update to the latest master and move this part to unreleased?

PatrikScully: Of course! It's done.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
```diff
@@ -43,6 +51,10 @@ public String getMethodSig() {
         return methodSig;
     }
 
+    public List<XField> getAttributes() {
+        return Collections.unmodifiableList(attributes);
```
JuditKnoll: As far as I can see, the items in attributes are not not modified. This wrapping into an unmodifiableList could be done when assigning the value to attributes in the constructor, not every time the getter is called.

Also, if `attributes` is null, `Collections.unmodifiableList` throws an NPE. Currently in the code it's not null, but there could be a null check to avoid issues in the future.

JuditKnoll (changes requested):
Can you please update to the latest master?

One small nitpicking, otherwise LGTM.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
PatrikScully: I updated the change and rebased the branch.

CHANGELOG.md
```diff
@@ -61,6 +61,12 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Add relevant CWE ids to bugs and refer the CWEs in the bug messages ([#3354](https://github.com/spotbugs/spotbugs/pull/3354)).
 - Replace `LOCAL_VARIABLE_UNKNOWN` with exact method name for `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` ([#3485](https://github.com/spotbugs/spotbugs/pull/3485))
 
+### Added
+
+- New detector `AtomicOperationsCombinedDetector` and introduced new bug types:
+  - `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC` is reported in case of combined atomic operations are not synchronized.
+  - `AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION` is reported when an atomic operation is not synchronized, but should be because of thread safety.
```
JuditKnoll: Right now this is under release 4.9.4. Since then there were a few releases. Can you please update to the latest master and move this to the added section there?

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
```diff
@@ -18,17 +18,25 @@
...
+
 public class Call {
     private final String className;
 
     private final String methodName;
 
     private final String methodSig;
 
-    public Call(String className, String methodName, String methodSig) {
+    private final List<XField> attributes;
```
JuditKnoll, line 33: The `attributes` field is not checked in `equals` and `hashCode`, and `CallList`'s merge depends on equality check, so Calls with different attributes might be merged.

JuditKnoll (changes requested):

CHANGELOG.md
PatrikScully: Done, moved the entry to the Unreleased section in commit c72cb93aa.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
PatrikScully: Done, attributes are now included in both equals() and hashCode() as of commit c72cb93aa.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,324 @@
...
+            BugPrototype bugPrototype = new BugPrototype(classContext, method, location);
+            InstructionHandle handle = location.getHandle();
+            Instruction instruction = handle.getInstruction();
+            int pc = handle.getPosition();
+
+            if (instruction instanceof MONITORENTER) {
+                synchronizedBlock = true;
+            } else if (instruction instanceof MONITOREXIT) {
+                synchronizedBlock = false;
+            } else if (instruction instanceof PUTFIELD && !synchronizedBlock && !MethodAnalysis.isDuplicatedLocation(methodDescriptor, pc)) {
```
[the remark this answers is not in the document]
PatrikScully: Good catch. Replaced the manual MONITORENTER/MONITOREXIT tracking entirely with LockDataflow, which is SpotBugs' proper dataflow analysis for lock tracking. This correctly handles nested synchronized blocks, exception handler control flow paths, and all edge cases that a simple counter would miss.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
```diff
@@ -43,18 +51,22 @@ public String getMethodSig() {
...
         Call other = (Call) obj;
         return this.className.equals(other.className) && this.methodName.equals(other.methodName)
-                && this.methodSig.equals(other.methodSig);
+                && this.methodSig.equals(other.methodSig) && this.attributes.equals(other.attributes);
     }
 
     @Override
     public int hashCode() {
-        return className.hashCode() + methodName.hashCode() + methodSig.hashCode();
+        return className.hashCode() + methodName.hashCode() + methodSig.hashCode() + attributes.hashCode();
```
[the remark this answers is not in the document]
PatrikScully: This was an intentional change requested by the reviewer. The attributes are part of the call identity for this detector's purposes, so including them in equals/hashCode is the correct behavior here.

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
     - `USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION` is reported when the backing collection of a lock can be altered by subclasses.
       (See [SEI CERT rule LCK00-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code) and [SEI CERT rule LCK04-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible))
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
+- New detector `AtomicOperationsCombinedDetector` and introduced new bug types:
+  - `AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC` is reported in case of combined atomic operations are not synchronized.
```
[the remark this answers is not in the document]
PatrikScully: Fixed in commit 92ea00069.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,324 @@
...
+            call.getAttributes().stream()
+                    .filter(fieldsForAtomicityCheck::contains)
+                    .forEach(involvedFields::add);
+        }
+        return involvedFields;
+    }
+
+    private void removePresynchronizedPrivateMethodCalls() {
+        localVariableInvocations.entrySet().removeIf(entry -> entry.getKey().isPrivate()
+                && unsynchronizedPrivateMethods.stream().noneMatch(privMethod -> equalMethods(entry.getKey(), privMethod)));
```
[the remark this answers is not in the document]
PatrikScully: Valid point. The unsynchronizedPrivateMethods set is only populated from the field-based detection path, so private methods operating only on atomic local variables could slip through. For now this is a known limitation since we have no concrete false positive test case to demonstrate the need. Will revisit if real-world usage surfaces a problem.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,324 @@
...
+
+    private Optional<XField> findFieldRequiringAtomicityCheck(OpcodeStack stack) {
+        if (stack.getStackDepth() > 0 && stack.getStackItem(0).getReturnValueOf() != null) {
+            return Optional.empty();
+        }
+        return IntStream.range(0, stack.getStackDepth())
+                .mapToObj(stack::getStackItem)
+                .map(OpcodeStack.Item::getXField)
+                .filter(fieldsForAtomicityCheck::contains)
+                .findFirst();
```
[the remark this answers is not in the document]
PatrikScully: Fixed. The early return was too broad. Instead of removing it entirely (which caused false positives when unrelated fields sit on the stack below the consumed range), the method now scans only stack items consumed by the current InvokeInstruction via consumeStack(cpg). This correctly detects autoboxed arguments like syncList.add(Integer.valueOf(1)) while ignoring fields that just happen to be on the stack for a later instruction.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,324 @@
...
+
+    private void analyzeFieldsForAtomicityViolations(ClassContext classContext, Method method) throws CheckedAnalysisException {
+        if (Const.CONSTRUCTOR_NAME.equals(method.getName()) || Const.STATIC_INITIALIZER_NAME.equals(method.getName()) || method.isSynchronized()) {
+            return;
+        }
+
+        CFG cfg = classContext.getCFG(method);
+        ConstantPoolGen cpg = classContext.getConstantPoolGen();
+        CallListDataflow callListDataflow = classContext.getCallListDataflow(method);
+        JavaClass javaClass = classContext.getJavaClass();
```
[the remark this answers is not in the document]
PatrikScully: Done, callListDataflow is now computed lazily, only when isAtomicOperationsCombined() is true.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,324 @@
...
+            Instruction instruction = handle.getInstruction();
+
+            if (instruction instanceof PUTFIELD) {
+                OpcodeStack stack = OpcodeStackScanner.getStackAt(classContext.getJavaClass(), method, handle.getPosition());
+                if (stack.getStackDepth() > 0) {
+                    OpcodeStack.Item stackItem = stack.getStackItem(0);
+                    if (isAtomicField(stackItem.getReturnValueOf())) {
+                        fieldsForAtomicityCheck.add(XFactory.createXField((FieldInstruction) instruction, cpg));
+                    }
+                }
```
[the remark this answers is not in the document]
PatrikScully: Acknowledged. Refactoring to a single-pass OpcodeStackDetector would fix the O(n^2) behavior but requires a significant restructuring of the detector. Deferring this to a follow-up since it is a performance optimization, not a correctness issue. Several other detectors in the codebase (e.g. FindNullDeref, FindReturnRef) use the same getStackAt pattern.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/CallListAnalysis.java
```diff
@@ -20,49 +20,84 @@
...
-                Call call = new Call(inv.getClassName(cpg), inv.getName(cpg), inv.getSignature(cpg));
+                OpcodeStack stack = OpcodeStackScanner.getStackAt(classContext.getJavaClass(), method, handle.getPosition());
+                List<XField> attributes = IntStream.range(0, stack.getStackDepth())
+                        .mapToObj(stackOffset -> stack.getStackItem(stackOffset).getXField())
+                        .filter(Objects::nonNull)
+                        .collect(Collectors.toList());
+
+                Call call = new Call(inv.getClassName(cpg), inv.getName(cpg), inv.getSignature(cpg), attributes);
                 callMap.put(handle, call);
             }
```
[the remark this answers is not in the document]
PatrikScully: Same situation as above. The getStackAt pattern in CallListAnalysis.buildCallMap is consistent with how other analysis factories in the codebase work. Optimizing this would be a cross-cutting change affecting all CallListDataflow consumers, so it is better handled separately.

spotbugs/etc/findbugs.xml
```diff
@@ -756,6 +758,8 @@
 
           <BugPattern abbrev="AT" type="AT_OPERATION_SEQUENCE_ON_CONCURRENT_ABSTRACTION" category="MT_CORRECTNESS" cweid="362"/>
           <BugPattern abbrev="AT" type="AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD" category="MT_CORRECTNESS" cweid="362"/>
+          <BugPattern abbrev="AT" type="AT_COMBINED_ATOMIC_OPERATIONS_ARE_NOT_ATOMIC" category="MT_CORRECTNESS"/>
+          <BugPattern abbrev="AT" type="AT_ATOMIC_OPERATION_NEEDS_SYNCHRONIZATION" category="MT_CORRECTNESS"/>
```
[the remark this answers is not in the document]
PatrikScully: Fixed, added cweid="362" to both new AT bug patterns in commit 92ea00069.

spotbugsTestCases/src/java/atomicMethods/UnsafeSynchronizedNavigableMap.java
```diff
@@ -0,0 +1,15 @@
+package atomicMethods;
+
+import java.util.Collections;
+import java.util.Map;
+
+public final class UnsafeSynchronizedNavigableMap {
+
+    private final Map<String, Integer> nums = Collections.synchronizedSortedMap(Collections.emptyNavigableMap());
+
```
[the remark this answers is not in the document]
PatrikScully: Fixed, changed to Collections.synchronizedNavigableMap() with NavigableMap field type in commit 92ea00069.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/CallListAnalysis.java
```diff
@@ -20,49 +20,84 @@
...
-    private static Map<InstructionHandle, Call> buildCallMap(CFG cfg, ConstantPoolGen cpg) {
+    private static Map<InstructionHandle, Call> buildCallMap(ClassContext classContext, Method method) {
         Map<InstructionHandle, Call> callMap = new HashMap<>();
+        CFG cfg;
+        try {
+            cfg = classContext.getCFG(method);
+        } catch (CFGBuilderException e) {
+            AnalysisContext.logError("Error generating derefs for " + method, e);
+            throw new IllegalStateException(e);
+        }
```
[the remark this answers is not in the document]
PatrikScully: This wrapping was introduced in an earlier commit in this PR. It follows the same pattern used in getDepthFirstSearch above it. Changing the error handling strategy for CallListAnalysis is out of scope for this change.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/Call.java
```diff
@@ -43,18 +51,22 @@ public String getMethodSig() {
...
     }
```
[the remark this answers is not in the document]
PatrikScully: Already addressed in a previous review round. This was an intentional change requested by the reviewer.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,423 @@
...
+                && unsynchronizedPrivateMethods.stream()
+                        .noneMatch(m -> m.getName().equals(entry.getKey().getName())
+                                && m.getSignature().equals(entry.getKey().getSignature())));
+    }
+
+    // --- Phase 4: Report accumulated bugs ---
+
+    private void reportFieldBugs() {
+        Set<XField> fieldsWithMultiMethodAccess = fieldAccessBugPrototypes.values().stream()
+                .flatMap(map -> map.entrySet().stream())
```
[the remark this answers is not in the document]
PatrikScully: The name was misleading. The set actually collects fields that have multiple accesses within at least one method (i.e. combined operations somewhere), not fields accessed from multiple methods. The logic itself is correct for the detector's purpose: if a field has combined operations in method A, then even a single access in method B should be synchronized. Renamed to fieldsWithCombinedAccess.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AtomicOperationsCombinedDetector.java
```diff
@@ -0,0 +1,423 @@
...
+
+    // --- Phase 3: Filter false positives from pre-synchronized private methods ---
+
+    private void removePresynchronizedPrivateMethodCalls() {
+        localVariableInvocations.entrySet().removeIf(entry -> entry.getKey().isPrivate()
+                && unsynchronizedPrivateMethods.stream()
+                        .noneMatch(m -> m.getName().equals(entry.getKey().getName())
+                                && m.getSignature().equals(entry.getKey().getSignature())));
+    }
+
```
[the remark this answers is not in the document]
PatrikScully: Known limitation, same as the previous review round. No concrete false negative test case exists to justify the additional complexity.

spotbugsTestCases/src/java/atomicMethods/UnsafeSynchronizedCollection.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Collection;
+import java.util.Collections;
+
+public final class UnsafeSynchronizedCollection {
+
+    private final Collection<Object> nums = Collections.synchronizedCollection(Collections.emptyList());
+
+    public void addAndPrintNumbers(Integer number) {
+        nums.add(number);
+        Integer[] numberCopy = nums.toArray(new Integer[0]);
```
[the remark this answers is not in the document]
PatrikScully: This compiles fine. Collection.toArray(T[]) accepts any array type. The test only needs to compile to bytecode for the detector to analyze it, runtime behavior is irrelevant.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ca/CallListAnalysis.java
```diff
@@ -20,49 +20,84 @@
...
             Instruction ins = handle.getInstruction();
 
             if (ins instanceof InvokeInstruction) {
                 InvokeInstruction inv = (InvokeInstruction) ins;
-                Call call = new Call(inv.getClassName(cpg), inv.getName(cpg), inv.getSignature(cpg));
+                OpcodeStack stack = OpcodeStackScanner.getStackAt(classContext.getJavaClass(), method, handle.getPosition());
+                List<XField> attributes = IntStream.range(0, stack.getStackDepth())
+                        .mapToObj(stackOffset -> stack.getStackItem(stackOffset).getXField())
+                        .filter(Objects::nonNull)
+                        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
PatrikScully: Same as the other CallListAnalysis comment, this is from an earlier commit in this PR and not related to the current changes.

PatrikScully:
The detector previously called `OpcodeStackScanner.getStackAt()` inside loops over CFG locations, once per PUTFIELD/PUTSTATIC and once per InvokeInstruction. Since `getStackAt` rescans bytecode from the method entry up to the target PC on every call, this resulted in O(n^2) complexity per method where n is the number of instructions.

Replaced this with a `MethodStackScanner` (extends `OpcodeStackDetector`) that walks each method's bytecode once and captures operand stack data at all relevant instruction sites into a map keyed by PC. The CFG iteration phase then looks up pre-computed results instead of rescanning. This brings the per-method cost down to O(n).

The same O(n^2) pattern exists in `CallListAnalysis.buildCallMap`, which also calls `getStackAt` in a loop. That one is left as-is since it is shared infrastructure used by other consumers beyond this detector.

PR #2925, opened 2024-04-08
Fetch Git history when performing a release

gtoison:
The release workflow does not fetch the Git history so looking for the previous tag only gets the latest tag and `requarks/changelog-action` says that it cannot update the Changelog (because from/to tags are the same)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-
+        with:
+          fetch-depth: 0
       - name: Get previous tag for the change log
         id: previousTag
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
         uses: requarks/changelog-action@v1
         with:
           token: ${{ github.token }}
-          toTag: ${{ github.ref }}
+          toTag: ${{ github.ref_name }}
           fromTag: ${{ env.previousTag }}
           writeToFile: true
         continue-on-error: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,8 +30,8 @@ jobs:
         uses: requarks/changelog-action@v1
         with:
           token: ${{ github.token }}
-          toTag: ${{ github.ref_name }}
-          fromTag: ${{ env.previousTag }}
+          toTag: ${{ env.previousTag }}
+          fromTag: ${{ github.ref_name }}
           writeToFile: true
         continue-on-error: true
 


JuditKnoll:
Just tried to run the release gha on my fork (see [running it](https://github.com/JuditKnoll/spotbugs/actions/runs/8602402519/job/23571891406) on `tag_master2` tag on branch `master`**), and I think there is one more issue here.
In the `Get previous tag for the change log` step, in the `$(git --no-pager tag --sort=creatordate --merged ${{ github.ref_name }} | tail -2 | head -1)` command the `${{ github.ref_name }}` is evaluated to the name of the tag, not the branch. In case of tags, the branch should be `${{ github.event.base_ref }}`.

** Please, don't let that bother you, that the build fails, it's because of the missing gpg key.

JuditKnoll:
I'm sorry, I tried to merge your change to my fork to try it out, not the other way around. I'll fix it in a few.

gtoison:
No problem, but I must admit that I'm even more confused now

JuditKnoll:
Tried this change on my fork (see the [run here](https://github.com/JuditKnoll/spotbugs/actions/runs/8603066484/job/23574068377)). I was experimenting with some things on my master, so it has a few other tags, but finding the previous tag seems to be successful. However, finding the commits between the latest and the previous tags still doesn't work properly. In this case there are two commits between the tags (not including either of the commits with the tags).
I think the problem may be because the tag range is between `previous-tag` and `refs/tags/latest-tag`, with the latest having the `refs/tags/` prefix, and the previous not.

hazendaz:
how about 'git describe --abbrev=0 --tags' to get the previous tag?  That would replace all of '$(git --no-pager tag --sort=creatordate --merged ${{ github.ref_name }} | tail -2 | head -1)'.

ThrawnCA (approved):

gtoison:
I tried `git describe --abbrev=0 --tags` but it gives me the latest tag, we want to get the previous one.

With the change in this PR this seems to work a bit better but still not quite there, in that run the previous tag is picked up correctly but it still says it cannot find any commit:
https://github.com/gtoison/spotbugs/actions/runs/8611439632/job/23598621569
```
Run requarks/changelog-action@v1
Using tag range: 4.8.4 to refs/tags/gtoison_test_2921_2
Error: Couldn't find any commits between latest and previous tags.
```

gtoison:
I've added your fixes @JuditKnoll

hazendaz:
> I tried `git describe --abbrev=0 --tags` but it gives me the latest tag, we want to get the previous one.
> 
> With the change in this PR this seems to work a bit better but still not quite there, in that run the previous tag is picked up correctly but it still says it cannot find any commit: https://github.com/gtoison/spotbugs/actions/runs/8611439632/job/23598621569
> 
> ```
> Run requarks/changelog-action@v1
> Using tag range: 4.8.4 to refs/tags/gtoison_test_2921_2
> Error: Couldn't find any commits between latest and previous tags.
> ```

Long term we should change the release process so we have a GHA that does all the work and requires nothing manual.  In doing so, that command does work, we have used it at my day job for years with maven in fully automated way on cloud.  Part of the process is to capture that first so its the last one released, then running the prepare, then looking at the diff between the commit tags for listing release notes.  Anyway food for thought as its rather manual + automation we have to remember every time.

hazendaz (approved):

PR #2937, opened 2024-04-10
Fixed FindDeadLocalStores to handle LOOKUPSWITCH and TABLESWITCH

gtoison:
A Java switch might be compiled into a LOOKUPSWITCH or TABLESWITCH. The corresponding BCEL classes both extends Select so we can use that class for the test.

PR #2979, opened 2024-05-06
Check for LOOKUPSWITCH or TABLESWITCH instructions

gtoison:
When looking for a switch the instruction might be `LOOKUPSWITCH` or `TABLESWITCH`. Look for `Select` which is the base class for both.
See #2782
See #2937


hazendaz:
@gtoison Can you move the change log up to the unreleased section?  I pushed out 4.8.5 on Friday.

hazendaz:
looks good otherwise.

hazendaz (approved):

PR #2986, opened 2024-05-15
Add 'VNA04-J. Ensure that calls to chained methods are atomic' detector with test cases

PatrikScully:
Method chaining allows multiple method calls on the same object in a single statement by having methods return the this reference. This lets the next method call be made on the return value of the previous one. While individual methods can be atomic, the chain itself is not. Therefore, callers must ensure proper locking to guarantee the entire chain of invocations is atomic, as relying on independently atomic methods is not sufficient.

See [https://wiki.sei.cmu.edu/confluence/display/java/VNA04-J.+Ensure+that+calls+to+chained+methods+are+atomic](https://wiki.sei.cmu.edu/confluence/display/java/VNA04-J.+Ensure+that+calls+to+chained+methods+are+atomic)

- [x] TODO for me: If https://github.com/spotbugs/spotbugs/pull/2919 merged, put the marked methods into a shared utility class


ThrawnCA (approved):

gtoison (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/CollectionAnalysis.java
```diff
@@ -0,0 +1,30 @@
...
+
+import edu.umd.cs.findbugs.ba.ClassMember;
+
+/**
+ * Utility class for analyzing collections.
+ */
+public final class CollectionAnalysis {
+
+    private CollectionAnalysis() {
+    }
```
JuditKnoll, line 15: Can you please add a comment that it's empty because it's a utility class, so the Sonar analysis won't mark it as problematic?

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/CollectionAnalysis.java
```diff
@@ -0,0 +1,30 @@
+package edu.umd.cs.findbugs.util;
```
JuditKnoll, line 1: Can you please add the SpotBugs license header here?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ResourceInMultipleThreadsDetectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    void testUnsafeFieldUsages() {
+        performAnalysis("commonResources/UnsafeFieldUsage.class",
+                "commonResources/Vehicle.class",
+                "commonResources/Vehicle$Builder.class");
+        assertBugTypeCount("AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD", 5);
+        assertBugInMethodAtLine("AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD", "UnsafeFieldUsage", "lambda$new$0", 8);
+        assertBugInMethodAtLine("AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD", "UnsafeFieldUsage", "lambda$new$0", 8);
```
JuditKnoll, line 32: This line is the same as the previous one.

JuditKnoll (changes requested):

JuditKnoll (approved):

hazendaz (approved):

hazendaz:
Should we wait on the todo issue to be resolved first (ie get the other PR in ready state and merged before finishing this off?)

JuditKnoll:
> Should we wait on the todo issue to be resolved first (ie get the other PR in ready state and merged before finishing this off?)

I don't think it's necessary to wait. The todo task is checked, the mentioned methods are moved to utility classes (MethodAnalysis, CollectionAnalysis, and some into MutableClasses) in both PRs, so they don't rely on each other anymore.

PR #3000, opened 2024-05-28
Adjust priority since relaxed mode reports even `IGNORED_PRIORITY`

gtoison:
Adjust the priority in case is was `>= IGNORED_PRIORITY` as it might be the case in "relaxed" mode.
I understand that this is expected because there's a similar adjustment at: https://github.com/spotbugs/spotbugs/blob/861f4324c1c62c99b87f5d63b9e636f8df990261/spotbugs/src/main/java/edu/umd/cs/findbugs/ProjectStats.java#L277

Not sure how to reproduce the problem and this seem fairly straightforward so I did not add a tautological test, let me know if you think this is required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,14 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Do not report BC_UNCONFIRMED_CAST for Java 21's type switches when the switch instruction is TABLESWITCH ([#2782](https://github.com/spotbugs/spotbugs/issues/2782))
 - Do not throw exception when inspecting empty switch statements ([#2995](https://github.com/spotbugs/spotbugs/issues/2995))
+- Adjust priority since relaxed mode reports even `IGNORED_PRIORITY` ([#2994]https://github.com/spotbugs/spotbugs/issues/2994)
+
 ## 4.8.5 - 2024-05-03
 ### Fixed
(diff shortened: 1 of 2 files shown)


ThrawnCA (approved):

hazendaz:
Looks like a very old bug with a few of those links that could still have something to test the condition.  Guess one was hitting mybatis spring boot project, and I don't recall this at all beyond my comments.  The fix seems good, though.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Do not report BC_UNCONFIRMED_CAST for Java 21's type switches when the switch instruction is TABLESWITCH ([#2782](https://github.com/spotbugs/spotbugs/issues/2782))
 - Do not throw exception when inspecting empty switch statements ([#2995](https://github.com/spotbugs/spotbugs/issues/2995))
+- Adjust priority since relaxed mode reports even `IGNORED_PRIORITY` ([#2994]https://github.com/spotbugs/spotbugs/issues/2994)
```
JuditKnoll, line 12: It looks like there is one `(` missing for the correct formatting after the `]`. Can you please add it?

gtoison: Good catch @JuditKnoll 
I have made #3012 to fix other errors I've made

PR #3006, opened 2024-06-06
Removed log4j2 prefer() dependency constraint

gtoison:
The constraint causes gradle to add log4j-core-2.22.0.jar and log4j-core-2.23.1.jar to the distribution.
Removed the constraint so only the latter version is included.


hazendaz (approved):

stu3139 (approved):

stu3139 (approved):

stu3139 (changes requested):
Compatibility

PR #3012, opened 2024-06-13
Incorrect formatting for changelog links

gtoison:
Apparently I cannot write these changelog entries reliably... sorry for the mess!


JuditKnoll (approved):
Thanks for the fix 😄

Issue #2995, opened 2024-05-22
Index out of bounds for empty switch


zbynek:
Since update to 4.8.4 I'm getting this
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0
      At edu.umd.cs.findbugs.SwitchHandler.enterSwitch(SwitchHandler.java:90)
      At edu.umd.cs.findbugs.detect.FindDeadLocalStores.analyzeMethod(FindDeadLocalStores.java:283)
      At edu.umd.cs.findbugs.detect.FindDeadLocalStores.visitClassContext(FindDeadLocalStores.java:198)
      At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
      At edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$1(FindBugs2.java:1108)
      At java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
      At edu.umd.cs.findbugs.CurrentThreadExecutorService.execute(CurrentThreadExecutorService.java:86)
      At java.base/java.util.concurrent.AbstractExecutorService.invokeAll(AbstractExecutorService.java:242)
      At edu.umd.cs.findbugs.FindBugs2.analyzeApplication(FindBugs2.java:1118)
      At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:309)
      At edu.umd.cs.findbugs.FindBugs.runMain(FindBugs.java:395)
      At edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1231)
 ```
 when analyzing a generaated class that contains
 ```
 switch(jjstateSet[--i]) {
   default : break;
}
```
and
```
switch(jjmatchedKind) {
      default :
         break;
   }
```
It could make sense to optimize the code generator, but I think it's still worth fixing the OutOfBounds problem in SpotBugs.

Possibly related to https://github.com/spotbugs/spotbugs/pull/2828

gtoison:
Thanks for reporting the issue, I tried reproducing it but it does not seem to crash for me with javac (OpenJDK21U-jdk_x64_windows_hotspot_21_35)
What compiler are you using?

zbynek:
Java 11. I also checked with a minimal class that has just one empty switch and couldn't reproduce it. I'll try to create a minimal testcase (currently the problematic class has 450 lines).

zbynek:
[...]

PR #3013, opened 2024-06-14
Update spotbugs plugin to 6.0.16

zbynek:
Replacement for https://github.com/spotbugs/spotbugs/pull/3009 fixing a build failure.
The latest Gradle plugin fails to find the stylesheet if the `spotbugs` gradle configuration was changed to include a *project* instead of an *external jar* file.

While this could be also fixed in the Gradle plugin, I don't think it makes sense to add complexity to the plugin for a very unique use-case of running SpotBugs on self.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 plugins {
   id "org.sonarqube" version "5.0.0.4638"
   id "org.gradle.crypto.checksum" version "1.4.0"
-  id "com.github.spotbugs" version "6.0.15"
+  id "com.github.spotbugs" version "6.0.16"
   id "io.github.gradle-nexus.publish-plugin" version "2.0.0"
 }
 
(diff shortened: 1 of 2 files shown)


hazendaz:
Thanks @zbynek

zbynek:
@hazendaz what is the plan for 4.8.6 release? (This was the last open item in the milestone.)

hazendaz:
Hi @zbynek I wanted to do a bit of code cleanup so I opened #3014 then was going to release.   Only concern noted there was some protected as unsure if any callers were using the internals in some odd way. So I'll likely relax that part back but likely take it.  However, I'm going to be unavailable after tomorrow until next week.   So its not critical that even makes the cut.  We could just release as is and follow that later.  If we do that, I can release tonight and get the maven plugin out shortly after but at the moment I cannot guarantee I can do so.  So let me think about that a bit.   If its is critical you have this out sooner than later I'll try my best tonight. Let me know what you think.

zbynek:
Hi, no rush, finishing the cleanup before release sounds reasonable. My only issue with the current version is the slightly noisy log because of #2995 but that can certainly wait a few weeks.

PR #3014, opened 2024-06-16
[cleanup] Rough pass through code cleaning up sonarlint, initialization issues, and others

hazendaz:
This effort was manually applied.  Easily this could be double the size as I gave up.  So I plan to do this a few times over next month or so.  Most likely we cut a release in between efforts like this since a lot to look through.  There are 101 files in this set.  The bulk are very straight forwards and this is very long overdue especially when point of library is to tell users to write better code yet our code base looks like we missed the point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2024-??-??
+### Cleanup
+- Rough pass through code cleaning up sonarlintt, initialization issues, and others including one deprecation and regex changes ([#3014](https://github.com/spotbugs/spotbugs/issues/3014))
 
 ## 4.8.6 - 2024-06-17
 ### Fixed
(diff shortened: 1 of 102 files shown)


hazendaz:
I do not believe there are any breaking changes in regards here but if there is any concern their might be, this could move to 4.9.0.

JuditKnoll (approved):
This PR restricts the visibility of several functions and fields. I haven't checked whether any of these are accessible outside of SpotBugs and used in any of the plugins. @hazendaz, did you check this?

Otherwise it is good to go in the next version, LGTM. Thank you for your work!

zbynek:
Does this actually reduce any visibilities? Most of what looks like visibility change is removing `public` from interfaces where it has no effect anyway. In a few cases constructors of abstract classes are made protected, but that's also not changing the scope where they can be used (is there some weird reflection trick that works with public but not protected?)
Only (?) field where visibility did change is `uiNameToCodeMap` but it's not used by any projects on GitHub other than forks of this repo: https://github.com/search?q=uiNameToCodeMap&type=code
Another potentially breaking change is removing the type parameter from `Plugin::addFindbugsMain`, but that also seems unused.

hazendaz:
Lets do this just for sake of doing it.  I'll hold off here and go to 4.9.0 with this.  I don't think it will break anything and thank you @zbynek for poking around.  Instead I'll cut 4.8.6 as-is now so we can get real stuff out.  While this has merits of getting code cleaned up it doesn't really do anything for us other than possibly fix thread interruption issues so that alone is worth 4.9.0.

hazendaz:
I'm going to close this for now.  I'll be away for a week or so.  When I get back to this, I'll break this up into parts so the non issues will be far separated so it can more easily be reviewed by users if something breaking is in consideration as pointed out there.  Thanks.

hazendaz:
finally getting back to this and reworking it to bring in items separately so its not overloading what its doing and easier to understand.

PR #3033, opened 2024-07-03
Add new detector for 'LCK00-J' and 'LCK04-J'

norbo03:
### LCK00-J. Use private final lock objects to synchronize classes that may interact with untrusted code
Shared mutable variables can be synchronized using either method or block synchronization, both of which use the object's intrinsic lock as a monitor. An attacker could exploit this by holding the intrinsic lock indefinitely, causing a DoS through contention and deadlock. To prevent such situations, ensure that internal locking mechanisms are not exposed.

### LCK04-J. Do not synchronize on a collection view if the backing collection is accessible
When wrapper implementations are used for locking, they share similar vulnerabilities since the backing collection's lock can be leaked in several ways. To avoid these situations, ensure that the backing collection itself is used for synchronization.

See [LCK00-J. Use private final lock objects to synchronize classes that may interact with untrusted code](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code)
and [LCK04-J. Do not synchronize on a collection view if the backing collection is accessible](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible)

### Known limitations of the detector

The following situations are not handled by the detector:

- inner classes: The detector cannot handle inner classes for _LCK00-J_ and _LCK04-J_ neither
- primitive collections and reference arrays: For _LCK04-J_ the detector doesn't support primitive collections like `int[]` or reference arrays such as `Object[]` as backing collections.
- chained wrappers: For _LCK04-J_ the detector cannot detect exposure of backing collections through chaining. For example, it cannot recognize the exposure of the backing collection of a backing collection etc. :
```java
    public List<Integer> backingColl1= new ArrayList<>();
    private List<Integer> backingColl2 = Collections.synchronizedList(backingColl1);
    private List<Integer> view = Collections.synchronizedList(backingColl2);
```
- inheritance exposure: For _LCK04-J_ the detector fails to recognize cases where the collection wrapper, the backing collection, or the exposure are separated through multiple layers of inheritance.


ThrawnCA (approved):

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationWithAccessibleBackingCollectionsTest.java
```diff
@@ -0,0 +1,255 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.jupiter.api.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+public class FindImproperSynchronizationWithAccessibleBackingCollectionsTest extends AbstractIntegrationTest {
```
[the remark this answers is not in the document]
norbo03: Yes, of course. Thank you for bringing it to my attention.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperMixedSynchronizationTest.java
```diff
@@ -0,0 +1,153 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.jupiter.api.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+public class FindImproperMixedSynchronizationTest extends AbstractIntegrationTest {
```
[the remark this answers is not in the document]
norbo03: Yes, of course. Thank you for bringing it to my attention.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,807 @@
...
+                } else {
+                    if (xField.isPublic()) {
+                        badBackingCollections.add(xField);
+                    }
+
+                    if (xField.isProtected() && !getThisClass().isFinal()) {
+                        inheritableBackingCollections.add(xField);
+                    } else if (isPackagePrivate(xField)) {
+                        inheritableBackingCollections.add(xField);
+                    }
```
JuditKnoll, line 247: As far as I can see the body of these two `if`s are the same. Can you please combine them?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,807 @@
...
+     * To ensure no false positives, we compare only the outer class name if the lock object is contained in an inner class.
+     * Only checking equality is enough, since the lock object is a field declared in a class. That means if
+     * the declaring class of the lock object is the same as the current class, the lock object is not inherited.
+     * @param lockObject the lock object to check for inheritance
+     * @return true if the lock object is inherited from the current class, false otherwise
+     * @throws ClassNotFoundException if the class of the lock object cannot be found
+     */
+    private boolean isInherited(XField lockObject) throws ClassNotFoundException {
+        return !getDottedClassName().equals(lockObject.getClassName().split("\\$")[0]);
+    }
```
JuditKnoll, line 461: `ClassNotFoundException` cannot be thrown from this method implementation. Can you please remove it from the signature and the javadoc? Also, if I'm not mistaken, this may have an impact on the callsites as well (the try blocks become unnecessary or at least the body of them can be minimized).

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,807 @@
...
+                    .addMethod(synchronizedMethod)
+                    .addField(lock));
+        }
+    }
+
+    private void reportAccessibleBackingCollectionBug(XField lock, MultiMap<XField, BugInfo> possibleBugs,
+            XField backingCollection,
+            MultiMap<XField, XMethod> collectionAccessors) {
+        if (collectionAccessors.containsKey(backingCollection)) {
+            String exposingMethods = buildMethodsMessage(collectionAccessors.get(backingCollection));
```
JuditKnoll, line 731: This variable hides the field with the same name. Can you please rename it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,807 @@
...
+        for (JavaClass declaringClass : getThisClass().getSuperClasses()) {
+            for (Method possibleAccessorMethod : declaringClass.getMethods()) {
+                if (possibleAccessorMethod.getCode() == null) {
+                    continue;
+                }
+
+                if (Arrays.asList(ownMethods).contains(possibleAccessorMethod)
+                        || isInitializerMethod(possibleAccessorMethod.getName())) {
+                    continue;
+                }
```
JuditKnoll, line 589: Can you please combine these two `if`s, as the body of these are the same?

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperMixedSynchronizationTest.java
```diff
@@ -0,0 +1,153 @@
...
+import static org.hamcrest.Matchers.hasItem;
+
+class FindImproperMixedSynchronizationTest extends AbstractIntegrationTest {
+    private static final String OBJECT_BUG = "US_UNSAFE_OBJECT_SYNCHRONIZATION";
+    private static final String ACCESSIBLE_OBJECT_BUG = "US_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION";
+    private static final String INHERITABLE_OBJECT_BUG = "US_UNSAFE_INHERITABLE_OBJECT_SYNCHRONIZATION";
+    private static final String EXPOSED_LOCK_OBJECT_BUG = "US_UNSAFE_EXPOSED_OBJECT_SYNCHRONIZATION";
+    private final String BAD_BACKING_COLLECTION = "USBC_UNSAFE_SYNCHRONIZATION_WITH_BACKING_COLLECTION";
+    private final String ACCESSIBLE_BACKING_COLLECTION = "USBC_UNSAFE_SYNCHRONIZATION_WITH_ACCESSIBLE_BACKING_COLLECTION";
+    private final String INHERITABLE_BACKING_COLLECTION = "USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION";
```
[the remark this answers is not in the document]
norbo03: No. That was a mistake on my part.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperMixedSynchronizationTest.java
```diff
@@ -0,0 +1,153 @@
...
+        assertNumOfBugs(0, INHERITABLE_BACKING_COLLECTION);
+    }
+
+
+    private void assertNumOfBugs(int number, String bugType) {
+        final BugInstanceMatcher bugTypeMatcher = new BugInstanceMatcherBuilder().bugType(bugType).build();
+        assertThat(getBugCollection(), containsExactly(number, bugTypeMatcher));
+    }
+
+    private void assertObjectBugExactly(String clazz, String method, String field) {
```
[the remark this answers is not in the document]
norbo03: Yes, I will merge them ASAP.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
```diff
@@ -0,0 +1,418 @@
...
+                "synchronizationLocks/privateFinalLocks/UnsafeMethodSynchronizationWithClassExposingACollectionOfItself.class",
+                "synchronizationLocks/privateFinalLocks/UnsafeMethodSynchronizationWithClassExposingAMapOfItself.class",
+                "synchronizationLocks/privateFinalLocks/UnsafeMethodSynchronizationWithClassExposingMixedGenericOfItself.class");
+
+        assertNumOfBugs(5, METHOD_BUG);
+        assertNumOfBugs(0, STATIC_METHOD_BUG);
+        assertNumOfBugs(0, OBJECT_BUG);
+        assertNumOfBugs(0, ACCESSIBLE_OBJECT_BUG);
+        assertNumOfBugs(0, INHERITABLE_OBJECT_BUG);
+        assertNumOfBugs(0, EXPOSED_LOCK_OBJECT_BUG);
```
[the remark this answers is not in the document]
norbo03: Yes, I will add those checks too in every test case.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: You are right, this exception cannot be thrown from the function.
Generally, the callsites would also be affected, however, in these cases, other functions could throw the same exception.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: Yes!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    private void assertNumOfBugs(int number, String bugType) {
+        final BugInstanceMatcher bugInstance = new BugInstanceMatcherBuilder().bugType(bugType).build();
+        assertThat(getBugCollection(), containsExactly(number, bugInstance));
+    }
+
+    private void assertMethodBugExactly(String clazz, String method) {
```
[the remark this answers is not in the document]
norbo03: Yes, I will merge them ASAP.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationWithAccessibleBackingCollectionsTest.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    @Test
+    public void testSafeSynchronizationOnCollectionView() {
+        performAnalysis("synchronizationLocks/collectionViews/SafeSynchronizationOnCollectionView.class");
+
+        assertZeroBadSyncBugs();
+    }
+
+    private void assertPublicBackingCollectionBugExactly(String clazz, String method, String view) {
```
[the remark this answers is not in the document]
norbo03: Yes, I will merge them ASAP.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationWithAccessibleBackingCollectionsTest.java
```diff
@@ -0,0 +1,255 @@
...
+    private static final String ACCESSIBLE_BACKING_COLLECTION = "USBC_UNSAFE_SYNCHRONIZATION_WITH_ACCESSIBLE_BACKING_COLLECTION";
+    private static final String INHERITABLE_BACKING_COLLECTION = "USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION";
+
+    @Test
+    void testUnsafeSynWithAccessiblePublicBackingCollectionInPlace() {
+        performAnalysis("synchronizationLocks/collectionViews/UnsafeSyncWithAccessiblePublicBackingCollectionInPlace.class");
+
+        assertNumOfBugs(26, BAD_BACKING_COLLECTION);
+        assertNumOfBugs(0, ACCESSIBLE_BACKING_COLLECTION);
+        assertNumOfBugs(0, INHERITABLE_BACKING_COLLECTION);
```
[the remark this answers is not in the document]
norbo03: I added the other checks too.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: I combined them, thank you!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,800 @@
...
+        if (xMethod.isPublic() && xMethod.isSynchronized()) {
+            if (xMethod.isStatic()) {
+                bugReporter.reportBug(new BugInstance(
+                        this, "US_UNSAFE_STATIC_METHOD_SYNCHRONIZATION", NORMAL_PRIORITY)
+                        .addClassAndMethod(this));
+            }
+            synchronizedMethods.add(xMethod);
+        }
+
+        SignatureParser signature = new SignatureParser(xMethod.getSignature());
```
JuditKnoll, line 195: The `signature` field is only used first in line 205. Can you please move this field closer to the first usage of it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,800 @@
...
+        if ("readResolve".equals(obj.getName()) && "()Ljava/lang/Object;".equals(obj.getSignature())
+                && Subtypes2.instanceOf(getThisClass(), "java.io.Serializable")) {
+            return;
+        }
+
+        if (!(xMethod.isStatic() || xMethod.isPublic())) {
+            return;
+        }
```
JuditKnoll, line 203: Can you please change the order of these two `if`s? It would make the code a tiny bit more efficient.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: Yes, they can be simplified.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
```diff
@@ -0,0 +1,453 @@
...
+        assertNumOfBugs(0, ACCESSIBLE_BACKING_COLLECTION);
+        assertNumOfBugs(0, INHERITABLE_BACKING_COLLECTION);
+    }
+
+    private void assertNumOfBugs(int number, String bugType) {
+        final BugInstanceMatcher bugInstance = new BugInstanceMatcherBuilder().bugType(bugType).build();
+        assertThat(getBugCollection(), containsExactly(number, bugInstance));
+    }
+
+    private void assertBugExactly(String bugType, String clazz, String method) {
```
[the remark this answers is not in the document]
norbo03: Yes, there are cases when no field is added to the buginstance (in the case of method bugs). I will rename it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationWithAccessibleBackingCollectionsTest.java
```diff
@@ -0,0 +1,282 @@
...
+        assertBugExactly(INHERITABLE_BACKING_COLLECTION, "synchronizationLocks.collectionViews.UnsafeSyncWithExposedPackagePrivateBackingCollections",
+                "doStuff3", "view3"); /* Backed by: view1 -> collection1 */
+        assertBugExactly(INHERITABLE_BACKING_COLLECTION, "synchronizationLocks.collectionViews.UnsafeSyncWithExposedPackagePrivateBackingCollections",
+                "doStuff4", "view4"); /* Backed by: view4 -> collection2 */
+        assertBugExactly(INHERITABLE_BACKING_COLLECTION, "synchronizationLocks.collectionViews.UnsafeSyncWithExposedPackagePrivateBackingCollections",
+                "doStuff5", "view5"); /* Backed by: view5 -> collection3 */
+    }
+
+    @Test
+    public void testSafeSynchronizationOnCollectionView() {
```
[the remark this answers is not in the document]
norbo03: Yes! Thank you for bringing it to my attention, I have missed that one.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,800 @@
...
+        for (JavaClass declaringClass : getThisClass().getSuperClasses()) {
+            for (Method possibleAccessorMethod : declaringClass.getMethods()) {
+
+                if (possibleAccessorMethod.getCode() == null) {
+                    continue;
+                }
+
+                if (isInitializerMethod(possibleAccessorMethod.getName())) {
+                    continue;
+                }
```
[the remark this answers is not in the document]
norbo03: Yes, ofc.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,800 @@
...
+            XField lock = stack.getStackItem(0).getXField();
+
+            if (lock != null) {
+                XMethod xMethod = getXMethod();
+                try {
+                    if (isInherited(lock)) {
+                        inheritedLockObjects.put(lock, xMethod);
+                        if (isCollection(lock)) {
+                            inheritedCollectionLockObjects.add(lock, new BugInfo(this));
+                            try {
```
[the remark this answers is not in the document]
norbo03: I will restructure the `try` blocks to remove nested ones. I can simply handle both try blocks inside the `analyzeAssignmentsInHierarchy` with a minimal body. However, I have already restructured the other try block, so technically there are no nested `try` blocks, but I am still going to move them inside the function, because this way the driver code will be cleaner.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: Yes, definitively.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: Yes, I will optimize them.

JuditKnoll (changes requested):

norbo03:
> Can you please update it to the latest master, where in the meantime some assert functions got introduced to `AbstractIntegrationTest` and use those? I mentioned them at your test util functions.

Thank you very much for the comments! I will update to the latest master and bump my tests.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,801 @@
...
+                } catch (CFGBuilderException e) {
+                    AnalysisContext.logError("Error building CFG", e);
+                }
+            }
+        }
+    }
+
+    private void checkLockUsageInHierarchy(XField field, XMethod exposingMethod) {
+        if (isInherited(field)) {
+            if (declaredInheritedFieldAccessingMethods.get(field).isEmpty()) {
```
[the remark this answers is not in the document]
norbo03: Yes of course.

JuditKnoll (changes requested):
The build is failing because of some indentation issues. Can you please run gradlew spotlessApply to resolve them?

norbo03:
@JuditKnoll I fixed the build. Thank you!

JuditKnoll (changes requested):

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperMixedSynchronizationTest.java
```diff
@@ -0,0 +1,77 @@
...
+        assertBugTypeCount(OBJECT_BUG, 1);
+        assertBugTypeCount(ACCESSIBLE_OBJECT_BUG, 2);
+        assertBugTypeCount(INHERITABLE_OBJECT_BUG, 3);
+        assertBugTypeCount(EXPOSED_LOCK_OBJECT_BUG, 1);
+        assertBugTypeCount(BAD_BACKING_COLLECTION, 3);
+        assertBugTypeCount(ACCESSIBLE_BACKING_COLLECTION, 1);
+        assertBugTypeCount(INHERITABLE_BACKING_COLLECTION, 1);
+
+        assertBugInMethodAtField(BAD_BACKING_COLLECTION, "synchronizationLocks.MixedBadSynchronization", "doStuff1",
+                "view1"); /* Backed by: view1 -> collection1 */
```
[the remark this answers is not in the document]
norbo03: Oh, that adjustment is great! It will make my test much more readable.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,799 @@
...
+        WRAPPER_IMPLEMENTATIONS.put(new MethodDescriptor(juLists, "subList", "(II)Ljava/util/List;", false), 2);
+
+        IMMUTABLE_RETURNERS.addAll(Arrays.asList("Ljava/util/Arrays$ArrayList;",
+                "Ljava/util/Collections$UnmodifiableList;",
+                "Ljava/util/Collections$UnmodifiableMap;",
+                "Ljava/util/Collections$UnmodifiableSortedMap;",
+                "Ljava/util/Collections$UnmodifiableSet;",
+                "Ljava/util/Collections$UnmodifiableNavigableMap;",
+                "Ljava/util/Collections$UnmodifiableSortedSet;",
+                "Ljava/util/Collections$UnmodifiableNavigableSet;"));
```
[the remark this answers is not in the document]
norbo03: I will check if any of them can be added here too. However, at first glance, reusing them could pose some risks, as `MutableClasses.KNOWN_IMMUTABLE_CLASSES` was defined as private; I assume this was an intentional design choice.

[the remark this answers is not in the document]
norbo03: @JuditKnoll It appears that there are no additional classes from the standard library that I can utilize here. To reuse these values, we need to either change the modifier on the constant or refactor it later to place the constant in a globally accessible location. However, I believe the detectors typically require only a subset of these string values to function effectively, so the constants must be constructed or filtered at the class they are used in.

Can you please explain where should I use the utility methods you mentioned earlier? I am currently working with the signature that came from the stack items. Are you suggesting that it should be the other way around, like storing the class names instead of signatures and converting the found signature to the class name? Maybe, that might just simply work!

CHANGELOG.md
```diff
@@ -46,6 +46,19 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
...
+    - `US_UNSAFE_INHERITABLE_OBJECT_SYNCHRONIZATION` is reported when the lock used for synchronization is can be altered by subclasses,
+    - `US_UNSAFE_EXPOSED_OBJECT_SYNCHRONIZATION` is reported when the lock used for synchronization is later exposed in the subclasses.
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_BACKING_COLLECTION` is reported when the backing collection of a lock is visible from the outside,
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_ACCESSIBLE_BACKING_COLLECTION` is reported when the backing collection of a lock is made accessible, with methods that update or return the lock, to the outside,
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION` is reported when the backing collection of a lock can be altered by subclasses.
+      (See [SEI CERT rule LCK00-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code#:~:text=LCK00%2DJ%2DEX2,lock%20object%20inadvertently.) and [SEI CERT rule LCK04-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible))
+
```
JuditKnoll, line 61: There is already an Added block in the beginning of this release. Can you please merge the two?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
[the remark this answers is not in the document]
JuditKnoll: > I will check if any of them can be added here too. However, at first glance, reusing them could pose some risks, as MutableClasses.KNOWN_IMMUTABLE_CLASSES was defined as private; I assume this was an intentional design choice.

I think if necessary, the visibility modifier could be modified to reuse it in other detectors, if it would be an unmodifiable collection.

> It appears that there are no additional classes from the standard library that I can utilize here. 

What about the classes not from the JSL (e.g. from the `com.google.common.collect` package)?

> Can you please explain where should I use the utility methods you mentioned earlier? I am currently working with the signature that came from the stack items. Are you suggesting that it should be the other way around, like storing the class names instead of signatures and converting the found signature to the class name? Maybe, that might just simply work!

Yes, I meant converting the signatures to class names, as the collection I mentioned contains the class names and not the signature. But if you keep using the collection in the class, then it's better to not to convert.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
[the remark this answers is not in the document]
norbo03: @JuditKnoll I think I will keep the collection from the class for now.

I did not check any classes outside of the JSL, like classes from the `com.google.common.collect` package, as I think they were out of the scope of the initial problem/issue.

While incorporating additional classes from other packages could potentially improve the detector, doing so would require further testing and development. Unfortunately, I can't take on that work at the moment. I suggest proceeding with this PR as is, and we can revisit these improvements later if needed. It would also be beneficial to gather feedback from other projects before making further changes. What do you think?

ThrawnCA (approved):

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,798 @@
...
+    private static Integer getStackOffset(MethodDescriptor methodDescriptor) {
+        return WRAPPER_IMPLEMENTATIONS.get(methodDescriptor);
+    }
+
+    private static boolean isImmutableReturner(String signature) {
+        return IMMUTABLE_RETURNERS.contains(signature);
+    }
+
+    @Override
+    public void visit(Method obj) {
```
JuditKnoll, line 173: I think visiting synthethic methods can be skipped.

JuditKnoll (changes requested):
Can you please update to the latest master?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,798 @@
...
+
+                if (updatedField.isPublic()) {
+                    return;
+                }
+                declaredLockAccessors.add(updatedField, updateMethod);
+            }
+        }
+
+        if (seen == Const.ARETURN) {
+            OpcodeStack.Item returnValue = stack.getStackItem(0);
```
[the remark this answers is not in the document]
norbo03: Yes, ofc. Thank you for pointing this out. I missed this one!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: Yes, they can be skipped.

ThrawnCA (approved):

norbo03:
@JuditKnoll I've addressed all the feedback I could find, but GitHub still shows a requested change. Are there any outstanding comments or suggestions I might have missed?

JuditKnoll:
> @JuditKnoll I've addressed all the feedback I could find, but GitHub still shows a requested change. Are there any outstanding comments or suggestions I might have missed?

Oh, sorry, it's absolutely me, I missed the change here. I'll review it properly next week.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,803 @@
...
+            } catch (ClassNotFoundException e) {
+                AnalysisContext.reportMissingClass(e);
+            }
+            declaredInheritedFieldAccessingMethods.add(field, exposingMethod);
+        }
+    }
+
+    private Set<XMethod> findExposureInHierarchy(XField lockObject) throws ClassNotFoundException {
+        Set<XMethod> unsafeMethods = new HashSet<>();
+        Method[] ownMethods = getThisClass().getMethods();
```
JuditKnoll: As far as I can see, this field is only used in line 586 like this inside a nested for cycle:
```java
Arrays.asList(ownMethods).contains(possibleAccessorMethod)
```
The field could be an ArrayList type field, to not to create the ArrayList from the array every cycle, just once.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,803 @@
...
+                    }
+                }
+            }
+
+        }
+    }
+
+    private void analyzeAssignmentsInHierarchy(XField collectionLockObject) {
+        String lockDeclaringClassName = collectionLockObject.getClassName();
+        Method[] ownMethods = getThisClass().getMethods();
```
JuditKnoll: As far as I can see, this field is only used in line 555 like this inside a nested for cycle:

```java
if (ownMethods.length > 0 && Arrays.asList(ownMethods).contains(method) || method.getCode() == null) {
```
The field could be an ArrayList type field, to not to create the ArrayList from the array every cycle, just once.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,803 @@
...
+                    exposingMethods.add(xMethod);
+                }
+            }
+        }
+
+    }
+
+    @Override
+    public void visit(Field obj) {
+        XField xField = getXField();
```
JuditKnoll, line 237: Synthetic fields and other generated fields could be skipped. The `MemberUtils.isUserGenerated()` functions handle this.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,803 @@
...
+                    .addField(lock)
+                    .addReferencedField(FieldAnnotation.fromXField(assignedBackingCollection))
+                    .addSourceLine(sourceLineAnnotation);
+            for (String message : extraMessages) {
+                bugInstance.addString(message);
+            }
+
+            return bugInstance;
+        }
+    }
```
JuditKnoll: As far as I can see, this function is called three times, either with nothing as `extraMessages`, or with one item.
IMO it would make a code slightly easier to read, if it would look like something like this:

```suggestion
        private BugInstance createBugInstance(BytecodeScanningDetector detector, String bugType, XField lock, XField assignedBackingCollection) {
            return new BugInstance(detector, bugType, NORMAL_PRIORITY)
                    .addClassAndMethod(inMethod)
                    .addField(lock)
                    .addReferencedField(FieldAnnotation.fromXField(assignedBackingCollection))
                    .addSourceLine(sourceLineAnnotation);
        }
    }
    
    private BugInstance createBugInstance(BytecodeScanningDetector detector, String bugType, XField lock, XField assignedBackingCollection, String extraMessage) {
            BugInstance bugInstance = createBugInstance(detector, bugType, lock, assignedBackingCollection);
            bugInstance.addString(extraMessage);

            return bugInstance;
        }
    }
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,803 @@
...
+        return WRAPPER_IMPLEMENTATIONS.get(methodDescriptor);
+    }
+
+    private static boolean isImmutableReturner(String signature) {
+        return IMMUTABLE_RETURNERS.contains(signature);
+    }
+
+    @Override
+    public void visit(Method obj) {
+        if (obj.isSynthetic()) {
```
JuditKnoll: Can you please use one of the `MemberUtils.isUserGenerated()` functions here instead (with negating it ofc)? That one handles some additional generated methods, which could be skipped.

JuditKnoll (changes requested):
Just some really small requests, otherwise LGTM.

norbo03:
@JuditKnoll Thanks for the review! I’ve made the suggested changes. When you have a moment, could you take another look?

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
JuditKnoll: The `MemberUtils.isUserGenerated()` function returns true, if the code was written by a user, programmer and it's not synthetic or otherwise generated. Can you please negate the condition for the early return?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
JuditKnoll: I'm sorry, this could be a simple `List<Method>` type, it doesn't need to be wrapped into an `ArrayList`. Can you please modify it like this?
```java
List<Method> ownMethods = Arrays.asList(getThisClass().getMethods());
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
JuditKnoll: I'm sorry, this could be a simple `List<Method>` type, it doesn't need to be wrapped into an `ArrayList`. Can you please modify it like this?
```java
List<Method> ownMethods = Arrays.asList(getThisClass().getMethods());
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -785,16 +790,18 @@ private BugInfo(BytecodeScanningDetector detector) {
...
+            return new BugInstance(detector, bugType, NORMAL_PRIORITY)
                     .addClassAndMethod(inMethod)
                     .addField(lock)
                     .addReferencedField(FieldAnnotation.fromXField(assignedBackingCollection))
                     .addSourceLine(sourceLineAnnotation);
-            for (String message : extraMessages) {
-                bugInstance.addString(message);
-            }
+        }
+    }
```
JuditKnoll: It looks like I added an unnecessary `}` here in my earlier suggestion, which makes this code failing to compile. Can you please remove it?

JuditKnoll (changes requested):
Can you please run `.\gradlew spotlessApply` after the changes to make sure there are no formatting issues?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: @JuditKnoll Yes, of course, thank you, I missed the negation!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: @JuditKnoll After taking another look, it turns out wrapping it in an `ArrayList` is indeed unnecessary.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: @JuditKnoll After taking another look, it turns out wrapping it in an `ArrayList` is indeed unnecessary.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
norbo03: @JuditKnoll I've implemented the requested changes and updated the branch with the latest from master.

norbo03:
@JuditKnoll  I resolved the merge conflicts; it is ready for review. Could you please take a look?

norbo03:
@JuditKnoll I think I have addressed everything. Can you please take a look at it?

norbo03:
@JuditKnoll I believe I've addressed every comment from the review. GitHub still indicates a requested change status, which is blocking the PR.
Could you please verify the changes and either approve the PR or let me know if there's anything I need to fix?

CHANGELOG.md
```diff
@@ -11,6 +11,19 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
...
+    - `USO_UNSAFE_INHERITABLE_OBJECT_SYNCHRONIZATION` is reported when the lock used for synchronization is can be altered by subclasses,
+    - `USO_UNSAFE_EXPOSED_OBJECT_SYNCHRONIZATION` is reported when the lock used for synchronization is later exposed in the subclasses.
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_BACKING_COLLECTION` is reported when the backing collection of a lock is visible from the outside,
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_ACCESSIBLE_BACKING_COLLECTION` is reported when the backing collection of a lock is made accessible, with methods that update or return the lock, to the outside,
+    - `USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION` is reported when the backing collection of a lock can be altered by subclasses.
+      (See [SEI CERT rule LCK00-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code) and [SEI CERT rule LCK04-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible))
+
```
JuditKnoll, line 137: Can you please merge master and move this to latest unreleased section?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,810 @@
...
+     * @param classContext the class context of the currently analyzed class, in which the wrapping method call is located
+     * @param location the location of the wrapping method call
+     * @param handle the instruction handle of the wrapping method call
+     * @param assignedField the field to which the result of the wrapping method call is assigned
+     * @param stackOffsetForWrapperMethodParameter the stack offset of the wrapped collection parameter of the wrapping method call
+     */
+    private void analyzeWrappedField(Method currentMethod, ClassContext classContext, Location location,
+            InstructionHandle handle, XField assignedField, Integer stackOffsetForWrapperMethodParameter) {
+        InstructionHandle prevHandle = location.getBasicBlock().getPredecessorOf(handle);
+        Instruction prevInstruction = prevHandle.getInstruction();
```
JuditKnoll, line 482: As far as I can see these, both `prevHandle` and `prevInstruction` can be null.
Can you please check that these are not null to avoid NPE?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,810 @@
...
+                        analyzeAssignmentsInHierarchy(lock);
+                    }
+
+                    try {
+                        findExposureInHierarchy(lock).forEach(method -> lockAccessorsInHierarchy.add(lock, method));
+                    } catch (ClassNotFoundException e) {
+                        AnalysisContext.reportMissingClass(e);
+                    }
+
+                    if (lock.isProtected() || lock.isPublic() || !lock.isPrivate()) {
```
JuditKnoll: This could be simplified to `!lock.isPrivate()`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -0,0 +1,810 @@
...
+                    declaredLockObjects.put(lock, xMethod);
+                    if (isCollection(lock)) {
+                        declaredCollectionLockObjects.add(lock, new BugInfo(this));
+                    }
+
+                    if (lock.isPublic()) {
+                        potentialObjectBugContainingMethods.add(lock, xMethod);
+                    }
+
+                    if (lock.isProtected() && !getThisClass().isFinal() || isPackagePrivate(lock)) {
```
JuditKnoll: This expression relies on operator precedence. Can you please add an additional bracket to make this intention explicit for the maintenance like this?
```suggestion
                    if ((lock.isProtected() && !getThisClass().isFinal()) || isPackagePrivate(lock)) {
```

JuditKnoll (changes requested):
I'm sorry for taking so long and thank you for the ping. Requested some small improvements, but after these are fixed it's good to go for me. Thank you for your work and patience.

norbo03:
@JuditKnoll No worries! I’ve addressed all your comments—thank you for your detailed review! When you have a moment, could you please approve the changes so the PR can be merged after the CI jobs succeed?

JuditKnoll (approved):

ThrawnCA (approved):

hazendaz:
@norbo03 Can you fix the merge conflicts?  Think based on other responses we should be able to pick up once that is done.

norbo03:
@hazendaz Updated to the latest master; there shouldn’t be any conflicts.

jglick:
LCK00-J reclassifies straightforward Java code which has been working fine for three decades as buggy. Just because outside code _could_ abuse an accessible monitor does not mean this is a likely risk in an established codebase. Perhaps the severity level should be reconsidered.

Issue #3059, opened 2024-07-31
Issue during analysis of SpringCGLIB after Spring Boot Native build


klopfdreh:
We are currently testing Spring Boot Native builds and found out that during the analysis SpotBugs failed.

See the following `maven_plugin_logs.txt`

[maven_plugin_logs.txt](https://github.com/user-attachments/files/16436767/maven_plugin_logs.txt)

Initial lines of error:
```plain
[java]   Error analyzing mypackage.steps.prepare.StepS3FilesPrepareExecutionListener$$SpringCGLIB$$0.setCallbacks([Lorg/springframework/cglib/proxy/Callback;)V for unconditional deref training
[java]     edu.umd.cs.findbugs.ba.CFGBuilderException: Invalid stack at    2: dup2[92](1) when checking    8: putfield[181](3) 58
```

See a minimal example here: https://github.com/klopfdreh/native-cloud-config-test and add spotbugs maven plugin together with a configuration that creates a SpringCGLIB 

`suppression.xml` file:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<FindBugsFilter xmlns="https://github.com/spotbugs/filter/3.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0/spotbugs/etc/findbugsfilter.xsd">
    <!-- Ignore checking for Spring Boot AOT generated classes -->
    <Match>
        <Or>
            <Class name="~.*\$\$SpringCGLIB\$\$.*"/>
        </Or>
    </Match>
    <Match>
        <Or>
            <Class name="~.*__BeanDefinitions.*"/>
        </Or>
    </Match>
    <Match>
        <Or>
            <Class name="~.*__BeanFactoryRegistrations.*"/>
        </Or>
    </Match>
    <Match>
        <Or>
            <Class name="~.*RetryTemplateFactory.*"/>
        </Or>
    </Match>
    <Match>
        <Or>
            <Class name="~.*DefaultDataSourceConfiguration.*"/>
        </Or>
    </Match>
</FindBugsFilter>
```

Plugin Configuration
```xml
                <plugin>
                    <groupId>com.github.spotbugs</groupId>
                    <artifactId>spotbugs-maven-plugin</artifactId>
                    <version>4.8.5.0</version>
                    <executions>
                        <execution>
[...]

PR #3064, opened 2024-08-02
BetterCFGBuilder2 error when analyzing Spring AOT bytecode

gtoison:
`dup2` is not handled correctly when trying to analyze that `putfield` is not a PEI (Potentially Exception-throwing Instruction) in the following bytecode:
```
// Method descriptor #303 ([Lorg/springframework/cglib/proxy/Callback;)V
  // Stack: 5, Locals: 2
  public void setCallbacks(org.springframework.cglib.proxy.Callback[] arg0);
     0  aload_0 [this]
     1  aload_1 [arg0]
     2  dup2
     3  iconst_0
     4  aaload
     5  checkcast org.springframework.cglib.proxy.MethodInterceptor [63]
     8  putfield spotbugs.aot.test.SpotBugsAotComponent$$SpringCGLIB$$0.CGLIB$CALLBACK_0 : org.springframework.cglib.proxy.MethodInterceptor [51]
```

The target of `putfield` is effectively loaded from the first `aload_0` so `putfield` is not a PEI here ( because `aload_0 == this`: it is not assigning to a null object)


gtoison:
The `depth` variable keeps track of how many stack elements we need to track back to find the reference of putfield. We start with 2 and increase/decrease it when we see preceding instructions popping/pushing
elements.
When reaching a dup2 and depth comes to zero it means that the instructions preceding the putfield have cancelled each other (the stack is the same as before these instructions), so we reset depth to 2.

gtoison:
The `isPEI` method attemps to find cases when we _know_ that `putfield` is assigning to `this`, so its target is guaranteed to be non null and `putfield` won't throw an exception. This reduces a bit the size of the CFG.
Could not find a sensible way to handle `dup2` in a generic way so I added another when we encounter it and reach `depth=0`
This should fix #3059

hazendaz (approved):

PR #3065, opened 2024-08-03
Cleanup in `FindNullDeref` detector

JuditKnoll:
Small cleanup around the `FindNullDeref` detector, which shouldn't cause any change in functionality.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
@@ -199,21 +199,21 @@ public class FindNullDeref implements Detector, UseAnnotationDatabase, NullDeref
         }
 
         List<Method> methodsInCallOrder = classContext.getMethodsInCallOrder();
-        for (Method method : methodsInCallOrder) {
+        for (Method m : methodsInCallOrder) {
             try {
-                if (method.isAbstract() || method.isNative() || method.getCode() == null) {
+                if (m.isAbstract() || m.isNative() || m.getCode() == null) {
                     continue;
                 }
 
-                currentMethod = SignatureConverter.convertMethodSignature(jclass, method);
+                currentMethod = SignatureConverter.convertMethodSignature(jclass, m);
 
-                if (METHOD_NAME != null && !method.getName().equals(METHOD_NAME)) {
+                if (METHOD_NAME != null && !m.getName().equals(METHOD_NAME)) {
                     continue;
                 }
                 if (DEBUG || DEBUG_NULLARG) {
                     System.out.println("Checking for NP in " + currentMethod);
                 }
-                analyzeMethod(classContext, method);
+                analyzeMethod(classContext, m);
             } catch (MissingClassException e) {
                 bugReporter.reportMissingClass(e.getClassNotFoundException());
             } catch (DataflowAnalysisException e) {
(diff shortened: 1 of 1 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -199,21 +199,21 @@ public void visitClassContext(ClassContext classContext) {
         }
 
         List<Method> methodsInCallOrder = classContext.getMethodsInCallOrder();
-        for (Method method : methodsInCallOrder) {
+        for (Method m : methodsInCallOrder) {
```
JuditKnoll, line 202: Renaming because of hiding the field with the same name.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -299,10 +294,10 @@ private void analyzeMethod(ClassContext classContext, Method method) throws Data
      */
     private BitSet findPreviouslyDeadBlocks() throws DataflowAnalysisException, CFGBuilderException {
         BitSet deadBlocks = new BitSet();
-        ValueNumberDataflow vnaDataflow = classContext.getValueNumberDataflow(method);
-        for (Iterator<BasicBlock> i = vnaDataflow.getCFG().blockIterator(); i.hasNext();) {
+        ValueNumberDataflow valueNumberDataflow = classContext.getValueNumberDataflow(method);
```
JuditKnoll, line 297: Renaming because of hiding the field with the same name.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -1635,13 +1598,10 @@ public void foundGuaranteedNullDeref(@Nonnull Set<Location> assignedNullLocation
 
             if (sourceLocations == doomedLocations && assignedNullLocationSet.size() == 1) {
                 Location assignedNull = assignedNullLocationSet.iterator().next();
-                SourceLineAnnotation sourceLineAnnotation = SourceLineAnnotation.fromVisitedInstruction(classContext, method,
-                        assignedNull);
-                if (sourceLineAnnotation != null) {
```
JuditKnoll, line 1640: Removing the null check, since `SourceLineAnnotation.fromVisitedInstruction()` won't return null.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java
```diff
@@ -1492,9 +1461,6 @@ public void foundGuaranteedNullDeref(@Nonnull Set<Location> assignedNullLocation
         SortedSet<SourceLineAnnotation> knownNullLocations = new TreeSet<>();
         for (Location loc : sourceLocations) {
             SourceLineAnnotation sourceLineAnnotation = SourceLineAnnotation.fromVisitedInstruction(classContext, method, loc);
-            if (sourceLineAnnotation == null) {
-                continue;
-            }
```
JuditKnoll, line 1497: Removing the null check, since `SourceLineAnnotation.fromVisitedInstruction()` won't return null.

gtoison (approved):

hazendaz (approved):

Issue #3023, opened 2024-06-25
False positive EI_EXPOSE_REP when use Collections.emptySet, emptyList or emptyMap is used


danielfelgar:
This example return EI_EXPOSE_REP, but `emptyList()`, `emptySet()` and `emptyMap()` are immutable collections.

```java
public record UnmodifiableRecord(List<String> list, Set<String> set, Map<String, String> map) {
    public UnmodifiableRecord {
        list = list != null ? List.copyOf(list) : Collections.emptyList();
        set = set != null ? Set.copyOf(set) : Collections.emptySet();
        map = map != null ? Map.copyOf(map) : Collections.emptyMap();
    }
}

public class UnmodifiableClass {
    private final List<String> list;
    private final Set<String> set;
    private final Map<String, String> map;

    public UnmodifiableClass(List<String> list, Set<String> set, Map<String, String> map) {
        this.list = list != null ? List.copyOf(list) : Collections.emptyList();
        this.set = set != null ? Set.copyOf(set) : Collections.emptySet();
        this.map = map != null ? Map.copyOf(map) : Collections.emptyMap();
    }
    // getters
}
```

Here a example to reproduce with tests to show immutable of classes.
[spotBugsIssue.zip](https://github.com/user-attachments/files/15971459/spotBugsIssue.zip)

PR #3080, opened 2024-08-14
Partial fix for Issue #3023

JuditKnoll:
Partial fix for #3023. The problem is not with the `Collections.emptyList()`,  `Collections.emptySet()` and  `Collections.emptyMap()` functions, there are no FPs, if there is only one assignment to a field with any of these functions. The issue is rather about having multiple assignments to one field (e.g. with if-else or ternary operator).
This PR fixes the class example given at the issue and a slightly modified version of the record example as well, but the original record example with zero argument constructor still reports FPs if not disabled. I tried to solve it, but couldn't find a way to connect the assigned actual values to the field in that case. If anyone has good idea how to solve it, feel free to share it or take it up.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
@@ -19,13 +19,14 @@
 
 package edu.umd.cs.findbugs.detect;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.HashMap;
 
 import org.apache.bcel.Const;
-import org.apache.bcel.classfile.Code;
 import org.apache.bcel.classfile.JavaClass;
 import org.apache.bcel.classfile.Method;
 
@@ -50,6 +51,8 @@ public class FindReturnRef extends OpcodeStackDetector {
 
     private int parameterCount;
 
+    private boolean isFirstVisit = true;
+
     private XField fieldUnderClone = null;
     private OpcodeStack.Item paramUnderClone = null;
     private XField fieldCloneUnderCast = null;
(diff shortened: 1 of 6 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -104,15 +118,26 @@ public void visit(Method obj) {
     }
 
     @Override
-    public void visit(Code obj) {
-        if (check) {
-            super.visit(obj);
+    public void sawOpcode(int seen) {
+        if (isFirstVisit) {
```
gtoison, line 122: Are you sure this still detects true positives when the second assignment is modifiable?
I probably misunderstood but it looks like only the first value is considered, so it might miss true positives when only the first assignment was indeed unmodifiable. 
In any case, it would be beneficial to add tests for that case.

JuditKnoll: Thanks, I added some tests with if-elses. It should have found these, if the condition in line 217-218 would have been okay. I fixed, so now it works for the current tests, however when using the ternary operator, it does not work properly, I'm still looking into that issue.

JuditKnoll: The logic here is that the examined class is visited two times: the first time, every value assignment is collected. Then the bugs are reported with consideration of the collected data.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
gtoison: Thanks for the explanation and the unit tests showing the true positives!

gtoison (approved):
Thank you for the fix and unit tests!

ThrawnCA (approved):

Issue #608, opened 2018-04-05
Exception analyzing ... using detector edu.umd.cs.findbugs.detect.RedundantConditions


gargooza:
I'm getting a lot of the same exception:
```
 Exception analyzing com.redacted.MyTroublesomeClassToAnalyze using detector edu.umd.cs.findbugs.detect.RedundantConditions
   java.lang.IndexOutOfBoundsException: fromIndex: 1303 > toIndex: 1114
     At java.util.BitSet.checkRange(BitSet.java:366)
     At java.util.BitSet.set(BitSet.java:481)
     At edu.umd.cs.findbugs.classfile.engine.bcel.ValueRangeAnalysisFactory.analyze(ValueRangeAnalysisFactory.java:789)
     At edu.umd.cs.findbugs.classfile.engine.bcel.ValueRangeAnalysisFactory.analyze(ValueRangeAnalysisFactory.java:87)
     At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.analyzeMethod(AnalysisCache.java:369)
     At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.getMethodAnalysis(AnalysisCache.java:322)
     At edu.umd.cs.findbugs.detect.RedundantConditions.visitClassContext(RedundantConditions.java:67)
     At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
     At edu.umd.cs.findbugs.FindBugs2.analyzeApplication(FindBugs2.java:1079)
     At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:280)
     At edu.umd.cs.findbugs.FindBugs.runMain(FindBugs.java:401)
     At edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1187)
```

MyTroublesomeClassToAnalyze is not the only class for which this is a problem. Other classes throw the same exception, just with different indices for the IndexOutOfBoundsException:
```
MyTroublesomeClassToAnalyze.java: java.lang.IndexOutOfBoundsException: fromIndex: 1303 > toIndex: 1114
AnotherClass.java: java.lang.IndexOutOfBoundsException: fromIndex: 1078 > toIndex: 919
YetAnother.java: java.lang.IndexOutOfBoundsException: fromIndex: 1226 > toIndex: 1050
[...]

PR #3086, opened 2024-08-19
Generics enhancements

gtoison:
Variable types such as `<C extends Collection<String>>` are not analyzed correctly because they are not parsed from the method signature as noted in #608

### DIFF
--- a/classfile-backport/build.gradle
+++ b/classfile-backport/build.gradle
@@ -1,0 +1,33 @@
+apply from: "$rootDir/gradle/checkstyle.gradle"
+apply from: "$rootDir/gradle/javadoc.gradle"
+apply from: "$rootDir/gradle/maven.gradle"
+
+tasks.named('javadoc', Javadoc).configure {
+  options.with {
+    memberLevel = JavadocMemberLevel.PUBLIC
+    docTitle = 'SpotBugs Classfile API Backport Documentation'
+  }
+  doFirst {
+    // This is supposed to enable everything except "missing" but doesn't work with gradle
+    // See http://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html
+    options.addBooleanOption('Xdoclint:all,-missing', true)
+  }
+  doLast {
+    logger.warn('Javadoc: "missing" warnings are disabled, see #340!')
+  }
+}
+
+java {
+  withJavadocJar()
+  withSourcesJar()
+}
+
+publishing.publications.maven {
+  pom {
+    name = 'SpotBugs Classfile API Backport'
+    description = 'Backport of Java classfile API'
+  }
+}
+
+ext.moduleName = 'com.github.spotbugs.classfile-backport'
+apply from: "$rootDir/gradle/jigsaw.gradle"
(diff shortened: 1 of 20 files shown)


gtoison:
- Backported JEP's 457 (classfile API) signature parser to support more generics use cases, such as type variables
- The backported code is a subset of the classfile API (the signatures parser), most of the changes relate to records, switches and sealed interfaces
- The new signature parser does not handle invalid signatures produced by older Eclipse JDT versions
- Parsing a signature might require the method and type signatures to locate variable types, some of the code still uses the old parser and could be updated in further enhancements
- GenericObjectType.variable was used for the variable type identifier OR for the wildcard indicator (extends/super), added a wildcard indicator field
- Let GenericObjectType have several extensions (for instance <P extends InterfaceA & InterfaceB>)
- Modified GenericObjectType equals and toString to handle self-referencing types (e.g. <C extends Map<X, C>>)
- In particular GenericObjectType.toString() is modified to return the bytecode signature when available

gtoison:
I think the failed smoke tests are because spotbugs analyzes itself using the Gradle plugin, which isn't incorporating the changes to support `<C extends Collection<String>>`.
On my computer `gradlew clean test` shows the same error while `gradlew test`
I tried modifying the Gradle build to fix this but can't seem to understand how it works. Would anyone know?

gtoison:
Regarding the backport of the classfile API, the license seems to be GNU GPL2 so it seems allowed to me to modify the sources and include them in SpotBugs.
I've not looked much into the rest of the API but there might other useful things.
This is meant to be temporary as we will eventually require a >21 JDK version but realistically this backport would stay for several years.

iloveeclipse:
> the license seems to be GNU GPL2 so it seems allowed to me to modify the sources and include them in SpotBugs.

You can't and should not add *any* modified GPL licensed code to SpotBugs as it would require SpotBugs to use GPL and that in turn would be a breaking change because SpotBugs uses *LGPL* and so our users/adopters are *not* forced to open source all their code.

iloveeclipse (changes requested):
Adding GPL'ed code, especially the one from Oracle is not possible, sorry.

As much as I would like to see any enhancement, I can't agree on changing SpotBugs license to GPL.

gtoison:
What about using this backport of the classfile Api https://github.com/dmlloyd/jdk-classfile-preview ?

iloveeclipse:
> What about using this backport of the classfile Api https://github.com/dmlloyd/jdk-classfile-preview ?

Same license as before, same Oracle origin GPL code.

gtoison:
But doesn't it fall under the classpath exception? After all we are using the base java library under that exception. Or do you mean that the exception does not transfer to separately released modified code?

iloveeclipse:
See https://opensource.stackexchange.com/questions/1410/what-is-the-difference-between-gpl-classpath-exception-vs-lgpl

LGPL allows is to modify the original code **and** ship that, GPL + classpath doesn't.

iloveeclipse:
Beside GPL/LGPL license question, I'm not a lawyer, but I would not embed **any** modified Oracle code to SpotBugs in general, rememberimg Oracle vs Google lawsuit and how Oracle lawyers argued about trivial code snippets etc.

PR #1802, opened 2021-11-11
Issue-637

ramraval:
Addresses Issue #637 by adding new detector & associated test case classes for bad combinations of format strings used with SimpleDateFormat

(Apologies for any confusion, I know @arnab44 had been assigned to this issue about a month ago, but I am a first-time contributor who took up this issue for a university project)

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

Issue #637, opened 2018-05-14
SimpleDateFormat string checks


Hansyperman:
There is a discussion about date format strings on
https://news.ycombinator.com/item?id=17066739 

Can  spotbugs check if bad combinations of date format flags exist in the formatting string?

Some rules, from that discussion:

* %Y can not appear with %M and %d unless also %w appears.  You need probably %y
* %h or %K requires %a, or %H or %k forbids %a.

ThrawnCA:
It’s certainly possible to do. For new, experimental detectors, you might have more success at fb-contrib (https://github.com/mebigfatguy/fb-contrib). Or you could try your hand at writing it, and submit a pull request :)

ThrawnCA:
Sure, but you don't actually need to be assigned. It's more of a nudge than anything.

gayatrib:
Hello, team, just checking in? Is this issue still open and up for grabs for first timers like me?

ThrawnCA:
@gayatrib It's been assigned to @arnab44 so I'd recommend discussing it with arnab first. Otherwise, sure.

PR #3088, opened 2024-08-26
Changes for Issue-637

stt08:
Original PR that is being taken over is #1802 
due to not seeing any activity there lately.

Addresses Issue https://github.com/spotbugs/spotbugs/issues/637 by adding new detector & associated test case classes for bad combinations of format strings used with SimpleDateFormat, DateTimeFormatter and FastDateFormat.
Co-authored-by: @ramraval

Requested changes are applied, latest changes from master are collected as well.


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue637Test.java
```diff
@@ -0,0 +1,71 @@
...
+     *  - was called in a specified function (parameter)
+     *  - and caught on a proper line (parameter)
+     */
+    @Test
+    public void testClassWithErrors() {
+        performAnalysis("ghIssues/Issue637Errors.class");
+        assertNumOfBug637(25);
+
+        int line;
+        for (line = 28; line <= 32; line++)
```
ThrawnCA, line 35: Multiline `for` loops should use braces to avoid the risk of "goto fail" errors.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DateFormatStringChecker.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /**
+         * Checks if string does not contain a
+         * forbiddenFlag and has all requiredFlags
+         * @param dateFormatString - string to be checked
+         * @return {@code true} if code equals to one of the elements
+         * */
+        boolean verify(String dateFormatString) {
+            if (this.forbiddenFlag != null && dateFormatString.contains(this.forbiddenFlag))
```
ThrawnCA, line 47: Use braces to avoid the risk of "goto fail" errors.

ThrawnCA (dismissed):
`for` and `if` statements should use braces even if the language doesn't strictly require it, to reduce the risk of introducing errors by adding lines in future.

gtoison (changes requested):
Thank you for picking this up!
Can you please merge in the changelog so it does not look like your removing older entries?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue637Test.java
stt08: Fixed by calling each assert manually, line-by-line

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue637Test.java
```diff
@@ -1,71 +1,127 @@
...
-    public void testClassWithErrors() {
-        performAnalysis("ghIssues/Issue637Errors.class");
-        assertNumOfBug637(25);
+    void testClassWithErrors() {
+        String className = "Issue637Errors";
+        performAnalysis("ghIssues/issue637/" + className + ".class");
+        assertBugTypeCount(DESIRED_BUG_TYPE, 80);
+
+        assertBugInMethodAtLine(DESIRED_BUG_TYPE, className, "oneArgConstructorTest", 31);
+        assertBugInMethodAtLine(DESIRED_BUG_TYPE, className, "oneArgConstructorTest", 31);
```
ThrawnCA, line 28: This duplicates the previous line.

stt08: changes applied

JuditKnoll (approved):

gtoison (approved):

PR #3102, opened 2024-09-04
Extends `@SuppressFBWarnings` to allow regex and exact matches

gtoison:
Extend the bug suppression, allowing exact (not prefix) bug type match, and regex bug type match

### DIFF
--- a/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressMatchType.java
+++ b/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressMatchType.java
@@ -3,20 +3,25 @@ package edu.umd.cs.findbugs.annotations;
 public enum SuppressMatchType {
 
     /**
+     * Default bug suppression using a mixed prefixed / case insensitive match depending on the criterion.
      * Suppress bugs matching any of:
-     * <li/> the given bug type {@link String#startsWith(String)}
-     * <li/> the given bug category {@link String#equalsIgnoreCase(String)}
-     * <li/> the given bug abbreviation {@link String#equalsIgnoreCase(String)}
+     * <li> the given bug type with: {@link String#startsWith(String)}
+     * <li> the given bug category with: {@link String#equalsIgnoreCase(String)}
+     * <li> the given bug abbreviation with: {@link String#equalsIgnoreCase(String)}
      */
     DEFAULT,
 
     /**
-     * Suppress bugs matching the exact (case sensitive) given type.
+     * Exact (case sensitive match).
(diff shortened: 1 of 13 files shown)


spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressMatchType.java
```diff
@@ -0,0 +1,22 @@
...
+     */
+    DEFAULT,
+
+    /**
+     * Suppress bugs matching the exact (case sensitive) given type.
+     */
+    EXACT,
+
+    /**
+     * Suppress bug whose type match the given regular expression.
```
JuditKnoll, line 19: IMO it's rather connected to the `value` field that to which properties is it connected to (type, category and abbreviation), so even if it's regex or exact match, both the type, the category and the abbreviation could be considered. Or if not, the comment at the `SuppressFBWarnings.value` to should show this change.

JuditKnoll (changes requested):
Spotless didn't like this PR, can you please run `gradlew spotlessApply`?
Can you please add a few tests for this?
Before merging this to master, it would be great to update the documentation about this as well.

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressMatchType.java
gtoison: I guess what we're aiming for here is the least surprise principle, so I _think_ it's going to be less surprising if all matchTypes use the same criteria (type, category, abbreviation)
I've changed the code to match either of the 3 for all match types

JuditKnoll (changes requested):

JuditKnoll (approved):
It might be a good idea to add some info about this feature in the [documentation](https://github.com/spotbugs/spotbugs/blob/master/docs/annotations.rst#eduumdcsfindbugsannotationssuppressfbwarnings) as well, but otherwise LGTM.

gtoison:
I looked for mentions to `@SuppressFBWarnings` in the documentation but couldn't find any, I guess the javadocs _are_ the docs for now :)

JuditKnoll:
> I looked for mentions to `@SuppressFBWarnings` in the documentation but couldn't find any, I guess the javadocs _are_ the docs for now :)

First I thought so as well. But it differs a bit. All I could find about suppressing warnings is in the [annotations.rst](https://github.com/spotbugs/spotbugs/blob/master/docs/annotations.rst#eduumdcsfindbugsannotationssuppressfbwarnings) file.

ThrawnCA (approved):

PR #3106, opened 2024-09-10
Adding tests for already existing code examples

JuditKnoll:
Quite a large number of the code examples has no test at all associated with it (neither `AbstractIntegrationTest` nor tests with `@ExpectWarning` and similar annotations). This PR creates tests for some of them. A lot of the BugIdeas code examples would still fail if had tests, as there are no relevant detectors or the existing detectors doesn't cover the issue the code example describes. Also, these examples have minimal comments, and it's not always easy to find out what was the programmer's intent with the code example.
All of these tests should be successful. The purpose of this PR is to increase the coverage of the examples and document the current behavior of the detectors. This PR does not provide a full coverage of the existing code examples, even with this PR just in the [bugIdeas](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java/bugIdeas) package, there is still a huge number of files without any tests at all.

Since this PR doesn't change the actual code of the project, only adds tests, I haven't added a changelog entry.

For an easy review session, here are the tested classes in the order of tests:
- [ArrayEqualsNegativeCases](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/com/google/errorprone/bugpatterns/ArrayEqualsNegativeCases.java) and [ArrayEqualsPositiveCases](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/com/google/errorprone/bugpatterns/ArrayEqualsPositiveCases.java),
- [Bug1864046](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/sfBugs/Bug1864046.java),
- [Ideas_2008_11_14](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2008_11_14.java),
- [Ideas_2009_01_05](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_01_05.java),
- [Ideas_2009_02_05](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_02_05.java),
- [Ideas_2009_03_14](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_03_14.java),
- [Ideas_2009_04_30](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_04_30.java),
- [Ideas_2009_05_04](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_05_04.java),
- [Ideas_2009_05_26](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_05_26.java),
- [Ideas_2009_05_30](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_05_30.java),
- [Ideas_2009_06_03](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_06_03.java),
- [Ideas_2009_06_16](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_06_16.java),
- [Ideas_2009_07_26](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_07_26.java),
- [Ideas_2009_10_05](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_10_05.java),
- [Ideas_2009_10_06](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_10_06.java),
- [Ideas_2009_10_07](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_10_07.java),
- [Ideas_2009_10_10](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2009_10_10.java),
- [Ideas_2011_07_28](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2011_07_28.java),
- [Ideas_2011_11_18](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2011_11_18.java),
- [Ideas_2012_05_21](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2012_05_21.java),
- [Ideas_2013_04_29](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/bugIdeas/Ideas_2013_04_29.java).

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
@@ -1,0 +1,37 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.jupiter.api.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+class RegressionIdeas20090105Test extends AbstractIntegrationTest {
+    @Test
+    void test() {
+        performAnalysis("bugIdeas/Ideas_2009_01_05.class");
+        assertBugNum(2);
+        assertExactBug("test", 25);
+        assertExactBug("test2", 29);
+    }
+
+    private void assertBugNum(int num) {
+        BugInstanceMatcher matcher = new BugInstanceMatcherBuilder()
+                .bugType("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS")
+                .build();
+        assertThat(getBugCollection(), containsExactly(num, matcher));
+    }
+
+    private void assertExactBug(String method, int line) {
+        BugInstanceMatcher matcher = new BugInstanceMatcherBuilder()
+                .bugType("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS")
+                .inClass("Ideas_2009_01_05")
+                .inMethod(method)
+                .atLine(line)
+                .build();
+        assertThat(getBugCollection(), hasItem(matcher));
+    }
+}
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
@@ -1,37 +1,15 @@
 package edu.umd.cs.findbugs.detect;
 
 import edu.umd.cs.findbugs.AbstractIntegrationTest;
-import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcher;
-import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
 import org.junit.jupiter.api.Test;
 
-import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.hasItem;
 
 class RegressionIdeas20090105Test extends AbstractIntegrationTest {
     @Test
     void test() {
         performAnalysis("bugIdeas/Ideas_2009_01_05.class");
-        assertBugNum(2);
-        assertExactBug("test", 25);
-        assertExactBug("test2", 29);
-    }
-
-    private void assertBugNum(int num) {
-        BugInstanceMatcher matcher = new BugInstanceMatcherBuilder()
-                .bugType("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS")
-                .build();
-        assertThat(getBugCollection(), containsExactly(num, matcher));
-    }
-
-    private void assertExactBug(String method, int line) {
-        BugInstanceMatcher matcher = new BugInstanceMatcherBuilder()
-                .bugType("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS")
-                .inClass("Ideas_2009_01_05")
-                .inMethod(method)
-                .atLine(line)
-                .build();
-        assertThat(getBugCollection(), hasItem(matcher));
+        assertBugTypeCount("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS", 2);
+        assertBugInMethodAtLine("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS", "Ideas_2009_01_05", "test", 25);
+        assertBugInMethodAtLine("NP_NULL_PARAM_DEREF_ALL_TARGETS_DANGEROUS", "Ideas_2009_01_05", "test2", 29);
     }
 }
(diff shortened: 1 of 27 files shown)


ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20091010Test.java
```diff
@@ -0,0 +1,42 @@
...
+        assertExactBug("NP_NONNULL_PARAM_VIOLATION", 15);
+        // FNs in line 17, 18, 19
+    }
+
+    private void assertBugNum(String bugtype, int min, int max) {
+        BugInstanceMatcher matcher = new BugInstanceMatcherBuilder()
+                .bugType(bugtype)
+                .build();
+        long exactMatches = getBugCollection().getCollection().stream().filter(matcher::matches).count();
+        assertTrue(exactMatches >= min);
```
ThrawnCA, line 29: `assertTrue` is not very descriptive by default, can this have a failure message added?

JuditKnoll: You are right, in this case a fail wouldn't have a helpful msg. Instead of repeating a failure message several times, I added `containsBetween()` to `CountMatcher` and used that.

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
```diff
@@ -0,0 +1,37 @@
...
+class RegressionIdeas20090105Test extends AbstractIntegrationTest {
+    @Test
+    void test() {
+        performAnalysis("bugIdeas/Ideas_2009_01_05.class");
+        assertBugNum(2);
+        assertExactBug("test", 25);
+        assertExactBug("test2", 29);
+    }
+
+    private void assertBugNum(int num) {
```
gtoison, line 21: I wonder if we shouldn't move these `assertXyzBug` methods in `AbstractIntegrationTest` because we seem to be duplicating them all over the place.
In theory they would only wrap `BugInstanceMatcher`'s fluent API but I think that it would be nicer to have built-in methods for the most frequent cases

JuditKnoll: Yes, there are code a lot of code repetition with these asserts, but there are already a lot of similar code duplication. IMO it would be best if the already existing similar code would use these functions, but that would make this PR too big and messy, and would be out of scope of this PR. I'll create a new PR in a few days adding them, and after it's merged, make this PR using that feature.

gtoison: Sounds good to me, thank you for the improvements!

gtoison:
Sorry I marked this PR as ready for review by mistake, and now I can't seem to convert it to draft again...

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/RegressionIdeas20090105Test.java
JuditKnoll: Created https://github.com/spotbugs/spotbugs/pull/3120 to solve this issue.

spotbugsTestCases/src/java/bugIdeas/Ideas_2009_10_07.java
```diff
@@ -17,7 +17,7 @@ static int f(boolean b1, boolean b2, Object x) {
         return -x.hashCode();
     }
 
-    static int f(boolean b1, boolean b2) {
+    static int f2(boolean b1, boolean b2) {
```
JuditKnoll, line 20: Renamed this function, so it doesn't have the same name as the one in line 5, and the bug can be checked easily.

test-harness/src/main/java/edu/umd/cs/findbugs/test/CountMatcher.java
```diff
@@ -33,7 +35,25 @@ public CountMatcher(int count, Matcher<T> matcher) {
...
+
+    /**
+     * Creates a matcher for {@link Iterable}s that only matches if at least {@code minCount} and at most {@code maxCount}
+     * items match the specified {@code matcher}.
+     *
+     * @param matcher
+     *            A non-{@code null} matcher that must match at least {@code minCount} and at most {@code maxCount} times.
+     * @param <T>
+     *            matcher item type
+     * @param minCount
```
gtoison, line 49: In some contexts the upper (or lower) bound of a "between" might be exclusive, could you please add to the comment that they are both inclusive?

gtoison (changes requested):

test-harness/src/main/java/edu/umd/cs/findbugs/test/CountMatcher.java
JuditKnoll: Thanks, added this info to the javadoc.

gtoison: Thank you!

gtoison (approved):

Issue #2969, opened 2024-04-24
Stylesheet not working in 4.8.4


esabol:
We use `stylesheet="fancy-hist.xsl"` to generate our SpotBugs reports in our CI/CD pipeline, which builds using ant. Our `build.xml` has the following:
```xml
    <taskdef
        resource="edu/umd/cs/findbugs/anttask/tasks.properties"
        classpath="${spotbugs.home}/lib/spotbugs-ant.jar"/>
 
    <target name="spotbugs">
        <spotbugs home="${spotbugs.home}"
                  output="html"
                  outputFile="${dist.dir}/spotbugs.html"
                  stylesheet="fancy-hist.xsl" >
            <auxclasspath path="${basedir}/../libs" />
            <sourcePath path="${basedir}/src/java" />
            <class location="${build.web.dir}/WEB-INF/classes" />
        </spotbugs>
    </target>
```
This all worked for us just fine with SpotBugs 4.8.3. After upgrading to 4.8.4, however, the SpotBugs report generated by the same exact `build.xml` is clearly not using `fancy-hist.xml`. And, yes, `spotbugs.home` is defined properly in `project.properties` (which hasn't changed) and we are using the `spotbugs-ant.jar` from the 4.8.4 release.

Any idea as to what the problem is or what the solution might be? Thanks!

gtoison:
Hello @Vogel612, I'm not sure how this was working but I was wondering if this could be a side effect of #2894

JuditKnoll:
Maybe it can be because of https://github.com/spotbugs/spotbugs/pull/2915. Can you please share the `fancy-hist.xml`? Is there any error logged about this?

gtoison:
I think the xsl is part of SpotBugs: https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/xsl/fancy-hist.xsl
I was wondering about the version change for the filter xsd too, that might be the problem

esabol:
Yes, the `fancy-host.xsl` we are using is the one provided by SpotBugs, as @gtoison indicated. We haven't modified it or moved it.

esabol:
My colleague additionally tested it with SpotBugs 4.8.2, and the stylesheet worked.

Also tested 4.8.4 with `stylesheet="${spotbugs.home}/src/xsl/fancy-hist.xsl"` instead of `stylesheet="fancy-hist.xsl"`. Still didn't work.

4.8.3 works with either of those two `stylesheet` attribute values.

[...]

PR #3115, opened 2024-09-20
Perform Bug-Reporter deduplication based on target files

### LINKED ISSUE #2047: Specifying the same report multiple times results in output issues

Vogel612:
This is attempt number two to fix #2047, after the initial attempt broke stylesheet behavior and was reverted.

The first attempt at this was in PR #2894, subsequently reverted in #3038 to fix issue #2969

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -114,8 +114,8 @@ class TextUICommandLineTest {
         Path altFile = tmpdir.resolve("sibling2").resolve("..").resolve(file.getFileName());
         TextUICommandLine commandLine = new TextUICommandLine();
         try (FindBugs2 findbugs = new FindBugs2()) {
-            commandLine.handleOption("-xml", "=" + file.toString());
-            commandLine.handleOption("-xml", "=" + altFile.toString());
+            commandLine.handleOption("-xml", "=" + file);
+            commandLine.handleOption("-xml", "=" + altFile);
 
             commandLine.configureEngine(findbugs);
             var configuredReporter = findbugs.getBugReporter();
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -83,7 +83,7 @@ class TextUICommandLineTest {
 
     @Test
     void sharedOutputFilesAreNotDuplicated() throws Exception {
-        Path file = Files.createTempFile("spotbugs", ".xml");
+        Path file = Files.createTempFile("spotbugs", ".html");
         TextUICommandLine commandLine = new TextUICommandLine();
         try (FindBugs2 findbugs = new FindBugs2()) {
             commandLine.handleOption("-html", "fancy-hist.xsl=" + file.toFile().getAbsolutePath());
@@ -103,5 +103,28 @@ class TextUICommandLineTest {
         // we don't even *really* need to finish analysis, but it's better to explicitly check
         String html = Files.readString(file, StandardCharsets.UTF_8);
         assertThat(html, containsString("#historyTab"));
+    }
+
+    @Test
+    void deduplicationUsesCanonicalPaths() throws Exception {
+        Path tmpdir = Files.createTempDirectory("spotbugs");
+        Files.createDirectory(tmpdir.resolve("sibling1"));
+        Files.createDirectory(tmpdir.resolve("sibling2"));
+        Path file = Files.createTempFile(tmpdir.resolve("sibling1"), "tempFile", ".xml");
+        Path altFile = tmpdir.resolve("sibling2").resolve("..").resolve(file.getFileName());
+        TextUICommandLine commandLine = new TextUICommandLine();
+        try (FindBugs2 findbugs = new FindBugs2()) {
+            commandLine.handleOption("-xml", "=" + file.toString());
+            commandLine.handleOption("-xml", "=" + altFile.toString());
+
+            commandLine.configureEngine(findbugs);
+            var configuredReporter = findbugs.getBugReporter();
+            // this is a horrible hack: This would be a BugReportDispatcher if there was more than one reporter
+            assertThat(configuredReporter, not(isA(BugReportDispatcher.class)));
+            configuredReporter.finish();
+        }
+        // we don't even *really* need to finish analysis, but it's better to explicitly check
+        String xml = Files.readString(file, StandardCharsets.UTF_8);
+        assertThat(xml, containsString("BugCollection"));
     }
 }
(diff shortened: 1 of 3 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
```diff
@@ -78,4 +80,28 @@ void htmlReportWithOption() throws IOException {
         assertTrue(xdocsFile.toFile().isFile());
         assertTrue(textFile.toFile().isFile());
     }
+
+    @Test
+    void sharedOutputFilesAreNotDuplicated() throws Exception {
```
JuditKnoll, line 85: Can you please add a test where not the same path is added, but two paths refer to the same location? E.g. the first path is `tempFile.xml` and the second one is `tempDir/../tempFile.xml` (obviously without calling `getAbsolutePath()` or canonizing the paths in the test.)

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
Vogel612: What would be the expected behavior for that case? The original intention of the Bugreport was to make this use only a single outputstream, i.e. only a single reporter, which would imply canonicalizing the path in the underlying code...

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -816,4 +818,19 @@ private void handleAnalyzeFromFile(String filePath) throws IOException {
     private UserPreferences getUserPreferences() {
         return project.getConfiguration();
     }
+
+    /**
+     * Adds a reporter to the aggregating list of reporters, assuming it does not duplicate a known reporter.
```
gtoison, line 823: I think the "assuming" bit could be rephrased, because it is ignoring the reporter in case it is deemed to be a duplicate

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
JuditKnoll: IMO the same as if the same file is added, so that, it is not duplicated.

Vogel612: Added test case `deduplicationUsesCanonicalPaths`

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUIBugReporter.java
```diff
@@ -47,6 +47,8 @@ public abstract class TextUIBugReporter extends AbstractBugReporter implements C
 
     private boolean applySuppressions = false;
 
+    private String deduplicationKey = null;
```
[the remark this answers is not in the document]
Vogel612: renamed to `outputTarget`... not sure if that's better, but it's slightly closer to what you want I think?

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
Vogel612: reworded

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
```diff
@@ -78,4 +80,51 @@ void htmlReportWithOption() throws IOException {
...
+    void deduplicationUsesCanonicalPaths() throws Exception {
+        Path tmpdir = Files.createTempDirectory("spotbugs");
+        Files.createDirectory(tmpdir.resolve("sibling1"));
+        Files.createDirectory(tmpdir.resolve("sibling2"));
+        Path file = Files.createTempFile(tmpdir.resolve("sibling1"), "tempFile", ".xml");
+        Path altFile = tmpdir.resolve("sibling2").resolve("..").resolve(file.getFileName());
+        TextUICommandLine commandLine = new TextUICommandLine();
+        try (FindBugs2 findbugs = new FindBugs2()) {
+            commandLine.handleOption("-xml", "=" + file.toString());
+            commandLine.handleOption("-xml", "=" + altFile.toString());
```
JuditKnoll, line 118: These two `toString()`s are unnecessary.

JuditKnoll (approved):
One small issue, but otherwise LGTM.

ThrawnCA (approved):

JuditKnoll (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUIBugReporter.java
hazendaz, line 1: Null is default on class variable so remove that.  On check on not null flip thr checks so null check is not necessary.  Instead of paths get.  Use path of.  Java replaced that and path.get just calls path.of.

 All else looks ok.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUIBugReporter.java
```diff
@@ -240,6 +242,19 @@ protected void checkBugInstance(BugInstance bugInstance) {
         }
     }
 
+    public String getOutputTarget() {
+        return this.outputTarget;
```
ThrawnCA, line 246: Why use `this` when there is no local variable to confuse it with?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
```diff
@@ -78,4 +80,51 @@ void htmlReportWithOption() throws IOException {
         assertTrue(xdocsFile.toFile().isFile());
         assertTrue(textFile.toFile().isFile());
     }
+
+    @Test
+    void sharedOutputFilesAreNotDuplicated() throws Exception {
+        Path file = Files.createTempFile("spotbugs", ".html");
+        TextUICommandLine commandLine = new TextUICommandLine();
+        try (FindBugs2 findbugs = new FindBugs2()) {
+            commandLine.handleOption("-html", "fancy-hist.xsl=" + file.toFile().getAbsolutePath());
```
ThrawnCA, line 89: Could extract the absolute path to a local variable.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUIBugReporter.java
hazendaz: I don't see harm there, personally I prefer objects to be qualified as its easier to understand where its defined.

Issue #3123, opened 2024-09-25
FI_NULLIFY_SUPER mapped to CWE 586, but should be 568


davewichers:
This rule flags: "Finalizer nullifies superclass finalizer", i.e., doesn't call super.finalize().

But it is mapped to: CWE-586: Explicit Call to Finalize().

However, I think this is the correct mapping: CWE-568: finalize() Method Without super.finalize()

UPDATE: The rule: FI_MISSING_SUPER_CALL, which says: "Finalizer does not call superclass finalizer" should be mapped to 568 as well.

They both talk about finalizers but I think its mapped to the wrong one.

Also, where are these rules defined in the codebase and/or the CWE mappings defined? I am aware of some other rules that I don't think are properly mapped as well.  If there is a single mapping file, I could provide feedback on multiple rules that are either missing CWE mappings, but should have them, or are mapped to CWEs MITRE has marked as DEPRECATED/PROHIBITED, like this one as an example: https://cwe.mitre.org/data/definitions/218.html. There is a spotbugs rule that is mapped to CWE 218. 218 says: it is a duplicate of CWE-493, so that rule should be mapped to CWE 493 instead. This rule is MS_PKGPROTECT.

Ideally, someone would go through all the CWE mappings and move them to non-deprected, non-prohibited CWEs.

JuditKnoll:
The rules are defined as Bug Patterns in the [findbugs.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugs.xml#L722) file. The CWE can be given by `cweid`. The CWE can also be given at the Bug Codes [in the same file](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugs.xml#L692), which seems to be the case in the examples you mentioned, as there is [no CWE given to FI_NULLIFY_SUPER](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugs.xml#L861) or to [MS_PKGPROTECT](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugs.xml#L947), but there are to their bug codes.
The description of the Bug Patterns are in the [messages.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/messages.xml#L1842) file.
Feel free to contribute to the project 😄

gtoison:
@davewichers I've submitted PR #3124 to show what the fixes would look like

gtoison:
[...]

PR #3124, opened 2024-09-26
CWE mapping fixes

gtoison:
Updated some incorrect CWE mappings reported in #3123


davewichers:
@gtoison - Thanks for this! Looks good. However, there must be a mapping of the rule FI_NULLIFY_SUPER to CWE-586. Can we find that and either remove it or change it to match this change? I'm guessing it's in the definition/implementation of the rule?

gtoison:
`FI_NULLIFY_SUPER` was CWE-586 because the `FI` bugs were all CWE-586 by default:
https://github.com/spotbugs/spotbugs/blob/c4b3fa473d6df297539aa20c57770b90cb9aff52/spotbugs/etc/findbugs.xml#L704

davewichers:
OK. Makes sense. It might make sense to review every FI rule and make sure that rule belongs in either 568, or 586, or maybe neither?? Not sure how many FI rules there are or what they each do.

gtoison:
Thanks, I have mapped FI_MISSING_SUPER_CALL to CWE 568.
Let me know if you find others!

davewichers:
I have reviewed the other FI_ rules and they look fine to me. So I think you can submit this PR for merging by the spotbugs team. If I find other rules that are missing CWEid, or have them wrong, I'll create a separate issue for those, but I think this PR is good to go.

hazendaz (approved):

Issue #3159, opened 2024-10-14
Incorrect method call order


topolik:
Hi guys,

do you know why there is SelfMethodCalls#interestingSignature method? https://github.com/spotbugs/spotbugs/blob/d999c35c134f5310b688f2b4513e3689fb1439bc/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/SelfMethodCalls.java#L44 ?

This ruins the correct order of method calls as derived by [ClassContext#getMethodsInCallOrder](https://github.com/spotbugs/spotbugs/blob/d999c35c134f5310b688f2b4513e3689fb1439bc/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ClassContext.java#L247) -> [ClassInfo#getXMethodsInCallOrder()](https://github.com/spotbugs/spotbugs/blob/d999c35c134f5310b688f2b4513e3689fb1439bc/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/analysis/ClassInfo.java#L245) -> [SelfMethodCalls#interestingSignature](https://github.com/spotbugs/spotbugs/blob/d999c35c134f5310b688f2b4513e3689fb1439bc/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/SelfMethodCalls.java#L65)

Having this example class:

```Test.java
public class Test {

	static {
		staticMethod1();
	}

	public static void staticMethod1(){
		staticMethod2();
	}
	public static void staticMethod2(){
		staticMethod3();
	}
	public static void staticMethod3(){
		staticMethod4();
	}
	public static void staticMethod4(){
		new Test().method1();
	}

	public void method1() {
		method2();
	}
	public void method2() {
		method3();
	}
	public void method3() {
		method4();
	}
	public void method4() {
		System.out.println("HERE WE GO!");
	}
}
```

The call order is derived as:
```
=========
= ClassContext.getMethodsInCallOrder() 
=========
staticMethod2
<init>
staticMethod1
staticMethod4
staticMethod3
method4
method3
method2
<clinit>
method1
```

Commenting out the private static thing then the topologically sorted call tree looks correct:
```
=========
= FIXED: 
=========
<init>
method4
method3
method2
method1
staticMethod4
staticMethod3
staticMethod2
staticMethod1
<clinit>
```

topolik:
Actually there's another bug. 

[...]

PR #3160, opened 2024-10-14
Fixing #3159 - Method call graph ignores constructors and non-parametric void methods

topolik:
Looks like the problem is there [from the very beginnings](https://github.com/spotbugs/spotbugs/commit/a5b3c2de4045e2f94858ccac2a4543f480265f3b) where the logic ignored all non-parametric methods as "not interesting". 

Omitting those methods from the MultiMap doesn't give TopologySort any information about the graph orientation and these methods are randomly misplaced when calling `ClassContext#getMethodsInCallOrder`. 

Part of the PR is a test that has a simple call stack from the default `static{}` constructor `<clinit>()V` through several static methods to a non-parametric class constructor, one more constructor and then instantiated class methods which creates a very simple DAG that any topological sorting algorithm should have no problem analyzing.

After fixing the `interestingSignature` method from ` return !"()V".equals(signature);` to `return true;`, it works! :) 

FYI: required for proper FindSecurityBugs taint analysis of derived methods. Without proper order they are skipped by the analysis engine as unknown.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/classfile/MethodsCallOrderTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/classfile/MethodsCallOrderTest.java
@@ -88,4 +88,35 @@ class MethodsCallOrderTest {
 
         assertEquals(expected, actual);
     }
+
+    @Test
+    void testShortcutPathTest() {
+        try {
+            load("../spotbugsTestCases/build/classes/java/main/MethodsCallOrder$ShortcutPathTest.class");
+        } catch (CheckedAnalysisException | IOException | InterruptedException e) {
+            fail("Unable to add MethodsCallOrder$SimpleTest.class to the build: " + e.getMessage(), e);
+        }
+
+        ClassDescriptor classDescriptor = DescriptorFactory.createClassDescriptor("MethodsCallOrder$ShortcutPathTest");
+
+        List<String> actual = new ArrayList<>();
+
+        try {
+            ClassContext classContext = Global.getAnalysisCache().getClassAnalysis(ClassContext.class, classDescriptor);
+            for (Method method : classContext.getMethodsInCallOrder()) {
+                actual.add(method.getName() + method.getSignature());
+            }
+        } catch (CheckedAnalysisException e) {
+            fail("Unable to get ClassContext analysis for MethodsCallOrder: " + e.getMessage(), e);
+        }
+
+        List<String> expected = Arrays.asList(new String[] {
+            "methodC()V",
+            "methodB()V",
+            "methodA()V",
+            "<init>()V"
+        });
+
+        assertEquals(expected, actual);
+    }
 }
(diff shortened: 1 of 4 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/classfile/MethodsCallOrderTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * @author Tomas Polesovsky
+ */
+public class MethodsCallOrderTest {
```
JuditKnoll, line 21: In Junit Jupiter the test classes have package private visibility. Can you please remove the `public` to not overwrite this?

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/engine/SelfMethodCalls.java
```diff
@@ -41,7 +41,7 @@
 public class SelfMethodCalls {
 
     static private boolean interestingSignature(String signature) {
-        return !"()V".equals(signature);
+        return true;
     }
```
JuditKnoll, line 45: As far as I can see, this function is only used once, in line 65. Since it will always return true, IMO it's better to remove this part of the condition.

JuditKnoll (changes requested):
Thank you for your contribution!
Can you please add one more test, where one of the functions called multiple times in a way that it could have affect on the order?

Can you please add an entry to the `CHANGELOG.md` file's unreleased part?

topolik:
> Can you please add one more test, where one of the functions called multiple times in a way that it could have affect on the order?

Hi @JuditKnoll , just to confirm my understanding, I can think of 2 different kinds that impacts the order.

First case is when `A -> C` and `B -> C`. In such scenario we call C twice. C goes first and  then A and B go next as second together, so we don't care about the order of A and B.

Another case is `A -> C` and `A -> B -> C`. In this scenario C goes first, then B and then A. We call C twice but the order in the call graph is fixed.

Thank you

JuditKnoll:
> > Can you please add one more test, where one of the functions called multiple times in a way that it could have affect on the order?
> 
> Hi @JuditKnoll , just to confirm my understanding, I can think of 2 different kinds that impact the order.
> 
> First case is when `A -> C` and `B -> C`. In such scenario we call C twice. C goes first and then A and B go next as second together, so we don't care about the order of A and B.
> 
> Another case is `A -> C` and `A -> B -> C`. In this scenario C goes first, then B and then A. We call C twice but the order in the call graph is fixed.
> 
> Thank you

I was thinking about the second scenario.

topolik:
Hi @JuditKnoll,

please see the new commits.

Thanks.

JuditKnoll (changes requested):
Can you please add an entry to the `CHANGELOG.md` file's unreleased part?

topolik:
Ops, forgot. Thank you @JuditKnoll.

topolik:
There was a conflict in CHANGELOG.md. Rebased.

JuditKnoll (approved):

gtoison (approved):

hazendaz (approved):

hazendaz (approved):

PR #3187, opened 2024-11-06
Small fixes in BugInstance and BugInstanceMatcher

JuditKnoll:
This PR should not cause any change in functionality.
I haven't added any changelog entry, since I don't think this type of fix needs it.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
@@ -157,7 +157,7 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
      */
     private static boolean adjustExperimental;
 
-    private static Set<String> missingBugTypes = Collections.synchronizedSet(new HashSet<String>());
+    private static final Set<String> missingBugTypes = Collections.synchronizedSet(new HashSet<>());
 
     public static class NoSuchBugPattern extends IllegalArgumentException {
         private static final long serialVersionUID = 1L;
@@ -202,12 +202,6 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
     public static DateFormat firstSeenXMLFormat() {
         return new SimpleDateFormat("M/d/yy h:mm a", Locale.ENGLISH);
     }
-
-    /*
-    private boolean isFakeBugType(String type) {
-        return "MISSING".equals(type) || "FOUND".equals(type);
-    }
-     */
 
     private void boundPriority() {
         priority = boundedPriority(priority);
@@ -540,7 +534,7 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
                 + " must contain at least one class, method, or field annotation");
     }
 
-    public Collection<? extends SourceLineAnnotation> getAnotherInstanceSourceLineAnnotations() {
+    public Collection<SourceLineAnnotation> getAnotherInstanceSourceLineAnnotations() {
         // Highest priority: return the first top level source line annotation
         Collection<SourceLineAnnotation> result = new ArrayList<>();
         for (BugAnnotation annotation : annotationList) {
@@ -595,7 +589,7 @@ public class BugInstance implements Comparable<BugInstance>, XMLWriteable, Clone
     /**
      * Get an Iterator over all bug annotations.
      */
-    public List<? extends BugAnnotation> getAnnotations() {
+    public List<BugAnnotation> getAnnotations() {
         return annotationList;
     }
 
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -1044,10 +1038,7 @@ public BugInstance addSuperclass(PreorderVisitor visitor) {
      *
      * <p>
      * For information on type descriptors, <br>
-     * see http://java.sun.com/docs/books/vmspec/2nd-edition/html/ClassFile.doc.
-     * html#14152 <br>
```
JuditKnoll, line 1048: This page doesn't exist anymore.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -1044,10 +1038,7 @@ public BugInstance addSuperclass(PreorderVisitor visitor) {
...
-     * or http://www.murrayc.com/learning/java/java_classfileformat.shtml#
-     * TypeDescriptors
+     * see <a href="https://www.murrayc.com/permalink/1998/03/13/the-java-class-file-format/#TypeDescriptors">TypeDescriptors</a>.
```
JuditKnoll, line 1041: This link forwards to the new link.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -1617,11 +1608,7 @@ public BugInstance addSourceLine(SourceLineAnnotation sourceLine) {
      */
     @Nonnull
     public BugInstance addSourceLine(BytecodeScanningDetector visitor, int pc) {
-        SourceLineAnnotation sourceLineAnnotation = SourceLineAnnotation.fromVisitedInstruction(visitor.getClassContext(),
-                visitor, pc);
-        if (sourceLineAnnotation != null) {
-            add(sourceLineAnnotation);
-        }
+        add(SourceLineAnnotation.fromVisitedInstruction(visitor.getClassContext(), visitor, pc));
```
JuditKnoll, line 1611: All of these cases, the variable can never be null, so the null check is not necessary.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -1880,7 +1842,7 @@ public String getMessageWithoutPrefix() {
     }
 
     String getLongDescription() {
-        return getBugPattern().getLongDescription().replaceAll("BUG_PATTERN", type);
+        return getBugPattern().getLongDescription().replace("BUG_PATTERN", type);
```
JuditKnoll, line 1845: The first argument doesn't contain any regex, so we can use `replace()` instead of `replaceAll()`.

test-harness/src/main/java/edu/umd/cs/findbugs/test/matcher/BugInstanceMatcher.java
```diff
@@ -109,8 +109,8 @@ public boolean matches(Object obj) {
                 String fullName = classAnn.getClassName();
                 int startDot = fullName.lastIndexOf(".") + 1;
                 int endDollar = fullName.indexOf('$');
-                String simpleName = fullName.substring(startDot != -1 ? startDot : 0, endDollar != -1 ? endDollar : fullName.length());
-                String simpleNameInner = fullName.substring(startDot != -1 ? startDot : 0, fullName.length());
+                String simpleName = fullName.substring(startDot, endDollar != -1 ? endDollar : fullName.length());
+                String simpleNameInner = fullName.substring(startDot);
```
JuditKnoll, line 113: The value of `startDot` can never be -1, since the smallest value of `lastIndexOf()` is -1, but there is a +1.

hazendaz (approved):

hazendaz:
> This PR should not cause any change in functionality. I haven't added any changelog entry, since I don't think this type of fix needs it.

Since appendText of comma is a fix to the description, I think a change log for that would be nice.  Otherwise agree rest is just cleanup.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -2551,13 +2512,9 @@ public List<BugAnnotation> getAnnotationsForMessage(boolean showContext) {
...
 
         for (BugAnnotation b : getAnnotations()) {
-            if (primaryAnnotations.contains(b)) {
-                continue;
-            }
-            if (b instanceof LocalVariableAnnotation && !((LocalVariableAnnotation) b).isNamed()) {
-                continue;
-            }
-            if (b instanceof SourceLineAnnotation && ((SourceLineAnnotation) b).isUnknown()) {
+            if (primaryAnnotations.contains(b)
```
ThrawnCA, line 2515: Alternatively, flip the logic to "if not any of these then add".

ThrawnCA (approved):

ThrawnCA (approved):

hazendaz (approved):

PR #3193, opened 2024-11-11
Fix FP UUF_UNUSED_FIELD with MethodHandles (#2749)

### LINKED ISSUE #2749: False positive UUF_UNUSED_FIELD with pure VarHandles use
The change in #2326 ends up reporting "false" positives when a private field is solely used via VarHandles, for example [here](https://github.com/opendaylight/yangtools/blob/master/common/concepts/src/main/java/org/opendaylight/yangtools/concepts/AbstractRegistration.java), which reports no warnings with SpotBugs-4.7.3.

Unfortunately SpotBugs cannot see the `@SuppressWarnings("unused")` annotation due to it being `RetentionPolicy.SOURCE`, leading to the need to issue a `@SuppressFBWarnings("UUF_UNUSED_FIELD")`.

That not only seems redundant, but also has a class file impact.

Recognize references via `Lookup.findVarHandle()` and `Atomic{Reference,...}FieldUpdater.newUpdater()` arguments as valid references as long as the returned value is not ignored.

rovarga:
Flag fields accessed via Lookup.findVarHandle(), .findGetter() and
.findSetter() as being reflective access -- just like ARFU et al.


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2749Test.java
```diff
@@ -0,0 +1,24 @@
...
+class Issue2749Test extends AbstractIntegrationTest {
+    @Test
+    @DisabledOnJre(JRE.JAVA_8)
+    void testIssue() {
+        performAnalysis(
+                "../java11/ghIssues/Issue2749$WithMethodHandles.class",
+                "../java11/ghIssues/Issue2749$WithVarHandle$Value.class",
+                "../java11/ghIssues/Issue2749$WithVarHandle.class",
+                "../java11/ghIssues/Issue2749.class");
+        assertThat(getBugCollection(), is(emptyIterable()));
```
JuditKnoll, line 22: Instead of an empty iterable can you please specify the relevant bugcode(s)? For this there are some assert functions in AbstractIntegrationTest, e.g. [assertNoBugType()](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java#L158). 
This would also help in the build failing because of the failing tests.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2749Test.java
rovarga: Understood, I will be pushing more patches to update the implementation and tests.

Nevertheless, I want to keep this assertion in place to the extent possible. At the end of the day, I want SpotBugs to to acknowledge parts of Issue2749 as fully-compliant code -- i.e. as a downstream user I want to know that this code example is fully valid and SpotBugs won't start complaining about it in the future.

I understand that is not how the test suite works right now, but as a downstream user I need SpotBugs to start providing these sorts of guarantees for Real Code(tm).

PR #3209, opened 2024-11-28
Use SignatureParser to extract the arguments of a method

### LINKED ISSUE #3208: Sonar scanner raises IllegalArgumentExceptions for 20 findbug detectors on one class
## Issue Description

Sonar scanner raises IllegalArgumentException for 20 findbugs detectors for one class. The error message "can't push void" does not help me to see the reason. Therefore I cannot fix the class source code.

See log extract.
[sonar-scanner-java-findbugs-exception-log.txt](https://github.com/user-attachments/files/17916506/sonar-scanner-java-findbugs-exception-log.txt)

## Environment

| Component          | Version |
| ------------------ | ------- |
| SonarQube          | Community Edition - Version 9.9.3 (build 79811)  |
| sonar scanner     | 5.0.0.4389 |
| Sonar-FindBugs     | findbugs 4.2.9   |
| Maven              | mvnw 3.9.9   |
| Java               | RedHat_JDK17.0.5.0.8   |

gtoison:
ClassInfo.Builder.arguments() was incorrectly extracting the arguments from a method signature: for instance the output for "(Xyz)V" was "(" instead of the expected "Xyz".
As a result the ClassInfo.Builder.build() was incorrectly matching methods with same name (but different arguments) as bridge methods. This caused errors ("Can't push void") in
TypeFrameModelingVisitor.pushValue() when the mismatched method had a void return type.

I tried making a unit test reproducing the problem in #3208 but was not able to extract the problematic case.
This should fix #3208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix call graph, include non-parametric void methods ([#3160](https://github.com/spotbugs/spotbugs/pull/3160))
 - Added missing comma between line number and confidence when describing matching and mismatching bugs for tests ([#3187](https://github.com/spotbugs/spotbugs/pull/3187))
 - Fixed method matchers with array types ([#3203](https://github.com/spotbugs/spotbugs/issues/3203))
+- Fixed an error in the detection of bridge methods causing analysis crashes ([#3208](https://github.com/spotbugs/spotbugs/issues/3208))
 
 ### Cleanup
 - Cleanup thread issue and regex issue in test-harness ([#3130](https://github.com/spotbugs/spotbugs/issues/3130))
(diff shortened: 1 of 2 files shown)


JuditKnoll:
This change looks like a good fix, however I didn't go into so deep to verify that it fixes the mentioned issue.
I would love to see some kind of test for this to show the fixed behavior (testing the `build()` method of `ClassInfo` with two functions with the same name, but different parameters).

ThrawnCA (approved):
This seems like an improvement over the current behaviour even if it turns out to need further work.

gtoison:
> This change looks like a good fix, however I didn't go into so deep to verify that it fixes the mentioned issue. I would love to see some kind of test for this to show the fixed behavior (testing the `build()` method of `ClassInfo` with two functions with the same name, but different parameters).

I tried a lot of things but could not seem to reproduce the issue as a unit test. I'm suspecting the order of the classes might be a factor here, but clearly I don't 100% understand how the abstract bridge methods are processed.
The current behaviour was clearly off though, and while debugging the reproducer I could see how that created the issue.
The method `void toSerializable(final LogEvent event, final Writer writer)` was incorrectly added as a bridge method of `T toSerializable(LogEvent event)` and since the return type was void (instead of `T`), the type analysis was trying to push that void onto the stack.

JuditKnoll (approved):

Issue #3099, opened 2024-08-30
The MacOS builds fail


gtoison:
It seems that all the MacOs builds fail ater 360 minutes, the last step is:
```
> Task :eclipsePlugin:generateCandidateP2Metadata
signing eclipsePlugin/build/site/eclipse-candidate/plugins/com.github.spotbugs.plugin.eclipse_4.9.0.202408291622-326ffe1.jar
signing eclipsePlugin/build/site/eclipse-candidate/features/com.github.spotbugs.plugin.eclipse_4.9.0.202408291622-326ffe1.jar
Error: The operation was canceled.
```
So apparently there's something blocking the build progress (maybe a popup needs to be clicked???)
If someone who has a Mac could try investigating this it would be great!

nrayburn-tech:
Followed the steps from build.yml and below is the popup I ended up receiving.  I ran `./gradlew spotlessCheck build smoketest --no-daemon --scan` with Java 21 on my path (tried with Java 11 too).  I tried configuring `eclipsePlugin/build.gradle` to use a different version within the `eclipse.jdt` section, but I couldn't get that to change anything.

![image](https://github.com/user-attachments/assets/d3bc2400-1323-450e-8f3b-8023b4cf7305)

nrayburn-tech:
Searching around, it seems like Eclipse searches for the JVM to use.  It looks like the Eclipse version used for CI tests is a bit older, so maybe just simply updating to the latest would resolve this.  Not sure if there's a reason to test against the older build here or not.  Another option would be to specify which JVM eclipse should use and do that for CI (not sure how to do this, my attempts didn't work).

hazendaz:
Reason we are on older eclipse is due to newer needing higher jdk.  Error message is odd.  Although I thought mac os dropped java 8.  Many thanks though showing what it's facing now.  Maybe we just upgrade now.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hazendaz:
@nrayburn-tech I've moved the eclipse portions to java 11 now but seems the same problem, is it possible you could test our master again to see if the issue is the same?

hazendaz:
While PR #3258 was expected to fix the issue, its still getting stuck in same place.  Reopened for now as this is continued to look at.

iloveeclipse:
Is this the reason why I constantly see following on PR's:

[...]

PR #3214, opened 2024-12-02
Change macos-latest to macos-13 in GHA

JuditKnoll:
The current macos-latest gha is macos-14 with the softwares of [macos-14-arm64](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) according to the github actions [runner images repo](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images). This image does [not contain java 8](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#java), which is currently still seems to be necessary for the build of the eclipse plugin.
The [macos-13](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#java) image has java 8. 
With using macos-13 instead of macos-latest the [build job is successful](https://github.com/JuditKnoll/spotbugs/actions/runs/12121006111/job/33790996092), not only because of timeout.

The [macos-14](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#java) image also has java 8, which is the macos-latest-large or macos-14-large image, but "[larger runners are only available for organizations and enterprises using the GitHub Team or GitHub Enterprise Cloud plans](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners)" and not free.

This is a workaround, a temporary fix for https://github.com/spotbugs/spotbugs/issues/3099.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-13]
         java: [11, 17, 21, 23]
       max-parallel: 6
     runs-on: ${{ matrix.os }}


hazendaz:
Lets hang on this only because I moved Eclipse back to be compliant with java 8 but there is nothing preventing moving it back forwards.  Just have to look up which version was java 11 based.  I'll try to do that tonight so we won't need this one.  If I'm able to do that will close this, if something is blocking, will go with this.

hazendaz:
@JuditKnoll converted to draft and will remove once mac issue is fully figured out.  I put the build to java 11 in eclipse area but that didn't do the trick.  I'm going to clear all my caches tomorrow and try to see what else is missed.  I've already found some items.  If this cannot be resolved by Thursday night release, the release will move forwards anyways.  The impact of this is that it really puts pressure on our project on github, I've went through, rebased everything and subsequently cleared our caches here up to last few builds and cancelled all running builds so we are a mostly clean slate for now.

JuditKnoll:
> @JuditKnoll converted to draft and will remove once mac issue is fully figured out. I put the build to java 11 in eclipse area but that didn't do the trick. I'm going to clear all my caches tomorrow and try to see what else is missed. I've already found some items. If this cannot be resolved by Thursday night release, the release will move forwards anyways. The impact of this is that it really puts pressure on our project on github, I've went through, rebased everything and subsequently cleared our caches here up to last few builds and cancelled all running builds so we are a mostly clean slate for now.

This is only a workaround, a temporary fix, not a proper solution. When there is a proper solution, feel free to close this PR. Thank you for looking into the issue. 😄

iloveeclipse:
Let merge this please. Each PR I've seen at Spotbugs hangs for no reason building Mac and finally fails on Mac. I have no Mac, will never have one and will never be able to debug or fix that. If the provided PR fixes the issue but somewhat not ideal (why?), let's merge to make every PR happy and search for a better solution afterwards.

hazendaz:
Ok, merging.  I tried to resolve it.  Unclear why it fails.  There are definitely some general setup issues throughout the build that need some tlc.  This does no harm, only good for now.  Will revisit upgrades later after reviewing other incorrect gradle issues I'm seeing, such as annotations package claiming java 5 compiled, but it's byte code disagrees, so probably same here with something still dragging java 8.

hazendaz (approved):

hazendaz:
To more clearly answer as another user did debug for us.  It's popping java 8 not found prompt, but Mac latest doesn't have it any longer.  Since it's a ui to user, the action locks in non interactive mode until timeout.  Mac 13 still has it.  Why it's needed is strange since eclipse version used is java 11.  Maybe it goes away on its own when we bump up to 17.

PR #3217, opened 2024-12-03
Cleanup: use Boolean.parseBoolean for string-to-boolean conversion

JuditKnoll:
Fixing "use `Boolean.parseBoolean()` for string-to-boolean conversion" issues.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
@@ -773,7 +773,7 @@ public class PluginLoader implements AutoCloseable {
 
                 try {
                     String propertiesLocation = componentNode.valueOf("@properties");
-                    boolean disabled = Boolean.valueOf(componentNode.valueOf("@disabled"));
+                    boolean disabled = Boolean.parseBoolean(componentNode.valueOf("@disabled"));
 
                     Node filterMessageNode = findMessageNode(messageCollectionList,
                             "/MessageCollection/PluginComponent[@id='" + componentId + "']",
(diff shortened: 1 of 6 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
```diff
@@ -851,7 +851,7 @@ private void loadPluginComponents()
                 String reports = detectorNode.valueOf("@reports");
                 String requireJRE = detectorNode.valueOf("@requirejre");
                 String hidden = detectorNode.valueOf("@hidden");
-                if (speed == null || speed.isEmpty()) {
+                if (speed == null || speed.length() == 0) {
```
ThrawnCA, line 854: Why replace `isEmpty`? It seems like the better choice, O(1) instead of O(n).

JuditKnoll: Oops, this got back when merging. Thanks for highlighting it!

hazendaz (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java
```diff
@@ -851,7 +851,7 @@ private void loadPluginComponents()
                 String reports = detectorNode.valueOf("@reports");
                 String requireJRE = detectorNode.valueOf("@requirejre");
                 String hidden = detectorNode.valueOf("@hidden");
-                if (speed == null || speed.isEmpty()) {
```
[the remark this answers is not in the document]
JuditKnoll: When I merged master there was some conflict with the changes from #3219 in this file, I didn't pay enough attention and this change creeped back, so I modified this once more to be in sync with master.

PR #3220, opened 2024-12-03
Cleanup: remove unnecessary throws declarations from methods

JuditKnoll:
Removing unnecessary throws declarations from methods, where the method body doesn't throw the listed exception, or a more general exception already listed.

See the issues in [SonarCloud](https://sonarcloud.io/project/issues?rules=java%3AS1130&issueStatuses=OPEN%2CCONFIRMED&id=com.github.spotbugs.spotbugs&open=AVzpu0ArZm7fF7LkQ57c).

### DIFF
--- a/eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
@@ -367,8 +367,6 @@ public class FindBugsWorker {
             // props
             st.newPoint("storeBugCollection");
             FindbugsPlugin.storeBugCollection(project, resultCollection, monitor);
-        } catch (IOException e) {
-            FindbugsPlugin.getDefault().logException(e, "Error performing SpotBugs results update");
         } catch (CoreException e) {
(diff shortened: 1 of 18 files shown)


eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
```diff
@@ -367,8 +367,6 @@ private void updateBugCollection(Project findBugsProject, Reporter bugReporter,
             // props
             st.newPoint("storeBugCollection");
             FindbugsPlugin.storeBugCollection(project, resultCollection, monitor);
-        } catch (IOException e) {
-            FindbugsPlugin.getDefault().logException(e, "Error performing SpotBugs results update");
         } catch (CoreException e) {
             FindbugsPlugin.getDefault().logException(e, "Error performing SpotBugs results update");
         }
```
JuditKnoll, line 372: The IOException was catched from the `FindbugsPlugin.storeBugCollection()` method, which does not throw this exception. Also, the two catch blocks are the same.

Issue #1515, opened 2021-04-16
[JEP 411] Deprecate `DoInsideDoPrivileged` (`DP_DO_INSIDE_DO_PRIVILEGED` + `DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED`)


jglick:
Assuming [JEP 411](https://openjdk.java.net/jeps/411) is accepted, `DoInsideDoPrivileged` will be issuing recommendations to insert API calls which are slated to be officially deprecated and either deleted or made no-ops. Suggest this detector be removed, or disabled by default. https://github.com/spotbugs/spotbugs/blob/43001a94202c4d041880c40224002b2f2b758a95/spotbugs/etc/messages.xml#L2386-L2418

pfirmstone:
Um, please keep this, we don't know what the replacement Security API is yet, my project uses Java Security API's as well as Spotbugs to identify bugs in the build.


People can ignore the issue themselves.

jglick:
> we don't know what the replacement Security API is yet

JEP 411 does not propose a replacement. The whole point is that the concept is outdated and more trouble than it is worth.

To support projects which are remaining on current Java versions indefinitely, it would make sense to merely disable this by default. Or perhaps it could be disabled when the Java target (bytecode) level is 17+ (or whenever JEP 411 lands)?

pfirmstone:
I understand if it's too much trouble for you, however we are still developing for this and all you need to do is use the <excludeFilterFile> and it won't bother you again.   The impact of missing this bug matters to us.

A Principle of Least Privilege (POLP) authentication layer is not an outdated concept.   We are experiencing increasing demand.  The way that OpenJDK documents the way to use Security policy files is an outdated concept however, they only have trust completely or not at all, so that trusted code is granted AllPermission and they try and sandbox untrusted code, so it doesn't constrain their trusted code at all.   We don't allow untrusted code into the system, but we limit trusted code (and roles) to only what they need to do, if we have a bug in our software, something that's not obvious, POLP limits the damage an attacker can do.

We also have highly scalable security manager, policy implementations, and tools to generate POLP policy files, the only argument in JEP 411 that applies to us is the maintenance burden.  All other arguments are based on the Java implementations which are crap and old outdated ideas like trusted code with AllPermission.

POLP policy files only grant permissions which are absolutely necessary to perform intended functions.  Nothing else is allowed, so this has a big benefit to security.

[...]

Issue #28, opened 2016-11-15
Make SpecialKind handling thread-safe


KengoTODA:
Legacy way to handle `SpecialKind` in `OpcodeStack.Item` is not thread-safe. Delete these deprecated API and make new one (added by #27) thread-safe.

## TODO

- [ ] delete `OpcodeStack.Item#specialKindNames` static field, which is public and non-thread safe
- [ ] delete `OpcodeStack.Item#defineNewSpecialKind(String)` static method, which returns wrong value
- [ ] add a `ConcurrentMap` private static field to `OpcodeStack.Item`
- [ ] use new static field in  `OpcodeStack.Item#defineSpecialKind(String)` and  `OpcodeStack.Item#getSpecialKindName(int)` to improve thread safety

refs: #23, #27

Issue #25, opened 2016-11-13
Replace logo within source


jsotuyod:
Create proper size copies of the logo to replace the FindBugs one as per comment: https://github.com/spotbugs/spotbugs/pull/24/files#r87721532

PR #3225, opened 2024-12-10
Do not report bugs advising using deprecated doPrivileged on Java 17+

JuditKnoll:
The `java.security.AccessController` class got deprecated in Java 17 (see [JEP 411](https://openjdk.org/jeps/411)), and will be removed in Java 24 (see [JEP 486](https://openjdk.org/jeps/486)). The `DP_DO_INSIDE_DO_PRIVILEGED`, `DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED` and `USC_POTENTIAL_SECURITY_CHECK_BASED_ON_UNTRUSTED_SOURCE` bugs advise using the `java.security.AccessController::doPrivileged` methods. This PR modifies the relevant detectors to not report these bugs, if the class is targeting Java 17 or above.

The `VSC_VULNERABLE_SECURITY_CHECK_METHODS` bug is also reported in connection with Security Manager - which got deprecated the same time - , but only if the code already uses the class `java.lang.SecurityManager`, so it doesn't cause FPs at our users.

This PR fixes https://github.com/spotbugs/spotbugs/issues/1515.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fixed `FI_FINALIZER_NULLS_FIELDS` FPs for functions called finalize() but not with the correct signature. ([#3207](https://github.com/spotbugs/spotbugs/issues/3207))
 - Fixed an error in the detection of bridge methods causing analysis crashes ([#3208](https://github.com/spotbugs/spotbugs/issues/3208))
 - Fixed detector `ThrowingExceptions` by removing false positive reports, such as synthetic methods (lambdas), methods which inherited their exception specifications and methods which call throwing methods ([#2040](https://github.com/spotbugs/spotbugs/issues/2040))
+- Do not report `DP_DO_INSIDE_DO_PRIVILEGED`, `DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED` and `USC_POTENTIAL_SECURITY_CHECK_BASED_ON_UNTRUSTED_SOURCE` in code targeting Java 17 and above, since it advises the usage of deprecated method ([#1515](https://github.com/spotbugs/spotbugs/issues/1515)).
 
 ### Cleanup
 - Cleanup thread issue and regex issue in test-harness ([#3130](https://github.com/spotbugs/spotbugs/issues/3130))
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -2582,33 +2582,42 @@ will need to be changed in order to compile it in later versions of Java.</p>
     <ShortDescription>Method invoked that should be only be invoked inside a doPrivileged block</ShortDescription>
     <LongDescription>Invocation of {2}, which should be invoked from within a doPrivileged block, in {1}</LongDescription>
     <Details>
-<![CDATA[
-  <p> This code invokes a method that requires a security permission check.
-  If this code will be granted security permissions, but might be invoked by code that does not
-  have security permissions, then the invocation needs to occur inside a doPrivileged block.</p>
-]]>
+      <![CDATA[
+        <p> This code invokes a method that requires a security permission check.
+        If this code will be granted security permissions, but might be invoked by code that does not
+        have security permissions, then the invocation needs to occur inside a doPrivileged block.</p>
+        The <code>java.security.AccessController<\code> class, which contains the <code>doPrivileged</code> methods,
+        got deprecated in Java 17 (see <a href="https://openjdk.org/jeps/411">JEP 411</a>), and removed in Java 24 (see <a href="https://openjdk.org/jeps/486">JEP 486</a>).
+        For this reason, this bug isn't reported in classes targeted Java 17 and above.
+      ]]>
     </Details>
   </BugPattern>
   <BugPattern type="DP_DO_INSIDE_DO_PRIVILEDGED"> <!-- misspelled for backward compatibility -->
     <ShortDescription>Method invoked that should be only be invoked inside a doPrivileged block</ShortDescription>
     <LongDescription>Invocation of {2}, which should be invoked from within a doPrivileged block, in {1}</LongDescription>
     <Details>
-<![CDATA[
-  <p> This code invokes a method that requires a security permission check.
-  If this code will be granted security permissions, but might be invoked by code that does not
-  have security permissions, then the invocation needs to occur inside a doPrivileged block.</p>
-]]>
+      <![CDATA[
+        <p> This code invokes a method that requires a security permission check.
+        If this code will be granted security permissions, but might be invoked by code that does not
+        have security permissions, then the invocation needs to occur inside a doPrivileged block.</p>
+        The <code>java.security.AccessController<\code> class, which contains the <code>doPrivileged</code> methods,
+        got deprecated in Java 17 (see <a href="https://openjdk.org/jeps/411">JEP 411</a>), and removed in Java 24 (see <a href="https://openjdk.org/jeps/486">JEP 486</a>).
+        For this reason, this bug isn't reported in classes targeted Java 17 and above.
+      ]]>
     </Details>
   </BugPattern>
   <BugPattern type="DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED">
     <ShortDescription>Classloaders should only be created inside doPrivileged block</ShortDescription>
     <LongDescription>{1} creates a {2} classloader, which should be performed within a doPrivileged block</LongDescription>
     <Details>
-<![CDATA[
-  <p> This code creates a classloader,  which needs permission if a security manage is installed.
-  If this code might be invoked by code that does not
-  have security permissions, then the classloader creation needs to occur inside a doPrivileged block.</p>
-]]>
+      <![CDATA[
+        <p> This code creates a classloader,  which needs permission if a security manage is installed.
+        If this code might be invoked by code that does not
+        have security permissions, then the classloader creation needs to occur inside a doPrivileged block.</p>
+        The <code>java.security.AccessController<\code> class, which contains the <code>doPrivileged</code> methods,
+        got deprecated in Java 17 (see <a href="https://openjdk.org/jeps/411">JEP 411</a>), and removed in Java 24 (see <a href="https://openjdk.org/jeps/486">JEP 486</a>).
+        For this reason, this bug isn't reported in classes targeted Java 17 and above.
+      ]]>
     </Details>
   </BugPattern>
   <BugPattern type="JCIP_FIELD_ISNT_FINAL_IN_IMMUTABLE_CLASS">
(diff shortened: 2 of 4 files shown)


gtoison (approved):

ThrawnCA (approved):

hazendaz (approved):

pfirmstone:
Is there a way to enable the bug checks for Java 17 and above?

pfirmstone:
So if I was to submit a patch which added that functionality, what would be a good convention to use?

Would it be something like this?

<FindBugsFilter>
    <Match>
        <Bug category="SECURITY-DEPRECATED"/>
    </Match>
</FindBugsFilter>

Or something else?

Thanks,
Peter.

We couldn't remove SecurityManager as it was integrated into our software architecture at a foundational level, we use it for authorization,  which includes limiting privileges to trusted code and authenticated user principals (privileges are not granted to either alone), we don't use it for sandboxing untrusted client code, rather than abandon 20 years of software development, we decided to fork OpenJDK mainline development instead, since it's a lot less work, our alternative was to rearchitect from the ground up, so far it's working out quite well, we've been able to address a lot of issues we couldn't previously...   Now we have the first JVM that prevents loading of untrusted or unsigned code, we've fixed the many race conditions in Permission implementation classes (had to remove Serialization support) and have a non blocking SecurityManager that uses Executors to compute permission checks in parallel and contains a non blocking cache to prevent repeating the same permission checks (eg in executor tasks).

Before anyone suggests it (we get this a lot), using ClassLoader's or Agents for isolation doesn't work for us.  It's funny how we've gone from years of strictly only using Java's public API, to now hacking its implementation code.  OpenSearch and ElasticSearch are still trying to decide how to replace SM, deprecation and removal have happened in a very short timeframe, leaving some developers scrambling for alternatives.   Postgreql's JDBC developers are also still trying to figure out what to do about it.   We simply just couldn't remove it an leave a plethora of security holes open for exploitation.

hazendaz:
@pfirmstone Wow you must have an incredible team that wants to fork java and write your own way.  Is that public?  If its not public, I'm not quite sure how or why spotbugs should support unknown usage like that.  We are always open for pull requests though but don't want you to waste your time on some feature that won't benefit others.  I don't think I personally would have forked java but rather just reused a few classes here and there.  I get it, their decisive just abandon on some features makes it tough.  As long as you think whatever you plan to do will benefit all not using your jvm too (assuming not public) then sure feel free to bring a PR. It may not land in 4.9.0 though.  I think since we got over last thing we needed we probably will be releasing this weekend (that is unless you can get something together in next 48 hours).  So maybe a 5.0.0 thing.

pfirmstone:
5.0.0's good.

Yeah, it's public, we tried and exhausted all other options first, but as it turns out, git makes forking relatively easy.   We have a master branch, which we perform reversions of SM removal commits in, make compatibility fixes and perform testing on to minimise conflicts, then once that passes we have a trunk branch where we have our own developments, so upstream changes are merged in like it's just another member of the development team.

We realised if we didn't do it now, it wouldn't be an option in 5 years when all the existing releases that support SM are no longer be available.   We're not trying to sandbox untrusted code, so it's not as difficult a task as what OpenJDK had set for themselves.  But we also wanted to make it much easier to use, high scaling and performant, and we've got that done already.  Most of the work is in maintaining permission hooks and doPrivileged calls throughout the JVM.   Since OpenJDK is making breaking changes, it's actually an opportunity to fix a few things that backward compatibility would have prevented.   So far we're 434 commits ahead of OpenJDK master with 3,088 files changed.

hazendaz:
ok feel free to submit a patch.  If you can point us to your openJdk fork that might be additionally helpful.  I'm certainly curious with so many fixes there they couldn't figure out ;)

pfirmstone:
I sent an email reply, but not sure if it was received:

I don't think it was a case of couldn't figure it out, more likely they were constrained by backward compatibility and budget, many of the implementation classes were from Java 1.2, prior to improvements in memory models, serialized form was already baked in, they maintained the original serial form, but needed different implementation code, this means you can't use final fields and all the good stuff that supports immutability, they had to make compromises, the original PermissionCollection implementations were heavily synchronized, so immutability didn't matter so much as race conditions weren't exposed.   To make SecurityManager and PolicyFile performant would have required completely rewriting all the Permission and PermissionCollection implementations and breaking Serialization backward compatibility.   It only came to light recently with JEP411 that OpenJDK developers thought SecurityManager was only used for Applets, have a look at some of our conversations on OpenJDK mail lists after they announced JEP411, they didn't think what we were doing was even possible, perhaps that's why it remained this way for so long.

When we first implemented ConcurrentPolicyFile in 2010, we ignorantly accepted the spec that stated Permission's were immutable, and shared them among threads as though they were immutable, this resulted in permission's not implying permissions they should and other problems like UnresolvedPermission not resolving, so we changed our code to call methods on Permission classes to trigger lazy initializers  etc, before publishing them to other threads.  We thread confined PermissionCollection's so they wouldn't cause contention.

CombinerSecurityManager had a timer based, weakly referenced non blocking cache of previously checked call stacks, it also broke permission checks up into multiple tasks for each ProtectionDomain, submitted to an ExecutorService, as permissions came in for checking, we had to ensure they were initialized properly before they were dispatched to pool threads.

When Policy was released in Java 1.2, you had to hand edit policy files, but this was a futile exercise, you had to try and figure out by trial and error the permissions required, never mind syntax errors etc, so people would just grant AllPermission to their "trusted code", but if code's granted AllPermission, then you can't limit user permission.  Some companies, like OpenSearch, ElasticSearch, Gigaspaces, JBoss etc had decent policy file implementations.   We generate policy files thousands of lines long based on least privilege principles, so the code or the user by themselves are not privileged, only when the user is using our trusted code will either gain privileges, then you just audit and make any changes to the policy files and deploy.

Modules are great, they've broken the JVM up into many ProtectionDomain's, significantly reducing the size of the trusted Java platform, provided AllPermission declarations are removed from <<JAVA_HOME>>/lib/security/java.policy

Here are some build artifacts, if you'd like to try one:

Linux x64 release artifacts https://github.com/pfirmstone/jdk-with-authorization/actions/runs/12327366841/artifacts/2320639575
Macos x64 release artifacts https://github.com/pfirmstone/jdk-with-authorization/actions/runs/12327366841/artifacts/2320634999
Macos-aarch64 release artifacts https://github.com/pfirmstone/jdk-with-authorization/actions/runs/12327366841/artifacts/2320606004
Windows x64 release artifacts https://github.com/pfirmstone/jdk-with-authorization/actions/runs/12327366841/artifacts/2320649800

New Policy and SecurityManager implementation code:

[jdk-with-authorization/src/java.base/share/classes/au/zeus/jdk at trunk · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/tree/trunk/src/java.base/share/classes/au/zeus/jdk)

Some other changes / improvements (not limited to):

[Refactor Permission and PermissionCollection classes · Issue #32 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/32)

[Replace java.security.Permissions with ConcurrentPermissions · Issue #25 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/25)

[Throw IllegalArgumentException if System::setSecurityManager argument is null. · Issue #34 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/34)

[Allow Policy to restrict loading of classes to Signed or message digest checked jar files. · Issue #5 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/5)

[ForkJoinPool threads - enable Permission support for SM. · Issue #28 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/28)

[Remove static permission assignment in platform code. · Issue #21 · pfirmstone/jdk-with-authorization](https://github.com/pfirmstone/jdk-with-authorization/issues/21)

Example of a generated policy file:

[JGDMS/qa/harness/policy/defaultsecuresharedvm.policy at trunk · pfirmstone/JGDMS](https://github.com/pfirmstone/JGDMS/blob/trunk/qa/harness/policy/defaultsecuresharedvm.policy)

Cheers,

Peter.

hazendaz:
Wow!  Thank you for this information.  It all makes sense now.  Thanks for taking the time to explain all this in such detail.

pfirmstone:
You're welcome to join our discord https://discord.gg/yhfGmaMu

ElasticSerach was able to avoid the recent Log4j vulnerability with good security policy.

We've made it fast and easy to use and configure, so anyone can use it.

The first step is to remove AllPermission from <<JAVA_HOME>>/lib/security/java.policy , we will remove them in future.

Policy is generated with property substitution, so you can stage in a test environment and then deploy to production, or just deploy for a short period of time with -Djava.security.manager=polpAudit it will continue to add permissions to policy as required, exercise your software's normal execution, if your jar's are signed and users are logging on, these are recorded in policy files, then review / audit policy and restart using -Djava.security.manager=default 

It will prevent loading of unauthorized code, unauthorized use of MethodHandles.Lookup.defineClass, agents, unused transitive dependencies, it will contain a whitelist of serialized classes, URL's, socket addresses, types of connections allowed, access to properties and permitted file access.

We're still working on adding some additional permissions for the Foreign Function API, but for now properties can be used to switch it off, however permissions will allow more flexibility for authorized use only.

It locks down the JVM by turning off unused features.  Even if an attacker manages to use an injection attack, it's unlikely to succeed as it will invoke unused execution paths that will trigger new permission checks, we still need to be scrutinizing data passed through to databases or native software, it doesn't defend against an SQL injection attack from an authorized user (eg if someone's user account was stolen).

hazendaz:
@JuditKnoll Do you recall why the motivation to not report this?  To me it seems we should not look for it where it no longer exists.  Until them I don't see why it needs to be removed.  The code is still valid and itself is flagged as deprecated.  This matters at least to junit 6 coming in.  I've got it working but the test for `PotentialSecurityCheckBasedOnUntrustedSource` fails due to this.  I can disable the test but then that is rather pointless and the code is in fact still valid so I don't see why we don't report the issue.  If anything, I think we should just indicate there is an issue and that while using it, it needs to be updated as detected or rewritten entirely (something like that).

hazendaz:
I'm going to push up on the junit 6 and will ping you over there as well as set to draft until figured out.

JuditKnoll:
> @JuditKnoll Do you recall why the motivation to not report this? To me it seems we should not look for it where it no longer exists. Until them I don't see why it needs to be removed. The code is still valid and itself is flagged as deprecated. This matters at least to junit 6 coming in. I've got it working but the test for `PotentialSecurityCheckBasedOnUntrustedSource` fails due to this. I can disable the test but then that is rather pointless and the code is in fact still valid so I don't see why we don't report the issue. If anything, I think we should just indicate there is an issue and that while using it, it needs to be updated as detected or rewritten entirely (something like that).

As far as I remember it was because of https://github.com/spotbugs/spotbugs/issues/1515 and https://openjdk.org/jeps/411. The relevant rules - `DP_DO_INSIDE_DO_PRIVILEGED` and `DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED` - recommend using deprecated methods over Java 17. AFAIK there is no replacement for doPrivileged in official Java API, nor a plan for it.
Since under Java 17, it can be still useful, and SpotBugs can analyze code under Java 17, the bugs were not reported from that Java version.
I had the impression that these bugs are annoying for more people than useful, so if these are reported on Java 17 and over, the detector could be disabled by default. (Especially that the `chooseVisitors` option is supported by both maven and gradle plugins, so it could be enabled if necessary.)

PR #3226, opened 2024-12-11
fix double reporting

### LINKED ISSUE #3185: Some bugs are counted twice in the class/package/project summary
Using Spotbugs 4.8.2.
When using the simple "println" checker from @iloveeclipse ([FindPrintlnDetector.java](https://github.com/iloveeclipse/SimpleSpotbugsDetectors/blob/master/SimpleSpotBugsDetectors/src/de/loskutov/fb/detectors/FindPrintlnDetector.java)), the bugs reported by the checker are reported twice under certain circumstances.

This happens, for example, when using spotbugs on the command line with XML output. In that case, the FindBugsSummary, PackageStats and ClasStats all count the bugs twice.
Curiously, this does not happen when using the Eclipse plugin and showing the XML generated by Eclipse.

Furthermore, it seems to only happen with checkers that extend `edu.umd.cs.findbugs.asm.ClassNodeDetector`.

Here is an example java file:
```java
class Test {
    public void doIt() {
        System.out.println("foobar");
    }
}
```

And here is an excerpt from the generated XML file:
```XML
<FindBugsSummary timestamp="Tue, 5 Nov 2024 15:19:21 +0100" total_classes="1" referenced_classes="15" total_bugs="2" total_size="6" num_packages="1" java_version="17.0.6" vm_version="17.0.6+10" cpu_seconds="3.04" clock_seconds="0.69" peak_mbytes="227.31" alloc_mbytes="7960.00" gc_seconds="0.00" priority_2="2">
    <PackageStats package="" total_bugs="2" total_types="1" total_size="6" priority_2="2">
      <ClassStats class="Test" sourceFile="Test.java" interface="false" size="6" bugs="2" priority_2="2"/>
    </PackageStats>
```

Note the total_bugs="2" and "priority_2="2" entries.

ttho:
Fixes #3185 

While reporting identical errors multiple times certainly is a bug of the client that is adding a BugInstance, nevertheless it should not cause a mismatch of statistics and the bug collection (the latter being a set).
See also https://github.com/iloveeclipse/SimpleSpotbugsDetectors/issues/1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR for fields initialized in method annotated with TestNG's @BeforeClass. ([#3152](https://github.com/spotbugs/spotbugs/issues/3152))
 - Fixed detector `FindReturnRef` not finding references exposed from nested and inner classes ([#2042](https://github.com/spotbugs/spotbugs/issues/2042))
 - Fix call graph, include non-parametric void methods ([#3160](https://github.com/spotbugs/spotbugs/pull/3160))
+- Fix multiple reporting of identical bugs messing up statistics ([#3185](https://github.com/spotbugs/spotbugs/issues/3185))
 
 ### Cleanup
 - Cleanup thread issue and regex issue in test-harness ([#3130](https://github.com/spotbugs/spotbugs/issues/3130))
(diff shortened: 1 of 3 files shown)


JuditKnoll (changes requested):
Can you please add a few simple tests showing the issue and an entry to the changelog?

gtoison (changes requested):
The changelog was modified since you forked off the project, can you please merge your change?

hazendaz (approved):

PR #3239, opened 2024-12-25
Correct syntax so the bug descriptions aren't corrupted

TimothyJones:
I noticed that the [bug descriptions](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html) are corrupt from [this point on](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#dp-method-invoked-that-should-be-only-be-invoked-inside-a-doprivileged-block-dp-do-inside-do-privileged).  It looks like this is because there's a `code` tag which is incorrectly terminated with `<\code>` instead of `</code>`.

This PR fixes the syntax.

I looked for the same error in the other translations, but it seems like they don't have `<\code>` in the `messages` files (only the correctly formed `</code>`).

### DIFF
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -2586,7 +2586,7 @@ will need to be changed in order to compile it in later versions of Java.</p>
         <p> This code invokes a method that requires a security permission check.
         If this code will be granted security permissions, but might be invoked by code that does not
         have security permissions, then the invocation needs to occur inside a doPrivileged block.</p>
-        The <code>java.security.AccessController<\code> class, which contains the <code>doPrivileged</code> methods,
+        The <code>java.security.AccessController</code> class, which contains the <code>doPrivileged</code> methods,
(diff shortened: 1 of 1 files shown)


TimothyJones:
It [looks much better](https://spotbugs--3239.org.readthedocs.build/en/3239/bugDescriptions.html#dp-classloaders-should-only-be-created-inside-doprivileged-block-dp-create-classloader-inside-do-privileged), but those entries might still be a little malformed as they're running into the heading of the next bug description.

TimothyJones:
Oh, I was just trying to follow what looked like the convention ([conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)) in the most recent commit messages. Can reopen with a different message if needed - on a closer look it looks like the conventional commits to spotbugs are mostly automated rather than humans.

iloveeclipse:
Just change message, amend commit and force push.

TimothyJones:
Done!

iloveeclipse (approved):
Thanks !

PR #3258, opened 2025-01-15
Move Eclipse to java 11 to match rest of the project

### LINKED ISSUE #3099: The MacOS builds fail

hazendaz:
fixes #3099 amd replaces #3214.  This was missed in our upgrade to java 11.

Given, our change log already states we moved to java 11, did not indicate this.  Further, we were already using java 11 based Eclipse but had it compiling its portion to java 8.  Other than fact we target one set of tests to java 8, I believe all of the build is otherwise property targeting java 11.

### DIFF
--- a/eclipsePlugin-test/build.gradle
+++ b/eclipsePlugin-test/build.gradle
@@ -34,7 +34,7 @@ dependencies {
 javadoc {
   doFirst {
     // This is supposed to enable everything except "missing" but doesn't work with gradle
-    // See http://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html
+    // See https://docs.oracle.com/en/java/javase/11/tools/javadoc.html
     options.addBooleanOption('Xdoclint:all,-missing', true)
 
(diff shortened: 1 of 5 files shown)


iloveeclipse (approved):

hazendaz:
Still locks up on mac, have a request out to user that was able to confirm the build issue with java 8 to see if that is still the same.  This otherwise is moving right direction.

PR #2072, opened 2022-05-27
Fix for Issue 2040

baloghadamsoftware:
This PR fixes issue #2040 by limiting the errors reported by the detector `ThrowingExceptions` to cases where
`java.lang.Exception` or `lava.lang.Throwable` appears in the exception specification of th method but it is neither
inherited from an overridden method nor is is coming from another method that the method invokes. Syntethic
methods are also omitted as well as methods which throw generic exceptions.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [X] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

Issue #2040, opened 2022-05-05
False positive for THROWS_METHOD_THROWS_CLAUSE_THROWABLE


alexpdp7:
Hi,

Updating to 4.7.0 ( https://github.com/alexpdp7/zqxjkcrud/pull/315 ) is causing the following warning for me:

```
[ERROR] Medium: Method lists Throwable in its throws clause. [net.pdp7.zqxjkcrud.dao.CatalogRepository] At CatalogRepository.java:[line 23] THROWS_METHOD_THROWS_CLAUSE_THROWABLE
```

But the line it's pointing to doesn't make sense?

https://github.com/alexpdp7/zqxjkcrud/blob/c247e3be1daf02dce1340eee436889b7c2fd1786/src/main/java/net/pdp7/zqxjkcrud/dao/CatalogRepository.java#L23

Cheers,

Álex

ThrawnCA:
I suspect that this is because Spotbugs doesn't handle lambdas well.

KengoTODA:
The [`ConnectionCallable#run()` method throws `Throwable`](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/ConnectionCallable.html#run(java.sql.Connection)) so the lambda in your code also declares that it throws `Throwable`.

I believe it is a problem of the interface provided by jOOQ. As a workground, you can filter or suppress the warning reported by SpotBugs, or replace the lamdba with an anonymous class like below:

```java
new ConnectionCallable<YourClass>() {
  @Override
  public YourClass run(Connection connection) /* no `throws Throwable` here */ {
    // your code
  }
}
```

Bananeweizen:
Labeling this as a third party bug seems wrong to me. Spotbugs should rather not raise that issue for lambdas used in third party methods. I get that warning hundreds of times for trivial AssertJ assertions like this:
```
assertThatExceptionOfType(SomeException.class).isThrownBy(() -> foo());
```
just because the "isThrownBy(...)" method _of course_ must handle every kind of throwable. I would expect the same to be reported for JUnit, Hamcrest and similar assertion libraries, if people start upgrading to spotbugs 4.7.0. Any change on my side to avoid the spotbugs issue just makes the code worse to read, and doesn't improve the security/fault tolerance or any other criteria.

KengoTODA:
I understand, and in the case of AssertJ, it is understandable to throw `Throwable`.

[...]

PR #3268, opened 2025-01-19
ThrowingExceptions detector crash when parsing method signature containing nested generics

### LINKED ISSUE #3267: 4.9.0 Regression: IllegalStateException in ThrowingExceptions detector when parsing method signature containing nested generics
After attempting to upgrade to 4.9.0, I started observing a regression throwing an ISE when analyzing some of our files.  Best guess is that this has to do with nested generics and the compilation settings of the file.

The existing stack trace is:

> The following errors occurred during analysis:
>   Exception analyzing {filename} using detector edu.umd.cs.findbugs.detect.ThrowingExceptions
>     java.lang.IllegalStateException: Invalid method signature: (Ljava/util/Map<Ljava/lang/String;Ljava/util/List<Ljava/lang/String;>;>;Ljava/io/PrintWriter;)V
>       At edu.umd.cs.findbugs.ba.SignatureParser$ParameterSignatureIterator.next(SignatureParser.java:132)
>       At edu.umd.cs.findbugs.ba.SignatureParser$ParameterSignatureIterator.next(SignatureParser.java:83)
>       At edu.umd.cs.findbugs.ba.SignatureParser.getParameterOffset(SignatureParser.java:59)
>       At edu.umd.cs.findbugs.ba.SignatureParser.getNumParameters(SignatureParser.java:208)
>       At edu.umd.cs.findbugs.detect.ThrowingExceptions.signatureMatches(ThrowingExceptions.java:170)
>       At edu.umd.cs.findbugs.detect.ThrowingExceptions.lambda$parentThrows$3(ThrowingExceptions.java:134)
>       At java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
>       At java.base/java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:1034)
>       At java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
>       At java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
>       At java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
>       At java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
>       At java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
>       At java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
>       At java.base/java.util.stream.ReferencePipeline.findAny(ReferencePipeline.java:652)
>       At edu.umd.cs.findbugs.detect.ThrowingExceptions.parentThrows(ThrowingExceptions.java:135)
>       At edu.umd.cs.findbugs.detect.ThrowingExceptions.visit(ThrowingExceptions.java:70)
>       At edu.umd.cs.findbugs.visitclass.BetterVisitor.visitMethod(BetterVisitor.java:351)
>       At org.apache.bcel.classfile.Method.accept(Method.java:115)
>       At edu.umd.cs.findbugs.visitclass.PreorderVisitor.doVisitMethod(PreorderVisitor.java:320)
>       At edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitJavaClass(PreorderVisitor.java:405)
>       At org.apache.bcel.classfile.JavaClass.accept(JavaClass.java:249)
>       At edu.umd.cs.findbugs.BytecodeScanningDetector.visitClassContext(BytecodeScanningDetector.java:38)
>       At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
>       At edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$1(FindBugs2.java:1100)
>       At java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
>       At edu.umd.cs.findbugs.CurrentThreadExecutorService.execute(CurrentThreadExecutorService.java:86)
>       At java.base/java.util.concurrent.AbstractExecutorService.invokeAll(AbstractExecutorService.java:247)
>       At edu.umd.cs.findbugs.FindBugs2.analyzeApplication(FindBugs2.java:1110)
>       At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:309)
>       At edu.umd.cs.findbugs.FindBugs.runMain(FindBugs.java:399)
>       At edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1223)
 
The signature of the method is:

`public void execute(Map<String, List<String>> parameters, PrintWriter output) throws Exception {...}`,

and it appears that the parser is complaining about the `Map<String, List<String>>` construct in the first parameter.

This file was able to be parsed in 4.8.6 without any errors being surfaced.

gtoison:
As reported in #3267 ThrowingExceptions causes errors with some methods with generics.

See #2072 
See #2040

This should fix #3267

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
@@ -33,6 +33,8 @@ import org.apache.bcel.generic.InvokeInstruction;
  * A simple class to parse method signatures.
  *
  * @author David Hovemeyer
+ *
+ * @deprecated This signature parser does not support generics, use GenericSignatureParser instead.
  */
 public class SignatureParser {
     private int totalArgumentSize;
@@ -112,7 +114,6 @@ public class SignatureParser {
                     break;
 
                 case 'L':
-                case 'T':
                     int semi = signature.indexOf(';', index + 1);
                     if (semi < 0) {
                         throw new IllegalStateException("Invalid method signature: " + signature);
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
@@ -36,6 +36,7 @@ import org.apache.bcel.generic.InvokeInstruction;
  *
  * @deprecated This signature parser does not support generics, use GenericSignatureParser instead.
  */
+@Deprecated
 public class SignatureParser {
     private int totalArgumentSize;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased - 2024-??-??
+## Unreleased - 2025-??-??
+
+### Fixed
+- Fixed the parsing of generics methods in `ThrowingExceptions` ([#3267](https://github.com/spotbugs/spotbugs/issues/3267))
 
 ## 4.9.0 - 2025-01-15
 ### Added
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3267Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3267Test.java
@@ -1,0 +1,15 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3267Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue3267.class",
+                "ghIssues/Issue3267BaseClass.class");
+
+        assertBugTypeCount("THROWS_METHOD_THROWS_CLAUSE_BASIC_EXCEPTION", 1);
+    }
+}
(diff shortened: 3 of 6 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
```diff
@@ -33,6 +33,8 @@
  * A simple class to parse method signatures.
  *
  * @author David Hovemeyer
+ *
+ * @deprecated This signature parser does not support generics, use GenericSignatureParser instead.
```
hazendaz, line 37: Can you also add deprecated annotation to match javadoc?  Or if that only applies to generics, it should us deprecated javadoc annotation.

gtoison: I've added the annotation.

This is a digression but I think we should consider requiring JDK soon (when it's GA), so we can use [JEP-484](https://openjdk.org/jeps/484)'s classfile API: among other things it has a signature parser and it would be a good replacement for these limited parsers.

hazendaz (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/SignatureParser.java
iloveeclipse: > I've added the annotation.
> 
> This is a digression but I think we should consider requiring JDK soon (when it's GA), so we can use [JEP-484](https://openjdk.org/jeps/484)'s classfile API: among other things it has a signature parser and it would be a good replacement for these limited parsers.

We just updated to Java 11 requirement, let us bump it to Java 17 and then to 21 first. After that we might consider to require Java 25 (next LTS), which is however not released yet... So I don't expect this API be available soon for SpotBugs.

PR #3281, opened 2025-01-29
StackOverFlowError in MultipleInstantiationsOfSingletons

### LINKED ISSUE #3280: MultipleInstantiationsOfSingletons.hasSynchronized() can produce stack overflow
When running latest spotbugs version 4.9.0 against my code I get the Exception below. I did not have time yet to find out which exact code in my code base is triggering this so I cannot give a reproducer yet.

```
`Exception in thread "main" edu.umd.cs.findbugs.ba.AnalysisException: Exception was thrown during analysis
	at edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$3(FindBugs2.java:1117)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at edu.umd.cs.findbugs.FindBugs.runMain(FindBugs.java:399)
	at edu.umd.cs.findbugs.FindBugs2.main(FindBugs2.java:1223)
Caused by: java.util.concurrent.ExecutionException: java.lang.StackOverflowError
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$3(FindBugs2.java:1112)
	... 5 more
Caused by: java.lang.StackOverflowError
	at java.base/java.util.ArrayList.indexOf(ArrayList.java:318)
	at java.base/java.util.ArrayList.contains(ArrayList.java:307)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:280)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
	at edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletons.hasSynchronized(MultipleInstantiationsOfSingletons.java:287)
```

gtoison:
- Break the recursion in `MultipleInstantiationsOfSingletons`, this should fix #3280 
- Reordered some conditions so the trivial clauses are evaluated first

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
@@ -10,5 +10,6 @@ class Issue3280Test extends AbstractIntegrationTest {
         performAnalysis("ghIssues/Issue3280.class");
 
         assertNoBugType("SING_SINGLETON_IMPLEMENTS_CLONEABLE");
+        assertBugInClass("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue3280");
     }
 }
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
@@ -1,0 +1,14 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3280Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue3280.class");
+
+        assertNoBugType("SING_SINGLETON_IMPLEMENTS_CLONEABLE");
+    }
+}
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
@@ -9,7 +9,7 @@ class Issue3280Test extends AbstractIntegrationTest {
     void testIssue() {
         performAnalysis("ghIssues/Issue3280.class");
 
-        assertNoBugType("SING_SINGLETON_IMPLEMENTS_CLONEABLE");
-        assertBugInClass("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue3280");
+        assertNoBugInClass("SING_SINGLETON_IMPLEMENTS_CLONEABLE", "Issue3280");
+        assertBugInClassCount("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue3280", 1);
     }
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
@@ -243,7 +243,7 @@ public class MultipleInstantiationsOfSingletons extends OpcodeStackDetector {
             }
         }
 
-        if (instanceGetterMethod != null && !hasSynchronized(instanceGetterMethod) && isInstanceFieldLazilyInitialized) {
+        if (instanceGetterMethod != null && !hasSynchronized(instanceGetterMethod, new HashSet<>()) && isInstanceFieldLazilyInitialized) {
             bugReporter.reportBug(new BugInstance(this, "SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", NORMAL_PRIORITY).addClass(this)
                     .addMethod(instanceGetterMethod));
         }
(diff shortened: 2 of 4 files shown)


JuditKnoll (approved):
I'm sorry for the issue. Thank you for looking into and fixing it. You are super fast 😄

gtoison:
No problem 😄 
I wonder how we could avoid these issues because the at the same time that problem was rather obvious in retrospect but also tricky to anticipate

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
```diff
@@ -0,0 +1,14 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3280Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue3280.class");
+
+        assertNoBugType("SING_SINGLETON_IMPLEMENTS_CLONEABLE");
```
iloveeclipse, line 12: You could also add
`assertBugInClass("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue3280");`

gtoison: Good point, I have added this check as well!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
```diff
@@ -243,7 +243,7 @@ public void visitAfter(JavaClass javaClass) {
             }
         }
 
-        if (instanceGetterMethod != null && !hasSynchronized(instanceGetterMethod) && isInstanceFieldLazilyInitialized) {
+        if (instanceGetterMethod != null && !hasSynchronized(instanceGetterMethod, new HashSet<>()) && isInstanceFieldLazilyInitialized) {
```
iloveeclipse, line 246: Looking at this line, the call to the offending method would not be needed if `&& isInstanceFieldLazilyInitialized` condition would be evaluated first.

JuditKnoll: I think the `hasSynchronized()` call is necessary here, as it checks whether the instance getter method is synchronized, and the `isInstanceFieldLazilyInitialized` doesn't contain this info. 
You are right about the order of different conditions. Thank you for highlighting this problem.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
```diff
@@ -0,0 +1,15 @@
...
+        assertBugInClass("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue3280");
```
JuditKnoll, line 13: The number of hits for the bugtype could be also checked: `assertBugTypeCount("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", 1);`

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
gtoison: I have made the proposed reorderings with a small change on that one because the closing parenthesis after `isInstanceFieldLazilyInitialized` apparently got misplaced:
```
@@ -227,7 +228,5 @@
         if (!(hasSingletonPostFix
-                || (!javaClass.isAbstract() && !javaClass.isInterface() && !javaClass.isRecord()
-                        && isInstanceAssignOk
-                        && hasNoFactoryMethod
-                        && instanceGetterMethod != null
-                        && (isInstanceFieldEagerlyInitialized || isInstanceFieldLazilyInitialized)))) {
+                || (isInstanceAssignOk && hasNoFactoryMethod && instanceGetterMethod != null
+                        && (isInstanceFieldEagerlyInitialized || isInstanceFieldLazilyInitialized)))
+                        &&!javaClass.isAbstract() && !javaClass.isInterface() && !javaClass.isRecord()) {
```

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java
gtoison: You're right, I've made a change to assert the bug count and the type!

JuditKnoll (approved):

iloveeclipse:
Thanks all.

PR #3285, opened 2025-01-30
Fixed NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE false positive when add edu.umd.cs.findbugs.annotations.Nullable Annotation

komuro-hiraku:
I created a pull request to correct the behavior that differs from what is described in the documentation.

Specifically, the behavior when adding the edu.umd.cs.findbugs.annotations.Nullable annotation differs from the definition in the documentation.
https://spotbugs.readthedocs.io/en/stable/annotations.html#edu-umd-cs-findbugs-annotations-nullable

See https://github.com/spotbugs/spotbugs/issues/3243

thank you.

Issue #3243, opened 2025-01-06
NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE false positive when add edu.umd.cs.findbugs.annotations.Nullable 


komuro-hiraku:
Hello, I have found an issue that appears to be a bug.

## Description

After updating to SpotBugs 4.8.6, the method annotated with edu.umd.cs.findbugs.annotations.Nullable is incorrectly flagged with NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE as a false positive.

## Java version
-  OpenJDK Runtime Environment Corretto-21.0.1.12.1 (build 21.0.1+12-LTS)

## Change Version

- SpotBugs: 4.8.1 to 4.8.6
 
## minimum example

```java
package com.sample.realworld;

import edu.umd.cs.findbugs.annotations.Nullable;

import javax.annotation.Nonnull;

public class Main {
 
    public static void main(String[] args) {
        var hoge = new Hoge();
        System.out.println(hoge.sample);
    }
 
    public static class Hoge {
        @Nonnull public String sample = getNullable();
 
        @Nullable String getNullable() {
            return null;
        }
    }
}
```

Build script

```groovy
plugins {
    id 'java'
    id("com.github.spotbugs") version "6.0.26"
}
 
group = 'com.sample.realworld'
version = '1.0-SNAPSHOT'
 
repositories {
    mavenCentral()
    gradlePluginPortal()
}
 
dependencies {
    testImplementation platform('org.junit:junit-bom:5.10.0')
    testImplementation 'org.junit.jupiter:junit-jupiter'
 
    implementation 'com.diffplug.spotless:spotless-plugin-gradle:6.22.0'
    implementation 'com.github.spotbugs.snom:spotbugs-gradle-plugin:6.0.22'
    spotbugs("com.github.spotbugs:spotbugs:4.8.6")
//    spotbugs("com.github.spotbugs:spotbugs:4.8.1")
 
    compileOnly "com.github.spotbugs:spotbugs-annotations:${spotbugs.toolVersion.get()}"
}
 
test {
    useJUnitPlatform()
}
```

### Run SpotBugs with version 4.8.1:
 
```bash
./gradlew -Duser.language=en -Duser.country=US clean spotbugsMain
 
BUILD SUCCESSFUL in 1s
3 actionable tasks: 3 executed
```

### Run SpotBugs with version 4.8.6:
 
```bash
[...]

PR #3288, opened 2025-01-31
Fixed NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE false positive when add edu.umd.cs.findbugs.annotations.Nullable Annotation

komuro-hiraku:
I created a pull request to correct the behavior that differs from what is described in the documentation.

Specifically, the behavior when adding the edu.umd.cs.findbugs.annotations.Nullable annotation differs from the definition in the documentation.
https://spotbugs.readthedocs.io/en/stable/annotations.html#edu-umd-cs-findbugs-annotations-nullable

See https://github.com/spotbugs/spotbugs/issues/3243

thank you.

FYI: recreated PR #3285

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue3243Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue3243Test.java
@@ -1,0 +1,14 @@
+package edu.umd.cs.findbugs.ba;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+public class Issue3243Test extends AbstractIntegrationTest {
+
+    @Test
+    public void testBugSample() {
+        performAnalysis("ghIssues/Issue3243.class");
+
+        assertNoBugType("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE");
+    }
+}
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
@@ -68,7 +68,8 @@ public class TypeQualifierResolver {
     };
 
     private static final ClassDescriptor[] NULLABLE_CLASS_DESCRIPTORS = new ClassDescriptor[] {
-        DescriptorFactory.createClassDescriptor("edu/umd/cs/findbugs/annotations/Nullable"),
+        // intentionally ignored, semantics is different to other Nullable annotations, see <a href="https://github.com/spotbugs/spotbugs/issues/3243">#3243</a>
+        // DescriptorFactory.createClassDescriptor("edu/umd/cs/findbugs/annotations/Nullable"),
         DescriptorFactory.createClassDescriptor("android/support/annotation/Nullable"),
         DescriptorFactory.createClassDescriptor("androidx/annotation/Nullable"),
         DescriptorFactory.createClassDescriptor("com/google/common/base/Nullable"),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Added execute file permission to files in the distribution archive ([#3274](https://github.com/spotbugs/spotbugs/issues/3274))
 - Fixed a stack overflow in  `MultipleInstantiationsOfSingletons` when a singleton initializer makes recursive calls ([#3280](https://github.com/spotbugs/spotbugs/issues/3280))
 - Fixed NPE in  `FindReturnRef` on inner class fields ([#3283](https://github.com/spotbugs/spotbugs/issues/3283))
+- Fixed NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE false positive when add edu.umd.cs.findbugs.annotations.Nullable ([#3243](https://github.com/spotbugs/spotbugs/issues/3243))
 
 ## 4.9.0 - 2025-01-15
 ### Added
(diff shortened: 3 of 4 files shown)


hazendaz:
Because the naming seems off, even though all info points back here, can you add a comment where it's removed stating not to add back pointing to same conversation.  That way it's less likely someone thinks it's a mistake and naturally tries to clean up again.  Better to be too explicit here!

This will be otherwise included in 4.9.1hopefully put this weekend.

iloveeclipse:
> can you add a comment where it's removed stating not to add back pointing to same conversation

Do you mean, from which point in time we had the regression / change in behavior? That is https://github.com/spotbugs/spotbugs/pull/2694.

Or do you mean to add comment **in the code** stating that we omit `edu.umd.cs.findbugs.annotations.Nullable` on purpose? That would be good. @komuro-hiraku : could you please instead of removing the line `DescriptorFactory.createClassDescriptor("edu/umd/cs/findbugs/annotations/Nullable")` comment it out and put a comment above `// intentionally ignored, semantics is different to other Nullable annotations, see #3243`

> This will be otherwise included in 4.9.1hopefully put this weekend.

Yes, this would be good.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
```diff
@@ -68,7 +68,6 @@ public class TypeQualifierResolver {
     };
 
     private static final ClassDescriptor[] NULLABLE_CLASS_DESCRIPTORS = new ClassDescriptor[] {
-        DescriptorFactory.createClassDescriptor("edu/umd/cs/findbugs/annotations/Nullable"),
```
iloveeclipse, line 71: Comment it out instead and add a comment line above
// intentionally ignored, semantics is different to other Nullable annotations, see #3243

gtoison:
There are dozens of `@Nullable` annotations out there, I think we should add some kind of documentation explaining why `edu.umd.cs.findbugs.annotations.Nullable` is handled differently.
We should also explain why it exists (I do not know). 

If we decide that the reference is https://spotbugs.readthedocs.io/en/stable/annotations.html#edu-umd-cs-findbugs-annotations-nullable

> The annotated element could be null under some circumstances. This is treated the same way as not being annotated.

Then we should align the javadoc of `@Nullable` which only has the 1st sentence:

> The annotated element could be null under some circumstances.

komuro-hiraku:
@iloveeclipse Thank you for your feedback. I have fixed, could you please check them?

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
komuro-hiraku: fixed and add comment https://github.com/spotbugs/spotbugs/pull/3288/commits/6db15f5b03d37b186c54b3502f974ff301e375c0

iloveeclipse (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue3243Test.java
```diff
@@ -0,0 +1,17 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.emptyIterable;
+
+public class Issue3243Test extends AbstractIntegrationTest {
+
+    @Test
+    public void testBugSample() {
+        performAnalysis("ghIssues/Issue3243.class");
+
+        assertThat(getBugCollection(), emptyIterable());
```
JuditKnoll, line 15: Please list the specific bugtype(s), and not `emptyIterable`. Feel free to use the assert functions from `AbstractIntegrationTest` (e.g. `assertNoBugType()`).

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
```diff
@@ -68,7 +68,8 @@ public class TypeQualifierResolver {
...
+        // intentionally ignored, semantics is different to other Nullable annotations, see #3243
```
JuditKnoll, line 71: Can you please add a link to the issue as well or at least mention before the issue number that it's an issue and not a PR?

iloveeclipse:
> There are dozens of `@Nullable` annotations out there, I think we should add some kind of documentation explaining why `edu.umd.cs.findbugs.annotations.Nullable` is handled differently. 

=> https://github.com/spotbugs/spotbugs/issues/3289

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
iloveeclipse: Not sure I understood, #3243 is an issue, not a PR and the comment refers to the issue.

iloveeclipse:
@komuro-hiraku : please if you update the code, don't create new commit but squash all commits to one, **rebase** on latest master commit and force push to this branch. You can also mark your branch editable for others so others could push fixes for trivial things directly to your branch and update the PR.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/jsr305/TypeQualifierResolver.java
JuditKnoll: I meant to either update like this:
```suggestion
        // intentionally ignored, semantics is different to other Nullable annotations, see issue #3243
```
Or something like this:
```suggestion
        // intentionally ignored, semantics is different to other Nullable annotations, see <a href="https://github.com/spotbugs/spotbugs/issues/3243">#3243</a>
```

As from the number alone in the code it's not 100% clear whether it's an issue or a PR.

JuditKnoll (approved):

iloveeclipse:
Thanks @komuro-hiraku !

hazendaz:
@komuro-hiraku Thanks so much!

PR #3300, opened 2025-02-04
Remove deprecated bugpatterns

gtoison:
SpotBugs currently has 6 deprecated bug patterns, and they are never emitted.
They have been deprecated for a long while so it's time to remove then, as discussed in https://github.com/spotbugs/sonar-findbugs/pull/1147 this will remove them for the rules set in the SonarQube plugin.


JuditKnoll:
There are a few more deprecated bugtypes: `IS_INCONSISTENT_SYNC`, `BC_IMPOSSIBLE_CAST_PRIMITIVE_ARRAY`, `NP_METHOD_PARAMETER_RELAXING_ANNOTATION` (I haven't investigated them further). Were these left in intentionally?

gtoison:
Hum, good question!
It looks like the problematic bug patterns are deprecated AND experimental. 
I'm starting to suspect that there's an issue in the Groovy script transforming SpotBugs bug patterns into Sonar rules

gtoison:
I've found the explanation: the SonarQube plugins looks for the deprecated flag in `messages.xml`, not in `findbugs.xml`.
So for instance `IS_INCONSISTENT_SYNC` is not deprecated in `messages.xml` but it is in `findbugs.xml`.

We probably want to make this clearer by aligning the deprecated flags in the `messages.xml` file and the translations, but I propose that we make this a separate PR.
The 4 bug patterns removed in that PR have between deprecated for a while (in `messages.xml` and `findbugs.xml`) so there was ample time for the users to notice.

JuditKnoll:
There is also one more bugpattern deprecated quite a long time ago (https://github.com/spotbugs/spotbugs/commit/e309cc6a0e4bc920a6ffec3126dbc9a5294b91ec) in `messages.xml`, `messages_fr.xml` and `messages_ja.xml`, but missing from `findbugs.xml`: `RCN_REDUNDANT_CHECKED_NULL_COMPARISON` (according to the comment `deprecated in favor of two separate RCN_ patterns`, which is not deleted. 
Doesn't it cause issues in the SonarQube plugin?

Also, there is one more deprecated bugpattern in `messages_fr.xml`: `NP_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`. This seems to be missing from both `messages.xml`, `messages_ja.xml` and `findbugs.xml`.

gtoison:
I saw that `RCN_REDUNDANT_CHECKED_NULL_COMPARISON` is missing yes, so in the SonarQube plugin a rule is created but none the "profiles" contain that rule, because the profiles are created from `findbugs.xml`.
We should indeed remove and since this PR is about removing bug patterns, it would be sensible to remove it here as well as `NP_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2025-??-??
 
+### Removed
+- Removed `TLW_TWO_LOCK_NOTIFY`, `LI_LAZY_INIT_INSTANCE`, `BRSA_BAD_RESULTSET_ACCESS` and `BC_NULL_INSTANCEOF` deprecated bug patterns.
```
JuditKnoll, line 11: Maybe in the changelog entry `NP_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` and `RCN_REDUNDANT_CHECKED_NULL_COMPARISON` also could be mentioned.

JuditKnoll (approved):
> I saw that `RCN_REDUNDANT_CHECKED_NULL_COMPARISON` is missing yes, so in the SonarQube plugin a rule is created but none the "profiles" contain that rule, because the profiles are created from `findbugs.xml`. We should indeed remove and since this PR is about removing bug patterns, it would be sensible to remove it here as well as `NP_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`

Thank you for the explanation 😄

CHANGELOG.md
gtoison: You're right, I've updated the changelog

ThrawnCA (approved):
Is there corresponding code that also needs to be removed?

JuditKnoll:
> Is there corresponding code that also needs to be removed?

`BC_NULL_INSTANCEOF` is mentioned in a few tests (`Bug3107916Test`, `Bug3277814Test`, `RegressionIdeas20111219Test`, `IssueRequireNotNullTest`), other than this, I couldn't find any other usages for these bugtypes.

hazendaz:
should the tests that is mentioned in be removed?

gtoison:
> should the tests that is mentioned in be removed?

I've removed the references to `BC_NULL_INSTANCEOF` in the tests

JuditKnoll (approved):

JuditKnoll:
> > should the tests that is mentioned in be removed?
> 
> I've removed the references to `BC_NULL_INSTANCEOF` in the tests

@gtoison You removed the references from two tests, `Bug3107916Test` and `IssueRequireNotNullTest`, but not from `Bug3277814Test` and `RegressionIdeas20111219Test`. Can you please share why?
To me all of these seem quite similar.

gtoison:
I don't remember how I checked but clearly I missed references to `BC_NULL_INSTANCEOF `, I've removed them now

JuditKnoll (approved):

Issue #3305, opened 2025-02-06
False positive in HSM_HIDING_METHOD for "switch on enums" code


iloveeclipse:
Regression in 4.9.1, via https://github.com/spotbugs/spotbugs/pull/2467

For the code below compiled with Java 21 eclipse compiler, SpotBugs reports false positive
```
The method '$SWITCH_TABLE$SwitchTableBug$X()' in class 'B' hides a method in class 'A'
```
with the `HSM_HIDING_METHOD` warning.
This is wrong. synthetic `SWITCH_TABLE` generated by compiler should never be considered as a coding issue.

```
public class SwitchTableBug {
	class A {
		public void hello() {
			// Switch on local variable
			X x = X.Z;
			switch (x) {
			case Y: {
				System.out.println("Y");
				break;
			}
			default: break;
			}
		}
	}

	class B extends A {
		X x = X.Z;
		@Override
		public void hello() {
			// Switch on a field with same name
			switch (x) {
			case Y: {
				System.out.println("Y");
				break;
			}
			default: break;
			}
		}
	}

	enum X { Y, Z }
}
```

PR #3306, opened 2025-02-06
Synthetic method incorrectly picked up by `FindHiddenMethod`

gtoison:
As reported in #3305 ECJ can generate a synthetic method for an enum switch.


iloveeclipse (approved):

iloveeclipse:
Thanks!

PR #1054, opened 2020-01-04
Issue641: reporting useless @SuppressFBWarnings

mafor:
Implementation of the ticket #641, or more precisely of the first half it:
reporting `@SuppressFBWarnings` annotations which are not "hit" by any bugs.

Test class:
```java
@SuppressFBWarnings(value = "UUF_UNUSED_FIELD") // <- useless suppression
public class Issue641 {

    @SuppressFBWarnings(value = "UUF_UNUSED_FIELD") // <- useless suppression
    private String field;

    @SuppressFBWarnings(value = "UUF_UNUSED_FIELD") // <- valid suppression
    private String field2;

    public String getField() {
        return field;
    }

    public void setField(/* useless suppression -> */ @SuppressFBWarnings(value = "UUF_UNUSED_FIELD") String val) {
        field = val;
    }
}
```
Output:
```
[INFO] BugInstance size is 3
[INFO] Error size is 0
[INFO] Total bugs: 3
[ERROR] Suppressing annotation on the class com.chaosonic.Issue641 is unnecessary [com.chaosonic.Issue641] At Issue641.java:[lines 6-20] US_USELESS_SUPPRESSION_ON_CLASS
[ERROR] Suppressing annotation on the field com.chaosonic.Issue641.field is unnecessary [com.chaosonic.Issue641] In Issue641.java US_USELESS_SUPPRESSION_ON_FIELD
[ERROR] Suppressing annotation on the parameter 2 of the method com.chaosonic.Issue641.setField(String) is unnecessary [com.chaosonic.Issue641] At Issue641.java:[lines 19-20] US_USELESS_SUPPRESSION_ON_METHOD_PARAMETER
[INFO]
```
----

Make sure these boxes are checked before submitting your PR -- thank you!

- [x] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #3307, opened 2025-02-07
Report useless `@SuppressFBWarnings` annotations

### LINKED ISSUE #641: A checker that for unnecessary @SuppressFBWarnings ?
It would be nice to have a checker for unnecessary SuppressFBWarnings annotation uses.

the annotation is used mostly to work around false positives, but as those false positives get fixed in upcoming versions of SB, the code remains littered with useless suppression... wold be nice to eliminate them...

Additionally it would be nice to have a check for allowing only explicit use of @suppressfbwarnings("CHECK...")... to force the devs to explicitly specify which check to ignore...

gtoison:
I'm not sure why but we never answered #1054 and the PR got abandoned.
Reporting useless `@SuppressFBWarnings` annotations would be useful indeed and this would fix #641 
This PR merges the original work from @mafor with the latest version.


spotbugs/etc/messages.xml
```diff
@@ -9442,6 +9442,61 @@ Using floating-point variables should not be used as loop counters, as they are
      ]]>
     </Details>
   </BugPattern>
+
+  <BugPattern type="US_USELESS_SUPPRESSION_ON_CLASS">
+    <ShortDescription>Useless suppression on a class</ShortDescription>
+    <LongDescription>Suppressing annotation on the class {0} is unnecessary</LongDescription>
```
JuditKnoll, line 9448: The `@SuppressFBWarnings` annotation and the unnecessarily suppressed bug(s) also could be mentioned.

JuditKnoll (changes requested):

spotbugs/etc/messages.xml
gtoison: I've added the name of the annotation to the bug description, I'm not sure if I can easily extract the unnecessarily suppressed bug(s) because it might be suppressing based on prefix or regex

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue641Test.java
```diff
@@ -0,0 +1,25 @@
...
+        performAnalysis("suppression/Issue641.class");
+        SortedBugCollection bugCollection = (SortedBugCollection) getBugCollection();
+
+        assertBugAtField("US_USELESS_SUPPRESSION_ON_FIELD", "suppression.Issue641", "field");
+        assertBugInClass("US_USELESS_SUPPRESSION_ON_CLASS", "suppression.Issue641");
+        assertBugInMethod("US_USELESS_SUPPRESSION_ON_METHOD_PARAMETER", "suppression.Issue641", "setField");
+
+        assertThat("Expected three warnings",
+                bugCollection.getCollection().size(),
+                CoreMatchers.equalTo(3));
```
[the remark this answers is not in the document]
gtoison: I've added more checks and remove that assertion

spotbugsTestCases/src/java/suppression/Issue641.java
```diff
@@ -0,0 +1,21 @@
+package suppression;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+@SuppressFBWarnings(value = "UUF_UNUSED_FIELD") // <- useless suppression
+public class Issue641 {
```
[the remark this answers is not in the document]
gtoison: I've added more checks, they do not cover all cases but I think it's a bit better now

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/WarningSuppressorTest.java
```diff
@@ -14,15 +14,21 @@ class WarningSuppressorTest extends AbstractIntegrationTest {
     void testIssue() {
         performAnalysis("suppress/SuppressedBugs.class");
 
+        // We're also getting many (true positive) US_USELESS_SUPPRESSION_ON_METHOD for this sample code
```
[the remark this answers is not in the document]
gtoison: I've added checks that the suppression with incorrect matching now raise a bug that they are useless. This should alert the user that they have an unnecessary suppression

spotbugs/etc/messages.xml
JuditKnoll: Maybe also the number of each bugtype hits could be checked.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue641Test.java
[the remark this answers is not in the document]
JuditKnoll: Maybe also the number of each bugtype hits could be checked.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Mock
+    private BugReporter upstreamReporter;
+
+    @InjectMocks
+    SuppressionMatcherBugReporter reporter;
+
+    @Test
+    void shouldNotCallUpstreamReporterIfBugSuppressed() {
+        // given
+        BugInstance suppressed = mock(BugInstance.class);
```
ThrawnCA, line 26: Why mix annotation-based and method-based mocks? This could become another annotated field.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    void shouldMatchBugInstance() {
+        // given
+        when(suppressor.match(bugInstance)).thenReturn(true);
+        SuppressionMatcher matcher = new SuppressionMatcher();
+        // when
+        matcher.addSuppressor(suppressor);
+        // then
+        assertTrue(matcher.match(bugInstance), "Should match BugInstance");
```
ThrawnCA, line 46: Error messages should give more detail about the failed value(s) if possible, especially for `assertTrue` (which doesn't give much detail natively)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
```diff
@@ -0,0 +1,69 @@
...
+        matcher.validateSuppressionUsage(bugReporter);
+        // then
+        verifyNoInteractions(bugReporter);
+    }
+
+    @Test
+    void shouldNotMatchBugInstance() {
+        // given
+        when(suppressor.match(bugInstance)).thenReturn(false);
+        when(suppressor.buildUselessSuppressionBugInstance()).thenReturn(uselessSuppressionBugInstance);
```
ThrawnCA, line 57: Should this be in `init`?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Test
+    void shouldMatchClassLevelSuppression() {
+        // given
+        matcher.addSuppressor(new ClassWarningSuppressor("UUF_UNUSED_FIELD", SuppressMatchType.DEFAULT, CLASS_ANNOTATION));
+        BugInstance bug = new BugInstance("UUF_UNUSED_FIELD", 1).addClass(CLASS_NAME);
+        // when
+        boolean matched = matcher.match(bug);
+        // then
+        assertThat("Should match the bug", matched, is(true));
```
ThrawnCA, line 48: This ultimately doesn't give any more information than using `assertTrue`. If possible the error message should say more about the bug value.

spotbugs/etc/messages.xml
```diff
@@ -9442,6 +9442,61 @@ Using floating-point variables should not be used as loop counters, as they are
...
+    <Details>
+      <![CDATA[<p>
+      Suppressing annotations <code>&amp;SuppressFBWarnings</code> should be removed from the source code as soon as they are no more needed.
```
ThrawnCA, line 9451: Perhaps this message could be adjusted to make it clear why the annotation is no longer needed, ie the code no longer contains that bug type.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
gtoison: There are three tests, one calls the bug instance `suppressed`, the second calls it `reported` and the third one does not create a bug instance, so I think it makes sense that @mafor used a local variable here

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
gtoison: The two tests mock objects differently:
`when(suppressor.match(bugInstance)).thenReturn(true);`
and:
`when(suppressor.match(bugInstance)).thenReturn(false);`

So I think this can't be factored in `init`

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue641Test.java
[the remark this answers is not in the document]
gtoison: I've added assert methods with the expected bugs count

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
ThrawnCA: Those lines are different, yes, but making the suppressor return the instance is the same in both.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
ThrawnCA: There is no need for the tests to use different names for the instance. As I see it, we just have 2/3 tests creating an instance mock, so it makes sense to be consistent with other values by making it an annotated field.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue641Test.java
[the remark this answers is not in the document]
JuditKnoll: Sorry, I wasn't clear enough. I meant asserting only the bugtypes and the number of hits for all the relevant bugtypes e.g. `assertBugTypeCount(US_USELESS_SUPPRESSION_ON_FIELD, 1)`.

This way even if there are several hits with the same bugtype at the same place, or there are hits at an unexpected place, it's checked, but the testcase won't break if a new bugtype is introduced.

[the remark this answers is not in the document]
gtoison: I have added the expected bug count for each type

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
gtoison: I have made the change to create the instance with `@Mock`

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
gtoison: I've looked into this and do not see a practical way for the matcher to return _why_ the instance wasn't matched.
We're using the term matcher for two separate things: 
- the suppression matchers we're using to filter out detected bugs
- the hamcrest matchers we're using in unit tests (these matchers do provide messages)

I think it is fine to leave the test like that, it is short enough so we can reason about it

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherTest.java
gtoison: I'm not sure how to extract the reason _why_ the instance was matched or not, I think the test is fine like that

ThrawnCA (approved):

spotbugs/etc/messages.xml
ThrawnCA: Just something to the effect that the the bug type is no longer raised by that code.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
ThrawnCA: Line 57, specifically, is duplicated across multiple tests, and could be extracted to the setup method.

    when(suppressor.buildUselessSuppressionBugInstance()).thenReturn(uselessSuppressionBugInstance);

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
ThrawnCA: The message could at least specify which suppression matcher is being applied? It could be deduced anyway by tracing the test, but seeing it at a glance is better.

JuditKnoll (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
gtoison: The suppressor is mocked here, so there's no real reason for the instance to be suppressed besides the suppressor being mocked to return false.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherMockedTest.java
gtoison: Sorry I'm still lost, I don't see duplications for that line in `SuppressionMatcherMockedTest`

spotbugs/etc/messages.xml
gtoison: I've expanded the message for these unnecessary suppressions, let me know if that could be improved!

JuditKnoll (approved):

hazendaz (approved):

uhafner:
This is a very nice new feature! Is there a way to disable this for a specific case? I have a case where the new detector does not work correctly (I will create a new bug report tomorrow). 

Example: 
When I do not remove the suppression in [LineRangeList](https://github.com/uhafner/codingstyle/blob/315a08836f8c22ba994a56dce5cd7ff68df13658/src/main/java/edu/hm/hafner/util/LineRangeList.java#L244) in I get a new warnings of type `US_USELESS_SUPPRESSION_ON_METHOD`, see 
https://github.com/uhafner/codingstyle/actions/runs/13753858193/job/38458244783

But if I remove the suppression I get a warning of `IT_NO_SUCH_ELEMENT`, see https://github.com/uhafner/codingstyle/actions/runs/13753897007/job/38458341372

PR #3313, opened 2025-02-10
`AT_STALE_THREAD_WRITE_OF_PRIMITIVE` false positives fixes

gtoison:
Merge of #3314 and #3317
This should fix #3310 and #3316

PR #3314, opened 2025-02-12
Fix `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` FP

### LINKED ISSUE #3310: AT_STALE_THREAD_WRITE_OF_PRIMITIVE false positive
```console
Shared primitive variable "written" in one thread may not yield the value of the most recent write from another thread

Bug type AT_STALE_THREAD_WRITE_OF_PRIMITIVE
In class com.github.benmanes.caffeine.jcache.CacheProxy
In method com.github.benmanes.caffeine.jcache.CacheProxy.lambda$putNoCopyOrAwait$8(Object, boolean, Object, CacheProxy$PutResult, Object, Expirable)
Field com.github.benmanes.caffeine.jcache.CacheProxy$PutResult.written
At CacheProxy.java:[line 380]
Another occurrence at CacheProxy.java:[line 394]
```

This case is unfortunately a [bit complex](https://github.com/ben-manes/caffeine/blob/2befdb17e793848da0fba95683d4a041ada5c5e0/jcache/src/main/java/com/github/benmanes/caffeine/jcache/CacheProxy.java#L380), but there is only a single thread able to read and write to this value. The caller is using a capturing lambda in a Map's `compute` to complete with more information about what happened when inside this atomic operation. This way while the map stores a value, additional facts during the operation are communicated to the outside scope.

```java
var result = ...
map.compute(key, (k, oldValue) -> {
    ...
    if (published) {
      result.written = true
    }
    return newValue;
});
return result;
```

JuditKnoll:
Fix #3310 by only reporting `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` (also `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` and `AT_NONATOMIC_64BIT_PRIMITIVE` from the same detector), if the field is defined in the visited class.
This way, we loose some true positives (e.g. if the field is defined in a super or an inner class and used incorrectly), but handling these edge cases would lead to unnecessary complexity. I also checked a few other detectors working with fields, and several of them only cares about the field of the visited class.

Thanks @gtoison for the reproducer at https://github.com/spotbugs/spotbugs/pull/3313. I hope you don't mind, that I created this fix PR on top of it.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fixed an `HSM_HIDING_METHOD` false positive when ECJ generates a synthetic method for an enum switch ([#3305](https://github.com/spotbugs/spotbugs/issues/3305))
 - Fix `AT_UNSAFE_RESOURCE_ACCESS_IN_THREAD` false negatives, detector depending on method order.
 - Fix `THROWS_METHOD_THROWS_CLAUSE_THROWABLE` reported in a method calling `MethodHandle.invokeExact` due to its polymorphic signature ([#3309](https://github.com/spotbugs/spotbugs/issues/3309))
+- Fix `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` false positive in inner class ([#3310](https://github.com/spotbugs/spotbugs/issues/3310)).
 
 ## 4.9.1 - 2025-02-02
 ### Added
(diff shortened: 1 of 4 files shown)


gtoison:
> Thanks @gtoison for the reproducer at #3313. I hope you don't mind, that I created this fix PR on top of it.

I do not mind at all, thank you for the quick fix!

ThrawnCA (approved):

gtoison (approved):

gtoison:
Merged via #3313

Issue #3316, opened 2025-02-13
AT_STALE_THREAD_WRITE_OF_PRIMITIVE false positive on enum switches


iloveeclipse:
Regression in 4.9.0.
For the code below compiled with Java 21 Eclipse compiler, SpotBugs reports false positive:
`Shared primitive variable "field" in one thread may not yield the value of the most recent write from another thread`
with the `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` warning.

```
public class MyClass  {
    private boolean field;

    public static enum MyEnum {
    }

    void method(MyEnum myEnum) {
        field = false;
        switch (myEnum) {
        default:
        }
    }

    public void print() {
        System.out.println(field);
    }
}
```

The analysis is wrong.

For the switch on enum value, ecj generates this field as volatile:
`private static volatile synthetic int[] $SWITCH_TABLE$MyClass$MyEnum`

Here, `volatile` modifier that is put by compiler on synthetic `SWITCH_TABLE` field (generated by compiler) should never qualify the class as potentially multi-threaded.

Fix is similar to https://github.com/spotbugs/spotbugs/pull/3306.

Synthetic fields shouldn't be considered for analysis.

I have a trivial patch that fixes the problem.

iloveeclipse:
Fixed via https://github.com/spotbugs/spotbugs/pull/3313

PR #3317, opened 2025-02-13
Fix `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` false positive for ECJ compiled

iloveeclipse:
Don't check synthetic fields.

Fixes https://github.com/spotbugs/spotbugs/issues/3316


JuditKnoll:
I think https://github.com/spotbugs/spotbugs/pull/3314 should already fix this, however it doesn't check everywhere that the field is not synthetic, only when the bug is reported.

iloveeclipse:
> I think #3314 should already fix this, however it doesn't check everywhere that the field is not synthetic, only when the bug is reported.

Nope, it doesn't, just tried.

gtoison:
Merged via #3313

Issue #3324, opened 2025-02-17
False negative with NP_METHOD_RETURN_RELAXING_ANNOTATION using @NonNullByDefault


ghernadi:
Although the title already says all, here is a MWE:
```java 
// file: package-info.java

@NonNullByDefault
package ghIssues.issue3324;

import org.eclipse.jdt.annotation.NonNullByDefault;
```

```java
package ghIssues.issue3324;

import javax.annotation.Nonnull;

public interface DummyInterface{
    String getNameDefaultNonnull();

    @Nonnull
    String getNameExplicitNonnullFromInterface();
}
```

```java
package ghIssues.issue3324;

public abstract class DummyAbs {
    public abstract String getNameDefaultNonnullFromAbs();
}

```

```java
package ghIssues.issue3324;

import javax.annotation.Nullable;

public class DummyImpl extends DummyAbs implements DummyInterface {
    @Override
    public @Nullable String getNameDefaultNonnull() {
        return null;
    }

    @Override
    public @Nullable String getNameExplicitNonnullFromInterface() {
        return null;
    }

    @Override
    public @Nullable String getNameDefaultNonnullFromAbs() {
        return null;
    }
}
```

First I was not sure if SpotBugs can detect issues that involve multiple classes / interfaces, but since the above MWE shows 1 `NP_METHOD_RETURN_RELAXING_ANNOTATION`, I guess I was wrong, SpotBugs can indeed detect such issues 🎉 

However, in the above example I would have expected 3 reports of `NP_METHOD_RETURN_RELAXING_ANNOTATION` instead of 1.

PR #3325, opened 2025-02-17
Reproducer for issue 3324

ghernadi:
Reproducer for https://github.com/spotbugs/spotbugs/issues/3324


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/Issue3324Test.java
```diff
@@ -0,0 +1,28 @@
...
+            "ghIssues/issue3324/DummyAbs.class",
+            "ghIssues/issue3324/DummyInterface.class",
+            "ghIssues/issue3324/DummyImpl.class"
+        );
+
+        BugInstanceMatcher bugTypeMatcher = new BugInstanceMatcherBuilder()
+            .bugType("NP_METHOD_RETURN_RELAXING_ANNOTATION")
+            .build();
+
+        assertThat(getBugCollection(), containsExactly(3, bugTypeMatcher));
```
JuditKnoll, line 26: There is an assert function inherited from `AbstractIntegrationTest` which does the same: `assertBugTypeCount()`. Can you please use that instead?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/Issue3324Test.java
```diff
@@ -0,0 +1,28 @@
...
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import org.junit.jupiter.api.Test;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class Issue3324Test extends AbstractIntegrationTest
+{
+    @Test
+    void testIssue()
```
JuditKnoll, line 14: Can you please specify where you expect the bugs?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/Issue3324Test.java
```diff
@@ -0,0 +1,28 @@
...
+    {
+        performAnalysis(
+            "ghIssues/issue3324/DummyAbs.class",
+            "ghIssues/issue3324/DummyInterface.class",
+            "ghIssues/issue3324/DummyImpl.class"
+        );
```
JuditKnoll, line 20: Can you please add the `package-info.java` file to the analysis as well? Otherwise it won't be included in the analysis.

JuditKnoll (changes requested):
The build is indeed failing, but because of spotless. Can you please run `gradlew spotlessApply` to resolve the formatting issues?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/Issue3324Test.java
JuditKnoll: Sorry, I wasn't clear enough. Can you please add asserts about where do you expect the bugs?
Feel free to use the assert functions from `AbstractIntegrationTest`.

JuditKnoll:
> @JuditKnoll Thank you for the review. Am I supposed to mark the requested changes as "resolved" or is this the job of the reviewer (for example to verify if the changes indeed resolve the things mentioned)?

If you are absolutely sure, that the question is resolved, feel free to mark it, but you don't need to. I marked them as resolved.

Would you like to work on a fix for this issue, or you just wanted to report it with the reproducer?

ghernadi:
> If you are absolutely sure, that the question is resolved, feel free to mark it, but you don't need to. I marked them as resolved.

Thanks, will (try to) remember for the next time! :) 

> Would you like to work on a fix for this issue, or you just wanted to report it with the reproducer?

In theory I would like to, but in practice I have no idea if I find time for that. I am more just a user of SpotBugs, have debugged it once and failed quite miserably in identifying the root cause. So I would be more than happy if someone more knowledgeable than me could take a look at fixing this issue.

PR #3326, opened 2025-02-18
Avoid calling `File.getCanonicalPath` twice to improve performance

gtoison:
On my computer calling `File.getCanonicalPath` twice from `ClassFactory` and then from `FilesystemCodeBaseLocator` adds ~15 seconds to the full build time.
According to JProfiler `File.getCanonicalPath` is the method where most of the time is spent, so halving the number of calls is significant.
Avoiding that double path normalization should improve SpotBugs' performance when analyzing folders of classes

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
@@ -83,6 +83,7 @@ public class ClassFactory implements IClassFactory {
         // It's not fatal if we can't.
         try {
             pathName = new File(pathName).getCanonicalPath();
+            return FilesystemCodeBaseLocator.fromCanonicalPath(pathName);
         } catch (IOException e) {
             // Ignore
         }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
@@ -79,15 +79,6 @@ public class ClassFactory implements IClassFactory {
      */
     @Override
     public ICodeBaseLocator createFilesystemCodeBaseLocator(String pathName) {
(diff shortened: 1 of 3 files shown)


JuditKnoll (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/impl/ClassFactory.java
```diff
@@ -83,6 +83,7 @@ public ICodeBaseLocator createFilesystemCodeBaseLocator(String pathName) {
         // It's not fatal if we can't.
         try {
             pathName = new File(pathName).getCanonicalPath();
+            return FilesystemCodeBaseLocator.fromCanonicalPath(pathName);
```
iloveeclipse, line 86: shouldn't this be a call with "false" argument?

iloveeclipse, line 86: shouldn't the try/catch be replaced with the proper  FilesystemCodeBaseLocator constructor that does the checks? In worst case getCanonicalPath is still called twice on IO errors.

gtoison: Good point! Instead of adding a flag to say if we want to canonicalize the path I should have simply removed the first call 🫤

iloveeclipse (approved):
Thanks.

iloveeclipse:
> adds ~15 seconds to the full build time.

And the overall build time is ...?

gtoison:
The overall build (`gradlew clean spotlessApply test`) now take 2 minutes 16 seconds on my computer.

iloveeclipse:
> The overall build (`gradlew clean spotlessApply test`) now take 2 minutes 16 seconds on my computer.

So this PR saves ~10%, which seem to be very high number for such a small change. I assume you are on Windows?

gtoison:
Yes I'm on Windows, also bear in mind that the tests are about half of the build time, so these 10% come from that half only.

PR #3349, opened 2025-03-05
Introduce UselessSuppressionDetector to report the useless suppressions

### LINKED ISSUE #3348: New US_USELESS_SUPPRESSION_ON_* warnings are not shown in Eclipse
Follow up on https://github.com/spotbugs/spotbugs/pull/3307

The `NoteSuppressedWarnings` detector which is a `NonReportingDetector` reports now warnings.
This is inconsistent, and as every inconsistency this has side effects.

In Eclipse we used `BugInstance.getDetectorFactory()` to retrieve information about the spotbugs plugin that contributes this detector, and if not available (which is the case for `NonReportingDetector`), we were trying to scan the `DetectorFactory for the matching bug patterns. However, we've restricted that to only factories which match to `DetectorFactory.isReportingDetector()` condition. Again, this is not the case for `NonReportingDetector`:

https://github.com/spotbugs/spotbugs/blob/b5c7686b671237be5227c1ab312b8198f5592b45/eclipsePlugin/src/de/tobject/findbugs/reporter/MarkerReporter.java#L203-L218

We can add a workaround in Eclipse plugin like below, but it would be better if either the bug instances would have the factory added or the `NoteSuppressedWarnings` would be a reporting detector. Later seem to be a problem as the comment says it has to be executed before all "reporting" detectors and so simply removing `implements NonReportingDetector` from it causes a cycle & analysis crash.

```
diff --git a/eclipsePlugin/src/de/tobject/findbugs/reporter/MarkerReporter.java b/eclipsePlugin/src/de/tobject/findbugs/reporter/MarkerReporter.java
index 70259a2..e006d6c 100644
--- a/eclipsePlugin/src/de/tobject/findbugs/reporter/MarkerReporter.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/reporter/MarkerReporter.java
@@ -59,2 +59,3 @@
 import edu.umd.cs.findbugs.config.UserPreferences;
+import edu.umd.cs.findbugs.detect.NoteSuppressedWarnings;
 
@@ -215,3 +216,6 @@
                 if (!df2.isReportingDetector()) {
-                    continue;
+                    // NoteSuppressedWarnings reports bugs even if it says it doesn't ...
+                    if (!NoteSuppressedWarnings.class.getName().equals(df2.getFullName())) {
+                        continue;
+                    }
                 }
```

@gtoison : would be nice if you could look at it.

gtoison:
Let the new `UselessSuppressionDetector` report the useless suppressions instead of `NoteSuppressedWarnings` which is a `NonReportingDetector`
This should fix #3348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+### Added
+
+- Introduced `UselessSuppressionDetector` to report the useless annotations instead of `NoteSuppressedWarnings` ([#3348](https://github.com/spotbugs/spotbugs/issues/3348))
 
 ## 4.9.2 - 2025-03-01
 ### Added
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/SuppressionMatcherBugReporterTest.java
@@ -6,6 +6,7 @@ import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
+
 import static org.mockito.Mockito.*;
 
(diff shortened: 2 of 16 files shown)


iloveeclipse (approved):
Looks clean & nice, thanks!
Warning is now shown in Eclipse, note properly reported detector too:
image

gtoison:
Great, thanks for testing!

Issue #3350, opened 2025-03-10
Do not detect US_USELESS_SUPPRESSION_ON_* warnings on Generated code.


SimSonic:
In my project I have lombok which is set to add `@SuppressFBWarnings` onto all generated constructors/methods thought it's configuration:

```
// root lombok.config
lombok.extern.findbugs.addSuppressFBWarnings = true
```

After updating to SpotBugs 4.9.3 javac throws medium warnings `US_USELESS_SUPPRESSION_ON_*` on generated code.

I believe we should have some possibility to prevent generated code from being scanned.

May be disable all warnings including these ones on all code which annotated as `@Generated` (javax/jakarta?).

fenuks:
SpotBugs have undocumented (it is not mentioned in
https://spotbugs.readthedocs.io/en/latest/filter.html) Annotation filter.
Unfortunately, as far I can see, it only works with annotations on classes.
Lombok adds `@lombok.Generated` on generated methods, so I think extending
[AnnotationMatcher](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/filter/AnnotationMatcher.java) to support annotations on methods would be one way to solve this problem.

gtoison:
Thank you for reporting the issue
@SimSonic could you please share a few example where the issue happens? This will help making sure that the problem is understood correctly

SimSonic:
@gtoison yes, here is a minimal example with maven, Java 21, based on Spring Initialized application:

[issue-3350-demo.zip](https://github.com/user-attachments/files/19244992/issue-3350-demo.zip)

Just run `mvn clean verify` to see this:

```
[ERROR] Medium: Suppressing annotation on the method new com.example.demo.DemoApplication$BrokenClass() is unnecessary [com.example.demo.DemoApplication$BrokenClass] At DemoApplication.java:[line 10] US_USELESS_SUPPRESSION_ON_METHOD
[ERROR] Medium: Suppressing annotation on the method com.example.demo.DemoApplication$BrokenClass.hashCode() is unnecessary [com.example.demo.DemoApplication$BrokenClass] At DemoApplication.java:[line 10] US_USELESS_SUPPRESSION_ON_METHOD
[...]

PR #3364, opened 2025-03-21
Extended annotation filtering

fenuks:
Pull request https://github.com/spotbugs/spotbugs/pull/2395 added option to filter out bugs based on annotation on class. This patch generalizes the feature to include annotations applied to methods and fields as well. It is particularly useful for generated code; e.g. lombok can be configured to apply `@lombok.Generated` to methods it generates.

Generalized annotation filtering can be used to fix issue https://github.com/spotbugs/spotbugs/issues/3350 with change to the `lombok.config` file:

```diff
+ lombok.addLombokGeneratedAnnotation = true
- lombok.extern.findbugs.addSuppressFBWarnings = true
```

and `<Annotation name="lombok.Generated" />` addition to the `spotbugsExclude.xml`.


hazendaz (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -2068,18 +2069,42 @@ private void addJavaAnnotationNames(BugAnnotation annotation) {
...
-            pma.setJavaAnnotationNames(javaAnnotationNames);
+            if (annotation instanceof ClassAnnotation) {
+                AnnotationEntry[] annotationEntries = javaClass.getAnnotationEntries();
+                // map annotation entry type to dotted class name, for example
+                // Lorg/immutables/value/Generated; -> org.immutables.value.Generated
+                addAnnotationNames(pma, annotationEntries);
+            } else if (annotation instanceof MethodAnnotation) {
+                MethodAnnotation ma = (MethodAnnotation) annotation;
+                for (Method method : javaClass.getMethods()) {
+                    if (method.getName().equals(ma.getMethodName())) {
```
JuditKnoll, line 2080: Are you sure that this check is enough? What if there are several methods with the same name, but with different parameter list? Can you please a testcase like this?

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/AnnotationMatcherTest.java
```diff
@@ -148,6 +149,23 @@ void testPerformAnalysis(SpotBugsRunner spotbugs) {
         assertEquals(4, numberOfMatchedBugs);
     }
 
+    @Test
+    void testFilteringWithAnnotationOnClassMembers(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/ghIssues/issue543/GeneratedOnClassMembers.class"));
+
+        assertEquals(2, bugCollection.getCollection().size());
```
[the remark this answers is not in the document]
fenuks: I will try to make it less dependent on a total number of bugs, but overall
I think it's unlikely to happen. If a new bug is detected in method or field
(there is no other code in test class),
then it should be filtered out due to @Generated annotation, and test should
assert that at the beginning that before filtering these bugs were detected
– change in a test code will be needed anyway.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
fenuks: Yes, that would be a problem. I'll add overload method to test class.

fenuks:
@JuditKnoll Can you please take another look at the updated patch?

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/Generated.java
```diff
@@ -3,7 +3,7 @@
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
 
-@Target(ElementType.TYPE)
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.TYPE})
```
JuditKnoll, line 6: The [original](https://github.com/immutables/immutables/blob/master/value-annotations/src/org/immutables/value/Generated.java) `org.immutables.value.Generated` annotation is only for `ElementType.TYPE`, so it can't be applied to fields or methods and as such, changing the fakeannotation isn't be enough, outside of SpotBugs, developers won't be able to apply this annotation on fields or methods.

As far as I can see, the [lombok's](https://projectlombok.org/api/lombok/Generated) `@Generated` annotation is defined for these targets as well, and that could be used and recognized instead.

JuditKnoll (changes requested):

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/Generated.java
fenuks: I don't know if I understand. I changed fake `org.immutables.value.Generated`
annotation to not add another one. Developers can use any annotation annotation
they want in `<Annotation>` filter, granted is is available at class
level – it is irrelevant what annotation SpotBugs uses in testing code. If you want, I can add fake `lombok.Generated` annotation, but I don't think it's really necessary.

JuditKnoll: IMO it's better if we keep the fakeAnnotations compatible with the original versions. Can you please add another annotation instead? It doesn't need to be a fake `lombok.Generated`, but it can be any annotation.

fenuks:
Is there anything left to be done before it can be merged?

hazendaz (approved):

hazendaz:
> Is there anything left to be done before it can be merged?

I fixed merged conflict on change log, rebuilding now.  If all builds ok will merge.  I'm thinking we probably release this coming weekend.

Issue #3320, opened 2025-02-14
IllegalArgumentException below FindNoSideEffectMethods.getMethodCall()


iloveeclipse:
Running latest SpotBugs Eclipse nightly build 4.9.2.202502140028-6b58d85. No small reproducer yet, except to load this project in Eclipse https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.ui and run SpotBugs on it.

The `org.eclipse.pde.ui` project has dependency to `biz.aQute.bndlib` library where SpotBugs seem to have trouble analyzing `aQute.lib.converter.Converter.newInstance` method call.

```
eclipse.buildId=4.35.0.I20250213-1800
java.version=21.0.6-7
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -os linux -ws gtk -arch x86_64 -data /data/4x_platform_workspace

com.github.spotbugs.plugin.eclipse
Error
Fri Feb 14 07:24:26 CET 2025
Can't get stack offset 1 from [< Ljava/lang/invoke/MethodHandle;, return value from java.lang.invoke.MethodHandles$Lookup.findConstructor(Ljava/lang/Class;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/MethodHandle; >] @ 10 in aQute.lib.converter.Converter.newInstance : (Ljava.lang.Class;)Ljava.lang.Object;

java.lang.IllegalArgumentException: 1 is not a value stack offset
	at edu.umd.cs.findbugs.OpcodeStack.getStackItem(OpcodeStack.java:3280)
	at edu.umd.cs.findbugs.detect.FindNoSideEffectMethods.getMethodCall(FindNoSideEffectMethods.java:587)
	at edu.umd.cs.findbugs.detect.FindNoSideEffectMethods.sawOpcode(FindNoSideEffectMethods.java:577)
	at edu.umd.cs.findbugs.visitclass.DismantleBytecode.visit(DismantleBytecode.java:884)
	at edu.umd.cs.findbugs.detect.FindNoSideEffectMethods.visit(FindNoSideEffectMethods.java:485)
	at edu.umd.cs.findbugs.visitclass.BetterVisitor.visitCode(BetterVisitor.java:218)
	at edu.umd.cs.findbugs.visitclass.PreorderVisitor.visitCode(PreorderVisitor.java:251)
[...]

Issue #2696, opened 2023-11-13
Eclipse plugin certificate expired


EnderCrypt:
It seems eclipse certificate expired

attempting to install SpotBugs via the eclipse marketplace
(SpotBugs Eclipse plugin 3.1.5.r201806132012-cbbf0a5)
or via downloading the plugin directly from the [official website](https://spotbugs.github.io/eclipse-candidate/) results in:

![image](https://github.com/spotbugs/spotbugs/assets/5500996/fdde9b44-26cf-429c-8123-61672550b3ad)

_i did search for relevant issues and found quite a few about eclipse plugin certificate expiring in the past, though they were all old and closed._

if its not too much to ask, i would appreciate an estimate for when this might get fixed, thank you and have a pleasant day

uwesinha:
I'm facing the same problem with the same certificate in the current version of the Eclipse plugin (4.8.3.r202312121044-1e42fc9). The certificate has been issued to "spotbugs.kengo-toda.jp" and has expired on July 9, 2022. 

Simply trusting the ISRG Root X1 cert allows you to continue installing, but it's still somewhat annoying...

PR #3369, opened 2025-03-26
Run a build on JDK 24 instead of JDK 23

gtoison:


gtoison:
This will need to wait until Gradle supports JDK 24: https://docs.gradle.org/current/userguide/compatibility.html

hazendaz:
for additional tracking, https://github.com/gradle/gradle/issues/32290

gtoison:
I think that `Task :spotbugs-annotations:spotbugsMain` fails because it is analyzing the project with an older version of SpotBugs

hazendaz:
This simply needs rebased then it should now work as gradle 8.14.2 added jdk 24 support

gtoison:
I have rebased but it's a chicken/egg problem: the build uses the last version of SpotBug (through the spotbugs gradle plugin) to analyse itself, but that version wasn't compatible with Java 24.
We'd need to release a new version and then verify that it is indeed compatible with Java 24

srdo:
You might know this already, but you could build your branch of spotbugs locally and use the Gradle task that publishes to the local Maven repo, then update your branch to use the snapshot version of spotbugs. That'd let you verify locally whether the tests pass using the latest spotbugs, before you release anything.

That's what I did to check whether the latest master commit would work for Java 24 when building Apache Kafka.

hazendaz:
Try merging master in again.  We are using gradle plugin tied to spotbugs 4.9.3 as of last night!

hazendaz:
@gtoison looks like a different issue now.

gtoison:
This is #3320 crashing on the sample reproducing this issue.
We should try analyzing SpotBugs with itself (instead of analyzing it with the latest release) as @srdo said

srdo:
@gtoison I took a look at doing this, here's what I did in case you want to try it:

I made the changes in https://github.com/srdo/spotbugs/tree/java-24-snapshot (two lines, adding the maven local repo and setting the plugin to use 4.9.4-SNAPSHOT)

Edit the toolVersion back to 4.9.3, then run `./gradlew publishToMavenLocal`

Edit the toolVersion to 4.9.4-SNAPSHOT. You should then be able to run whatever tests you need. I ran for example `./gradlew spotbugsMain`, and the analysis succeeds on Java 24 with 4.9.4-SNAPSHOT where it didn't with 4.9.3. You can probably run the commands CI would usually run to check if they pass.

gtoison:
Thanks, that's very helpful!
I tried that but unfortunately I think it still has the chicken/egg problem because it won't publish the snapshot version to the local repository until it has analysed itself.

I'm trying already to speed up the build and the problem here is that we're using the latest release to analyse the bug samples (from the `spotBugsTestCases` sub project). There will frequently be samples crashing previous versions of SpotBugs, as it is the case here... 
Analysing all these bug samples takes super long too and nobody looks at the result because the samples are analysed again in unit tests.

So the solution would be to stop analysing the samples. We currently do it because the SpotBugs analysis produces a file (`"...\spotbugs\spotbugsTestCases\build\spotbugs\auxclasspath\spotbugsMain"`) with all the dependencies, and we need that file when running the unit tests.
It would be great if instead we could let Gradle write that dependencies list file instead, I tried to do that a while back but couldn't get Gradle to do it (I'm more used to Maven)

srdo:
> I tried that but unfortunately I think it still has the chicken/egg problem because it won't publish the snapshot version to the local repository until it has analysed itself.

Right, but the spotbugs you put in the Maven repo doesn't need to be built with Java 24.

If you use Java 21 to publish the snapshot (which I did), you can then switch to Java 24 after that (I can recommend https://sdkman.io/ for this kind of thing) and see if it works.

As an alternative, you could disable the spotbugs scan for the command that publishes to the local repo. I'm sure there's a flag that can be passed to skip spotbugs.

srdo:
I went and ran the build workflow commands locally for Java 24

https://github.com/spotbugs/spotbugs/blob/master/.github/workflows/build.yml

There is one test failure, which I've fixed here https://github.com/spotbugs/spotbugs/pull/3488

Other than that, the build passes on Java 24, once spotbugs is upgraded to use a snapshot of current master.

So I think the steps to get this PR in should be something like this:

* Review and merge the linked PR fixing the test
* Release Spotbugs 4.9.4
* Upgrade this PR to Spotbugs 4.9.4. The tests should ideally then pass.

On a side note, the CI build has been broken for a few months, builds of master fail in "Pre-build setup" with 

```
Run echo "keystorepass=${KEYSTORE_PASS}" >> gradle.properties
gpg: decryption failed: Bad session key
Error: Process completed with exit code 2.
```

hazendaz:
releasing 4.9.4 is not necessary, the prior release was already java 24 compliant.  In fact, this already built ok other than the test you fixed.  I'll be merging that then I'll just merge this since that was the only failure.

hazendaz:
for the gpg issue, we are aware, @KengoTODA changed it and its been broken since.  Either he has to fix it or we are going to revamp the entire process including all new keys since we don't have the account data for what is there.

srdo:
> releasing 4.9.4 is not necessary, the prior release was already java 24 compliant.

4.9.3 is affected by https://github.com/spotbugs/spotbugs/issues/3320, which triggers when running Spotbugs on this project with Java 24. If I run `./gradlew check` in this repo on master using a Java 24 JDK, I get

```
The following errors occurred during analysis:
  Can't get stack offset 1 from [< Ljava/lang/invoke/MethodHandle;, 0:ghIssues.Issue3320$MethodHandleFactory.methodHandle Ljava/lang/invoke/MethodHandle; >] @ 4 in ghIssues.Issue3320$MethodHandleFactory.invoke : ()LghIssues.Issue3320$Alpha;
    java.lang.IllegalArgumentException: 1 is not a value stack offset
```
as part of the output.

So it's maybe not a huge deal to merge this before 4.9.4, but it'll break the spotbugs scan of spotbugs itself, and it's not entirely accurate that 4.9.3 works on Java 24, at least not in all cases.

hazendaz:
I'm not too concerned with outstanding bugs in 4.9.3.  Clearly those will be fixed when we do release.  We haven't released mainly because gpg is broken and we need the original author to fix it before we release anything.  The build does fully work otherwise which was confirmed last night when I merged the test you fixed and this PR.

gtoison:
@hazendaz, I'm not sure what might have been discussed with Kengo but he said he wanted to retire from the project.
I don't think he will put his personal gpg key and it might be time to sign the project with a new key.
For all we know that key was potentially mishandled and shouldn't be used anymore.

hazendaz:
hi @gtoison I meant back in March he changed the values in github actions which is why the current break on builds exists since.  I've got zero way to fix any of that as they physical jks.gpg sitting in the repo is owned by him.  One of us would need to figure out correct way to do this and it certainly is not sticking a gpg key behind jks key inside the repo as is done now.  All of those items are his and marked as such.  Following the same pattern IMO is another security risk because we are handing hackers files to try to attack.  I'm sure its exposed, there is no way it isn't, but its locked into @KengoTODA entirely so the entire thing needs gutted and replaced.  I presume many have noticed eclipse stopped releasing here as a result.

Anyway, there is a plugin for gradle that does this which you could find pretty quickly on google (don't have it handy).  It does it with pure setup in github secrets without any file involved.  We need to rewrite half of what is going on as well as there is zero reason to be signing aggressively as is done now.  I don't think we would even notice master failing as much as we do every single renovate PR because the signing is so aggressive and not needed there at all.

Anyway, I've looked at trying to fix what was in place thinking windows line endings got into what he had done in March but that doesn't appear the case so it all needs gutted.

gtoison:
Unless I misunderstood there are two separate issues:
- The GPG key used to sign the jars needs to be replaced. That one is for the sonatype requirement that jars must be signed.
- The keystore used to sign the Eclipse plugin needs to be replaced. That one is because Eclipse shows warning when running unsigned plugins, however the code signing certificate sitting in the keystore has expired a long time ago, so it was still showing warnings. From what I remember the keystore itself contains a certificate from letsencrypt.
See: https://dev.to/kengotoda/sign-eclipse-plugin-by-gradle-4md8 and #2696 

Regarding the solution to replace the keystore, I agree that it is not ideal to have it in a public repo.
The keystore has a passphrase and is encrypted with the GPG key. The GPG key (and its passphrase) are Github secrets.

One way to move forward without revamping everything would to store the keystore as a secret.
I can give it a try but would need access to update the project secrets.
We also need to decide whose GPG key we use going forward.

gtoison:
Github actually recommends to use gpg encrypted secret for large files so, all things considered, that encrypted keystore is not too awful security-wise

PR #3376, opened 2025-03-31
fix IllegalArgumentException in `FindNoSideEffectMethods` detector

### LINKED ISSUE #3320: IllegalArgumentException below FindNoSideEffectMethods.getMethodCall()

JuditKnoll:
Fixes #3320.
It looks like `XMethod` and `MethodDescriptor` behave a bit differently with varargs, and this causes the issue.
In the reproducer the [MethodHandle.invoke(Object... args)](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandle.html#invoke-java.lang.Object...-) method is defined with a single vararg parameter, and called without any parameter at all.
`XMethod` translates this to a call of a method with an array parameter - and the `MethodDescriptor` generated from this has the same parameters -, while `MethodDescriptor` to a function without parameters.

----


gtoison:
Hello and thank you @JuditKnoll,
I haven't had time to look into it just yet but this reminds me of https://github.com/spotbugs/spotbugs/pull/3311
Could it be the same issue with polymorphic signature?

ThrawnCA (approved):

JuditKnoll:
> Hello and thank you @JuditKnoll, I haven't had time to look into it just yet but this reminds me of #3311 Could it be the same issue with polymorphic signature?

Thank you for the hint. It seems quite similar, but it differs slightly. 

The relevant field is used in `changesOnlyNewObjects()`, `getMethodCall()` and `sawCall()` functions. (I haven't investigated the `sawCall()` in detail, but it looks like this change should not really cause an issue there.)

Both the `changesOnlyNewObjects()` and the `getMethodCall()` functions (see the definitions below) only use the signatures to get the relevant elements from the stack at the callsite.
In this case the info we need is how many arguments are handed over at the callsite, and only knowing that the function has a polymorphic signature is not enough.
All other usages of these functions either use `getMethodDescriptorOperand()`, or use a predefined `MethodDescriptor` instance. I think in this case using `getMethodDescriptorOperand()` instead of creating the `MethodDescriptor` object from the return value of `getXMethodOperand()` is better. To tell the truth, I'm not sure why it was implemented like this, but it's like this since the class was introduced. This solution fixes the `IllegalArgumentException` throw and does not break any of the tests.

Please let me know, if you see any possible issues with this solution or have a better idea to fix it.

The `changesOnlyNewObjects()` function:
https://github.com/spotbugs/spotbugs/blob/5a78827d415fd91a0398041c72842d4e4d8ece5d/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNoSideEffectMethods.java#L699-L706

The `getMethodCall()` function:
https://github.com/spotbugs/spotbugs/blob/5a78827d415fd91a0398041c72842d4e4d8ece5d/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNoSideEffectMethods.java#L586-L606

ThrawnCA (approved):

gtoison:
The test for this issue passes without the change from #3311:
https://github.com/spotbugs/spotbugs/pull/3311/files#diff-2451c8016f998a3a08f898d163e928c709d00570d4d4bb0c3bcfa9b410355c12

So this seems to be a regression caused by #3311
There a probably other places affected by this; so I'll try to understand what changed exactly

JuditKnoll:
https://github.com/spotbugs/spotbugs/pull/3401 superseeds this one and provides a better solution, so closing this issue. Thanks @gtoison !

PR #3384, opened 2025-04-03
small cleanup in messages

JuditKnoll:
Small (mostly grammar) cleanup in `messages.xml`. I haven't added a changelog entry, since I don't think this small change needs it.

### DIFF
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -5267,7 +5267,7 @@ Probably something else was meant or the condition can be removed.</p>
     <Details>
 <![CDATA[
 <p>Our analysis shows that this object is useless.
-It's created and modified, but its value never go outside the method or produce any side effect.
+It's created and modified, but its value never goes outside the method or produces any side effect.
 Either there is a mistake and object was intended to be used, or it can be removed.</p>
 <p>This analysis rarely produces false-positives. Common false-positive cases include:</p>
 <p>- This object used to implicitly throw some obscure exception.</p>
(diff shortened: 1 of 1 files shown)


spotbugs/etc/messages.xml
```diff
@@ -9486,7 +9485,7 @@ Using floating-point variables should not be used as loop counters, as they are
   <BugCode abbrev="SC">Constructor invokes Thread.start()</BugCode>
   <BugCode abbrev="MS">Mutable static field</BugCode>
   <BugCode abbrev="ME">Mutable enum field</BugCode>
-  <BugCode abbrev="EI">Method returning array may expose internal representation</BugCode>
+  <BugCode abbrev="EI">Method may expose internal representation</BugCode>
```
JuditKnoll, line 9488: Not all `EI` bug is about arrays.

spotbugs/etc/messages.xml
```diff
@@ -5267,8 +5267,8 @@ Probably something else was meant or the condition can be removed.</p>
     <Details>
 <![CDATA[
 <p>Our analysis shows that this object is useless.
-It's created and modified, but its value never go outside of the method or produce any side-effect.
-Either there is a mistake and object was intended to be used or it can be removed.</p>
+It's created and modified, but its value never go outside the method or produce any side effect.
```
ThrawnCA, line 5270: "value never go" -> "value never goes"

"or produce any" -> "or produces any"

JuditKnoll: Good catch! Thanks, fixed it.

ThrawnCA (approved):

hazendaz (approved):

PR #3385, opened 2025-04-03
add `CWO_CLOSED_WITHOUT_OPENED` bugtype to `FindUnreleasedLock` detector

JuditKnoll:
It looks like https://github.com/spotbugs/spotbugs/pull/2055 got stale and the original author didn't react for quite some time, so I'm creating this PR on top of it.
[Here](https://github.com/gonczmisi/spotbugs/compare/lck-08...JuditKnoll:spotbugs:lck-08) is the comparison with the original PR's branch. 

Co-authored-by: @gonczmisi


spotbugs/etc/findbugs.xml
```diff
@@ -502,7 +502,8 @@
           <Detector class="edu.umd.cs.findbugs.detect.FindUselessControlFlow" speed="fast"
                     reports="UCF_USELESS_CONTROL_FLOW,UCF_USELESS_CONTROL_FLOW_NEXT_LINE"/>
           <Detector class="edu.umd.cs.findbugs.detect.FindUnreleasedLock" speed="moderate"
-                    requirejre="1.5" reports="UL_UNRELEASED_LOCK,UL_UNRELEASED_LOCK_EXCEPTION_PATH"/>
+                    requirejre="1.5"
```
ThrawnCA, line 505: I think we're far past the point of compatibility with Java 1.4.

ThrawnCA (approved):

spotbugs/etc/findbugs.xml
JuditKnoll: That's true. I only modified the formatting of this part because of the added bugtype. I think removing the old, outdated require JRE values is out of the scope of this PR. I haven't checked the detector from the POV whether using an older Java version can break anything in it, so I prefer not to remove this.

hazendaz: You can remove old java 5 stuff.  No one using this project can target compilation back that far.  But I agree separate pr for legacy cleanup.  For that anything java 7 and before should be considered.  I would not worry if anyone that old is affected.  Non of that is even supported.

ThrawnCA (approved):

hazendaz (approved):

PR #3399, opened 2025-04-11
Multi-threaded code checks can be skipped by declaring `@NotThreadSafe`

### LINKED ISSUE #3390: Mark classes / code as not threadsafe
Hi,

We use spotbugs in a largeish project and there are a number of false positives where spotbugs reports problems with thread (AT_NONATOMIC_64BIT_PRIMITIVE, AT_STALE_THREAD_WRITE_OF_PRIMITIVE, AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE). 

The problem is that the places where this is reported have been explicitly documented as not threadsafe. So problems may exist but it does not matter because using this code from multiple threads is a problem in general.

Is there a way to mark such classes explicitly as not threadsafe so that spotbugs does not try to find multithreading bugs in code that is not intended to be used from multiple threads?

stevenschlansker:
Fixes #3390

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report `US_USELESS_SUPPRESSION_ON_METHOD` on generated methods ([#3350](https://github.com/spotbugs/spotbugs/issues/3350))
 - Rewrite some member in `ResourceValueFrame.java` to Enum ([#2061](https://github.com/spotbugs/spotbugs/issues/2061))
 
+### Added
+- Multi-threaded code checks can be skipped with `@NotThreadSafe` ([#3390](https://github.com/spotbugs/spotbugs/issues/3390))
+
 ## 4.9.3 - 2025-03-14
 ### Added
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MultiThreadedCodeIdentifierUtils.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/MultiThreadedCodeIdentifierUtils.java
@@ -57,9 +57,6 @@ public class MultiThreadedCodeIdentifierUtils {
 
     public static boolean isPartOfMultiThreadedCode(ClassContext classContext) {
         JavaClass javaClass = classContext.getJavaClass();
-        if (notThreadSafe(javaClass.getAnnotationEntries())) {
-            return false;
-        }
         if (Subtypes2.instanceOf(javaClass, JAVA_LANG_RUNNABLE) ||
                 Stream.of(javaClass.getFields()).anyMatch(MultiThreadedCodeIdentifierUtils::isFieldIndicatingMultiThreadedContainer)) {
             return true;
@@ -155,8 +152,13 @@ public class MultiThreadedCodeIdentifierUtils {
         return false;
     }
 
-    private static boolean notThreadSafe(AnnotationEntry[] annotationEntries) {
-        for (AnnotationEntry annotationEntry : annotationEntries) {
+    /**
+     * @return <code>true</code> if the class is explicitely annotated with <code>NotThreadSafe</code> to document that it is not thread sage
+     */
+    public static boolean isNotThreadSafe(ClassContext classContext) {
+        JavaClass javaClass = classContext.getJavaClass();
+
+        for (AnnotationEntry annotationEntry : javaClass.getAnnotationEntries()) {
             if (annotationEntry.getAnnotationType().endsWith("/NotThreadSafe;")) {
                 return true;
             }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report `US_USELESS_SUPPRESSION_ON_METHOD` on generated methods ([#3350](https://github.com/spotbugs/spotbugs/issues/3350))
 - Rewrite some member in `ResourceValueFrame.java` to Enum ([#2061](https://github.com/spotbugs/spotbugs/issues/2061))
 
+### Added
+- Multi-threaded code checks can be skipped with `@NotThreadSafe`
```
gtoison, line 16: Can you please add a link to #3390 ?

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MultiThreadedCodeIdentifierUtils.java
```diff
@@ -57,6 +57,9 @@ private MultiThreadedCodeIdentifierUtils() {
 
     public static boolean isPartOfMultiThreadedCode(ClassContext classContext) {
         JavaClass javaClass = classContext.getJavaClass();
+        if (notThreadSafe(javaClass.getAnnotationEntries())) {
+            return false;
+        }
```
gtoison, line 62: I think it would be better to introduce a new method here: now the name of the method does not match with what it does

gtoison (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MultiThreadedCodeIdentifierUtils.java
stevenschlansker: Hi @gtoison , I am not sure I understand your request here. From my understanding, "multi-threaded code" here refers to code that is in use from multiple threads and should be checked.
The `@NotThreadSafe` annotation is an explicit declaration from the user - "this code is not intended to be used from multiple threads". Therefore, any code marked `@NotThreadSafe` is definitionally not "part of multi-threaded code", as the user has told us it is not.

Do you have a better method name to propose?

stevenschlansker: @JuditKnoll previously indicated this would be a good place for the logic to live - perhaps they have a suggestion of the best way to resolve this

JuditKnoll: Currently this function is only used in the [SharedVariableAtomicityDetector](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SharedVariableAtomicityDetector.java#L89-L96) to decide whether the class should be visited and the bugs reported. To my best knowledge, there are no ongoing work (open PRs in SpotBugs) depending on this function.
From the detector's POV, it's true, it would prevent Fps, however there could be another usecase, where this annotation check is not necessary. (It's true that it could be easily refactored, if it's necessary, if the code depending on it is inside SpotBugs.)
Introducing a new method with the @NotThreadSafe annotation check and other functionalities of the function leaves open this option (of course with calling the new method from the SharedVariableAtomicityDetector). 
Maybe `isThreadSafeMultiThreadedClass` or `shouldBeThreadSafeMultiThreadedClass` names could reflect this check as well?

gtoison:
I've pushed a commit to illustrate what I meant, IMO the `isPartOfMultiThreadedCode()` logic is complicated enough already and we should avoid adding a somewhat contradictory case in there.
IMO it is slightly better to break things down a bit, let me know if you think there's a better way

ThrawnCA (approved):

JuditKnoll (approved):

spotbugsTestCases/src/fakeAnnotations/javax/annotation/concurrent/NotThreadSafe.java
```diff
@@ -0,0 +1,8 @@
+package javax.annotation.concurrent;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
```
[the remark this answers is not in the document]
stevenschlansker: Ok I updated this

stevenschlansker:
Sounds good to me. Let me know if I should rebase / squash the commit or anything.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MultiThreadedCodeIdentifierUtils.java
```diff
@@ -151,4 +151,18 @@ public static boolean isLocked(Method currentMethod, CFG currentCFG, LockDataflo
         }
         return false;
     }
+
+    /**
+     * @return <code>true</code> if the class is explicitely annotated with <code>NotThreadSafe</code> to document that it is not thread sage
```
hgschmie, line 156: typo

stevenschlansker: fixed

gtoison (approved):
Thank you very much for the contribution!

PR #3403, opened 2025-04-13
Add the unnecessary annotation to US_USELESS_SUPPRESSION_ON_* messages

### LINKED ISSUE #3395: US_USELESS_SUPPRESSION_ON_* should report which annotation is useless
Using spotbugs-maven-plugin 4.9.3.0 , I get warnings like:

```
[ERROR] Medium: Suppressing annotation on the method ... is unnecessary [...] At ...:[lines 74-77] US_USELESS_SUPPRESSION_ON_METHOD
```

However, if there is more than one suppressed annotation, it's a guessing game which annotation was actually useless.

Can the error include the useless annotation(s) please?

gtoison:
Adding the suppressed bug pattern to the message should fix #3395

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report `US_USELESS_SUPPRESSION_ON_METHOD` on generated methods ([#3350](https://github.com/spotbugs/spotbugs/issues/3350))
 - Rewrite some member in `ResourceValueFrame.java` to Enum ([#2061](https://github.com/spotbugs/spotbugs/issues/2061))
 
+### Added
+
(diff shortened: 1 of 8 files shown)


ThrawnCA (approved):

JuditKnoll (approved):
The earlier introduced [WarningSuppressorTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/WarningSuppressorTest.java) could be improved with this added info, but otherwise LGTM. Thank you for your work!

If someone else start to wonder as well: if one @SuppressFBWarning annotation suppresses multiple bugs uselessly (like at [SuppressedBugs.nonSuppressedNpeExactDifferentCase()](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/suppress/SuppressedBugs.java#L37)), for each unnecessary value there is a separate `US_USELESS_SUPPRESSION_ON_*` bug reported.

hazendaz (approved):

PR #3404, opened 2025-04-14
Cache the Eclipse download

gtoison:
- use actions/cache to avoid downloading Eclipse on each build
- on Windows remove the `curl->Invoke-WebRequest` alias since Invoke-WebRequest is significantly slower and has different parameters (~5 minutes on Invoke-WebRequest vs. ~30s on curl)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,10 @@ on: [push, pull_request]
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+env:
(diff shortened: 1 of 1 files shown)


JuditKnoll (approved):
At the first run `Cache Eclipse` step the `Cahce not found the input keys`.
Since it does seem to work on your fork, rerun the jobs, now it works properly on this fork as well. Thanks :)

hazendaz (approved):

Issue #3396, opened 2025-04-10
US_USELESS_SUPPRESSION_ON_* is too opinionated


asheldon:
The documentation reads:
> Leaving them may result in accidental warnings suppression. The annotation was probably added to suppress a warning raised by SpotBugs, but now SpotBugs does not report the bug anymore. Either the bug was solved or SpotBugs was updated, and the bug is no longer raised by that code.

One scenario ignored by this is code generators, like Immutables, [Lombok](https://github.com/projectlombok/lombok/issues/3867), or any other scenario where the generator adds @SuppressFBWarnings to generated code to avoid SpotBugs warnings. In the view of a code generator's users, there's no value in findings in the generated code. However, a code generator author cannot know which defects SpotBugs may find in their generated code now, in the future, or with plugins. It makes sense for them to defensively suppress all warnings, even if they have none. However, if they have no warnings to suppress, they now have a US_USELESS_ warning.

This leads to my second scenario, when a bug is suppressed both by findbugs-exclude.xml and also by an annotation. In organizations of almost any size, it's not uncommon to have shared findbugs-exclude.xml files that are used across multiple Java projects. When the organization deems a particular detector should be disabled by default, they update their findbugs-exclude.xml and the "useless" annotations become a harmless form of historical documentation. No one needs to "fix" them because there's nothing to fix. 

Removing these annotations is work without benefit and makes it harder to adopt new versions of SpotBugs. Annotations aren't "useless" just because they're not currently suppressing something. The problem with "useless" annotations is not that they're useless, but that they may be documenting the existence and suppression of a bug that does not exist, and in generated code, not even that.

Between these two scenarios, it has become very difficult to avoid useless triggering of US_USELESS_SUPPRESSION_ON_ without shutting them off entirely. Even this is unusually difficult for this new detector as it cannot be disabled by modifying findbugs.xml and rebuilding the jar.

ThrawnCA:
Agreed. I can see the merit in saying, "Once you've fixed a bug, you don't need to hide it anymore in case it crops back up." But sometimes the reason for a suppression is because the nature of the code is that you simply don't care if a detector would find problems with it.

gtoison:
PR #3409 should help with scenario 1 by silencing `US_USELESS_SUPPRESSION_ON_*` when the code is annotated with `*.Generated`.

[...]

PR #3409, opened 2025-04-20
Do not report useless suppressions when annotated with *.Generated

gtoison:
As discussed in #3396 tools such as [immutables](https://immutables.github.io/) might create false positives for `US_USELESS_SUPPRESSION_ON_*`
This PR extends the special case of `lombok.Generated` to all `*.Generated` annotations.

I've also updated the unit test from #3364 so the fake 'Generated' annotation doesn't cause the suppressions to be ignored.


ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3396Test.java
```diff
@@ -0,0 +1,14 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3396Test extends AbstractIntegrationTest {
+
+    @Test
+    void testSuppression() {
+        performAnalysis("ghIssues/Issue3396.class");
+
+        assertBugInClassCount("HE_EQUALS_USE_HASHCODE", "ghIssues.Issue3396", 0);
```
JuditKnoll, line 12: As far as I can see #3396 is about US_USELESS_SUPPRESSION_ON_* bugtypes. Can you please add asserts with those bugtypes as well?

spotbugsTestCases/src/java/ghIssues/Issue3396.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.annotation.Generated;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+public class Issue3396 {
+    
+    @Generated("test")
+    @SuppressFBWarnings
+    public boolean equals(Object x) {
```
JuditKnoll, line 12: Maybe the suppressed `HE_EQUALS_USE_HASHCODE` could be mentioned in a comment.

JuditKnoll (changes requested):
I couldn't find a changelog entry added to the `CHANGELOG.md`. Was it left out intentionally or you just forget it?

gtoison:
No, wait, this doesn't work as intented (and as @JuditKnoll noticed the unit test is broken).

`javax.annotation.processing.Generated` and `javax.annotation.Generated` are both `@Retention(SOURCE)` so this wouldn't work with these annotations (SpotBugs can't see them if they are not in the .class file).

On the other hand [`org.immutables.value.Generated`](https://github.com/immutables/immutables/blob/master/value-annotations/src/org/immutables/value/Generated.java) does not have a retention policy so this should default to `CLASS` and it should work with these annotations.

@asheldon would you know if ignoring based on `org.immutables.value.Generated` would help with #3396?

ThrawnCA (approved):

spotbugsTestCases/src/java/ghIssues/Issue3396.java
```diff
@@ -0,0 +1,14 @@
...
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+@Generated
+public class Issue3396 {
+    
+	@SuppressFBWarnings("XYZ")
+    public boolean test(Object x) {
+    	return x == this; 
+	}
```
JuditKnoll, line 13: The indentation is a bit strange here at line 10 and 13.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
```diff
@@ -140,8 +140,13 @@ private void suppressWarning(String pattern, SuppressMatchType matchType) {
...
             suppressionMatcher.addPackageSuppressor(new PackageWarningSuppressor(pattern, matchType, ClassName.toDottedClassName(getPackageName())));
         } else if (visitingMethod()) {
-            MethodWarningSuppressor suppressor = new MethodWarningSuppressor(pattern, matchType, clazz, MethodAnnotation.fromVisitedMethod(this),
-                    MemberUtils.isUserGenerated(getMethod()));
+            MethodWarningSuppressor suppressor = new MethodWarningSuppressor(
+                    pattern,
+                    matchType,
+                    clazz,
+                    MethodAnnotation.fromVisitedMethod(this),
+                    MemberUtils.isUserGenerated(getMethod()) && MemberUtils.isUserGenerated(getXClass()));
```
JuditKnoll, line 148: Wouldn't it be slightly more efficient if the two parts of this condition would be switched and the class was first checked?

spotbugs/src/main/java/edu/umd/cs/findbugs/bytecode/MemberUtils.java
```diff
@@ -29,16 +29,20 @@
...
  * Utility to analyze class members.
  */
 public final class MemberUtils {
 
-    private static final String LOMBOK_GENERATED_TYPE = "Llombok/Generated;";
-    @SlashedClassName
-    private static final String LOMBOK_GENERATED_NAME = "lombok/Generated";
+    /**
+     * This will capture annotations such as <code>org.immutables.value.Generated</code> or <code>lombok.Generated</code>.
+     * Note that <code>javax.annotation.Generated</code>, <code>javax.annotation.processing.Generated</code> only have source retention and are not visible to SpotBugs.
```
JuditKnoll, line 42: I think this part could be documented at the `US_SUPPRESSION_*` bugs description.

JuditKnoll (approved):

spotbugsTestCases/src/java/ghIssues/Issue3396.java
gtoison: There was a tab/space mix, this is now fixed

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
gtoison: I tried to let `MemberUtils.isUserGenerated(getMethod())` consider whether either the method or the class are generated but the method does not have a reference to the class descriptor unfortunately.

spotbugs/src/main/java/edu/umd/cs/findbugs/bytecode/MemberUtils.java
gtoison: I have updated the bug description 👍

gtoison:
Thanks for the review @ThrawnCA and @JuditKnoll 
I've realized that the previous version was only addressing the issue for unnecessary issues at method level, I made a change to also consider parameters, fields, classes and packages

ThrawnCA (approved):

Issue #3334, opened 2025-02-24
Do not report RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT with Mockito.doAnswer()


peterxcli:
## Description

```java
SnapshotDiffManager spy = spy(snapshotDiffManager);

doAnswer(invocation -> {
      String[] split = invocation.getArgument(0, String.class).split("/");
      String keyName = split[split.length - 1];
      return Integer.parseInt(keyName.substring(3)) % 2 == 0;
    }
).when(spy).isKeyInBucket(anyString(), anyMap(), anyString());
```

This code leads to:
```
M D RV: Return value of SnapshotDiffManager.isKeyInBucket(String, Map, String) ignored, but method has no side effect  At TestSnapshotDiffManager.java:[line 689]
```

In a `Mockito.doAnswer(...)` chained call it is OK to ignore the return value.

## Related link

- https://github.com/spotbugs/spotbugs/issues/872
- https://github.com/spotbugs/spotbugs/pull/2550

MateuszStefek:
The same for `Mockito.doReturn()`:

`Mockito.doReturn(aReturnValue).when(receiver).someMethod()`

I guess the solution for #2550 needs to be extended.

peterxcli:
Thanks for this fix!

PR #3410, opened 2025-04-22
Do not report `RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT` with Mockito doAnswer() and doReturn()

gtoison:
Followup to #2550 and fix for #3334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Rewrite some member in `ResourceValueFrame.java` to Enum ([#2061](https://github.com/spotbugs/spotbugs/issues/2061))
 - Ignore non-interpreted text when looking for `FS_BAD_DATE_FORMAT_FLAG_COMBO` ([#3387](https://github.com/spotbugs/spotbugs/issues/3387))
 - Fix IllegalArgumentException thrown from `FindNoSideEffectMethods` detector ([#3320](https://github.com/spotbugs/spotbugs/issues/3320))
+- Do not report `RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT` when part of a Mockito `doAnswer()` or `doReturn()` call ([#3334](https://github.com/spotbugs/spotbugs/issues/3334))
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


gtoison:
Good point and thanks for the link, I'll look into it!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MethodReturnCheck.java
```diff
@@ -66,6 +68,14 @@ private static enum State {
         INVOKE_OPCODE_SET.set(Const.INVOKEVIRTUAL);
     }
 
+    private static final List<String> MOCKITO_VOID_STUBBING_METHODS = Arrays.asList(
```
JuditKnoll, line 71: Since this collection is never modified and not intended to be modified, it could be immutable and since all elements are only contained once, it could be a Set instead of a List.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Rewrite some member in `ResourceValueFrame.java` to Enum ([#2061](https://github.com/spotbugs/spotbugs/issues/2061))
 - Ignore non-interpreted text when looking for `FS_BAD_DATE_FORMAT_FLAG_COMBO` ([#3387](https://github.com/spotbugs/spotbugs/issues/3387))
 - Fix IllegalArgumentException thrown from `FindNoSideEffectMethods` detector ([#3320](https://github.com/spotbugs/spotbugs/issues/3320))
+- Do not report `RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT` when part of a Mockito `doAnswer()` or `doReturn()` call ([#3334](https://github.com/spotbugs/spotbugs/issues/3334))
```
JuditKnoll, line 19: This changelog entry could be updated to not highlight only `doAnswer()` and `doReturn()`.

JuditKnoll (changes requested):
Just leaving the [Issue872Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue872Test.java) file link here for easy lookup.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MethodReturnCheck.java
gtoison: You're right, I've made the change 👍

CHANGELOG.md
gtoison: Indeed, I've updated the changelog

JuditKnoll (approved):

peterxcli:
Thanks @gtoison for fixing this, and @JuditKnoll for reviewing this!

Issue #3417, opened 2025-04-26
False positve CT_CONSTRUCTOR_THROW for "Compliant Solution (Java SE 6, Public and Private Constructors)"


garydgregory:
False positve CT_CONSTRUCTOR_THROW for "Compliant Solution (Java SE 6, Public and Private Constructors)"

To reproduce:

```
git clone https://github.com/apache/commons-imaging.git
cd commons-imaging
```

Apply:
```
diff --git a/src/conf/spotbugs-exclude-filter.xml b/src/conf/spotbugs-exclude-filter.xml
index c86373d..90ffdda 100644
--- a/src/conf/spotbugs-exclude-filter.xml
+++ b/src/conf/spotbugs-exclude-filter.xml
@@ -188,8 +188,8 @@
     <Bug pattern="SE_COMPARATOR_SHOULD_BE_SERIALIZABLE" />
   </Match>
   <!-- https://github.com/spotbugs/spotbugs/issues/2710 -->
-  <Match>
-    <Class name="~.*" />
-    <Bug pattern="CT_CONSTRUCTOR_THROW" />
-  </Match>
+<!--   <Match> -->
+<!--     <Class name="~.*" /> -->
+<!--     <Bug pattern="CT_CONSTRUCTOR_THROW" /> -->
+<!--   </Match> -->
 </FindBugsFilter>
```

Run:
```
mvn spotbugs:check -Dcommons.spotbugs.impl.version=4.9.3 -Dcommons.spotbugs.plugin.version=4.9.3.0
```

Which gives you:
```
[INFO] --- spotbugs:4.9.3.0:check (default-cli) @ commons-imaging ---
[INFO] BugInstance size is 3
[INFO] Error size is 0
[INFO] Total bugs: 3
[ERROR] Medium: Exception thrown in class org.apache.commons.imaging.formats.jpeg.segments.AbstractGenericSegment at new org.apache.commons.imaging.formats.jpeg.segments.AbstractGenericSegment(int, int, InputStream) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [org.apache.commons.imaging.formats.jpeg.segments.AbstractGenericSegment, org.apache.commons.imaging.formats.jpeg.segments.AbstractGenericSegment] At AbstractGenericSegment.java:[line 38]At AbstractGenericSegment.java:[line 38] CT_CONSTRUCTOR_THROW
[...]

PR #3424, opened 2025-04-29
Fix `CT_CONSTRUCTOR_THROW` FP in private and public ctor with private ctor first

JuditKnoll:
Small cleanup in the `ConstructorThrow` detector and fix https://github.com/spotbugs/spotbugs/issues/3417.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
@@ -121,7 +121,7 @@ public class ConstructorThrow extends OpcodeStackDetector {
         }
         if (isFirstPass) {
             collectExceptionsByMethods(seen);
-        } else if (isConstructor()) {
+        } else if (Const.CONSTRUCTOR_NAME.equals(getMethodName())) {
             reportConstructorThrow(seen);
         }
     }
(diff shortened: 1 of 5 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
```diff
@@ -99,6 +98,12 @@ private static boolean hasFinalFinalizer(JavaClass jc) {
                 .anyMatch(m -> "finalize".equals(m.getName()) && "()V".equals(m.getSignature()) && m.isFinal());
     }
 
+    @Override
+    public void visit(Method obj) {
+        hadObjectConstructor = false;
+        super.visit(obj);
+    }
+
```
JuditKnoll, line 106: The fix to the issue is this function and deleting line 89. The other changes in this class are only cleanup. The bug is reported if `hadObjectConstructor` field is true in the second visit in `reportConstructorThrow`, which is set in the second visit, but it was only set back to false during the first visit.

spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest18.java
```diff
@@ -8,6 +8,12 @@
  * @see <a href="https://github.com/spotbugs/spotbugs/issues/2710">GitHub issue</a>
  */
 public class ConstructorThrowNegativeTest18 {
+    private static boolean verify() throws IOException {
+        // Returns true if data entered is valid, else throws an Exception
+        // Assume that the attacker just enters invalid data, so this method always throws the exception
+        throw new IOException("Invalid data!");
+    }
+
```
JuditKnoll, line 16: It looks like this code example and [ConstructorThrowNegativeTest16](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest16.java) were completely the same. So I'm changing it slightly.

ThrawnCA (approved):

hazendaz (approved):

Issue #3428, opened 2025-05-03
False positive `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` for long in private method called by synchronized method


garydgregory:
Hi All:
CC @JuditKnoll 

I think I am getting a false positive `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` for a long instance varibale update in a `private` method only called by `synchronized` methods.

To reproduce:
```
git clone https://github.com/apache/commons-pool.git
cd commons-pool
git checkout 5d41a5185a301e3d135a0b611e5cbc283a367654
mvn clean compile spotbugs:check -Dcommons.spotbugs.impl.version=4.9.3 -Dcommons.spotbugs.plugin.version=4.9.3.0
```

Produces:
```
[INFO] --- spotbugs:4.9.3.0:check (default-cli) @ commons-pool3 ---
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
[ERROR] Medium: Operation on the "destroyCount" shared variable in "SoftReferenceObjectPool" class is not atomic [org.apache.commons.pool3.impl.SoftReferenceObjectPool] At SoftReferenceObjectPool.java:[line 269] AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE
[INFO] 
```

The instance variable `org.apache.commons.pool3.impl.SoftReferenceObjectPool.destroyCount` is updated at line 269 in the `private` method `org.apache.commons.pool3.impl.SoftReferenceObjectPool.destroy(PooledSoftReference<T>)`, and that private method is only called from `synchronized` methods:
- addObject()
- borrowObject()
- invalidateObject((T)
- returnObject(T)

Once the `synchronized` methods exit, the updated `destroyCount` value is visible to other threads, which is why this looks like an FP.

The `toString()` method only reads the `destroyCount` value.

TY.

JuditKnoll:
I'll look into this one.
In the meantime in your next ticket, if you could provide [a minimal, complete and verifiable example](https://stackoverflow.com/help/minimal-reproducible-example) or at least link the class, that would be awesome.
FYI: The `-Dcommons.spotbugs.impl.version=4.9.3 -Dcommons.spotbugs.plugin.version=4.9.3.0` parts of the command are not recognized, if simply commons-pool or some other github repo is cloned, for these to work, some additional settings are necessary. (Last time I modified the pom.xml instead of using these variables.)

Issue #3363, opened 2025-03-20
False positive AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE


iloveeclipse:
`AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` is falsely reported for an absolutely valid (nonsense) code (extracted from a bigger JUnit test), referring to https://wiki.sei.cmu.edu/confluence/display/java/VNA02-J.+Ensure+that+compound+operations+on+shared+variables+are+atomic :

```
public class FalsePositiveTest {

    private volatile boolean hasCalled = false;

    public void test1() {
        assertEquals(false, hasCalled);
        // do something that changes the flag
        hasCalled = false; // Operation on the "hasCalled" shared variable in "FalsePositiveTest" class is not atomic
    }

    public void test2() {
        assertEquals(false, hasCalled);
        // do something that changes the flag
        hasCalled = false; // Operation on the "hasCalled" shared variable in "FalsePositiveTest" class is not atomic
    }
    
    public void snapshot() {
        FalsePositiveTest p = new FalsePositiveTest();
        p.hasCalled = hasCalled; // Operation on the "hasCalled" shared variable in "FalsePositiveTest" class is not atomic
        System.out.println(p);
    }
    
    public static void assertEquals(Object expected, Object actual) {
    }
}
```

The warning seem to assume the operations on the variable inside each method may somehow see wrong value? 

But this assumption here is not correct, the opposite is the case, the test expects the shared variable to change! 
Even if it would be not a test, the assumption that the shared field value should not change if used multiple times in the same method is wrong. It could be a problem but it also could be perfectly expected state like here.

Neither  `test1()` nor `test2()` need to see **same** `hasCalled` value at any time inside the method (the opposite is the case), and the read of the field is guaranteed to be atomic. Sadly enough, removing the `volatile` would mute the spotbugs warning but break the test that actually modifies the field in parallel which is why the field is marked as volatile!

I'm not sure if this is matching to https://wiki.sei.cmu.edu/confluence/display/java/VNA02-J.+Ensure+that+compound+operations+on+shared+variables+are+atomic explanation either, as there is only about compound operations on a shared variable like `++` or `flag=!flag`, not about using shared variable multiple times in the method body.

JuditKnoll:
[...]

PR #3431, opened 2025-05-14
Fix FPs in `SharedVariableAtomicityDetector`

JuditKnoll:
Fixes https://github.com/spotbugs/spotbugs/issues/3428. The testcase is minimized from [SoftReferenceObjectPool](https://github.com/apache/commons-pool/blob/5d41a5185a301e3d135a0b611e5cbc283a367654/src/main/java/org/apache/commons/pool3/impl/SoftReferenceObjectPool.java) class's `destroyCount` field and its usages, which is mentioned at the issue.
The cause of the issue was that the check for the methods visibility and the functions calling the method was only done for the reads/other usages of the relevant field, but not for the method which modifies the field.

----


ThrawnCA (approved):

gtoison:
Thanks for looking into this @JuditKnoll 
Does it also fix #3363 ?

JuditKnoll:
> Thanks for looking into this @JuditKnoll Does it also fix #3363 ?

No, that's a separate issue. I plan to look into that and fix it next week.

gtoison (approved):

PR #3443, opened 2025-05-26
Handle BDDMockito when looking for ignored return values

### LINKED ISSUE #3441: false positive RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT with BDDMockito.then
Exactly the same as [#872](https://github.com/spotbugs/spotbugs/issues/872) except for

`BDDMockito.then(converter).should(times(1)).convert(isA(SFDLBean.class));`

gtoison:
This should fix #3441

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue872Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue872Test.java
@@ -13,7 +13,7 @@ class Issue872Test extends AbstractIntegrationTest {
         assertBugTypeCount("RV_RETURN_VALUE_IGNORED", 1);
         assertBugTypeCount("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT", 1);
 
-        assertBugAtLine("RV_RETURN_VALUE_IGNORED", 12);
-        assertBugAtLine("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT", 18);
+        assertBugAtLine("RV_RETURN_VALUE_IGNORED", 13);
+        assertBugAtLine("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT", 19);
     }
 }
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MethodReturnCheck.java
```diff
@@ -251,6 +251,12 @@ private boolean isCallMockitoInvocation(XMethod method) {
                 && MOCKITO_VOID_STUBBING_METHODS.contains(method.getName());
     }
 
+    private boolean isCallBDDMockitoInvocation(XMethod method) {
+        return method.isStatic()
+                && "org.mockito.BDDMockito".equals(method.getClassName())
+                && "then".equals(method.getName());
```
JuditKnoll, line 257: Can you please also add the similar BDDMockito functions as well, which corresponds to Mockito functions added in https://github.com/spotbugs/spotbugs/pull/3410?
See BDDMockito's [javadoc](https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/BDDMockito.html).

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MethodReturnCheck.java
JuditKnoll: I was thinking about the methods starting with `will` (e.g. `will()`, `willAnswer()`, `willDoNothing()`), for which the javadoc just mentions the original `Mockito.do` methods (e.g. `doAnswer()`, `doNothing()`).  The corresponding Mockito methods are already added to the `MOCKITO_VOID_STUBBING_METHODS` fields to prevent false positives.

In the second example at BDDMockito's javadoc (showing how to use stubbing voids with throwables): 
```java
//given
willThrow(new RuntimeException("boo")).given(mock).foo();
```

In this example I think there would be a hit with the current implementation, which would be a false positive. However I did not test it.
Please let me know if I misunderstood something or there is another angle I don't see.

PR #3480, opened 2025-06-10
Small cleanup in `DumbMethods` detector

JuditKnoll:
It's a small cleanup in `DumbMethods` detector without any functional change: mostly formatting changes to make the code slightly easier to read and a few small fixes.

I haven't added a changelog entry, since this PR doesn't cause any change in the functionality.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
@@ -95,35 +95,37 @@ public class DumbMethods extends OpcodeStackDetector {
 
         @Override
         public void sawOpcode(int seen) {
-            if (seen == Const.INVOKESTATIC && getClassConstantOperand().equals("java/lang/Math")
-                    && (getMethodDescriptorOperand().getName().equals("max") || getMethodDescriptorOperand().getName().equals("min"))) {
-                Object const1 = stack.getStackItem(0).getConstant();
-                Object const2 = stack.getStackItem(1).getConstant();
-                Number n = null;
-                if (const1 != null ^ const2 != null) {
-                    n = (const1 instanceof Number) ? (Number) const1 : (Number) const2;
-                    if (getMethodDescriptorOperand().getName().equals("min")) {
-                        upperBound = n;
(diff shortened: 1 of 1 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
```diff
@@ -1085,15 +1077,6 @@ else if (special == OpcodeStack.Item.HASHCODE_INT) {
                 if (stack.getStackDepth() > 0 && stack.getStackItem(0).getSpecialKind() == OpcodeStack.Item.NON_NEGATIVE) {
                     OpcodeStack.Item top = stack.getStackItem(0);
                     if (top.getRegisterNumber() != -1 && getMaxPC() > getNextPC() + 6) {
-                        int jump1;
-                        int jump2;
```
JuditKnoll, line 1089: These two fields are only assigned values, but never read, so they can be deleted.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
```diff
@@ -95,9 +95,8 @@ public void initMethod(Method method) {
 
         @Override
         public void sawOpcode(int seen) {
-            if (seen == Const.INVOKESTATIC && getClassConstantOperand().equals("java/lang/Math") && (getMethodDescriptorOperand().getName().equals(
-                    "max")
-                    || getMethodDescriptorOperand().getName().equals("min"))) {
+            if (seen == Const.INVOKESTATIC && getClassConstantOperand().equals("java/lang/Math")
+                    && (getMethodDescriptorOperand().getName().equals("max") || getMethodDescriptorOperand().getName().equals("min"))) {
```
ThrawnCA: While we're refactoring, perhaps extract `getMethodDescriptorOperand().getName()` to a local variable? We access it several times.

JuditKnoll: Thanks. Extracted it and some other parts to local vars, where I found it reasonable. Where there was several conditions with specific invoke instructions, I didn't modify it

ThrawnCA (approved):

hazendaz (approved):

Issue #3481, opened 2025-06-10
NP_NULL_ON_SOME_PATH false negative with access method for method


mimers:
if i have code like below:
```java
public class Container {
    @Nullable
    private String name;

    public Container() throws Exception {
        Callable<String[]> shortTask = new Callable<String[]>() {

            @Override
            public String[] call() throws Exception {
                if (!name.isEmpty()) {
                    System.out.println("name = " + name);
                }
                return getShortName().split(",");
            }
        };
        shortTask.call();
    }

    @Nullable
    private String getShortName() {
        return name.length() < 5 ? null : name.split(",")[0];
    }
}
```
now spotbugs will report NP for `name` because field access which decorated by @Nullable annotation :
```xml
  <BugInstance type="NP_NULL_ON_SOME_PATH" priority="2" rank="8" abbrev="NP" category="CORRECTNESS">
    <Class classname="com.example.libdemo.Container$1">
      <SourceLine classname="com.example.libdemo.Container$1" start="12" end="20" sourcefile="Container.java" sourcepath="com/example/libdemo/Container.java"/>
    </Class>
    <Method classname="com.example.libdemo.Container$1" name="call" signature="()[Ljava/lang/String;" isStatic="false">
      <SourceLine classname="com.example.libdemo.Container$1" start="16" end="20" startBytecode="0" endBytecode="110" sourcefile="Container.java" sourcepath="com/example/libdemo/Container.java"/>
    </Method>
    <Field classname="com.example.libdemo.Container" name="name" signature="Ljava/lang/String;" isStatic="false" role="FIELD_CONTAINS_VALUE">
      <SourceLine classname="com.example.libdemo.Container" sourcefile="Container.java" sourcepath="com/example/libdemo/Container.java"/>
    </Field>
    <SourceLine classname="com.example.libdemo.Container$1" start="16" end="16" startBytecode="7" endBytecode="7" sourcefile="Container.java" sourcepath="com/example/libdemo/Container.java" role="SOURCE_LINE_DEREF"/>
    <SourceLine classname="com.example.libdemo.Container$1" start="16" end="16" startBytecode="4" endBytecode="4" sourcefile="Container.java" sourcepath="com/example/libdemo/Container.java" role="SOURCE_LINE_KNOWN_NULL"/>
  </BugInstance>
```
but not the `getShortName()` method access with same annotation.

PR #3482, opened 2025-06-10
inline access method for method

mimers:

----


mimers:
fix #3481

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Multi-threaded code checks can be skipped with `@NotThreadSafe` ([#3390](https://github.com/spotbugs/spotbugs/issues/3390))
 - New bug type `CWO_CLOSED_WITHOUT_OPENED` for locks that might be released without even being acquired. (See [SEI CERT rule LCK08-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK08-J.+Ensure+actively+held+locks+are+released+on+exceptional+conditions)) ([#2055](https://github.com/spotbugs/spotbugs/pull/2055))
   - Breaking change: changed values and new items in `ResourceValueFrame`.
+- Inline access method for method. ([#3481](https://github.com/spotbugs/spotbugs/pull/3481))))
```
JuditKnoll: There are two additional unnecessary `)` at the end.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/BetterCFGBuilder2.java
```diff
@@ -641,6 +654,15 @@ public void optimize(InstructionList instructionList) {
                                     + methodGen.getSignature());
                          */
 
+                    } else if (method != null) {
+                        Instruction replacement;
+                        int index = getIndex(method);
+                        if (method.isStatic()) {
+                            replacement = new INVOKESTATIC(index);
+                        } else {
+                            replacement = new INVOKEVIRTUAL(index);
```
JuditKnoll: Can this ever be invoke special or interface?

JuditKnoll (changes requested):
Can you please add the code example under [spotbugsTestCases](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java) (e.g. in [ghIssues](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java/ghIssues)), and a code testing it under [spotbugs-tests](https://github.com/spotbugs/spotbugs/tree/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs)? The https://github.com/spotbugs/spotbugs/pull/3237 is a good example how a fix is usually added.
There are several assert methods in [AbstractIntegrationTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java) you can use to assert the expected buginstances (e.g. [assertBugAtVar()](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java#L408)).

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/BetterCFGBuilder2.java
mimers: i just realized this can only be invoke special

CHANGELOG.md
JuditKnoll: As far as I can see, there are still some unnecessary `)`s.

spotbugsTestCases/src/java8/Issue3482.java
```diff
@@ -0,0 +1,18 @@
+import edu.umd.cs.findbugs.annotations.CheckForNull;
```
JuditKnoll: Can you please move this to the [java/ghIssues](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java/ghIssues) directory?

CHANGELOG.md
JuditKnoll: Can you please fix the link, as it should be `https://github.com/spotbugs/spotbugs/issues/3481` with `issues`. If you open it the link you provided with `pull` in it, it redirects to `issues` one, but it's better to prevent unnecessary redirects.

spotbugsTestCases/src/java8/Issue3482.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class Issue3482 {
```
JuditKnoll: #3482 is this PR, which fixes the issue #3481. Please rename this class to Issue3481 and the Issue3482Test class to Issue3481Test.

spotbugsTestCases/src/java8/Issue3482.java
```diff
@@ -0,0 +1,18 @@
...
+    public void foo(String[] args) {
+        new Runnable() {
```
JuditKnoll: Is it only problematic in a `Runnable` or some other anonymous classes? Can you please add some test cases without anonymous classes (feel free to add it in the same code example file)?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3482Test.java
```diff
@@ -0,0 +1,14 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3482Test extends AbstractIntegrationTest {
+
+    @Test
+    void testAccessMethodForMethod() {
+        performAnalysis("../java8/Issue3482.class", "../java8/Issue3482$1.class");
+
+        assertBugInMethod("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE", "Issue3482$1", "run");
```
JuditKnoll: Please add another assert which checks the expected number of bugs from the relevant bugtype.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/BetterCFGBuilder2.java
JuditKnoll: I'm not seeing it clearly why it can be only invoke special, and never invoke interface or invoke virtual. Can you please shine some light on it? Maybe it could be also explained in a comment in the code.

spotbugsTestCases/src/java8/Issue3482.java
mimers: according to the java compiler behavior, the access method for private member will produce only when compiler version<= java8

mimers: https://npearson.co.uk/2024-08-25-java-inner-classes/
here's the relative blog that described this difference.

spotbugsTestCases/src/java8/Issue3482.java
mimers: yes, it must be in a nested class which accessing private member of it's host class.

CHANGELOG.md
mimers: sorry, fixed.

spotbugsTestCases/src/java8/Issue3482.java
JuditKnoll: Thanks, I get it now 😄 
To be a bit more precise, under Java 11, since Java 9 and 10 also produces these access methods, and [JEP 181](https://openjdk.org/jeps/181) was introduced in Java 11 - which introduces nest-based access instead of `access$` and mentioned in the refered blog.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/BetterCFGBuilder2.java
JuditKnoll: With your link at https://github.com/spotbugs/spotbugs/pull/3482#discussion_r2154011402, I can see now why.

Here is the docs for reference:
[JEP 181: Nest-Based Access Control](https://openjdk.org/jeps/181) was introduced in Java 11 to replace `access$` methods when accessing inner and nested classes. On the page of the JEP this is mentioned:
> This relaxes the existing constraint that private interface methods must be invoked using invokespecial (JVMS 6.5) and more generally allows invokevirtual to be used for private method invocation, rather than adding to the complex usage rules surrounding invokespecial.

spotbugsTestCases/src/java8/Issue3482.java
JuditKnoll: The fix is for under Java 11, but the code example itself is relevant and a proper bug report should be present for Java 11 and over as well. So, I still think this example should be moved to the java folder from the java8 folder.

Can you please check it: does this work properly for Java 11 and over?

If not, the [NestedAccessUtil](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/util/NestedAccessUtil.java) class can be helpful resolving it for that as well. The fix doesn't need to be in this PR, but it's a helpful info whether it works. If it doesn't work, the test can disabled for newer JDKs with a comment explaining it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3482Test.java
JuditKnoll: Can you please add another assert function to check the expected number of bugs in the code/class itself like this?
```java
assertBugTypeCount("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE", 1);
assertBugInMethod("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE", "Issue3482$1", "run");
```

mimers: yes, it have been changed to `        assertBugInMethodCount("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE", "Issue3481$Nested", "foo", 1);
`

spotbugsTestCases/src/java8/Issue3482.java
mimers: There's no need to worry about java 11, and the bug will report properly with current version of spotbugs.
For more reason, spotbugs now see a plain invoke instruction of original method when compiling classes with java 11.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3482Test.java
JuditKnoll: The two are a bit different. The two asserts which I suggested check this:
- there is exactly one bug of type `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` reported during the analyses, regardless where it is,
- there is bug reported with type `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` in the class `Issue3482$1`, in the function `run`. (There can be several bugs like this.)
The first and the second check makes sure that there is only one bug, and it has the requested characteristics.

The one assert which you have right now checks this:
- there is exactly one bug with type `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` in the class `Issue3482$Nested`, in the function `foo`.

So it doesn't make sure that there are no other bugs reported with this type in the test code.
It's a good practice to check the number of relevant bugs, and list each of them with additional infos with the tests.
Can you please add the additional assert I requested?

spotbugsTestCases/src/java8/Issue3482.java
JuditKnoll: Okay, thank you. In this case, please move the code example to the java folder, as it should work not only under Java 11, and it should work the same way for all java versions. The java8 directory is reserved for those cases, where test does not work over Java 11.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3482Test.java
mimers: ok, i get it.

spotbugsTestCases/src/java8/Issue3482.java
mimers: If move the `Issue3481.java` to the java folder, the version of compiled class file will not be strict to 52 which rely on the [classesJava8 task](https://github.com/spotbugs/spotbugs/blob/10b4da37b016ad20e867bb2cd8ac138a2dd936b8/spotbugsTestCases/build.gradle#L70), then spotbugs will not see the access method at all.

JuditKnoll: Okay, I get it now. Thank you for the explanation. In this case it's okay like this.

JuditKnoll (approved):

mimers:
@JuditKnoll could you please merge this pull request?

Issue #3483, opened 2025-06-10
ASE_ASSERTION_WITH_SIDE_EFFECT_METHOD produce false positives


dougxc:
I've been trying SpotBugs on the Graal compiler code using a snapshot build, running on a JDK 25 early access build. The latter is required since Graal references some recent JDK internals (e.g. [CompilerThreadCanCallJavaScope](https://github.com/openjdk/jdk/blob/master/src/jdk.internal.vm.ci/share/classes/jdk/vm/ci/hotspot/CompilerThreadCanCallJavaScope.java)) and it seems as though SpotBugs relies on the underlying JDK to resolve and analyze such accesses.

However, it seems as though SpotBugs is producing some false positive in my setup. Since I don't know where the problem actually is, I can only provide steps for reproducing the issue.

```
# Set JAVA_HOME to JDK 17
./gradlew clean
./gradlew build
./gradlew distZip
unzip ./spotbugs/build/distributions/spotbugs-4.9.4-SNAPSHOT.zip
export SPOTBUGS_HOME=$PWD/spotbugs-4.9.4-SNAPSHOT
export MX_PYTHON=python3 # Ensure this is python 3.8 or later
git clone https://github.com/graalvm/mx.git
export PATH=$PATH:$PWD/mx
git clone https://github.com/oracle/graal.git
mx -p graal/compiler fetch-jdk labsjdk-ce-latest --to . --auto-alias --strip-contents-home
export JAVA_HOME=$PWD/labsjdk-ce-latest
mx -p graal/compiler build
mx -p graal/compiler spotbugs
```

This should produce output that includes warnings such as:
```
L S ASE: The method called in assertion at jdk.graal.compiler.libgraal.truffle.LibGraalTruffleEntryPoints.getCompilationId(JNI$JNIEnv, JNI$JClass, long, long) may produce a side effect. If assertions are disabled, the method invocation won't get executed and the result of the method might change.  At LibGraalTruffleEntryPoints.java:[line 330]
```

The above ASE warning cannot be right as there is no `assert` anywhere in [LibGraalTruffleEntryPoints.java](https://github.com/oracle/graal/blob/master/compiler/src/jdk.graal.compiler.libgraal/src/jdk/graal/compiler/libgraal/truffle/LibGraalTruffleEntryPoints.java).

gtoison:
Can you check (or share) the `.class` file?
I guess that the Graal compiler could potentially add some asserts for methods annotated with `@CEntryPoint`

[...]

PR #3489, opened 2025-06-16
`TestASM` crashes with constant field with bad naming convention

gtoison:
While investigating #3483 I ran into a crash of the `TestASM` detector.
It is disabled by default so most probably nobody runs it, except when running integration tests because we enable all detectors.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FieldNamingConventionTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FieldNamingConventionTest.java
@@ -1,0 +1,13 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class FieldNamingConventionTest extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("badNaming/FieldNamingConvention.class");
+        assertBugTypeCount("NM_FIELD_NAMING_CONVENTION", 1);
+    }
+}
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/asm/ClassNodeDetector.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/asm/ClassNodeDetector.java
@@ -21,10 +21,13 @@ package edu.umd.cs.findbugs.asm;
 
 import java.util.ArrayList;
 
+import javax.annotation.CheckForNull;
+
 import org.objectweb.asm.tree.ClassNode;
 
 import edu.umd.cs.findbugs.BugReporter;
 import edu.umd.cs.findbugs.Detector2;
+import edu.umd.cs.findbugs.ba.XClass;
 import edu.umd.cs.findbugs.classfile.CheckedAnalysisException;
 import edu.umd.cs.findbugs.classfile.ClassDescriptor;
 import edu.umd.cs.findbugs.classfile.Global;
@@ -40,6 +43,8 @@ public abstract class ClassNodeDetector extends ClassNode implements Detector2 {
 
     protected final BugReporter bugReporter;
 
+    protected XClass xclass;
+
     /**
      * Construct a ClassNodeDetector. The bugReporter is passed to the
      * constructor and stored in a protected final field.
@@ -54,18 +59,32 @@ public abstract class ClassNodeDetector extends ClassNode implements Detector2 {
 
     @Override
     public void visitClass(ClassDescriptor classDescriptor) throws CheckedAnalysisException {
+        xclass = getClassInfo(classDescriptor);
+        if (xclass != null) {
+            FBClassReader cr = Global.getAnalysisCache().getClassAnalysis(FBClassReader.class, classDescriptor);
+            this.interfaces = new ArrayList<>();
+            this.innerClasses = new ArrayList<>();
+            this.fields = new ArrayList<>();
+            this.methods = new ArrayList<>();
+            cr.accept(this, 0);
+        }
+    }
 
-        FBClassReader cr = Global.getAnalysisCache().getClassAnalysis(FBClassReader.class, classDescriptor);
-        this.interfaces = new ArrayList<>();
-        this.innerClasses = new ArrayList<>();
-        this.fields = new ArrayList<>();
-        this.methods = new ArrayList<>();
-        cr.accept(this, 0);
+    @CheckForNull
+    protected XClass getClassInfo(ClassDescriptor classDescr) {
+        if (classDescr == null) {
+            return null;
+        }
+        try {
+            return Global.getAnalysisCache().getClassAnalysis(XClass.class, classDescr);
+        } catch (CheckedAnalysisException e) {
+            bugReporter.reportMissingClass(classDescr);
+            return null;
+        }
     }
 
     @Override
     public void finishPass() {
         // do nothing
     }
-
 }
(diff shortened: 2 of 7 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/asm/ClassNodeDetector.java
```diff
@@ -54,18 +59,32 @@ protected ClassNodeDetector(BugReporter bugReporter) {
...
-        cr.accept(this, 0);
+    @CheckForNull
+    protected XClass getClassInfo(ClassDescriptor classDescr) {
+        if (classDescr == null) {
+            return null;
+        }
+        try {
+            return Global.getAnalysisCache().getClassAnalysis(XClass.class, classDescr);
+        } catch (CheckedAnalysisException e) {
+            bugReporter.reportMissingClass(classDescr);
```
JuditKnoll, line 81: Are you sure that the error will be that the class is missing? `MissingClassException` is a child of `CheckedAnalysisException`, but there are several others as well.
I can see that you moved this function here from the child `CheckRelaxingNullNessAnnotation`, so it comes from there.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FieldNamingConventionTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class FieldNamingConventionTest extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("badNaming/FieldNamingConvention.class");
+        assertBugTypeCount("NM_FIELD_NAMING_CONVENTION", 1);
```
JuditKnoll, line 11: The issue is about a crash, so the test should check whether the analysis throws an exception, not the reported bugs.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/TestASM.java
```diff
@@ -81,8 +82,10 @@ public void visitMethodInsn(int opcode, String owner, String invokedName, String
     public FieldVisitor visitField(int access, String name, String desc, String signature, Object value) {
         if ((access & Opcodes.ACC_STATIC) != 0 && (access & Opcodes.ACC_FINAL) != 0 && (access & Opcodes.ACC_PUBLIC) != 0
                 && !name.equals(name.toUpperCase())) {
+            @DottedClassName
+            String className = xclass.getClassDescriptor().getDottedClassName();
             bugReporter.reportBug(new BugInstance(this, "NM_FIELD_NAMING_CONVENTION", Priorities.LOW_PRIORITY).addClass(this)
-                    .addField(this.name, name, desc, access));
+                    .addField(className, name, desc, access));
```
gtoison, line 88: This change here is key: instead of using `this.name` which is the slashed class name from ASM we're using the dotted `xclass.getClassDescriptor().getDottedClassName()`

spotbugs/src/main/java/edu/umd/cs/findbugs/asm/ClassNodeDetector.java
gtoison: Sorry, I forgot to explain what was the issue: there was a dotted/slashed class name confusion.

I have moved the `xclass` field to the base class for ASM detectors (there are only two of them) so `TestASM` also has the dotted class name.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FieldNamingConventionTest.java
gtoison: I think we want to test the desired behaviour (not crashing AND detecting the bug) but in its current state the detector was crashing when detecting the bug.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/TestASM.java
gtoison: Here's the stacktrace of the crash:
```
 FieldNamingConventionTest > testIssue() FAILED
    java.lang.AssertionError: classname badNaming/FieldNamingConvention should be dotted
        at edu.umd.cs.findbugs.PackageMemberAnnotation.<init>(PackageMemberAnnotation.java:81)
        at edu.umd.cs.findbugs.PackageMemberAnnotation.<init>(PackageMemberAnnotation.java:57)
        at edu.umd.cs.findbugs.FieldAnnotation.<init>(FieldAnnotation.java:92)
        at edu.umd.cs.findbugs.FieldAnnotation.<init>(FieldAnnotation.java:121)
        at edu.umd.cs.findbugs.BugInstance.addField(BugInstance.java:1158)
        at edu.umd.cs.findbugs.detect.TestASM.visitField(TestASM.java:85)
        at org.objectweb.asm.ClassVisitor.visitField(ClassVisitor.java:356)
        at org.objectweb.asm.ClassReader.readField(ClassReader.java:1138)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:740)
        at edu.umd.cs.findbugs.asm.FBClassReader.accept(FBClassReader.java:44)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:425)
        at edu.umd.cs.findbugs.asm.ClassNodeDetector.visitClass(ClassNodeDetector.java:63)
        at edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$1(FindBugs2.java:1100)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at edu.umd.cs.findbugs.CurrentThreadExecutorService.execute(CurrentThreadExecutorService.java:86)
        at java.base/java.util.concurrent.AbstractExecutorService.invokeAll(AbstractExecutorService.java:242)
        at edu.umd.cs.findbugs.test.AnalysisRunner.run(AnalysisRunner.java:115)
        at edu.umd.cs.findbugs.test.AnalysisRunner.run(AnalysisRunner.java:89)
        at edu.umd.cs.findbugs.AbstractIntegrationTest.performAnalysis(AbstractIntegrationTest.java:149)
        at edu.umd.cs.findbugs.detect.FieldNamingConventionTest.testIssue(FieldNamingConventionTest.java:10)
```

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FieldNamingConventionTest.java
JuditKnoll: Thank you for the explanation. In this case it's okay like this.

spotbugs/src/main/java/edu/umd/cs/findbugs/asm/ClassNodeDetector.java
JuditKnoll: I have no issue of the moving the `getClassInfo()` function from the child  `CheckRelaxingNullnessAnnotation` to the parent `ClassNodeDetector`.
My problem is only when `CheckedAnalysisException` is handled, it always looks like a `MissingClassException`, which is only a child of it, and this may hide another exception, making a debug a bit more problematic.
What do you think of an exception handling similar to the one in `ResolveAllReferences`?
https://github.com/spotbugs/spotbugs/blob/ee7743983a6977db5a57399b40221cb9d1d7dc9f/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ResolveAllReferences.java#L61-L68

gtoison: Ah, I see, sorry I misunderstood your point.
Indeed better keep track of the exception.

What about another PR introducing `IErrorLogger.reportMissingClass(ClassDescriptor, CheckedAnalysisException)`?
This would hopefully reduce a bit the clutter and the code duplications.

gtoison:
I've added a changelog entry and started a draft PR to improve the error handling: #3490
Let me know if you had something else in mind. Thanks for reviewing!

hazendaz (approved):

PR #3490, opened 2025-06-16
Improve missing class error reporting

gtoison:
`analysisCache.getClassAnalysis(JavaClass.class, ...)` throws `CheckedAnalysisException` which could be a `MissingClassException`. Added a reporting method so we don't need to use a try/multi catch block

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
@@ -995,7 +995,7 @@ public class FindBugs2 implements IFindBugsEngine, AutoCloseable {
                     XClass info = Global.getAnalysisCache().getClassAnalysis(XClass.class, desc);
                     factory.intern(info);
                 } catch (CheckedAnalysisException | RuntimeException e) {
-                    AnalysisContext.logError("Couldn't get class info for " + desc, e);
+                    AnalysisContext.currentAnalysisContext().getLookupFailureCallback().reportMissingClass(desc, e);
                     badClasses.add(desc);
                 }
             }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
@@ -994,7 +994,7 @@ public class FindBugs2 implements IFindBugsEngine, AutoCloseable {
                 try {
                     XClass info = Global.getAnalysisCache().getClassAnalysis(XClass.class, desc);
                     factory.intern(info);
-                } catch (CheckedAnalysisException | RuntimeException e) {
+                } catch (CheckedAnalysisException e) {
                     AnalysisContext.currentAnalysisContext().getLookupFailureCallback().reportMissingClass(desc, e);
                     badClasses.add(desc);
                 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
@@ -47,14 +47,14 @@ public interface IErrorLogger {
      *
      * @param classDescriptor
      *            ClassDescriptor of a missing class
-     * @param exception
-     *            The exception thrown when trying to get the class
+     * @param throwable
+     *            The throwable thrown when trying to get the class
      */
-    public default void reportMissingClass(ClassDescriptor classDescriptor, CheckedAnalysisException exception) {
-        if (exception instanceof MissingClassException) {
+    public default void reportMissingClass(ClassDescriptor classDescriptor, Throwable throwable) {
+        if (throwable instanceof MissingClassException) {
             reportMissingClass(classDescriptor);
         } else {
-            logError("Could not find class " + classDescriptor.getDottedClassName(), exception);
+            logError("Could not find class " + classDescriptor.getDottedClassName(), throwable);
         }
     }
 
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
@@ -47,14 +47,14 @@ public interface IErrorLogger {
      *
      * @param classDescriptor
      *            ClassDescriptor of a missing class
-     * @param throwable
-     *            The throwable thrown when trying to get the class
+     * @param exception
+     *            The exception thrown when trying to get the class
      */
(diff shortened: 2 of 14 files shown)


gtoison:
Hello @JuditKnoll 
As discussed in #3489 I've tried to improve a bit the error reporting when a class is missing.
When a `CheckedAnalysisException` is caught it could be a `MissingClassException`, either way we want to keep as much details as possible.
In some cases I was not sure how to proceed because the code was not so straightforward, so I did not modify all the places where we call `Global.getAnalysisCache().getClassAnalysis(XClass.class, classDescriptor)`
Let me know if this makes sense, thank you!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/Naming.java
```diff
@@ -346,6 +346,7 @@ private static boolean mightInheritFromException(ClassDescriptor d) {
                 XClass classNameAndInfo = Global.getAnalysisCache().getClassAnalysis(XClass.class, d);
                 d = classNameAndInfo.getSuperclassDescriptor();
             } catch (CheckedAnalysisException e) {
+                bugReporter.reportMissingClass(d, e);
```
JuditKnoll, line 349: If `AnalysisContext.reportMissingClass(d, e);` is used can the function maybe remain static?

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
```diff
@@ -42,6 +42,22 @@ public interface IErrorLogger {
...
+     * @param classDescriptor
+     *            ClassDescriptor of a missing class
+     * @param throwable
+     *            The throwable thrown when trying to get the class
+     */
+    public default void reportMissingClass(ClassDescriptor classDescriptor, Throwable throwable) {
+        if (throwable instanceof MissingClassException) {
+            reportMissingClass(classDescriptor);
+        } else {
+            logError("Could not find class " + classDescriptor.getDottedClassName(), throwable);
```
JuditKnoll: This message could be refined a bit. What do you think of `"Error looking up class "` instead of `"Could not find class "`? Or something similar which is a bit broader?

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -995,7 +995,7 @@ private void analyzeApplication() throws InterruptedException {
                     XClass info = Global.getAnalysisCache().getClassAnalysis(XClass.class, desc);
                     factory.intern(info);
                 } catch (CheckedAnalysisException | RuntimeException e) {
-                    AnalysisContext.logError("Couldn't get class info for " + desc, e);
+                    AnalysisContext.currentAnalysisContext().getLookupFailureCallback().reportMissingClass(desc, e);
```
JuditKnoll, line 998: Is the `.currentAnalysisContext().getLookupFailureCallback()` part necessary? The new `reportMissingClass()` function seems to be defined on the same level as the priorly used `logError()` function.

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
```diff
@@ -42,6 +42,22 @@ public interface IErrorLogger {
...
 
+    /**
+     * Called to report a class lookup failure.
+     *
+     * @param classDescriptor
+     *            ClassDescriptor of a missing class
+     * @param throwable
+     *            The throwable thrown when trying to get the class
+     */
+    public default void reportMissingClass(ClassDescriptor classDescriptor, Throwable throwable) {
```
JuditKnoll: Are you sure you wanted this with every possible `Throwable`? Maybe this could be a slightly smaller subset, e.g. `CheckedAnalysisException`.

JuditKnoll (commented):
Thank you for your work @gtoison ! It looks a great way forward! 😄 

I haven't looked at all usages of `Global.getAnalysisCache().getClassAnalysis()` and `CheckedAnalysisException` yet, but will do that when I do a proper review.
I left some comments for you.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/Naming.java
gtoison: I thought it was better to avoid using static methods which are using thread local variables.
If we ever attempt again to make SpotBugs more multi-threaded we'll need to sort out these static calls.
When possible I used the bug reporter

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
gtoison: That method (`AnalysisContext.reportMissingClass()`) is a bit different, I think there's still room to improve on this but maybe in a further PR?

spotbugs/src/main/java/edu/umd/cs/findbugs/classfile/IErrorLogger.java
gtoison: The only place where it was not a CheckedAnalysisException was in the FindBug2 class.
I have changed it because that does not seem necessary (it does not look like a RuntimeException can be thrown)

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/Naming.java
JuditKnoll: Thanks for the explanation. In this case it's good like this.

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
JuditKnoll: To tell the truth, I haven't looked into the details and differences of the `reportMissingClass()` functions, just noticed the difference and wondered about it. I agree, no need to do everything in one PR.

gtoison: Thanks for the review @JuditKnoll 
Unless you disagree I think that this PR is a small improvement and can be merged at this point

JuditKnoll (approved):

ThrawnCA (approved):

iloveeclipse:
The commit 0828aed61eab8d0c0ed8f01c985fa686df219f30 breaks Java build, it has Java error in the code 
```
> Task :spotbugs:compileJava
/data/git/spotbugs/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InvalidJUnitTest.java:104: error: non-static variable bugReporter cannot be referenced from a static context
            bugReporter.reportMissingClass(sDesc, e);
            ^
```

PR #3498, opened 2025-06-19
spotbugs-annotations: Use `--release 8`

### LINKED ISSUE #3497: spotbugs-annotations 4.9.1+ prevents `--release 8` Gradle builds
The commit message for a9592576c6f436ad2e3580e818255ca27f06fe15 states:

```
Move spotbugs annotations to java 11 - it was already as it ignored the settings (#3294)

this makes it more clear, the 4.9.0 byte code is already java 11.
```

This is true as far as the _bytecode_ goes, but setting `options.release = 11` has had the unintended consequence of bumping the `org.gradle.jvm.version` requirement in the produced Gradle module metadata:

```
[0] ~/src/spotbugs # ./gradlew clean pTML &>/dev/null
[0] ~/src/spotbugs # jq '.variants[].attributes."org.gradle.jvm.version"' spotbugs-annotations/build/publications/maven/module.json
11
11
null
null
[0] ~/src/spotbugs # perl -i -pe 's/options.release = 11/options.release = 8/g;' spotbugs-annotations/build.gradle
[0] ~/src/spotbugs # ./gradlew clean pTML &>/dev/null
[0] ~/src/spotbugs # jq '.variants[].attributes."org.gradle.jvm.version"' spotbugs-annotations/build/publications/maven/module.json
8
8
null
null
```

This makes it impossible to use spotbugs-annotations to build JDK8 code, even when using a JDK11+ toolchain. See, for example, https://github.com/apache/logging-log4j2/issues/3754.

rschmitt:
Fixes #3497.

### DIFF
--- a/spotbugs-annotations/build.gradle
+++ b/spotbugs-annotations/build.gradle
@@ -3,7 +3,7 @@ apply from: "$rootDir/gradle/javadoc.gradle"
 apply from: "$rootDir/gradle/maven.gradle"
 
 tasks.named('compileJava', JavaCompile).configure {
-    options.release = 11
+    options.release = 8
 }
 
 eclipse {


ThrawnCA (approved):

ppkarwasz:
Thanks everyone for implementing this fix — it definitely resolves the issue some of our users were experiencing (specifically, compile-time errors when using JDK 8, due to `spotbugs-annotations` being exposed via `compileOnlyApi`).

Just a small nit: even though this change doesn't impact consumers directly, would it make sense to also revert the changes made to the Eclipse plugin configuration and the OSGi manifest for consistency?

```groovy
eclipse {
  jdt {
    sourceCompatibility = 8
    targetCompatibility = 8
    javaRuntimeName = "JavaSE-1.8"
  }
}

def manifestSpec = java.manifest {
  attributes(
    'Bundle-Name': 'spotbugs-annotations',
    'Bundle-SymbolicName': 'spotbugs-annotations',
    'Bundle-Version': project.version.replace('-', '.'),
    'Export-Package': 'edu.umd.cs.findbugs.annotations',
    'Bundle-RequiredExecutionEnvironment': 'JavaSE-1.8',
    'Bundle-ManifestVersion': '2'
  )
}
```

hazendaz:
I don't think so.  Eclipse itself requires java 21 for a while now.  We are on an older copy to align with java 11 until we upgrade further.

gtoison:
@hazendaz the issue is that the `spotbugs-annotations` jar can become a transitive dependency for some projects running on older versions of java, in particular for log4j users.
I'm not sure what the `eclipe/jdt` thing is here for, but the OSGi manfinest is currently:
```
Manifest-Version: 1.0
Bundle-Name: spotbugs-annotations
Bundle-SymbolicName: spotbugs-annotations
Bundle-Version: 4.9.3
Export-Package: edu.umd.cs.findbugs.annotations
Bundle-RequiredExecutionEnvironment: JavaSE-11
Bundle-ManifestVersion: 2
Automatic-Module-Name: com.github.spotbugs.annotations
```
We should change it to state that the annotations only require Java 8 (or even 5?) at runtime.

hazendaz:
No java 5.  Anyone on java 5 would not be using latest anything.

Ok, since annotation bit, yes, that one can go back to 8.

gtoison:
I've submitted #3500 to make that change

hazendaz:
> I've submitted #3500 to make that change

merged now.  thanks.

PR #3512, opened 2025-07-03
Fix compilation error introduced in 0828aed

iloveeclipse:
See https://github.com/spotbugs/spotbugs/pull/3490


gtoison:
Sorry about that and thank you for the fix. This looks good to me

iloveeclipse:
> Sorry about that and thank you for the fix. This looks good to me

I just wonder why github actions were all green?

PR #3517, opened 2025-07-07
[gha] Disable gpg signing on *nix for time being and change release tto run from windows

hazendaz:
There is nothing directly wrong with the various keys involved here.  However, the gpg key was generated from windows and seems to contain windows line endings.  Therefore it does not work on *nix.  So this just simply changes it.  Confirmation it was fine was done on the archetype project successfully.  It is still known the underlying keystore used for eclipse is expired and needs replaced or turned off.  The intent for time being is to see if this causes eclipse to release again at least as its not been releasing for some time.  Once that is confirmed, likely will just move to release and come back to the issues later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,18 +59,19 @@ jobs:
           cp -r /Volumes/Eclipse/Eclipse.app /Applications/
           hdiutil detach /Volumes/Eclipse
           echo "eclipseRoot.dir=/Applications/Eclipse.app" | tee eclipsePlugin/local.properties
-      - name: Pre-build setup
-        if: matrix.os != 'windows-latest'
-        run: |
-          echo "keystorepass=${KEYSTORE_PASS}" >> gradle.properties
-          git fetch --no-tags https://$GITHUB_TOKEN@github.com/spotbugs/spotbugs.git +refs/heads/master:refs/remotes/origin/master
-          if [ "$GPG_SECRET_PASSPHRASE" != "" ]; then
-            gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
-          fi
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GPG_SECRET_PASSPHRASE: ${{ secrets.GPG_SECRET_PASSPHRASE }}
-          KEYSTORE_PASS: ${{ secrets.KEYSTORE_PASS }}
+      # Disable gpg signing on *nix for time being until we get new keys established, checks will be on windows only for now
+      #- name: Pre-build setup
+      #  if: matrix.os != 'windows-latest'
+      #  run: |
+      #    echo "keystorepass=${KEYSTORE_PASS}" >> gradle.properties
+      #    git fetch --no-tags https://$GITHUB_TOKEN@github.com/spotbugs/spotbugs.git +refs/heads/master:refs/remotes/origin/master
+      #    if [ "$GPG_SECRET_PASSPHRASE" != "" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -59,18 +59,19 @@ jobs:
...
-          echo "keystorepass=${KEYSTORE_PASS}" >> gradle.properties
-          git fetch --no-tags https://$GITHUB_TOKEN@github.com/spotbugs/spotbugs.git +refs/heads/master:refs/remotes/origin/master
-          if [ "$GPG_SECRET_PASSPHRASE" != "" ]; then
-            gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
-          fi
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GPG_SECRET_PASSPHRASE: ${{ secrets.GPG_SECRET_PASSPHRASE }}
-          KEYSTORE_PASS: ${{ secrets.KEYSTORE_PASS }}
+      # Disable gpg signing on *nix for time being until we get new keys established, checks will be on windows only for now
```
gtoison, line 62: This disables the decryption of the keystore (which contains the expired letsencrypt certificate used to sign the Eclipse plugin), it does not disable the gpg signing, right?

hazendaz:
Nothing is really disabled.  I just stopped it running on Linux.  Note it has been fine on windows.  Only Linux was broken so this puts us to a working state.


However, the git logic on the release part isn't well as a result and I didn't get a chance last night to fix that logic yet so it's cross platform compliant.

The goal here is simply to be working as it was.  It's not to address keys themselves.

Longer term all those keys need to remain and be fixed even if it requires updates every few months.  The tokens require that now as is.  Eclipse does expect signing so it's going backwards if we gut that.  I just don't otherwise have time to fix right now and our keys there were always OK just a rush on updates broke us.

gtoison:
My remark is that the comment says it is disabling the gpg signing. 
In reality the commented-out code was doing the decryption of the keystore, so it's a bit confusing. 

Anyway, thank you very much for looking into this !

hazendaz:
no worries, timing couldn't be any worse though.  Sonatype sunset the old platform on 6/30 as scheduled.  They didn't provide a gradle plugin at all.  The built in one has some instructions to switch over.  I've got it switched over but getting 401.  I Updated the credentials in case that wasn't set around time the switch was done and still 401.  I'm trying to figure that out now.

Here is what is basically working currently.

- The initial changes here
- The git command had to change for powershell
- The logic for eclipse download had to change.  I borrowed the changes I think you did in the build file as that was far easier to follow.
- While gpg works fine in build, that same code is throwing two errors on release but not sure if that is the key its decrypting or the gpg itself.  But that part isn't the important part yet until it gets farther.
- The spotbugs annotations tries to release and that is where I'm stuck at the moment with 401.

I did add a proper name to the release action so it shows up at the bottom now named 'release'.  I've released probably 100 or more times to new central with maven projects so my credentials are good.  I even released other stuff tonight without issue and of course the spotbugs archetype last week.  So I'm missing something in the build.  All info I have thus far says that the same named credentials get passed to gradle and just the change out for the endpoints.  The endpoint does show the current central maven snapshots location so it has to be something simple.

couple links to help

https://central.sonatype.org/publish/publish-portal-gradle/
on there following the built in way to limit changes as I"m not familiar with any of the other possible suggestions.

That will eventually get to here
https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#overview

That talks about it how to now configure it and states that it works.  It does feel to me they have a rather large gap and there are complaints all around in what they have done with many repos having significant problems migrating.  Maven community sounds less than thrilled as well given the mailing lists.  I'll keep looking at this and hopefully get some further progress.  I have at least gotten pretty far and based on the history, our release had been failing since March so we are in a better state, not a good state.

Issue #3520, opened 2025-07-08
Differences between Eclipse and TextUICommandLine execution results


iloveeclipse:
We got a request from our testers, they wanted validate Eclipse results with CI that uses `TextUICommandLine` to run SpotBugs on same code. The problem they have observed is that the SpotBugs reported different warnings depending on the environment.
I plan to provide a fix.

gtoison:
Thanks for looking into it, this is possibly linked to #3292

iloveeclipse:
> Thanks for looking into it, this is possibly linked to [#3292](https://github.com/spotbugs/spotbugs/issues/3292)

Sounds similar, if `CLI` means `TextUICommandLine` and not maven or something else.

PR #3521, opened 2025-07-08
Synchronize Eclipse and TextUICommandLine execution to use same options

iloveeclipse:
`TextUICommandLine` supports now all options encoded in Eclipse preferences file, and Eclipse doesn't hard code analysis options.

Fixes https://github.com/spotbugs/spotbugs/issues/3520

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
@@ -62,6 +62,9 @@ import org.slf4j.LoggerFactory;
  * enable and disable detectors as requested).
  */
 public class TextUICommandLine extends FindBugsCommandLine {
+
+    private static final String USER_PREFS = "-userPrefs";
+
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     /**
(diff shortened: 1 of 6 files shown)


iloveeclipse:
@gtoison : do you like to review or should I merge directly?

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -751,6 +755,16 @@ public void configureEngine(IFindBugsEngine findBugs) throws IOException, Filter
...
+        // Ideally applied last, to make sure if the user preferences file is
+        // given, it overrides the values set by the command line options.
+        if (parsedOptions.containsKey(USER_PREFS)) {
+            // only override values if the user preferences file was given, to avoid re-setting to defaults
+            textuiBugReporter.setPriorityThreshold(userPreferences.getUserDetectorThreshold());
+            textuiBugReporter.setRankThreshold(userPreferences.getFilterSettings().getMinRank());
+            findBugs.setAnalysisFeatureSettings(userPreferences.getAnalysisFeatureSettings());
+            findBugs.setMergeSimilarWarnings(userPreferences.getMergeSimilarWarnings());
+        }
+
```
gtoison, line 778: Maybe we should show a warning when the command line disagree with the preferences file. IMO the command line options should take over, but this is debatable and I'm happy if you wrote it the other way.
But either way, it is going to surprise some users and warning that they're was a conflict would help clarifying the behaviour.

iloveeclipse: Added warning.

iloveeclipse:
I've rebased & added extra conversion of relative filter paths stored in the config, so that should be also now synchnronized.

PR #3522, opened 2025-07-08
[gha] Move git for previous tag to use powershell and better git logic

hazendaz:
The original is not very precise, drops to using system level tail/head that isn't on windows, this should be pure git and while still uses that should not delegate to the platform.


hazendaz:
will be a few more passes, just fixing as they pop to see what needs adjusted.

PR #3535, opened 2025-07-13
Unnecessary suppressions fix for records headers

### LINKED ISSUE #3471: SuppressFBWarnings for EI_EXPOSE_REP/REP2 on record fields flagged as useless starting in 4.9.2
I'm seeing some different behavior in SpotBugs 4.9.2 after upgrading from 4.9.1 (via maven-spotbugs-plugin 4.9.2.0 and 4.9.2.0.  I actually tried using 4.9.3.0, where I first saw the problem, and then tracked it down to the change from .1 to .2.) Here's a simple record that has EI_EXPOSE_REP and EI_EXPOSE_REP2 problems:

```java
package example;

import java.util.List;

public record MyRecord(String name, List<String> values) {}
```

## Behavior in 4.9.1

SpotBugs (via maven-spotbugs-plugin, 4.9.1.0) reports the EI_EXPOSE_REP, and misses the EI_EXPOSE_REP2:

```
Medium: example.MyRecord.values() may expose internal representation by returning MyRecord.values [example.MyRecord] At MyRecord.java:[line 5] EI_EXPOSE_REP
```

In 4.9.1, this can be resolved with by adding a `@SuppressFBWarnings` to the values field:

```java
public record MyRecord(String name, @SuppressFBWarnings("EI_EXPOSE_REP") List<String> values) {}
```

## Behavior in 4.9.2

However, in 4.9.2 (via maven-spotbugs-plugin, 4.9.2.0), the original code results in two warnings (which I think is an improvement, since EI_EXPOSE_REP2 should apply here, too, since `someRecord.values().add(...)` would change the internal representation):

```
[ERROR] Medium: example.MyRecord.values() may expose internal representation by returning MyRecord.values [example.MyRecord] At MyRecord.java:[line 5] EI_EXPOSE_REP
[ERROR] Medium: new example.MyRecord(String, List) may expose internal representation by storing an externally mutable object into MyRecord.values [example.MyRecord] At MyRecord.java:[line 5] EI_EXPOSE_REP2
```

Adding `@SuppressFBWarnings({"EI_EXPOSE_REP","EI_EXPOSE_REP2"})` to the original code doesn't satisfy SpotBugs 4.9.2, though. Instead, the annotation is recognizes as being on the field, the accessor method, and the constructor parameter, but as useless in all three places. 

```
[ERROR] Medium: Suppressing annotation on the field example.MyRecord.values is unnecessary [example.MyRecord] In MyRecord.java US_USELESS_SUPPRESSION_ON_FIELD
[ERROR] Medium: Suppressing annotation on the method example.MyRecord.values() is unnecessary [example.MyRecord] At MyRecord.java:[line 6] US_USELESS_SUPPRESSION_ON_METHOD
[ERROR] Medium: Suppressing annotation on the parameter 3 of the method new example.MyRecord(String, List) is unnecessary [example.MyRecord] At MyRecord.java:[line 6] US_USELESS_SUPPRESSION_ON_METHOD_PARAMETER
```

(Also note, in the last warning, that's it's reported as "parameter 3". I'd have expected either "parameter 2" or "parameter 1", depending on whether parameter indexing is 1-based or 0-based.)

I can handle the EI_EXPOSE_REP by manually defining the accessor method and annotating it, but that's obviously undesirable since part of the appeal of records is automatically-defined accessors:

```java
  @SuppressFBWarnings("EI_EXPOSE_REP")
  public List<String> values() {
    return values;
  }
```

Alternatively, I can address REP and REP2 by copying within a constructor, which is actually what I'd typically _want_ to do, but that doesn't allow for the case where I actually want internal representations exposed (even if it's risky):

```java
  public MyRecord {
    values = List.copyOf(values);
  }
```

Is there some way in 4.9.2 and onward to suppress the EI_EXPOSE_REP and EI_EXPOSE_REP2 warnings on the `values` field?

gtoison:
For records the header is compiled into fields, accessor methods and a canonical constructor.
A `SuppressWarnings` annotation in the header is applied to the corresponding field, accessor method and parameter of the canonical constructor.
In the end we want to report unnecessary suppressions, however when only one of the three suppressor is matched we do not want to report the two others as unnecessary.
We link the suppressors together so we can detect whether at least one of the was matched in `SuppressionMatcher.match`

This should fix #3471

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
@@ -1,9 +1,10 @@
 package edu.umd.cs.findbugs;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import edu.umd.cs.findbugs.annotations.SuppressMatchType;
@@ -22,7 +23,7 @@ public abstract class WarningSuppressor implements Matcher {
     protected final String bugPattern;
     protected final SuppressMatchType matchType;
 
-    private Collection<WarningSuppressor> alternateSuppressors = Collections.emptySet();
+    private Set<WarningSuppressor> alternateSuppressors = Collections.emptySet();
 
     protected WarningSuppressor(String bugPattern, SuppressMatchType matchType) {
         this.bugPattern = bugPattern;
@@ -101,7 +102,7 @@ public abstract class WarningSuppressor implements Matcher {
 
     public void addAlternateSuppressors(Collection<WarningSuppressor> additionalSuppressors) {
         if (alternateSuppressors.isEmpty()) {
-            alternateSuppressors = new ArrayList<>(additionalSuppressors);
+            alternateSuppressors = new HashSet<>(additionalSuppressors);
         } else {
             alternateSuppressors.addAll(additionalSuppressors);
         }
@@ -112,6 +113,6 @@ public abstract class WarningSuppressor implements Matcher {
      * See {@link NoteSuppressedWarnings}
      */
     public Collection<WarningSuppressor> getAlternateSuppressors() {
-        return alternateSuppressors;
+        return Collections.unmodifiableCollection(alternateSuppressors);
     }
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
@@ -1,9 +1,13 @@
 package edu.umd.cs.findbugs;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.regex.Pattern;
 
 import edu.umd.cs.findbugs.annotations.SuppressMatchType;
+import edu.umd.cs.findbugs.detect.NoteSuppressedWarnings;
 import edu.umd.cs.findbugs.detect.UselessSuppressionDetector;
 import edu.umd.cs.findbugs.filter.Matcher;
 import edu.umd.cs.findbugs.xml.XMLOutput;
@@ -17,6 +21,8 @@ public abstract class WarningSuppressor implements Matcher {
 
     protected final String bugPattern;
     protected final SuppressMatchType matchType;
+
+    private Collection<WarningSuppressor> alternateSuppressors = Collections.emptySet();
 
     protected WarningSuppressor(String bugPattern, SuppressMatchType matchType) {
         this.bugPattern = bugPattern;
@@ -82,12 +88,30 @@ public abstract class WarningSuppressor implements Matcher {
     /**
      * @return true if useless suppressions should be reported.
      */
-    public abstract boolean isUselessSuppressionReportable();
+    public boolean isUselessSuppressionReportable() {
+        return true;
+    }
 
     public abstract BugInstance buildUselessSuppressionBugInstance(UselessSuppressionDetector detector);
 
     @Override
     public void writeXML(XMLOutput xmlOutput, boolean disabled) throws IOException {
         // no-op; these aren't saved to XML
+    }
+
+    public void addAlternateSuppressors(Collection<WarningSuppressor> additionalSuppressors) {
+        if (alternateSuppressors.isEmpty()) {
+            alternateSuppressors = new ArrayList<>(additionalSuppressors);
+        } else {
+            alternateSuppressors.addAll(additionalSuppressors);
+        }
+    }
+
+    /**
+     * @return The alternate suppressors that might have been generated from a single <code>SuppressWarnings</code> annotation.
+     * See {@link NoteSuppressedWarnings}
+     */
+    public Collection<WarningSuppressor> getAlternateSuppressors() {
+        return alternateSuppressors;
     }
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
@@ -68,8 +68,8 @@ public class NoteSuppressedWarnings extends AnnotationVisitor implements Detecto
     /**
      * For records the header is compiled into fields, accessor methods and a canonical constructor.
      * A <code>SuppressWarnings</code> annotation in the header is applied to the corresponding field, accessor method and parameter of the canonical constructor.
-     * In the end we want to report unnecessary suppressions, however when only one of the three suppressor is matched we do not want to report the two others as unnecessary.
-     * We link the suppressors together so we can detect whether at least one of the was matched in {@link SuppressionMatcher#match(edu.umd.cs.findbugs.BugInstance)}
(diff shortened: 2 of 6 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
```diff
@@ -82,12 +88,30 @@ public boolean match(BugInstance bugInstance) {
...
+            alternateSuppressors.addAll(additionalSuppressors);
+        }
+    }
+
+    /**
+     * @return The alternate suppressors that might have been generated from a single <code>SuppressWarnings</code> annotation.
+     * See {@link NoteSuppressedWarnings}
+     */
+    public Collection<WarningSuppressor> getAlternateSuppressors() {
+        return alternateSuppressors;
```
JuditKnoll: If we don't want to modify this collection outside of the class, wouldn't it be better to return an unmodifiable copy of it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
```diff
@@ -40,18 +50,32 @@
...
 
 public class NoteSuppressedWarnings extends AnnotationVisitor implements Detector, NonReportingDetector {
 
     private final Set<String> packages = new HashSet<>();
 
+    /**
+     * For records the header is compiled into fields, accessor methods and a canonical constructor.
+     * A <code>SuppressWarnings</code> annotation in the header is applied to the corresponding field, accessor method and parameter of the canonical constructor.
+     * In the end we want to report unnecessary suppressions, however when only one of the three suppressor is matched we do not want to report the two others as unnecessary.
+     * We link the suppressors together so we can detect whether at least one of the was matched in {@link SuppressionMatcher#match(edu.umd.cs.findbugs.BugInstance)}
```
JuditKnoll: A small typo:
```suggestion
     * We link the suppressors together so we can detect whether at least one of them was matched in {@link SuppressionMatcher#match(edu.umd.cs.findbugs.BugInstance)}
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
```diff
@@ -40,18 +50,32 @@
...
 import edu.umd.cs.findbugs.visitclass.AnnotationVisitor;
 
 public class NoteSuppressedWarnings extends AnnotationVisitor implements Detector, NonReportingDetector {
 
     private final Set<String> packages = new HashSet<>();
 
+    /**
+     * For records the header is compiled into fields, accessor methods and a canonical constructor.
+     * A <code>SuppressWarnings</code> annotation in the header is applied to the corresponding field, accessor method and parameter of the canonical constructor.
+     * In the end we want to report unnecessary suppressions, however when only one of the three suppressor is matched we do not want to report the two others as unnecessary.
```
JuditKnoll:
```suggestion
     * In the end we want to report unnecessary suppressions, however when only one of the three suppressors is matched we do not want to report the two others as unnecessary.
```

spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
```diff
@@ -82,12 +88,30 @@ public boolean match(BugInstance bugInstance) {
...
     public void writeXML(XMLOutput xmlOutput, boolean disabled) throws IOException {
         // no-op; these aren't saved to XML
     }
+
+    public void addAlternateSuppressors(Collection<WarningSuppressor> additionalSuppressors) {
+        if (alternateSuppressors.isEmpty()) {
+            alternateSuppressors = new ArrayList<>(additionalSuppressors);
+        } else {
+            alternateSuppressors.addAll(additionalSuppressors);
+        }
```
JuditKnoll: It's a bit strange, that at the initialization a Set is created in line 25, but here it's a List. Can this be a Set every time?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
```diff
@@ -83,6 +109,44 @@ public void visitClassContext(ClassContext classContext) {
...
+
+    @Override
+    public void visitMethod(Method method) {
+        // When we visit a record constructor with args types identical to the fields it is the canonical constructor
+        visitingCanonicalRecordConstructor = isRecord
+                && isConstructor(method)
+                && Arrays.equals(method.getArgumentTypes(), recordComponents.stream().map(c -> c.field.getType()).toArray());
+    }
+
+    private boolean isConstructor(Method method) {
```
JuditKnoll: This method could be static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
```diff
@@ -162,6 +232,11 @@ private void suppressWarning(String pattern, SuppressMatchType matchType) {
                     MemberUtils.isUserGenerated(getMethod()) && MemberUtils.isUserGenerated(getXClass()));
 
             suppressionMatcher.addSuppressor(suppressor);
+
+            if (isRecord) {
```
JuditKnoll, line 236: Maybe the constructor, `toString()`, `hashCode()` and `equals(Object)` methods could be filtered out here.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
gtoison: I've made the returned collection unmodifiable

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
gtoison: typo fixed

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
gtoison: typo fixed

spotbugs/src/main/java/edu/umd/cs/findbugs/WarningSuppressor.java
gtoison: I've made the collection a set for clarity

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/NoteSuppressedWarnings.java
gtoison: Thanks for reviewing!
My understanding is that when an annotation is declarared on a record component in the header, it compiles to the equivalent at three places:
- the canonical constructor's argument
- the field
- the accessor method
This is when the annotation has the appropriate target so, for instance, when the target is `FIELD`, the annotation will only compile on the corresponding field.

The problem here is that the suppress annotation might be necessary for the accessor method, but not for the field or the constructor parameter.

Even though the `toString`, `equals` and `hashCode` are also generated they do not face the same problem: the annotations from the header are not compiled into these methods

JuditKnoll: Thanks for the explanation 😄 . In this case it's okay like this.

JuditKnoll (approved):

Issue #3449, opened 2025-05-27
SF_DEAD_STORE_DUE_TO_SWITCH_FALLTHROUGH_TO_THROW not raised


csrma:
The following code does not raise the `SF_DEAD_STORE_DUE_TO_SWITCH_FALLTHROUGH_TO_THROW` although the while loop does not use the `result` variable (making it a dead store) and the `switch` falls through to `throw`.
```java
class DeadStoreBug {
    public int showBug(int input) {
        int i = 0;
        int result = 0;
        switch (input) {
            case 1:
                result = 10;
                while (i < input) {
                    System.out.println(i);
                    i++;
                }
            case 2:
                throw new RuntimeException("Invalid input");
            default:
                result = input * 2;
        }
        return result;
    }
}
```

SpotBugs version: `4.9.2`
Java: `Temurin openjdk 21.0.7`
Ran from CLI as the following two commands:
```bash
export SPOTBUGS_JAR=/path/to/spotbugs.jar
```
```bash
javac DeadStoreBug.java && \
java -jar $SPOTBUGS_JAR \
    -low \
    -effort:max \
    DeadStoreBug.class
```

Expectation: raises `SF_DEAD_STORE_DUE_TO_SWITCH_FALLTHROUGH_TO_THROW`.

PR #3536, opened 2025-07-15
Dead store fix when switch case contains loops

### LINKED ISSUE #3530: SF_DEAD_STORE_DUE_TO_SWITCH_FALLTHROUGH not detected if loop follows dead store
Dead store due to switch fall through is not detected if a loop (for/while) follows the assignment operation, even though the dead store variable is not used in the loop. If other branching syntax (e.g., if-else) follows the assignment, the rule is detected correctly. It only misses if a loop follows.

Below is a reproducible example.
```java
class DeadStoreDueToSwitchFallThrough {

    public String showBug(int value) {
        String valueType = null;
        switch (value % 2) {
            case 0:
                valueType = "even";        // this falls through and is over-written by valueType = "odd";
                for (int i = 0; i < 5; i++) {
                    System.out.println("Nice even value: " + i);
                }
            case 1:
                valueType = "odd";
                break;
            default:
                valueType = "invalid number";
        }
        return valueType;
    }
}
```
**Expected Outcome:** Error raised on block [5-16].

Spotbugs: `4.9.3`
Java: `Jdk 17` and `Jdk 21`
Run Method: `CLI` as:
```bash
javac DeadStoreDueToSwitchFallThrough.java && \
    java -jar $SPOTBUGS_JAR -textui -effort:max *.class
```

gtoison:
When looking for dead stores due to switch fall through SpotBugs currently gives up when a switch case contains a branch instruction (such as a `break`).
We only care for breaks, because they target an instruction outside of the current case. For other instructions (such as the `goto` of a loop) we are remaining inside the current case.

This should fix #3530 and #3449

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3530Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3530Test.java
@@ -1,0 +1,18 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3530Test extends AbstractIntegrationTest {
+
+    @Test
+    void testFallthrough() {
+        performAnalysis("ghIssues/Issue3530.class");
+
+        assertBugInMethod("SF_DEAD_STORE_DUE_TO_SWITCH_FALLTHROUGH", "ghIssues.Issue3530", "switchForLoop");
+        assertBugInMethod("SF_SWITCH_FALLTHROUGH", "ghIssues.Issue3530", "switchForLoop");
+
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -183,8 +180,9 @@ public void sawOpcode(int seen) {
 
         }
 
-        if (isBranch(seen) || isSwitch(seen) || seen == Const.GOTO || seen == Const.ARETURN || seen == Const.IRETURN || seen == Const.RETURN
-                || seen == Const.LRETURN || seen == Const.DRETURN || seen == Const.FRETURN) {
```
gtoison, line 187: `isBranch(seen)` was redundant with `seen == Const.GOTO` and the various `seen == Const.?RETURN` are factored into `isReturn(seen)`.
The important part is that we're only considering branch targets outside of the current switch case (i.e. a switch break)

JuditKnoll (approved):

Issue #3460, opened 2025-05-28
IL_INFINITE_LOOP not detected in infinite loop


csrma:
The following code runs indefinitely as there is no stop condition in the loop. However, SpotBugs is not able to detect this. I can confirm that it also does not detect similar `for` loop (e.g., `for (;true;)`).
```
class InfiniteLoopExample {

    public void showBug() {
        int counter = 0;
        while (true) { // BUG: IL: An apparent infinite loop (IL_INFINITE_LOOP)
            System.out.println("Counter: " + counter);
            counter++;
        }
    }

    public static void main(String[] args) {
        InfiniteLoopExample example = new InfiniteLoopExample();
        example.showBug();
    }
}
```
This is similar to [issue-1866](https://github.com/spotbugs/spotbugs/issues/1866). However, it is not the same as here the variable inside the loop is local and the infinite loop is still not detected, unlike issue-1866.

SpotBugs version: `4.9.2`
Java: `Temurin openjdk 21.0.7`
Ran from CLI as the following two commands:
```
export SPOTBUGS_JAR=/path/to/spotbugs.jar
```
```
javac InfiniteLoopExample.java && \
java -jar $SPOTBUGS_JAR \
    -low \
    -effort:max \
    InfiniteLoopExample.class
```

Yuvarajpa:
Hi 👋,

I'm Yuvaraj, a beginner in open-source and excited to contribute to this project!  
I’ve gone through the issue and would love to work on it.  

Thanks! 🙌



class InfiniteLoopExample {

    public void showBug() {
        int counter = 0;
        while (counter < 10) {
            System.out.println("Counter: " + counter);
            counter++;
        }
    }

    public static void main(String[] args) {
        InfiniteLoopExample example = new InfiniteLoopExample();
        example.showBug();
    }
}

ThrawnCA:
> I’ve gone through the issue and would love to work on it.

Sure, pull requests are welcome. If you mention the issue by including `#3460` in the pull request description, then GitHub will automatically cross-link them.

PR #3537, opened 2025-07-15
Unconditional while loops no raising `IL_INFINITE_LOOP`

gtoison:
As reported by @csrma in #3460 unconditional while loops are not detected as infinite loops.
This was disabled in https://github.com/spotbugs/spotbugs/commit/60598e01c1151daaf9e22bc853ef6b519d569fa8 with comment: `don't report unconditional infinite loops, even at low priority`

I'm not sure if we want to "fix" this, or maybe document that there's a special case for unconditional loops

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3460Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3460Test.java
@@ -9,6 +9,7 @@ class Issue3460Test extends AbstractIntegrationTest {
     void testIssue() {
         performAnalysis("infiniteLoop/Issue3460.class");
 
-        assertBugTypeCount("IL_INFINITE_LOOP", 1);
+        // unconditional loops are assumed to be intentional
+        assertBugTypeCount("IL_INFINITE_LOOP", 0);
     }
 }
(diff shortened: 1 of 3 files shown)


iloveeclipse:
I assume the reason for code removal were too many false positives.
What would be possible is to seperate this code to a dedicated detector and disable it by default. Whoever wants can enable it and see whether it would make sense to use or not.

gtoison:
Or maybe a new bug code, something like `IL_UNCONDITIONAL_LOOP`?

iloveeclipse:
> Or maybe a new bug code, something like `IL_UNCONDITIONAL_LOOP`?

Bug code is OK, but if there would be too many false positives, how disable that? Therefore I believe it should not report issues by default and AFAIK the easiest way is to use disabled detector.

ThrawnCA:
> I assume the reason for code removal were too many false positives.

There are cases where something should run until an exception is thrown.

gtoison:
I think that the intersection between the users who are going to enable a disabled detectors and the users who are going to find this useful is rather small.
What about documenting that unconditional while loops are not considered to be bugs? This would set the expectations

gtoison:
This was reported again as a false negative (#3758) so I've updated the bug's message to document the exception for unconditional loops

JuditKnoll:
Maybe a changelog entry could be added, but it's okay without it as well.

JuditKnoll:
This doesn't change SpotBugs' code, only the docs, IMO this could be merged without issues.

PR #3539, opened 2025-07-15
Disable the SpotBugs analysis on sample code to speed up the build

gtoison:
The current build includes the analysis (through the SpotBugs gradle plugin) of the test cases, this poses several problems:
- The analysis of the sample code takes a long time
- It is on the critical path so it slows down the overall build
- It runs with an older version of SpotBugs, so it occasionally crashes when a new sample was added and the sample was reproducing a crash.
- 
The analysis produces a 'spotbugsMain' file containing all the dependencies and this is used in the integration test, replace that by a custom gradle task.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
@@ -129,7 +129,7 @@ public abstract class AbstractIntegrationTest {
             throw new UncheckedIOException(e);
         }
 
-        final Path dependencies = getFindbugsTestCases().resolve("build/spotbugs/auxclasspath/spotbugsMain");
+        final Path dependencies = getFindbugsTestCases().resolve("build/runtime-classpath.txt");
         try {
             final List<String> lines = Files.readAllLines(dependencies);
             for (String line : lines) {
--- a/spotbugsTestCases/build.gradle
+++ b/spotbugsTestCases/build.gradle
@@ -40,6 +40,25 @@ dependencies {
   implementation 'org.mockito:mockito-core:5.18.0'
 }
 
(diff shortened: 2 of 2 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java
```diff
@@ -129,7 +129,7 @@ protected void performAnalysis(@SlashedClassName final String... analyzeMe) {
             throw new UncheckedIOException(e);
         }
 
-        final Path dependencies = getFindbugsTestCases().resolve("build/spotbugs/auxclasspath/spotbugsMain");
+        final Path dependencies = getFindbugsTestCases().resolve("build/runtime-classpath.txt");
```
JuditKnoll, line 132: This same path is used in [SpoBugsExtensionTest](https://github.com/spotbugs/spotbugs/blob/master/test-harness-jupiter/src/test/java/edu/umd/cs/findbugs/test/SpotBugsExtensionTest.java#L39) as well. Should it be modified there as well?

gtoison: Thanks for raising this, I'd expect the test to fail but it doesn't, I'll look into it

gtoison: So the test uses the `spotbugsMain` and `spotbugsTest` files produced by the gradle plugin when running on the `test-harness-jupiter` module.
These files still exist as this PR only disable the spotbugs analysis on the test cases module; the test still works as expected.

JuditKnoll: @gtoison thank you for looking into it and for your explanation.

PR #3541, opened 2025-07-17
Small Cleanup

JuditKnoll:
This is a small cleanup, which should not change anything in the functionality, so I haven't added a changelog entry.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/OverridingEqualsNotSymmetrical.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/OverridingEqualsNotSymmetrical.java
@@ -19,7 +19,6 @@
 
 package edu.umd.cs.findbugs.detect;
 
-import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/OverridingEqualsNotSymmetrical.java
```diff
@@ -167,55 +165,43 @@ public void visit(Code obj) {
...
-    boolean sawStaticDelegate;
+    private boolean sawStaticDelegate;
 
-    boolean sawEqualsBuilder;
+    private boolean sawEqualsBuilder;
 
-    boolean isRecord;
-
-    private final EnumMap<EqualsKindSummary.KindOfEquals, Integer> count = new EnumMap<>(
-            EqualsKindSummary.KindOfEquals.class);
```
JuditKnoll, line 209: This field is only used in the `count(EqualsKindSummary.KindOfEquals k)` function below, which is only called once. The function only fills the map, but it's never really read.

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java
```diff
@@ -134,10 +134,6 @@ public abstract class DismantleBytecode extends AnnotationVisitor {
 
     private static final String NOT_AVAILABLE = SlashedClassName.NOT_AVAILABLE;
 
-    static String replaceSlashesWithDots(String c) {
-        return c.replace('/', '.');
-    }
-
```
JuditKnoll, line 140: `ClassName.toDottedClassName()` does basically the same as this one.

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java
```diff
@@ -1158,14 +1146,6 @@ public boolean isRegisterStore(int opcode) {
...
     }
 
-    /**
-     * @param referencedXClass
-     *            The referencedXClass to set.
-     */
-    private void setReferencedXClass(XClass referencedXClass) {
-        this.referencedXClass = referencedXClass;
-    }
-
```
JuditKnoll, line 1168: This setter is private and only used once, to set the field to null.

hazendaz (approved):

PR #3542, opened 2025-07-17
Remove the broken changelog update task

gtoison:
The changelog automation never really worked and currently fails with "Must provide either input tag OR (fromTag and toTag). None were provided!".
Remove this unused task to reduce the clutter

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,24 +31,6 @@ jobs:
           echo "previousTag: $name"
           echo "previousTag=$name" >> $GITHUB_ENV
 
-      - name: Update CHANGELOG
-        id: changelog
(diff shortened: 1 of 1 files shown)


hazendaz:
Isn't this how it's coded right now, in that it should only run during actual release on the tag build to produce the results?  I know it's been problematic, but if removed, are we then manually generating changelog?

gtoison:
Yes, it's looking for the differences between two tags, so this task should run when we release, after the new tag is created.
But we're already updating the changelog manually, and I think that the conflicts friction got more manageable now. This is in part because the PR tend to get merged faster.

hazendaz:
Oh I understand now.  Confused this with the github publishing part.

PR #3549, opened 2025-07-21
Narrow the definition of singletons

### LINKED ISSUE #2985: Another false positive SING_SINGLETON_IMPLEMENTS_SERIALIZABLE when class is not a singleton (e.g. protobuf)
https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java#L218 lists the criteria for recognizing a class as a Singleton. While already quite sophisticated, it wrongly recognizes some non-singleton classes as singleton. E.g. protobuf-generated classes that contain a builder to create an instance like this (`GeneratedMessageV3` is `Serializable`).

Here's a manually simplified version of such a class generated by the `protoc` compiler:

```java
public final class Person extends
    com.google.protobuf.GeneratedMessageV3 implements PersonOrBuilder {
  private static final long serialVersionUID = 0L;

  private static final Person DEFAULT_INSTANCE;
  static {
    DEFAULT_INSTANCE = new Person();
  }

  public static Person getDefaultInstance() {
    return DEFAULT_INSTANCE;
  }

  private Person(com.google.protobuf.GeneratedMessageV3.Builder<?> builder) {
    super(builder);
  }
  private Person() {
    name_ = com.google.protobuf.LazyStringArrayList.emptyList();
  }

  @java.lang.Override
  @SuppressWarnings({"unused"})
  protected java.lang.Object newInstance( <--- this method should already be an indicator that it is not a singleton
      UnusedPrivateParameter unused) {
    return new Person();
  }

  public static Builder newBuilder() {
    return DEFAULT_INSTANCE.toBuilder();
  }

  public static final class Builder extends
      com.google.protobuf.GeneratedMessageV3.Builder<Builder> implements PersonOrBuilder {
    public Person build() {
      return new Person(this);
    }
  }
}
```

gtoison:
Classes exposing a public constructor or a factory method are not considered as singletons

This should fix #2985


gtoison:
Hello @JuditKnoll 
As reported in #2985 some classes might look like singletons because they have some "default" instances but are not because the developer added a public constructor.
Let me know if you think this is reasonable

spotbugsTestCases/src/java/singletons/InnerChildAndMoreInstance.java
```diff
@@ -14,4 +14,7 @@ public static InnerChildAndMoreInstance getUnknown() {
     public static InnerChildAndMoreInstance instance() {
         return INSTANCE;
     }
+
+    protected InnerChildAndMoreInstance() {
+    }
```
JuditKnoll, line 19: I think if the public constructor is the default one, the bug should be still reported. In that case this change here and the one in the `InnerChildAndMoreInstanceReordered` class are not necessary.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2985Test.java
```diff
@@ -0,0 +1,17 @@
...
+class Issue2985Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue2985.class",
+                "ghIssues/Issue2985$NonStaticFactoryMethod.class",
+                "ghIssues/Issue2985$PublicConstructor.class");
+
+        assertBugInClassCount("SING_SINGLETON_IMPLEMENTS_SERIALIZABLE", "ghIssues.Issue2985$NonStaticFactoryMethod", 0);
+        assertBugInClassCount("SING_SINGLETON_IMPLEMENTS_SERIALIZABLE", "ghIssues.Issue2985$PublicConstructor", 0);
```
JuditKnoll, line 15: These could be simplified with a call of `assertNoBugType("SING_SINGLETON_IMPLEMENTS_SERIALIZABLE");`, but it's okay like this as well.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
```diff
@@ -219,16 +218,22 @@ public void visitAfter(JavaClass javaClass) {
         // a class is considered a singleton, if
         // - the user clearly indicated, that it's intended to be a singleton
         // OR
+        // - the class does not have a public constructor
```
JuditKnoll, line 355: I think the bug should still be reported if the public constructor is the default one. What do you think?

JuditKnoll, line 355: This special case could be mentioned in the `messages.xml` at the `SING_SINGLETON_HAS_NONPRIVATE_CONSTRUCTOR` bugtype's description.

JuditKnoll (changes requested):
Thank you for taking this up and for your solution.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Do not report `DM_DEFAULT_ENCODING` for classes compiled with target >= 18 ([#3866](https://github.com/spotbugs/spotbugs/pull/3866))
 - Fix `FS_BAD_DATE_FORMAT_FLAG_COMBO` not suppressed by field-level annotation ([#3838](https://github.com/spotbugs/spotbugs/issues/3838))
 - Fix `SF_SWITCH_FALLTHROUGH` false positives ([#3767](https://github.com/spotbugs/spotbugs/issues/3767))
+- Narrow the definition of singletons ([#3508](https://github.com/spotbugs/spotbugs/issues/3508))
```
JuditKnoll: The link is not correct. The issue is #2985 and this PR is #3549.

JuditKnoll (commented):
IMO this change would be useful with some small adjustments.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
gtoison: The message already has a mention for the default constructor, do you think this needs to be clarified?

> This class is using singleton design pattern and has non-private constructor **(please note that a default constructor might exist which is not private)**. Given that, it is possible to create a copy of the object, thus violating the singleton pattern.
        The easier solution would be making the constructor private.

gtoison:
Hello @JuditKnoll, I've made some adjustments; sorry for the time this is taking, I've been very busy.
Let me know is something needs to be adjustment, I lost a bit track of the outstanding points in the conversation

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
JuditKnoll: Looked into the code once more and there is a the `DefaultConstructor` ([src](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/singletons/DefaultConstructor.java)) code example, in which the bug is still correctly found (see the [test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletonsTest.java#L114)).
So it's good like this.

spotbugsTestCases/src/java/singletons/InnerChildAndMoreInstance.java
JuditKnoll: There is a separate test for that: `DefaultConstructor`, where the bug is still found. Resolving this comment.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
JuditKnoll: Checked the messages.xml once more, and it's not described there what does spotbugs consider a singleton, so it's okay like this.

JuditKnoll:
> Hello @JuditKnoll, I've made some adjustments; sorry for the time this is taking, I've been very busy. Let me know is something needs to be adjustment, I lost a bit track of the outstanding points in the conversation

Hi!
No worries, welcome back. Looked into it and my comments once more, and it's LGTM. Thank you for your work!

JuditKnoll (approved):

ThrawnCA (approved):

JuditKnoll (approved):

JuditKnoll (approved):

ThrawnCA (approved):

hazendaz (approved):

PR #3553, opened 2025-07-23
`SA_FIELD_SELF_ASSIGNMENT` false positive fix

### LINKED ISSUE #2258: [false-positive] SA_FIELD_SELF_ASSIGNMENT reported when value changed
In the [minimal testcase](https://github.com/spotbugs/spotbugs/blob/7238a7ca73dfbf52fef55ea377f7eaa638ee7f59/spotbugsTestCases/src/java/ghIssues/Issue2258.java)
the assignment to `next` is reported as self-assignment, though it's not the case (`orfl` changed since last assignment).

The test code is a simplified version of the [original class](https://github.com/geogebra/geogebra/blob/75890412d55a297ea7e988bb74e9de869b28082a/common/src/main/java/org/geogebra/common/util/clipper/OutRec.java#L32) that triggered this report.

@KengoTODA @baloghadamsoftware this comes from #2161

gtoison:
The `CustomUserValue` annotation disables the "iterative stack analysis" and this results in a false positive in the test case of this PR.
Apparently removing the annotation is compatible with still using "user values" on the stack items

This should fix #2258

### DIFF
--- a/spotbugsTestCases/src/java/ghIssues/Issue2258.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue2258.java
@@ -1,0 +1,24 @@
+package ghIssues;
+
+public class Issue2258 {
+	private boolean isHole;
+	private Issue2258 next;
+	
+	public void fixHoleLinkage() {
+		Issue2258 orfl = next;
+		
+		while (orfl != null && (orfl.isHole == isHole)) {
+			orfl = orfl.next;
+		}
+		
+		next = orfl;
+	}
+	
+	public void setNext(Issue2258 next) {
+		this.next = next;
+	}
+	
+	public void setHole(boolean isHole) {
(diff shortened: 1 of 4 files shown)


spotbugsTestCases/src/java/ghIssues/Issue2258.java
```diff
@@ -0,0 +1,24 @@
...
+	private Issue2258 next;
+	
+	public void fixHoleLinkage() {
+		Issue2258 orfl = next;
+		
+		while (orfl != null && (orfl.isHole == isHole)) {
+			orfl = orfl.next;
+		}
+		
+		next = orfl;
```
gtoison, line 14: Without the change in that PR, the state of the stack here is the same as if the while loop above wasn't there.

PR #3557, opened 2025-07-26
Added `DMI_MISLEADING_SUBSTRING` for calling `subString(0)` on a StringBuffer/StringBuilder

### LINKED ISSUE #1928: False negatives on DMI_USELESS_SUBSTRING
Hi,
I found false negatives on DMI_USELESS_SUBSTRING when using SpotBugs to analyze my project.
> DMI_USELESS_SUBSTRING: This code invokes substring(0) on a String, which returns the original value.
```
String testDMI_USELESS_SUBSTRING(int index){
    StringBuilder sb = new StringBuilder();
    sb.append("}");
    return sb.substring(0);
}
```
Similar to #1927, it seems that SpotBugs does not do the same analysis on the StringBuilder class.

gtoison:
As noted in #1928 on might expect to get `DMI_USELESS_SUBSTRING` on a StringBuilder or StringBuffer when calling `subString(0)`.
That call isn't as useless as String.subString(0), because it converts the builder/buffer to String, so it seemed to me that "useless" was not the best word for that case and I've added a new bug code.

This should fix #1928

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1928Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1928Test.java
@@ -1,0 +1,17 @@
+package edu.umd.cs.findbugs.detect;
+
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue1928Test extends AbstractIntegrationTest {
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue1928.class");
+
+        assertBugInMethod("DMI_MISLEADING_SUBSTRING", "ghIssues.Issue1928", "builder");
+        assertBugInMethod("DMI_MISLEADING_SUBSTRING", "ghIssues.Issue1928", "buffer");
+        assertBugInMethod("DMI_USELESS_SUBSTRING", "ghIssues.Issue1928", "simpleString");
+    }
+}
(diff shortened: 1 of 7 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1928Test.java
```diff
@@ -0,0 +1,17 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue1928Test extends AbstractIntegrationTest {
+    @Test
+    void testIssue() {
+        performAnalysis("ghIssues/Issue1928.class");
+
+        assertBugInMethod("DMI_MISLEADING_SUBSTRING", "ghIssues.Issue1928", "builder");
+        assertBugInMethod("DMI_MISLEADING_SUBSTRING", "ghIssues.Issue1928", "buffer");
+        assertBugInMethod("DMI_USELESS_SUBSTRING", "ghIssues.Issue1928", "simpleString");
```
JuditKnoll: Can you please also check the number of expected relevant bugtypes?

gtoison: Sure, it's done

JuditKnoll: I meant another asserts, so it's also checked that e.g. the `DMI_USELESS_SUBSTRING` is not reported in the `buffer()` function.

gtoison: ah sorry, I misunderstood, I've added more asserts

JuditKnoll (approved):

PR #3558, opened 2025-07-26
Track inner classes access methods to correctly report the bugs

### LINKED ISSUE #2029: False negative for RANGE_ARRAY_INDEX
Hi, I found a false negative about the rule RANGE_ARRAY_INDEX. Please read the example below:

```java
int getLiteral() {
  return 4;
}
public int getValue() {
    return array[getLiteral()];  // should report a warning here
}
```
Based on my experience, SpotBugs is a powerful tool to analyze the literal return value. Hence, I think this is a false negative. Thanks for your consideration.

SpotBugs Version: Latest

gtoison:
Consider the code from the #2029 issue report:
```java
public class Outer {
    public class Inner {
    	void bug(Outer value) throws Throwable {
    		value.finalize();
    	}
    }
}
```
When compiled, the `value.finalize()` is placed in a synthetic "access" method of the outer class (not in the Inner.bug method as one might expect).

Since that method is synthetic, the priority gets adjusted down two levels and the bug instance is filter out.
Even if that bug was not filtered out, it would be reported on that synthetic method, which is on the outer class, and at a weird source line (the `public class Outer` line).

When detecting an issue in an access method (so in the Outer class) we have no knowledge of what inner classes might be calling it, so we need to scan all the classes for such cases first.

This PR adds a new detector running in a first pass and builds a database of these access. Then when a bug instance is found it checks if it was would in an access method; in that case it adjust the class and method to the place where the original inner class called that access method.

This should fix #2029 and also 
- https://github.com/find-sec-bugs/find-sec-bugs/issues/719
- https://github.com/find-sec-bugs/find-sec-bugs/issues/718 
- https://github.com/find-sec-bugs/find-sec-bugs/issues/716


gtoison:
This PR is missing a changelog entry, I'll add it when the other PR I have sent earlier are merged

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/AnnotationDatabase.java
```diff
@@ -109,7 +109,7 @@ public boolean anyAnnotations(AnnotationEnum n) {
     public AnnotationEnum getResolvedAnnotation(Object o, boolean getMinimal) {
         if (o instanceof XMethod) {
             XMethod m = (XMethod) o;
-            if (m.getName().startsWith("access$")) {
+            if (m.isAccessMethod()) {
```
gtoison, line 112: I took the opportunity to create an `isAccessMethod()` to replace these hardcoded strings

JuditKnoll: There are also other usages of this hardcoded string with `org.apache.bcel.classfile.Method` and `org.apache.bcel.generic.InvokeInstruction`. These could be also replaced with util functions (not necessarily in this PR).

Issue #1874, opened 2021-12-14
New Java versions lead to FN about the rule DMI_INVOKING_TOSTRING_ON_ARRAY


Zustin:
Hi, I found a false negative about the rule [DMI_INVOKING_TOSTRING_ON_ARRAY](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#useless-string-invocation-of-tostring-on-an-array-dmi-invoking-tostring-on-array). This case is found by using SpotBugs CLI and IDE, but class files are generated by different Java versions. Please refer to the following case, the class file generated by `java-8` can be detected, but class files generated by `java-11,16,17` cannot be detected.

```java
public class Test {
    private static final String[] gargs = new String[] { "1", "2" };
    public void print() {
        System.out.println("" + gargs);  // should report a warning here
    }
}
```
**Expected Outcome:** A warning at line 4, [Outcomes Repo](https://github.com/Zustin/Issue1874)
SpotBugs version: 4.5.2 and run by CLI

PR #3562, opened 2025-07-30
Fixed `DMI_INVOKING_TOSTRING_ON_ARRAY` on newer JDK

### LINKED ISSUE #1147: [false-negative]DMI_INVOKING_TOSTRING_ON_ARRAY can not detect String automatic conversion when string concatenation
**Affects SpotBugs Version:** 
spotbugs-4.0.0-RC2

**Rule:**
DMI_INVOKING_TOSTRING_ON_ARRAY

**Description:**
Unable to detect String automatic conversion when string concatenation, for example: <code>"foo" + barArray</code>,
This rule is implemented in SpotBugs.jar\main\java\edu\umd\cs\findbugs\detect\FindPuzzlers.java.
This rule uses the following code to detect explicit string conversion:
```
if (seen == Const.INVOKEVIRTUAL
                && stack.getStackDepth() > 0
                && ("toString".equals(getNameConstantOperand()) && "()Ljava/lang/String;".equals(getSigConstantOperand())
                        || "append".equals(getNameConstantOperand())
                                && "(Ljava/lang/Object;)Ljava/lang/StringBuilder;".equals(getSigConstantOperand())
                                && "java/lang/StringBuilder".equals(getClassConstantOperand())
                        || "append".equals(getNameConstantOperand())
                                && "(Ljava/lang/Object;)Ljava/lang/StringBuffer;".equals(getSigConstantOperand())
                                && "java/lang/StringBuffer".equals(getClassConstantOperand()) || ("print".equals(getNameConstantOperand())
                                        || "println".equals(getNameConstantOperand()))
                                        && "(Ljava/lang/Object;)V".equals(getSigConstantOperand())))
```
which includes the following 5 situations: 1.<code>toString</code>, 2.<code>StringBuilder.append(Object)</code>, 3.<code>StringBuffer.append(Object)</code>, 4.<code>print(Object)</code>, 5.<code>println(Object)</code>.
The string conversion performed during string concatenation is ignored.

**Code Sample demonstrating the issue:**
```
public void setRetrieveAttributes(String[] retrieveAttributes)
{
	LOGGER.debug("retrieveAttributes::" + retrieveAttributes);
}
```

**Expected outcome:**
false-negative

**Running SpotBugs through:**
command line

gtoison:
Detect the `String.valueOf(array)` call during string concatenation.

This should fix #1147

### DIFF
--- a/spotbugsTestCases/src/java/ghIssues/Issue1147.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue1147.java
@@ -1,0 +1,9 @@
+package ghIssues;
+
+public class Issue1147 {
+
+	// Test for issues 1147 and 1874
+	public String append(Integer[] array) {
+		return "x" + array;
+	}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
   - Breaking change: changed values and new items in `ResourceValueFrame`.
 - Inline access method for method. ([#3481](https://github.com/spotbugs/spotbugs/issues/3481))
 - Added `DMI_MISLEADING_SUBSTRING` for calling `subString(0)` on a StringBuffer/StringBuilder ([#1928](https://github.com/spotbugs/spotbugs/issues/1928))
+- Fixed `DMI_INVOKING_TOSTRING_ON_ARRAY` on newer JDK ([#1147](https://github.com/spotbugs/spotbugs/issues/1147))
 
 ## 4.9.3 - 2025-03-14
 ### Added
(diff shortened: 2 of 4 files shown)


spotbugsTestCases/src/java/ghIssues/Issue1147.java
```diff
@@ -0,0 +1,9 @@
+package ghIssues;
+
+public class Issue1147 {
+
+	// Test for issues 1147 and 1874
```
JuditKnoll, line 5: #1874 isn't linked as fixed by this PR, but it seems to be a duplicate of #1147 to me. Was this intentionally left out?

JuditKnoll (approved):

spotbugsTestCases/src/java/ghIssues/Issue1147.java
gtoison: This should fix #1874 too indeed!

Issue #3563, opened 2025-07-31
Build fails on master with test failure in Issue1147Test


iloveeclipse:
Seem to be regression from https://github.com/spotbugs/spotbugs/pull/3562

```
./gradlew build
...
> Task :spotbugs-tests:test

Issue1147Test > testIssue() FAILED
    java.lang.AssertionError: 
    Expected: Iterable containing exactly <1> BugInstance with:
    bugType="DMI_INVOKING_TOSTRING_ON_ARRAY",
         but: The collection was empty
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at edu.umd.cs.findbugs.AbstractIntegrationTest.assertBugTypeCount(AbstractIntegrationTest.java:173)
        at edu.umd.cs.findbugs.detect.Issue1147Test.testIssue(Issue1147Test.java:12)

808 tests completed, 1 failed, 5 skipped
```

iloveeclipse:
@gtoison : could you pease check?

gtoison:
So the GHA build passes but fails on your computer?
Can you do `java -version`?

JuditKnoll:
It also fails on my machine with OpenJDK 17 on Windows. 

I get this with running the test:
```
ghIssues.Issue1147.<init> : ()V
  0 aload_0         []::[< LghIssues/Issue1147;, IP, r0 >]
  1 invokespecial   [< LghIssues/Issue1147;, IP, r0 >]::[< LghIssues/Issue1147;, IP, r0 >]
  4 return          []::[< LghIssues/Issue1147;, IP, r0 >]

ghIssues.Issue1147.append : ([Ljava.lang.Integer;)Ljava.lang.String;
  0 aload_1         []::[< LghIssues/Issue1147;, IP, r0 >, < [Ljava/lang/Integer;, IP, r1 >]
  1 invokedynamic   [< [Ljava/lang/Integer;, IP, r1 >]::[< LghIssues/Issue1147;, IP, r0 >, < [Ljava/lang/Integer;, IP, r1 >]
  6 areturn         [< Ljava/lang/String; >]::[< LghIssues/Issue1147;, IP, r0 >, < [Ljava/lang/Integer;, IP, r1 >]

CFGs generated into directory: .\cfg-12618863127270368415. Please do not forget to delete it.
Analyzing ghIssues.Issue1147.<init>()

[...]

PR #3566, opened 2025-07-31
`DMI_INVOKING_TOSTRING_ON_ARRAY` false negative on earlier versions of JDK 11 and 17

gtoison:
See #3563


iloveeclipse:
That's interesting. So you mean later versions changed generated bytecode? I'm using `17.0.6`, which one was on CI server?

gtoison:
The build picks up `17.0.16+8`

JuditKnoll (approved):

iloveeclipse (approved):

iloveeclipse:
Thanks!

PR #3574, opened 2025-08-01
Switch default case fix

### LINKED ISSUE #1148: [false-positive]SF_SWITCH_NO_DEFAULT detect false positive result: some switchs with default are detected.
**Affects SpotBugs Version:** 
spotbugs-4.0.0-RC2

**Rule:**
SF_SWITCH_NO_DEFAULT

**Description:**
Detect false positive result: some switchs with default are detected.
This rule is implemented in SpotBugs.jar\main\java\edu\umd\cs\findbugs\detect\SwitchFallthrough.java.
The rule uses a complex method based on bytecode transfer to detect switch without default, and there are special cases where this method is not applicable.

**Code Sample demonstrating the issue:**
```
example 1:
@Override
protected int updateData(QTBlockData targetBlockData, QuitType targetData, QuitType sourceData, int indexInTarget) {
	FanoutWithLocation fwl = targetBlockData.getFanout(indexInTarget);
	FanoutType ft = fwl.getType();
	EntryId id = fwl.getEntryId();
	int result = 0;
	switch (ft) {	//this line is marked
	case DO:
	case EXTRINSIC:
		CallType ct = targetData.getFanout(id);
		return ct.updateFromFanout(sourceData);
	case GOTO:
	case ASSUMED_GOTO:
		result = targetData.markQuitFromGoto(sourceData, fwl.getCodeLocation());
	default:
		result += targetData.updateCallTypes(sourceData);
	}
	return result;
}		

example 2:
public void setDamageParam(int i, double v){
	switch(i){	//this line is marked
		case 0:  TACKLE = v;
		case 1:  PUNCH = v;
		case 2:  KICK = v;
		case 3:  HEADBUTT = v;
		case 4:  MAGIC = v;
		default:
			break;
	}
}

example 3:
public void addData(UsageType _type, ArrayList<Integer> _list)
{
	switch (_type)	//this line is marked
	{
	case CPU:
		this.cpuData.addAll(_list);

	case DISK:
		this.diskData.addAll(_list);

	case MEMORY:
		this.memData.addAll(_list);

	default:
		break;
	}
}
```

**Expected outcome:**
false-positive

**Running SpotBugs through:**
command line

### LINKED ISSUE #3572: False positive with SF_SWITCH_NO_DEFAULT
Most curious false positive I saw before...
Spotbugs 4.9.4 reports following warning on the innocent code below:

```
Bug: Switch statement found in FalsePositiveOnSwitch.omg(FalsePositiveOnSwitch$MyEnum) where default case is missing
 
 This method contains a switch statement where default case is missing. Usually you need to provide a default case. 
Because the analysis only looks at the generated bytecode, this warning can be incorrect triggered if the default case is at the end of the switch statement and the switch statement doesn't contain break statements for other cases. 
See <a href="https://cwe.mitre.org/data/definitions/478.html">CWE-478: Missing Default Case in Multiple Condition Expression.  
 
Rank: Of Concern (19), 
confidence: Normal

Pattern: SF_SWITCH_NO_DEFAULT 

Type: SF, 
Category: STYLE (Dodgy code)
```

The point is: the warning is only reported if one of the enum cases is missing and existing cases are in order that not match enum definition order. So remove commented out "C" case - warning disappears. Leave "C" but remove "D" - warning disappears.
```
public class FalsePositiveOnSwitch {
    public Object omg(MyEnum e) {
        Object o = null;
        switch (e) {
        case A:
            o = "A";
            break;
        case B:
            o = "B";
            break;
//        case C:
//            o = "C";
//            break;
        case D:
            o = "D";
            break;
        default:
            break;
        }
        return o;
    }
    public enum MyEnum { A, B, C, D, E }
}
```

This is regression from https://github.com/spotbugs/spotbugs/pull/3536.

gtoison:
This should fix #3572 and fix #1148 (follow up of #3536)

### DIFF
--- a/spotbugsTestCases/src/java/ghIssues/Issue1148.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue1148.java
@@ -1,0 +1,50 @@
+package ghIssues;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Issue1148 {
+	private double TACKLE;
+	private double PUNCH;
+	private double KICK;
+	private double HEADBUTT;
+	private double MAGIC;
+	
+	public List<Integer> cpuData;
+	public List<Integer> diskData;
+	public List<Integer> memData;
+
+	public void setDamageParam(int i, double v){
+		switch(i){	//this line is marked
+			case 0:  TACKLE = v;
+			case 1:  PUNCH = v;
+			case 2:  KICK = v;
+			case 3:  HEADBUTT = v;
+			case 4:  MAGIC = v;
+			default:
+				break;
+		}
+	}
+
+	public void addData(UsageType _type, ArrayList<Integer> _list)
+	{
+		switch (_type)	//this line is marked
+		{
+		case CPU:
+			this.cpuData.addAll(_list);
+
+		case DISK:
+			this.diskData.addAll(_list);
+
+		case MEMORY:
+			this.memData.addAll(_list);
+
+		default:
+			break;
+		}
+	}
+	
+	public enum UsageType {
+		CPU, DISK, MEMORY
+	}
+}
--- a/spotbugsTestCases/src/java/ghIssues/Issue1148.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue1148.java
@@ -23,7 +23,7 @@ public class Issue1148 {
 			case 3:  HEADBUTT = v;
 			case 4:  MAGIC = v;
 			default:
-				break; // The compiler does not emit a GOTO for this so we can't tell if it was written, hence  the SF_SWITCH_NO_DEFAULT false positive
+				break; // The compiler does not emit a GOTO for this so we can't tell if it was written, hence the SF_SWITCH_NO_DEFAULT false positive
 		}
 	}
 
--- a/spotbugsTestCases/src/java/ghIssues/Issue1148.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue1148.java
@@ -14,6 +14,7 @@ public class Issue1148 {
 	public List<Integer> diskData;
 	public List<Integer> memData;
 
+	// False positive for SF_SWITCH_NO_DEFAULT
 	public void setDamageParam(int i, double v){
 		switch(i){	//this line is marked
 			case 0:  TACKLE = v;
@@ -22,7 +23,7 @@ public class Issue1148 {
 			case 3:  HEADBUTT = v;
 			case 4:  MAGIC = v;
 			default:
-				break;
+				break; // The compiler does not emit a GOTO for this so we can't tell if it was written, hence  the SF_SWITCH_NO_DEFAULT false positive
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


spotbugsTestCases/src/java/ghIssues/Issue3572.java
```diff
@@ -0,0 +1,49 @@
...
+		switch (e) {
+		case A:
+			o = "A";
+			break;
+		case B:
+			o = "B";
+			break;
+//        case C:
+//            o = "C";
+//            break;
```
JuditKnoll, line 19: If this commented out code is necessary here, can you please add a comment explaining it, so it's not removed later? If it's not necessary, can you please remove it?

spotbugsTestCases/src/java/ghIssues/Issue1148.java
```diff
@@ -0,0 +1,51 @@
...
+	private double KICK;
+	private double HEADBUTT;
+	private double MAGIC;
+	
+	public List<Integer> cpuData;
+	public List<Integer> diskData;
+	public List<Integer> memData;
+
+	// False positive for SF_SWITCH_NO_DEFAULT
+	public void setDamageParam(int i, double v){
```
JuditKnoll, line 18: Maybe there could be a disabled test or a commented out assert added for this.

spotbugsTestCases/src/java/ghIssues/Issue1148.java
```diff
@@ -0,0 +1,51 @@
...
+		switch(i){	//this line is marked
+			case 0:  TACKLE = v;
+			case 1:  PUNCH = v;
+			case 2:  KICK = v;
+			case 3:  HEADBUTT = v;
+			case 4:  MAGIC = v;
+			default:
+				break; // The compiler does not emit a GOTO for this so we can't tell if it was written, hence  the SF_SWITCH_NO_DEFAULT false positive
```
JuditKnoll: There is an unnecessary space between "hence" and "the". Can you please remove it?
```suggestion
				break; // The compiler does not emit a GOTO for this so we can't tell if it was written, hence the SF_SWITCH_NO_DEFAULT false positive
```

JuditKnoll (changes requested):

spotbugsTestCases/src/java/ghIssues/Issue1148.java
gtoison: I've added a disabled test in case we find time time to get back to it

spotbugsTestCases/src/java/ghIssues/Issue1148.java
gtoison: I've fixed the typo

spotbugsTestCases/src/java/ghIssues/Issue3572.java
gtoison: The test does not show the issue when the comment is uncommented, I have added a comment (from the issue report) explaining the cases showing the issue

JuditKnoll (approved):

iloveeclipse:
Thanks.

Issue #3564, opened 2025-07-31
Add Java 25 support


iloveeclipse:
As of today spotbugs [uses ASM 9.8](https://github.com/spotbugs/spotbugs/commit/37424c8842a872ce123d7b8bcfae1106d40f9d53) that [supports Java 25](https://asm.ow2.io/versions.html), but we also still use BCEL 6.10 that only knows Java 24.

BCEL 6.11 [would have Java 25 support](https://github.com/apache/commons-bcel/commit/70562832ce2cc8925787bacf53911615e6fbb2b7), but it is not yet released.

We should make sure there is a release available that can analyze Java 25 code.

I'm currently trying to build locally using 6.11 snapshot from https://repository.apache.org/content/repositories/snapshots/org/apache/bcel/bcel/6.11.0-SNAPSHOT/

iloveeclipse:
Follow up: https://github.com/spotbugs/spotbugs/issues/3569

iloveeclipse:
Reverted in https://github.com/spotbugs/spotbugs/pull/3585

iloveeclipse:
> Reverted in [#3585](https://github.com/spotbugs/spotbugs/pull/3585)

Because of concerns related to maven dependency management related to the snapshot build used, see https://github.com/spotbugs/spotbugs/discussions/3380.

Waiting on BCEL final 6.11...

mtrigueira:
Java 25 Generally Available today. https://openjdk.org/projects/jdk/25/

iloveeclipse:
Whoever is interested in Java 25 support, current blocker is this BCEL ticket https://issues.apache.org/jira/browse/BCEL-377
Feel free to help with that or even contribute a fix.

JuditKnoll:
BCEL [6.11.0-RC1](https://github.com/apache/commons-bcel/releases/tag/commons-bcel-6.11.0-RC1) got released! Thank you for your work there @iloveeclipse !

iloveeclipse:
Fixed via https://github.com/spotbugs/spotbugs/pull/3763

PR #3585, opened 2025-08-07
Revert "Use bcel snapshot to get bcel 6.11 that supports Java 25 (#3564)(#3568)"

gtoison:
As discussed in #3380 we cannot release with a dependency on a snapshot version of BCEL because users pulling SpotBugs from Maven would need to add the snapshot repository.

### DIFF
--- a/eclipsePlugin/build.gradle
+++ b/eclipsePlugin/build.gradle
@@ -90,7 +90,7 @@ dependencies {
 }
 
 tasks.named('clean', Delete).configure {
-    delete "lib", "build"
+    delete "lib", "build", "META-INF/MANIFEST.MF", "build.properties"
 }
 
 // This disables hundreds of javadoc warnings on missing tags etc, see #340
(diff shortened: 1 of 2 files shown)


eclipsePlugin/build.gradle
```diff
@@ -90,7 +90,7 @@ dependencies {
 }
 
 tasks.named('clean', Delete).configure {
-    delete "lib", "build", "META-INF/MANIFEST.MF", "build.properties"
+    delete "lib", "build"
```
iloveeclipse: Please keep this

spotbugs/build.gradle
```diff
@@ -99,7 +99,7 @@ dependencies {
 }
 
 clean {
-  delete ".libs", "META-INF/MANIFEST.MF"
+  delete ".libs"
```
iloveeclipse: Please keep this

JuditKnoll (approved):

spotbugs/build.gradle
gtoison: I have made the change

iloveeclipse (approved):

PR #3588, opened 2025-08-07
Fix some build issues affecting linux/mac and add signing note to release notes as removed from eclipse plugin

hazendaz:


hazendaz:
didn't solve the issue so I'll be doing more cleanup as copilot hasn't found exactly where the problem is but there are a lot of general problems throughout that will help fixing.

JuditKnoll:
Eclipse plugin signing was removed in https://github.com/spotbugs/spotbugs/pull/3582

iloveeclipse:
> Eclipse plugin signing was removed in #3582

I believe the signing  meant here is about spotbugs library used by sonatype/maven. We don't upload Eclipse plugins to maven/whatever else repo.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Inline access method for method. ([#3481](https://github.com/spotbugs/spotbugs/issues/3481))
 - Added `DMI_MISLEADING_SUBSTRING` for calling `subString(0)` on a StringBuffer/StringBuilder ([#1928](https://github.com/spotbugs/spotbugs/issues/1928))
 
+### Signing
+- Signing for Eclipse plugin has been removed at the current time due to signing keys being expired.  The expired key produced a warning during install, the same is true without signing.
```
JuditKnoll, line 53: @iloveeclipse As far as I can see, this is about the eclipse plugin signing, this is why I mentioned your PR for easy access.

iloveeclipse: OK, I've misunderstood after so many mails what is meant where :-) Sorry for noise.
I meant the build/release shouldn't be affected by the signing or not signing Eclipse bits, it shouldn't be relevant.

hazendaz: My intention in change logs was to simply make it more obvious to users we stopped signing eclipse in off chance someone asks since it's been asked a few times to fix the expired key.  I agree it's on par otherwise.

lincredibleJC:
Hi, the change from `compileOnly` to `implementation` for the junit-bom in this PR has re-introduced an issue that was previously fixed in #2255 / #2256.

**The problem:**
With `implementation platform("org.junit:junit-bom:...")`, the JUnit BOM is now exposed to all consumers of `spotbugs-annotations`. This silently upgrades downstream projects' JUnit versions when they update SpotBugs, which can cause unexpected test failures.

**Timeline:**
- **#2255** (Jan 2023): Reported that junit-bom was leaking to consumers
- **#2256** (Jan 2023): Fixed by changing `implementation` → `compileOnly`
- **#3588** (this PR): Changed `compileOnly` → `implementation`, re-introducing the leak

**Impact:**
- `spotbugs-annotations:4.9.3` POM: No junit-bom ✅
- `spotbugs-annotations:4.9.4` POM: Includes junit-bom ❌ (regression introduced here)

`spotbugs-annotations` is a compile-time-only annotation library—it shouldn't be influencing downstream projects' test framework versions.

**Suggested fix:**
Revert back to `compileOnly platform(...)` for the junit-bom.

See:
- Original issue: #2255
- Original fix: #2256
- 4.9.3 POM (no junit-bom): https://repo1.maven.org/maven2/com/github/spotbugs/spotbugs-annotations/4.9.3/spotbugs-annotations-4.9.3.pom
- 4.9.4 POM (has junit-bom): https://repo1.maven.org/maven2/com/github/spotbugs/spotbugs-annotations/4.9.4/spotbugs-annotations-4.9.4.pom

Tracking issue: #3908

PR #3589, opened 2025-08-07
build: Cleanup and combine dependencies block

hazendaz:
constraints probably isn't even necessary but it should have been in the same block.


hazendaz:
cleaned up gstring usage too.  Will have more coming after for gstring cleanup, but for reference purposes, double quotes in groovy should not be used unless actually needs to be a gstring.  Otherwise since quotes should be used so that is done here too.

PR #3590, opened 2025-08-07
build: Cleanup gstring usage

hazendaz:
First pass cleaning up unnecessary gstring usage


spotbugs/build.gradle
```diff
@@ -377,7 +377,7 @@ publishing.publications.maven {
   artifact distZip
   pom {
     name = 'SpotBugs'
-    description = "SpotBugs: Because it's easy!"
```
hazendaz, line 380: will switch to it is rather than it's to avoid escaping or gstring.

PR #3591, opened 2025-08-07
gradle build: Type correctly instead of relying on 'def'

hazendaz:
Partial updates for moving away from def which helps ensure better performance.


hazendaz:
will redo this as the plugin is too hard to get correct and original isn't really ideal.

PR #3592, opened 2025-08-07
build: Fix types to avoid def usage

hazendaz:


hazendaz:
removed attempt to cleanup the spotbugs gradle plugin, leaving just the other def cleanup for now.

PR #3593, opened 2025-08-07
build: Fix sonar resolution to avoid issues with java 21 on *nix

hazendaz:


hazendaz:
still no go on fixing issue.  Will continue poking around to find where problem is.

PR #3601, opened 2025-08-08
build: GHA Updates

hazendaz:
- adjust action names
- Add note to macos as to why we use -R instead of --location
- Do not run gpg on forks, only spotbugs
- No longer check if gpg key exists, its only on spotbugs, it exists
- Don't use line continuation for runs if only one line

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,8 @@ jobs:
         run: |
           curl --continue-at - --create-dirs --location 'https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-${{ env.eclipse-version }}-${{ env.eclipse-date }}/eclipse-SDK-${{ env.eclipse-version }}-macosx-cocoa-aarch64.dmg&mirror_id=1' --output eclipse-sdk/eclipse.dmg
           hdiutil attach eclipse-sdk/eclipse.dmg
-          cp --recursive /Volumes/Eclipse/Eclipse.app /Applications/
+          # Mac uses BSD cp which doesn't have --recursive therefore -R is required
+          cp -R /Volumes/Eclipse/Eclipse.app /Applications/
           hdiutil detach /Volumes/Eclipse
           echo "eclipseRoot.dir=/Applications/Eclipse.app" | tee eclipsePlugin/local.properties
       # Disable gpg signing on *nix for time being until we get new keys established, checks will be on windows only for now
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,15 +59,9 @@ jobs:
           cp -r /Volumes/Eclipse/Eclipse.app /Applications/
           hdiutil detach /Volumes/Eclipse
           echo "eclipseRoot.dir=/Applications/Eclipse.app" | tee eclipsePlugin/local.properties
-      # Disable gpg signing on *nix for time being until we get new keys established
-      #- name: Pre-build setup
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -56,43 +55,38 @@ jobs:
...
       #    ORG_GRADLE_PROJECT_keystorepass: ${{ secrets.KEYSTORE_PASS }}
       - name: Pre-build setup Windows
-        if: matrix.os == 'windows-latest'
+        if: github.repository_owner == 'spotbugs' && matrix.os == 'windows-latest'
         run: |
           git fetch --no-tags https://$GITHUB_TOKEN@github.com/spotbugs/spotbugs.git +refs/heads/master:refs/remotes/origin/master
-          if ( "$GPG_SECRET_PASSPHRASE" -ne "" ) {
-            gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
-          }
+          gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
```
gtoison: It shouldn't be necessary to decrypt the .jks since it contained the certficate used to sign the Eclipse plugin

hazendaz: yes I know, just cleaning this up.  build did fail, now I know why.  the gpg key @KengoTODA likely just zapped so its empty content now as this then ran like release which fails but has a subsequent call.  I'll comment out for now entirely though as we don't need this running at all currently as its not being used anywhere now.

hazendaz: ok its commented out now in both release / build.  The stubs otherwise are correct if/when we decide to try to fix that again.

hazendaz: I went ahead and also deleted the duplicate platform for gpg there as that was done initially only to lock to windows as not sure the issue with gpg.  Now that its clear the key was empty for pwd to force us to do something, it doesn't need a platform split so its cleaner now.  Just enough there if we try to address that later.  Eventually we should but for now should be good and its staged for later.

.github/workflows/build.yml
```diff
@@ -56,43 +55,27 @@ jobs:
         run: |
           curl --continue-at - --create-dirs --location 'https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/R-${{ env.eclipse-version }}-${{ env.eclipse-date }}/eclipse-SDK-${{ env.eclipse-version }}-macosx-cocoa-aarch64.dmg&mirror_id=1' --output eclipse-sdk/eclipse.dmg
           hdiutil attach eclipse-sdk/eclipse.dmg
+          # Mac uses BSD cp which doesn't have --recursive therefore -R is required
           cp -r /Volumes/Eclipse/Eclipse.app /Applications/
           hdiutil detach /Volumes/Eclipse
           echo "eclipseRoot.dir=/Applications/Eclipse.app" | tee eclipsePlugin/local.properties
-      # Disable gpg signing on *nix for time being until we get new keys established, checks will be on windows only for now
+      # Disable gpg signing for time being until we get new keys established
```
gtoison: To be clear, this part is not doing gpg **signing**, what it does is gpg **decryption** of the .jks keystore (which contains the certificate used to sign the Eclipse plugin.

hazendaz: will update comment

hazendaz: done.

PR #3624, opened 2025-08-10
Fix `SF_SWITCH_FALLTHROUGH` false positive when continuing a loop

### LINKED ISSUE #3617: [4.9.4] SF_SWITCH_FALLTHROUGH false positive
Upgrading from 4.9.3 and the rule misinterprets a `continue` as a fall-through scenario.

Image

The method [scheduleAfterWrite()](https://github.com/ben-manes/caffeine/blob/38dd616b945749edcc102642adb1b08274ae3a88/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L1591-L1620) has a `continue` on the `PROCESSING_TO_IDLE` case,

```java
/**
 * Conditionally schedules the asynchronous maintenance task after a write operation. If the
 * task status was IDLE or REQUIRED then the maintenance task is scheduled immediately. If it
 * is already processing then it is set to transition to REQUIRED upon completion so that a new
 * execution is triggered by the next operation.
 */
void scheduleAfterWrite() {
  @Var int drainStatus = drainStatusOpaque();
  for (;;) {
    switch (drainStatus) {
      case IDLE:
        casDrainStatus(IDLE, REQUIRED);
        scheduleDrainBuffers();
        return;
      case REQUIRED:
        scheduleDrainBuffers();
        return;
      case PROCESSING_TO_IDLE:
        if (casDrainStatus(PROCESSING_TO_IDLE, PROCESSING_TO_REQUIRED)) {
          return;
        }
        drainStatus = drainStatusAcquire();
        continue;
      case PROCESSING_TO_REQUIRED:
        return;
      default:
        throw new IllegalStateException("Invalid drain status: " + drainStatus);
    }
  }
}
```

gtoison:
When a `GOTO` inside a switch corresponds to a loop's `continue` the branch target is before the switch PC.
This should fix #3617


gtoison:
I'll add a changelog entry once #3623 gets merged

hazendaz (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -311,11 +312,24 @@ else if (seen == Const.PUTFIELD && stack.getStackDepth() >= 2) {
...
+     * <li> the branch target is after the next switch offset as it is the case for a switch break</li>
+     * <li> or the branch target is before the PC corresponding to the switch instruction</li>
+     * <li> or there's no next switch case</li>
+     * </ul>
      */
-    public boolean isBranchTargetAfterNextCaseOffset() {
-        return getBranchTarget() > switchHdlr.getNextSwitchOffset(this);
+    public boolean isBranchTargetOutsideOfNextCase() {
+        int branchTarget = getBranchTarget();
+        SwitchDetails nextwitchDetails = switchHdlr.getNextSwitchDetails(this);
```
JuditKnoll: There is a small typo in the name of this variable, the S is missing from the beginning of Switch. Can you please add it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -311,11 +312,24 @@ else if (seen == Const.PUTFIELD && stack.getStackDepth() >= 2) {
...
      * A GOTO might correspond to a <code>break</code> or to a do/while/for loop.
      * For loops the branch target will be before the offset of the next case, for breaks we're exiting the switch so the target is actually even after the end of the last case.
+     * The branch target might be before the switch when we're inside another structure such as a loop.
      *
-     * @return <code>true</code> if the branch target is after the next switch offset as it is the case for a switch break.
+     * @return <code>true</code> if:
+     * <ul>
+     * <li> the branch target is after the next switch offset as it is the case for a switch break</li>
+     * <li> or the branch target is before the PC corresponding to the switch instruction</li>
+     * <li> or there's no next switch case</li>
```
JuditKnoll: The edge case of the default case could also be mentioned.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
gtoison: Thanks for reviewing, I have corrected the typo

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
gtoison: I have added a comment about the default case (which falls under the "there's no next case" category)

JuditKnoll (approved):

PR #3632, opened 2025-08-11
`CWO_CLOSED_WITHOUT_OPENED` false positive

gtoison:
During the CFG analysis there's an exception edge where the resource was closed, so it shouldn't transition to "not opened"

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueAnalysis.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueAnalysis.java
@@ -106,7 +106,7 @@ public class ResourceValueAnalysis<Resource> extends FrameDataflowAnalysis<Resou
                 // If status is OPEN, downgrade to OPEN_ON_EXCEPTION_PATH
                 tmpFact = modifyFrame(fact, null);
                 tmpFact.setStatus(ResourceValueFrame.State.OPEN_ON_EXCEPTION_PATH);
-            } else {
+            } else if (fact.getStatus() != ResourceValueFrame.State.CLOSED) {
                 tmpFact = modifyFrame(fact, null);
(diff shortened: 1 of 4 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/ResourceValueAnalysis.java
```diff
@@ -106,7 +106,7 @@ public void meetInto(ResourceValueFrame fact, Edge edge, ResourceValueFrame resu
                 // If status is OPEN, downgrade to OPEN_ON_EXCEPTION_PATH
                 tmpFact = modifyFrame(fact, null);
                 tmpFact.setStatus(ResourceValueFrame.State.OPEN_ON_EXCEPTION_PATH);
-            } else {
+            } else if (fact.getStatus() != ResourceValueFrame.State.CLOSED) {
```
JuditKnoll, line 109: Is only `CLOSED` an exceptional value (and `OPEN` handled in line 105)?

gtoison: I'm not sure honestly, I don't fully understand the algorithm.
I stepped through the execution in debug and it goes through this code a dozen of times for this simple method.
Only the `CLOSED` case seemed problematic to me, maybe in more complex cases there will be further adjustments required.

JuditKnoll: Thank you 😄 
`OPEN_ON_EXCEPTION_PATH` and `CLOSED_WITHOUT_OPENED` seem to be quite similar to `OPEN` and `CLOSED` to me, but it would need further investigation sometime.
This change fixes the fp and doesn't break any of the tests, so it's good to go for me.

PR #3635, opened 2025-08-12
False Negative `NN_NAKED_NOTIFY` when loading a field

### LINKED ISSUE #3634: False Negative in SpotBugs Rule NN_NAKED_NOTIFY
Hi, I found a false negative bug about the rule NN_NAKED_NOTIFY. The minimized case is provided below. SpotBugs should have reported a warning at line 6 as the variable is visible to different threads and there are no modifications to mutable objects. This appears to be a false negative in the analysis.

```java
class Test {
    volatile String field = Test.class.getCanonicalName();
    static Test record = new Test();
    public static void foo() {
        synchronized (record.field) {
            record.field.notifyAll(); // should report a warning here, but no warnings
        }
    }
}
```

SpotBugs version: 4.9.3

gtoison:
`FindNakedNotify` does not handle the case when the lock object is loaded from a field, only the case when we're loading from a register.
This should fix #3634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
 - Documentation of -adjustPriority parameter
 - Functionality from DetectorFactory setEnabledButNonReporting(), getPriorityAdjustment() methods and BugInstance.adjustForDetector() is deprecated and moved to PriorityAdjuster ([#3753](https://github.com/spotbugs/spotbugs/issues/3753))
+- Improved `FindNakedNotify` to handle the case when the lock is loaded from a field ([#3634](https://github.com/spotbugs/spotbugs/issues/3634))
 
 ### Changed
 - Support for fully qualified class names for detectors in -adjustPriority parameter


gtoison:
I'm not sure if we want to merge that one in 4.9.5 or wait for another release: if we can expedite the release of 4.9.5 with only bug/regression fixes that would be a good/stable version.
I'm slightly wary that this PR could cause regressions or new false positives.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNakedNotify.java
```diff
@@ -67,41 +67,54 @@ public void visit(Code obj) {
...
-            stage = 2;
+        case MONITOR_ENTERED:
+            if (isRegisterLoad() || seen == Const.GETSTATIC || seen == Const.GETFIELD) {
+                stage = Stage.LOADED;
+            }
             break;
-        case 2:
-            if (seen == Const.INVOKEVIRTUAL
+        case LOADED:
+            if (isRegisterLoad() || seen == Const.GETSTATIC || seen == Const.GETFIELD) {
```
gtoison, line 81: The import change is here: we allow `stage` to remain at `LOADED` if we see another "load" instruction

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNakedNotify.java
```diff
@@ -36,7 +36,7 @@
 //   9:   monitorexit
 
 public class FindNakedNotify extends BytecodeScanningDetector implements StatelessDetector {
-    int stage = 0;
+    private Stage stage = Stage.START;
```
JuditKnoll, line 39: Thank you for improving the code with this enum and making it easier to read.

JuditKnoll (changes requested):
Do not forget to add the changelog entry before merging, otherwise LGTM.

PR #3637, opened 2025-08-13
Fix NP_GUARANTEED_DEREF missing when identical conditions are duplicated (#3455)

### LINKED ISSUE #3455: conflicting NP_GUARANTEED_DEREF detection in two semantically same code
The given code _correctly_ raises `NP_GUARANTEED_DEREF` when scanning the following code.

```java
class NullDereferenceExample {

    public Integer showBug(String input) {
        Integer result = null;

        // Simulating a condition where the result is null
        if (input != null && input.length() > 5) {
            result = 10;
        }

        if (result == null) {
            throw new NullPointerException("Result is null");
        }

        return result;
    }

    public static void main(String[] args) {
        NullDereferenceExample example = new NullDereferenceExample();
        example.showBug("testing");
    }
}
```
However, if I add the if-condition (`if (input != null && input.length() > 5)`) again as below (same code snippet repeated), suddenly the issue is not detected any longer.
```java
class NullDereferenceExample {

    public Integer showBug(String input) {
        Integer result = null;

        // Simulating a condition where the result is null
        if (input != null && input.length() > 5) {
            result = 10;
        }
        
        if (input != null && input.length() > 5) {
            result = 10;
        }

        if (result == null) {
            throw new NullPointerException("Result is null"); // BUG: NP: Null value is guaranteed to be dereferenced (NP_GUARANTEED_DEREF)
        }

        return result;
    }

    public static void main(String[] args) {
        NullDereferenceExample example = new NullDereferenceExample();
        example.showBug("testing");
    }
}
```
SpotBugs version: `4.9.2`
Java: `Temurin openjdk 21.0.7`
Ran from CLI as the following two commands:
```bash
export SPOTBUGS_JAR=/path/to/spotbugs.jar
```
```bash
javac NullDereferenceExample.java && \
java -jar $SPOTBUGS_JAR \
    -low \
    -effort:max \
    NullDereferenceExample.class
```
Expectation: Adding the same if-condition again makes no change in the code execution. So the `NP_GUARANTEED_DEREF` should be detected in the second code snippet as well.

ItsZetty:
Fixes #3455

### Summary
Fix a false-negative where NP_GUARANTEED_DEREF disappears when an identical null-check block is duplicated. Semantically equivalent code should produce consistent diagnostics.

### Changes
- In `spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNullDeref.java`:
  - Do not early-return solely because `doomedLocations` is empty.
  - Continue reporting if `derefLocationSet` is non-empty, using `assignedNullLocationSet` as source when needed.

### Rationale
Duplicate, identical null-checks should not suppress NP_GUARANTEED_DEREF. The previous early return on empty `doomedLocations` caused the false-negative even though dereference locations existed.

### Testing
- Built locally:
  - `./gradlew clean build -x javadoc`
- Manually verified with the example from the issue that both variants now consistently report NP_GUARANTEED_DEREF.

### Compatibility
- No public API/signature changes. Low risk; narrows a false-negative.

### Checklist
- [x] Added an entry to `CHANGELOG.md` under Unreleased → Fixed.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3455Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3455Test.java
@@ -1,0 +1,18 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3455Test extends AbstractIntegrationTest {
+    @Test
+    void testNoDup() {
+        performAnalysis("ghIssues/Issue3455.class");
+        assertBugTypeCount("NP_GUARANTEED_DEREF", 1);
+    }
+
+    @Test
+    void testWithDup() {
+        performAnalysis("ghIssues/Issue3455.class");
+        assertBugTypeCount("NP_GUARANTEED_DEREF", 1);
+    }
+}
--- a/spotbugsTestCases/src/java/ghIssues/Issue3455.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue3455.java
@@ -1,0 +1,13 @@
+package ghIssues;
+
+public class Issue3455 {
+    public int withDup(boolean b1, String input) {
+        Object x = (input == null ? new Object() : null);
+
+        if (input != null && input.length() > 5) { x = null; }
+        if (input != null && input.length() > 5) { x = null; } 
+
+        if (b1) return x.hashCode();   
+        return -x.hashCode();            
+    }
+}


gtoison (changes requested):
Hello, thank you very much for looking into this!
Could you please add a test showing this issue? 
The test should fail without the fix and it should pass with the fix
Here's an example: https://github.com/spotbugs/spotbugs/pull/3632/commits/e7789f87820288e7f91a5fff068502dcbe7cb048

JuditKnoll (changes requested):
Can you please add a test reproducing the issue? 

You can find the code examples in [spotbugsTestCases](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java) and the corresponding test in [spotbugs-tests](https://github.com/spotbugs/spotbugs/tree/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs).

ItsZetty:
Thanks for the review! I’ll add the test and update the PR!

gtoison:
You can run `gradlew spotlessApply` before committing to adjust the formatting

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3455Test.java
```diff
@@ -0,0 +1,18 @@
...
+class Issue3455Test extends AbstractIntegrationTest {
+    @Test
+    void testNoDup() {
+        performAnalysis("ghIssues/Issue3455.class");
+        assertBugTypeCount("NP_GUARANTEED_DEREF", 1);
+    }
+
+    @Test
+    void testWithDup() {
+        performAnalysis("ghIssues/Issue3455.class");
```
gtoison, line 15: Am I missing something or these two tests are checking the same thing?
`performAnalysis(...)` is relatively slow so better not spend too much time running it twice if you can avoid that.

In case you want to check two bits of codes you can use the methods asserting that the bug was detected (or absent) in a method: `assertBugInMethodCount(...)`

ItsZetty:
Hello!! :)

I've been looking into this issue, and my analysis points to the meetInto method in UnconditionalValueDerefAnalysis.java.
The analysis seems to lose dereference information when merging paths from duplicated code blocks. This appears to be because the merge logic calculates an intersection (mergeWith) of dereference locations, which results in an empty set when the locations differ across paths.

A naive attempt to change this to a union (unionWith) caused the analysis engine to fail, so it seems a more careful approach is needed.

Could you offer some guidance on the correct way to resolve this? Please let me know if I've misunderstood the core problem here.

Thank you!

PR #3656, opened 2025-08-28
Pre build verification

Pankraz76:

----


.github/workflows/build.yml
```diff
@@ -65,17 +65,19 @@ jobs:
...
         if: matrix.java == '21'
-        run: ./gradlew spotlessCheck build smoketest ${SONAR_LOGIN:+sonar} --no-daemon "-Dsonar.token=$SONAR_LOGIN" --scan
+        run: ./gradlew build --no-daemon --scan ${SONAR_LOGIN:+sonar} "-Dsonar.token=$SONAR_LOGIN"
         env:
           SONAR_LOGIN: ${{ secrets.SONAR_LOGIN }}
           ORG_GRADLE_PROJECT_keystorepass: ${{ secrets.KEYSTORE_PASS }}
       - name: Build JDK ${{ matrix.java }}, ${{ matrix.os }}
         # https://community.sonarsource.com/t/sonarcloud-now-not-updating-github-pr-and-checks/6595/17
         if: matrix.java != '21'
-        run: ./gradlew spotlessCheck build smoketest --no-daemon --scan
```
Pankraz76, line 78: could this redundancy be centralized?

.github/workflows/build.yml
```diff
@@ -65,17 +65,19 @@ jobs:
       #  run: gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
       #  env:
       #    GPG_SECRET_PASSPHRASE: ${{ secrets.GPG_SECRET_PASSPHRASE }}
+      - name: Pre-Build verification
+        run: ./gradlew spotlessCheck smoketest --no-daemon --scan
```
Pankraz76, line 69: this would also give clear indication if it fails.

So the build step can not fail anymore due to corruption, as done in dedicated step.

hazendaz:
I believe this would make the build execute twice which would not be an improvement.

Pankraz76:
yes need to investigate this. Some goal like rewrite would need some compile phase as well.

.github/workflows/build.yml
```diff
@@ -65,17 +65,19 @@ jobs:
...
       - name: Build with Sonar JDK ${{ matrix.java }}, ${{ matrix.os }}
         # https://community.sonarsource.com/t/sonarcloud-now-not-updating-github-pr-and-checks/6595/17
         if: matrix.java == '21'
-        run: ./gradlew spotlessCheck build smoketest ${SONAR_LOGIN:+sonar} --no-daemon "-Dsonar.token=$SONAR_LOGIN" --scan
+        run: ./gradlew build --no-daemon --scan ${SONAR_LOGIN:+sonar} "-Dsonar.token=$SONAR_LOGIN"
```
Pankraz76, line 73: if this is not an issue if its there like the second if does **prevent**, we could have only one job declaration, right?

Pankraz76, line 73: runtime:

```
 or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

> Task :spotbugs:smokeTest
[ant:spotbugs] Java Result: 1

[Incubating] Problems report is available at: file:///Users/vincent.potucek/IdeaProjects/spotbugs/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.0.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 40s
46 actionable tasks: 29 executed, 17 up-to-date

Publishing Build Scan...
https://gradle.com/s/z6o7hthzu7z5e

➜  spotbugs git:(fix-ModifierOrder) ./gradlew build --no-daemon --scan 
```

Pankraz76:
> I believe this would make the build execute twice which would not be an improvement.

no its not:


`./gradlew build --no-daemon --scan`

`./gradlew spotlessCheck smoketest --no-daemon --scan`

Runtime is way to different to be equal.


image

hazendaz:
> yes need to investigate this. Some goal like rewrite would need some compile phase as well.

do not want open rewrite, not at all.

Pankraz76:
> > yes need to investigate this. Some goal like rewrite would need some compile phase as well.
> 
> do not want open rewrite, not at all.

yes was just an example. As build is testing this seems to be the deal breaker. As compiling as example mentioned is not expensive.

It seems to be splittable like purposed.

gtoison:
I'm not sure what we're trying to improve here and I think we should let gradle do its optimizations by only running it once. I surely don't want the build to become slower.

Pankraz76:
> I surely don't want the build to become slower.

Agree. It will only be a little slower as we calling gradle twice now but the goals are just the same. Please consider the duplicated code. 

2 for 1 is always a great deal to make, assuming no code is the best code we can create, not duplicated one and the same for actually no reason:

Trading 2 times the overhead for the same functionality.

image


Also this is something individual all other projects out there have dedicated builds making the build only fail for some tests, not for checkstyle spotless foo. It helps to determine which phase has failed as currently build and validation is once. Its kind of one yes as gradle has designed it that way. Im just want to make an offer. 

Please think about adding pmd we not have to make one change twice for actually no reason, right?

JuditKnoll:
I am not so positive about this change, I don't find it particularly useful or necessary.
@Pankraz76 You are adding comments and questions, but resolve them without any reactions. Are you expecting to get an answer to them? If yes, why are you resolving them? If not, why do you even add them?

Pankraz76:
> I am not so positive about this change, I don't find it particularly useful or necessary. @Pankraz76 You are adding comments and questions, but resolve them without any reactions. Are you expecting to get an answer to them? If yes, why are you resolving them? If not, why do you even add them?

doing so is less noisy as in detail some reviewer can see them and still answer of course. As its just some note i like to resolve them not to bother everybody with some dangling detailed comments. If there is some conversation my initiative is the foundation if there is non im also fine with it.

Pankraz76:
was just an suggestion to avoid dry. There might be another way. Closing, thanks for review.

PR #2753, opened 2023-12-07
Sonar Analyses fixes

JuditKnoll:
I looked at the Sonar Cloud Analyses results of the [issues introduced this year](https://sonarcloud.io/project/issues?createdAfter=2023-01-01&resolved=false&id=com.github.spotbugs.spotbugs). This PR solves some of them.
I haven't added any changelog entry, since I don't think this type of fix needs it.

----

Make sure these boxes are checked before submitting your PR -- thank you!

- [ ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code

PR #3657, opened 2025-08-28
S1124: Modifiers should be declared in the correct order

Pankraz76:

----

Finishing: 

- https://github.com/spotbugs/spotbugs/pull/2753
- https://github.com/spotbugs/spotbugs/pull/2753/files#r2307322092
- https://sonarsource.github.io/rspec/#/rspec/S1124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Merge conditions of nested ifs ([#3231](https://github.com/spotbugs/spotbugs/pull/3231))
 - Use non deprecated 'getDottedClassName' instead of 'toDottedClassName'([#3251](https://github.com/spotbugs/spotbugs/pull/3251))
 - Use try with resources where possible ([#3253](https://github.com/spotbugs/spotbugs/pull/3253))
+- S1124: Modifiers should be declared in the correct order ([#3657](https://github.com/spotbugs/spotbugs/pull/3657))
 
 ### Changed
 - Bump up Java version to 11
--- a/eclipsePlugin/src/de/tobject/findbugs/properties/PathsProvider.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/properties/PathsProvider.java
@@ -156,7 +156,7 @@ abstract class PathsProvider extends SelectionAdapter implements IStructuredCont
         return dialog;
     }
 
-    abstract protected void configureDialog(FileDialog dialog);
+    protected abstract void configureDialog(FileDialog dialog);
 
     protected String openFileDialog(FileDialog dialog) {
         return dialog.open();
@@ -231,7 +231,7 @@ abstract class PathsProvider extends SelectionAdapter implements IStructuredCont
         viewer.refresh(true);
     }
 
-    abstract protected IStatus validate();
+    protected abstract IStatus validate();
 
     public void remove(IPathElement holder) {
         paths.remove(holder);
--- a/eclipsePlugin/src/edu/umd/cs/findbugs/plugin/eclipse/quickfix/BugResolution.java
+++ b/eclipsePlugin/src/edu/umd/cs/findbugs/plugin/eclipse/quickfix/BugResolution.java
@@ -64,7 +64,7 @@ import edu.umd.cs.findbugs.plugin.eclipse.quickfix.exception.BugResolutionExcept
  */
 public abstract class BugResolution extends WorkbenchMarkerResolution {
 
-    static private final String MISSING_BUG_INSTANCE = "This bug is no longer in the system. "
+    private static final String MISSING_BUG_INSTANCE = "This bug is no longer in the system. "
             + "The bugs somehow got out of sync with the memory representation. "
             + "Try running SpotBugs again. If that does not work, check the error log.";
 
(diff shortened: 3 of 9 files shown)


Pankraz76:
We can use PMD as well considering the plugin just an random impl. detail again as we used to ignore everything in the sake of abstraction, right?

In first place its about [S1124](https://sonarsource.github.io/rspec/#/rspec/S1124)

hazendaz:
@Pankraz76 Please create a new PR where you have fixes only, I don't want to see open rewrite added, I don't think it belongs here.  Even open rewrite would tell you that.  The changes are valid, in fact I had a branch that did most of these about a year ago but held off applying as I did a full cleanup using Eclipse which does a better job than open rewrite.  At any rate, open rewrite is generally a one time shot, so sure run that locally but just send the code changes instead of trying to change our github actions which is unrelated here or introducing more build steps at this point in time.  We will gladly accept the work, its worthy just not adding open rewrite.  It has far too many bugs to be part of any build chain and their release process is fanatic at best (broken / works / broken / works / span updates / etc).

.github/workflows/build.yml
```diff
@@ -65,17 +65,19 @@ jobs:
       #  run: gpg --quiet --batch --yes --decrypt --passphrase="$GPG_SECRET_PASSPHRASE" --output spotbugs.jks .github/workflows/spotbugs.jks.gpg;
       #  env:
       #    GPG_SECRET_PASSPHRASE: ${{ secrets.GPG_SECRET_PASSPHRASE }}
+      - name: Pre-Build verification
```
hazendaz: no on all this.

eclipsePlugin/src/de/tobject/findbugs/properties/PathsProvider.java
```diff
@@ -156,7 +156,7 @@ private FileDialog createFileDialog(Shell parentShell) {
         return dialog;
     }
 
-    abstract protected void configureDialog(FileDialog dialog);
```
hazendaz, line 159: Yes on the coding changes

hazendaz:
> We can use PMD as well considering the plugin just an random impl. detail again as we used to ignore everything in the sake of abstraction, right?
> 
> In first place its about [S1124](https://sonarsource.github.io/rspec/#/rspec/S1124)

Using PMD I would be for.  Using checkstyle I would be for.  Using open rewrite, I'm not for it.  The other two work well, open rewrite does not.  Keep in mind I keep stressing this because I'll end up being the one dealing with the fires it causes and after years of being burnt by open rewrite, I have zero appetite for using it like this.  All they did was re-invent the wheel that every IDE already did well and other tools did well, then they added a lot of non-sensible fixes that break far more often that work and over the years, its rare for open rewrite to be fully successful.  Sure the items you are using are mostly ok as they fail less often but I assure you some of the hidden gems it does are boarder line causing production issues.  I've seen first hand it change code that was entirely invalid but nuanced enough to pass through some checks only to get caught up in UAT testing as a major defect.  So lets stick to products that have a far better track record of working.  If you want to use open rewrite for your local work, save that stuff off to the side so you can run it successfully to clean things up without introducing it as a perm fixture to spotbugs.  I do use open rewrite daily, I just have zero appetite for full fledged usage given its poor track record.

.github/workflows/build.yml
Pankraz76: then use PMD? if its not done with some plugin we still not finished, just as before. This can only be done my CPU not by human.

Pankraz76:
> I've seen first hand it change code that was entirely invalid but nuanced enough to pass through some checks only to get caught up in UAT testing as a major defect

yes, all the changes need to be double-checked.

I see your argument, still some recipes work very well like demonstrated. And some stuff is 1o1 covered by check PMD and rewrite, each and own having its blindspots.

Recipes like add missing override annotation can do no harm so idk whats the big deal. PMD will do the same but imposing the burden to dev, rewrite is automating it reducing it to the review only.

.github/workflows/build.yml
Pankraz76: agreed upon PMD applying the same best practice ignoring the obsolete (impl) detail as going to change anytime soon. If PMD or rewrite fails the build does not matter. We only talking about dryRun which makes no harm at all. 

No one forces to apply the run goal which of course should be the natural outcome, but still is in the responsibility of each dev to check each commit.

Pankraz76:
speaking of bugs i can not make the PMD plugin work and finding the config. Lets discuss this in dedicated PR, providing mergeble fix upfront.

CHANGELOG.md
```diff
@@ -152,6 +152,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Merge conditions of nested ifs ([#3231](https://github.com/spotbugs/spotbugs/pull/3231))
 - Use non deprecated 'getDottedClassName' instead of 'toDottedClassName'([#3251](https://github.com/spotbugs/spotbugs/pull/3251))
 - Use try with resources where possible ([#3253](https://github.com/spotbugs/spotbugs/pull/3253))
+- S1124: Modifiers should be declared in the correct order ([#3657](https://github.com/spotbugs/spotbugs/pull/3657))
```
JuditKnoll, line 155: This is under 4.9.0. Please move it under Unreleased.

hazendaz:
I will pick up these changes separately without a change log.  This user has been blocked and notified as to why.

PR #3666, opened 2025-08-31
test: setup/teardown MutableClassesTest to remove unstableTests

gtoison:
When running a unit test as a gradle test the `unstableTests` is also executed with its two tests, which takes a few seconds; this makes iterating slower.
This also requires to merge the jacoco execs reports.

The `unstableTests` task was introduced because `MutableClassesTest` failed when running as part of the gradle build, this should fix the root problem.

### DIFF
--- a/spotbugs-tests/build.gradle
+++ b/spotbugs-tests/build.gradle
@@ -34,36 +34,13 @@ tasks.named('compileJava', JavaCompile).configure {
 }
 
 tasks.named('jacocoTestReport', JacocoReport).configure {
-  dependsOn('unstableTest')
   additionalSourceDirs.setFrom files(project(':spotbugs').sourceSets.main.java.srcDirs)
   additionalClassDirs.setFrom files(project(':spotbugs').sourceSets.main.output.classesDirs)
-  executionData.setFrom files(layout.buildDirectory.file('jacoco/unstableTest.exec'), layout.buildDirectory.file('jacoco/test.exec'))
-}
-
-// Tests below fail if executed with other tests
-// So we run them before all other tests are executed
-tasks.register('unstableTest', Test.class) {
-  testClassesDirs = testing.suites.test.sources.output.classesDirs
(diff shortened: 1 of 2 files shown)


hazendaz (approved):

hazendaz:
Tests fail

spotbugs-tests/build.gradle
```diff
@@ -34,37 +34,14 @@ tasks.named('compileJava', JavaCompile).configure {
...
-    includeTestsMatching 'PlaceholderTest'
-    includeTestsMatching 'MutableClassesTest'
-  }
-  testLogging {
-    events 'passed', 'skipped', 'failed'
-  }
 }
 
 test {
-  dependsOn ':spotbugsTestCases:build', 'unstableTest'
```
JuditKnoll, line 59: Are you sure that `dependsOn ':spotbugsTestCases:build'` can be deleted?

gtoison: Thanks, that's fixed now

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/util/MutableClassesTest.java
```diff
@@ -102,6 +113,9 @@ public static void setImmutable(Immutable imm) {
 
     @Test
     void testImmutable() {
+        ICodeBase codeBase = new SingleFileCodeBase(null, "build/classes/java/main/edu/umd/cs/findbugs/util/MutableClassesTest$Immutable.class");
+        //classPath.addCodeBase(codeBase);
```
JuditKnoll: Is it necessary here? `codebase` looks like an unused variable.

gtoison: Thanks, that was from a previous attempt at fixing the issue and I forgot to remove it

JuditKnoll (approved):

PR #3677, opened 2025-09-10
Fix missing classes needed for analyses for Tests

JuditKnoll:
Several tests were emitting the `The following classes needed for analyses were missing:` message to the standard error, even though it doesn't appear in the log when running the tests from gradle, and this does not make the tests fail, but this masks some problems.

The following tests are still missing some classes:
- [Issue1771Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1771Test.java) misses java.util.Collections$EmptyNavigableMap, java.util.Collections$EmptyNavigableSet;
   - It may be because this tests needs Java 11.
- [Issue2759Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2759Test.java), [Issue2864Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue2864Test.java), [Issue3059Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3059Test.java), [Issue3094Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3094Test.java), [Issue3350Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3350Test.java) and [LombokWithTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/LombokWithTest.java) miss several classes, but these are read from classSample class files with several outside dependencies. IMO we shouldn't add all of those dependencies, and it's okay like this.
- [ClassPathBuilderTest.nestedTraversalEnabled](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/classfile/impl/ClassPathBuilderTest.java#L29) misses MethodsCallOrder$SimpleTest, foo.Foo, com.bla.Parent, edu.umd.cs.findbugs.annotations.Confidence, edu.umd.cs.findbugs.annotations.CheckReturnValue, edu.umd.cs.findbugs.annotations.ExpectWarning, com.foo.Foo, org.jacoco.agent.rt.internal_43a39aa.Offline, MethodsCallOrder,  com.bla.Parent$Child, com.bla.Parent$Child$1, MethodsCallOrder$ShortcutPathTest,  edu.umd.cs.findbugs.annotations.DesireWarning.
  - A jar is analyzed, most likely it does not contain these classes, so IMO it's acceptable.
- [FindReturnRefTest.testFindReturnRefTestNegativeChecks](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java#L11), because the `PreferenceTreeNode` inner class in `FindReturnRefNegativeTest` was not added to the analysis in the test in https://github.com/spotbugs/spotbugs/pull/3284. With it, the test fails because there is an unexpected hit, I'll do it in a separate PR, the detector may need to be modified. (The original issue is about not throwing an NPE, not about an bug hit.)
- [WarningSuppressorTest.testIssue](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/WarningSuppressorTest.java#L14), [Issue1254Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue1254Test.java) miss some classes as well, but only when running with all tests, not when running them individually.
- [DetectorsTest.testAllRegressionFilesJavac](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/DetectorsTest.java#L100) misses org.apiguardian.api.API$Status, org.apiguardian.api.API, javax.annotation.PreDestroy, javax.annotation.PostConstruct, and also when running with all tests additionally misses some.UnknownClass, org.example.Bar.
   - I think it's because of other classes
- [Issue574Test.testRegisterExtensionAnnotation](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue574Test.java#L39) misses org.apiguardian.api.API, org.apiguardian.api.API$Status.
   - This annotation is on the parent of the analyzed class, which is coming from Junit Jupiter. I'm not sure about this one.

I haven't added a changelog entry, since I only modified test or sample code, and I don't think it's necessary for this type of change.

### DIFF
--- a/spotbugsTestCases/src/java/multithreaded/primitivewrite/SynchronizedBlockDoubleFromOtherMethod.java
+++ b/spotbugsTestCases/src/java/multithreaded/primitivewrite/SynchronizedBlockDoubleFromOtherMethod.java
@@ -18,7 +18,7 @@ public class SynchronizedBlockDoubleFromOtherMethod {
     }
 
     public void anotherSyncPrint() {
-        synchronized (SynchronizedBlockDouble.class) {
+        synchronized (SynchronizedBlockDoubleFromOtherMethod.class) {
             printValue();
         }
     }
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
@@ -52,7 +52,8 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
         setPriorityAdjustment(-1); // raise priority
         SystemProperties.setProperty("findbugs.refcomp.reportAll", "false");
 
-        performAnalysis("RefComparisons.class");
+        performAnalysis("RefComparisons.class", "RefComparisons$MyEnum.class", "RefComparisons$MyClass2.class",
+                "RefComparisons$MyClass1.class", "RefComparisons$MyClass3.class");
         assertBugTypeCount("RC_REF_COMPARISON", 1);
         assertBugInMethodAtLineWithConfidence("RC_REF_COMPARISON", "RefComparisons", "integerBadComparison", 5, Confidence.HIGH);
     }
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1148Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue1148Test.java
@@ -3,11 +3,24 @@ package edu.umd.cs.findbugs.detect;
 import edu.umd.cs.findbugs.AbstractIntegrationTest;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
 
 class Issue1148Test extends AbstractIntegrationTest {
 
     @Test
-    void testIssue() {
+    @EnabledForJreRange(max = JRE.JAVA_20)
+    void testIssueUnderJava21() {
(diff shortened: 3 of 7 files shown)


spotbugsTestCases/src/java/multithreaded/primitivewrite/SynchronizedBlockDoubleFromOtherMethod.java
```diff
@@ -18,7 +18,7 @@ synchronized void synchronizedLog() {
     }
 
     public void anotherSyncPrint() {
-        synchronized (SynchronizedBlockDouble.class) {
+        synchronized (SynchronizedBlockDoubleFromOtherMethod.class) {
```
JuditKnoll, line 21: This sample code is analyzed in [SharedVariableAtomicityDetectorTest.noBugSynchronizedBlockDoubleFromOtherMethod](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/SharedVariableAtomicityDetectorTest.java#L174). This is an copy-paste error in the sample code.

PR #3683, opened 2025-09-11
chore: Move to NIO for file handling

hazendaz:
I have not added a change log for this change.

These are considered obsolete

- FileNotFoundException
- FileInputStream
- FileOutputStream

Replaced all with modern NIO type.  For file not found exception, its just a io exception without need for separate named exception.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/SourceCharset.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/SourceCharset.java
@@ -22,7 +22,6 @@ package edu.umd.cs.findbugs.charsets;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -30,6 +29,8 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Writer;
 import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 /**
  * @author pugh
@@ -43,7 +44,7 @@ public class SourceCharset {
     }
 
     public static Writer fileWriter(File fileName) throws IOException {
-        return new OutputStreamWriter(new FileOutputStream(fileName), charset);
+        return new OutputStreamWriter(Files.newOutputStream(fileName.toPath()), Charset.defaultCharset());
     }
 
     public static PrintWriter printWriter(File fileName) throws IOException {
@@ -51,7 +52,7 @@ public class SourceCharset {
     }
 
     public static Writer fileWriter(String fileName) throws IOException {
-        return new OutputStreamWriter(new FileOutputStream(fileName), charset);
+        return new OutputStreamWriter(Files.newOutputStream(Path.of(fileName)), Charset.defaultCharset());
     }
 
     public static PrintWriter printWriter(String fileName) throws IOException {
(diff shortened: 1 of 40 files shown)


hazendaz:
I pulled this from some other changes with charsets I'm making.  In case I missed something here and build fails, I'll correct tomorrow.

spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/SourceCharset.java
```diff
@@ -43,15 +44,15 @@ public class SourceCharset {
     }
 
     public static Writer fileWriter(File fileName) throws IOException {
-        return new OutputStreamWriter(new FileOutputStream(fileName), charset);
+        return new OutputStreamWriter(Files.newOutputStream(fileName.toPath()), Charset.defaultCharset());
```
JuditKnoll, line 47: Was this modification (and the one in line 54/55) of charset intentional? If yes, why not modify in line 62/63 as well and delete the field?

spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/UserTextFile.java
```diff
@@ -45,7 +46,7 @@ public class UserTextFile {
     }
 
     public static Writer fileWriter(String fileName) throws IOException {
-        return new OutputStreamWriter(new FileOutputStream(fileName), charset);
+        return new OutputStreamWriter(Files.newOutputStream(Path.of(fileName)), Charset.defaultCharset());
```
JuditKnoll, line 49: Was this modification of charset intentional? If yes, why not modify in line 57/58 as well and delete the field?

spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/SourceCharset.java
hazendaz: as noted last night, I'm working changes with charset cleanup, its all part of same changes but trying to split them out.  That got picked up as files new output stream on same line.  The removal is coming, will send it on next PR.

spotbugs/src/main/java/edu/umd/cs/findbugs/charsets/UserTextFile.java
hazendaz: same.

spotbugs/src/main/java/edu/umd/cs/findbugs/io/IO.java
```diff
@@ -191,25 +191,6 @@ public static void close(@CheckForNull Closeable c) {
...
-            return;
-        }
-
-        try {
-            inputStream.close();
-        } catch (IOException e) {
-            // Ignore
-        }
-    }
-
```
JuditKnoll, line 212: Oops, this function is used outside of SpotBugs as well. It is sure used in FindSecBugs and it's causing some issues (https://github.com/find-sec-bugs/find-sec-bugs/issues/766).
The changes in `edu.umd.cs.findbugs.util.Util` class also may have some unintended impacts.

hazendaz: Ok, we should restore with comment of where usage found.  Then deprecate it as it should not have been used.  If no one gets to it, I'll do so tonight.  I was planning to cut next release this weekend after we get the eclipse feature.  Depreciation should be flagged for removal, set to 1 year.  That way it gives users time to upgrade their plugins and any tools that used ample time to upgrade.  I can send PR over to findsecbugs.

JuditKnoll: Did I just miss this change or it's not done yet?

hazendaz: Not done yet

JuditKnoll: Created a PR for it here: https://github.com/spotbugs/spotbugs/pull/3756

PR #3699, opened 2025-09-14
chore: Use Java 11 Path.of instead of Paths.get

hazendaz:


hazendaz:
spotbugs doesn't know about Path.of, looking through code on how to fix, I see 2 places potentially, jdkBaseUnconditionalDeref and BuildStringPassthruGraph.

PR #3709, opened 2025-09-15
Adjust priority for eclipse prefs

### LINKED ISSUE #3687: Eclipse misses configuration of adjustPriority command line parameter
TextUICommandLine supports adjustPriority parameter, but there is no such setting in the prefs for Eclipse.

ttho:
Added a text-field to the preferences page that allows to set the -adjustPriority command line parameter . Fixes #3687

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Fixed
 - Fix Eclipse not always using latest preferences file state ([#3740](https://github.com/spotbugs/spotbugs/issues/3740)) 
 - Fix exception throw when singleton implementing Cloneable has no clone() method ([#3727](https://github.com/spotbugs/spotbugs/issues/3727)) 
+- Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
+- Documentation of -adjustPriority parameter
+
+### Changed
+- Support for fully qualified class names for detectors in -adjustPriority parameter
+- Support for numerical and absolute priority adjustments
 
 ## 4.9.6 - 2025-09-16
 ### Fixed
--- a/eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
@@ -32,6 +32,7 @@ import java.util.TreeMap;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 
+import edu.umd.cs.findbugs.PriorityAdjuster;
 import org.dom4j.DocumentException;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
@@ -158,7 +159,9 @@ public class FindBugsWorker {
             bugReporter.setReportingStream(FindBugsConsole.getConsole().newOutputStream());
         }
         bugReporter.setPriorityThreshold(userPrefs.getUserDetectorThreshold());
-
+        if (userPrefs.getAdjustPriority() != null && !userPrefs.getAdjustPriority().isEmpty()) {
+            bugReporter.setPriorityAdjuster(new PriorityAdjuster(userPrefs.getAdjustPriority()));
+        }
         FindBugs.setHome(FindbugsPlugin.getFindBugsEnginePluginLocation());
 
         Map<IPath, IPath> outLocations = createOutputLocations();
@@ -503,8 +506,7 @@ public class FindBugsWorker {
         IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
         // path to the project without project name itself
         IClasspathEntry[] entries = javaProject.getResolvedClasspath(true);
-        for (int i = 0; i < entries.length; i++) {
-            IClasspathEntry classpathEntry = entries[i];
+        for (IClasspathEntry classpathEntry : entries) {
             if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_SOURCE) {
                 IPath outputLocation = ResourceUtils.getOutputLocation(classpathEntry, defaultOutputLocation);
                 if (outputLocation == null) {
--- a/eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
+++ b/eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
@@ -20,24 +20,32 @@ package de.tobject.findbugs.properties;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.FocusAdapter;
+import org.eclipse.swt.events.FocusEvent;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.program.Program;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Combo;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Link;
 import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Scale;
 import org.eclipse.swt.widgets.TabFolder;
 import org.eclipse.swt.widgets.TabItem;
+import org.eclipse.swt.widgets.Text;
 
 import de.tobject.findbugs.FindbugsPlugin;
 import de.tobject.findbugs.preferences.FindBugsConstants;
@@ -86,6 +94,8 @@ public class ReportConfigurationTab extends Composite {
 
     private final Button mergeSimilarWarnings;
 
+    private Text adjustPriorityText;
+
     public ReportConfigurationTab(TabFolder parent, FindbugsPropertyPage page, int style) {
         super(parent, style);
         this.propertyPage = page;
@@ -116,6 +126,7 @@ public class ReportConfigurationTab extends Composite {
 
         createBugCategoriesGroup(rankAndPrioGroup, page.getProject());
         createBugSeverityGroup(rankAndPrioGroup);
+        createAdjustPriorityGroup(this);
     }
 
     private void createBugSeverityGroup(Composite parent) {
@@ -225,6 +236,54 @@ public class ReportConfigurationTab extends Composite {
         rankValueLabel = new Label(prioGroup, SWT.NONE);
         rankValueLabel.setLayoutData(new GridData(SWT.CENTER, SWT.CENTER, true, false));
         updateRankValueLabel();
+    }
+
+    private void createAdjustPriorityGroup(Composite parent) {
+        Composite prioGroup = new Composite(parent, SWT.NONE);
+        prioGroup.setLayout(new GridLayout());
+
+        Label titleLabel = new Label(prioGroup, SWT.NONE);
+        titleLabel.setText(getMessage("property.adjustPriorityLabel"));
+        Link label = new Link(prioGroup, SWT.NONE);
+        label.setText("(One adjustment per line. Commas can be omitted. "
+                + "<a href=\"https://spotbugs.readthedocs.io/en/latest/running.html#detector-visitor-configuration-options\">Details...</a>)");
+        label.addSelectionListener(new SelectionListener() {
+            @Override
+            public void widgetSelected(SelectionEvent e) {
+                Program.launch(e.text);
+            }
+
+            @Override
+            public void widgetDefaultSelected(SelectionEvent e) {
+                // noop
+            }
+        });
+        adjustPriorityText = new Text(prioGroup, SWT.BORDER | SWT.MULTI | SWT.WRAP | SWT.V_SCROLL);
+
+        GridData gd = new GridData(SWT.FILL, SWT.FILL, true, true);
+        gd.heightHint = 160;
+        gd.widthHint = 400;
+        adjustPriorityText.setLayoutData(gd);
+
+        adjustPriorityText.addFocusListener(new FocusAdapter() {
+            @Override
+            public void focusLost(FocusEvent e) {
+                updateAdjustPriority();
+            }
+        });
+
+        updateAdjustPriorityText(getCurrentProps().getAdjustPriority());
+    }
+
+    protected void updateAdjustPriority() {
+        String adjustPriorityStr = adjustPriorityText.getText();
+        // convert line breaks to commas
+        adjustPriorityStr = adjustPriorityStr.replaceAll("\n", ",");
+        // merge multiple commas into one
+        adjustPriorityStr = adjustPriorityStr.replaceAll(",+", ",");
+        Map<String, String> adjustPriority = UserPreferences.parseAdjustPriorities(adjustPriorityStr);
+        getCurrentProps().setAdjustPriority(adjustPriority);
+        updateAdjustPriorityText(adjustPriority);
     }
 
     private void updateRankValueLabel() {
@@ -367,6 +426,14 @@ public class ReportConfigurationTab extends Composite {
         }
         syncSelectedCategories();
         mergeSimilarWarnings.setSelection(prefs.getMergeSimilarWarnings());
+        updateAdjustPriorityText(prefs.getAdjustPriority());
+    }
+
+    private void updateAdjustPriorityText(Map<String, String> adjustPriorityMap) {
+        adjustPriorityText.setText(adjustPriorityMap.entrySet()
+                .stream()
+                .map(entry -> entry.getKey() + "=" + entry.getValue())
+                .collect(Collectors.joining("\n")));
     }
 
     protected List<Button> getChkEnableBugCategoryList() {
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
@@ -1,20 +1,26 @@
 package edu.umd.cs.findbugs.detect;
 
+import edu.umd.cs.findbugs.config.UserPreferences;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import edu.umd.cs.findbugs.AbstractIntegrationTest;
-import edu.umd.cs.findbugs.DetectorFactory;
-import edu.umd.cs.findbugs.DetectorFactoryCollection;
 import edu.umd.cs.findbugs.SystemProperties;
 import edu.umd.cs.findbugs.annotations.Confidence;
 
+import java.util.Map;
+
 class FindRefComparisonTest extends AbstractIntegrationTest {
+
+    @BeforeEach
+    void beforeEach() {
+        userPreferences = UserPreferences.createDefaultUserPreferences();
+    }
 
     @AfterAll
     static void tearDown() {
         SystemProperties.removeProperty("findbugs.refcomp.reportAll");
-        setPriorityAdjustment(0); // reset priority
     }
 
     private void performAnalysisForRefComp() {
@@ -26,7 +32,6 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
     void testComparisonsNotEnabled() {
         // When there is no priority adjustment, and we enable ref comparison for
         // all, we only get an integer ref comparison bug.
-        setPriorityAdjustment(0); // reset priority
         SystemProperties.setProperty("findbugs.refcomp.reportAll", "true"); // enable ref comparison
 
         performAnalysisForRefComp();
@@ -37,7 +42,7 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
     @Test
     void testComparisonsWhenPropertyUnset() {
         // When raising the priority, we get all ref comparison bugs.
-        setPriorityAdjustment(-1); // raise priority
+        setPriorityAdjustment("-1"); // raise priority
 
         SystemProperties.removeProperty("findbugs.refcomp.reportAll");
         performAnalysisForRefComp();
@@ -49,7 +54,7 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
 
     @Test
     void testComparisonsWhenReportAllIsNotEnabled() {
-        setPriorityAdjustment(-1); // raise priority
+        setPriorityAdjustment("-1"); // raise priority
         SystemProperties.setProperty("findbugs.refcomp.reportAll", "false");
 
         performAnalysis("RefComparisons.class", "RefComparisons$MyEnum.class", "RefComparisons$MyClass2.class",
@@ -60,7 +65,7 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
 
     @Test
     void testComparisonsWhenReportAllIsEnabled() {
-        setPriorityAdjustment(-1); // raise priority
+        setPriorityAdjustment("-1"); // raise priority
         SystemProperties.setProperty("findbugs.refcomp.reportAll", "true");
 
         performAnalysisForRefComp();
@@ -70,9 +75,8 @@ class FindRefComparisonTest extends AbstractIntegrationTest {
         assertBugInMethodAtLineWithConfidence("RC_REF_COMPARISON", "RefComparisons", "myClass2BadComparison", 21, Confidence.LOW);
     }
 
-    private static void setPriorityAdjustment(int adjustment) {
-        DetectorFactory detFactory = DetectorFactoryCollection.instance().getFactory(FindRefComparison.class.getSimpleName());
-        detFactory.setPriorityAdjustment(adjustment);
+    private void setPriorityAdjustment(String adjustment) {
+        userPreferences.setAdjustPriority(Map.of(FindRefComparison.class.getSimpleName(), adjustment));
     }
 
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java
@@ -1,0 +1,89 @@
+/*
+ * SpotBugs - Find Bugs in Java programs
+ * Copyright (C) 2025, Timo Thomas
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+package edu.umd.cs.findbugs;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class PriorityAdjuster {
+    Map<String, PriorityAdjustment> adjustments;
+
+    public PriorityAdjuster(Map<String, String> adjustments) {
+        this.adjustments = new HashMap<>();
+        for (Map.Entry<String, String> entry : adjustments.entrySet()) {
+            String target = checkTarget(entry.getKey());
+            PriorityAdjustment priorityAdjustment = new PriorityAdjustment(entry.getValue());
+            this.adjustments.put(target, priorityAdjustment);
+        }
+    }
+
+    private static String checkTarget(String adjustmentTarget) {
+        DetectorFactoryCollection factoryCollection = DetectorFactoryCollection.instance();
+        DetectorFactory factory = factoryCollection.getFactoryByClassName(adjustmentTarget);
+        if (factory != null) {
+            // return the factory's FQCN
+            return factory.getClass().getName();
+        }
+        factory = factoryCollection.getFactory(adjustmentTarget);
+        if (factory != null) {
+            // return the factory's FQCN
+            return factory.getClass().getName();
+        }
+        BugPattern pattern = factoryCollection.lookupBugPattern(adjustmentTarget);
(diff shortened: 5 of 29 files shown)


ttho:
Sorry, no tests. Could not get the existing plugin tests to work :(. Eclipse never executes one.

ttho:
> With the latest commit some tests are failing causing the build to fail. Please make sure that the build is successful. If you are still working on this, you can set the PR to draft to signal this.

Oops, was too quick about it. Now fixed, and added a (non-gui) test as a bonus.

hazendaz:
@ttho Can you fix merge conflicts?

@iloveeclipse could you review this one?  If it looks and acts as expected, I can pull it into release tonight for 4.9.6 patch.  If needs more time, can push that to 4.9.7.  Either way snapshots would pick up early once merged.

iloveeclipse:
> could you review this one? 

It is on my list, in fact, I've aked Timo to provide a fix for that.

> I can pull it into release tonight for 4.9.6 patch. If needs more time, can push that to 4.9.7. 

I'm done for today, depending on my load I will try to review it later this week. So please don't wait for this PR for 4.9.6, especially it seems not that trivial as I've expected.

iloveeclipse:
@ttho : please rebase your branch on master and force push once again. Also maybe you can squash two last commits into one.

ttho:
> @ttho : please rebase your branch on master and force push once again. Also maybe you can squash two last commits into one.

Done

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## 4.9.5 - 2025-09-14
 ### Fixed
+- Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
```
JuditKnoll: Please move this to the unreleased part of the changelog.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
```diff
@@ -367,6 +441,14 @@ void refreshUI(UserPreferences prefs) {
         }
         syncSelectedCategories();
         mergeSimilarWarnings.setSelection(prefs.getMergeSimilarWarnings());
+        updateAdjustPriorityText(prefs.getAdjustPriority());
+    }
+
+    private void updateAdjustPriorityText(Map<String, String> adjustPriorityMap) {
+        adjustPriorityText.setText(adjustPriorityMap.entrySet() //
+                .stream() //
+                .map(entry -> entry.getKey() + "=" + entry.getValue()) //
```
JuditKnoll: Can you please remove these empty comments?

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -505,6 +505,53 @@ public void setUserPreferences(UserPreferences userPreferences) {
...
+                }
+
+                DetectorFactory factory = DetectorFactoryCollection.instance().getFactoryByClassName(adjustmentTarget);
+                if (factory == null) {
+                    factory = DetectorFactoryCollection.instance().getFactory(adjustmentTarget);
+                }
+                if (factory != null) {
+                    factory.setPriorityAdjustment(adjustmentAmount);
+                } else {
+                    //
```
JuditKnoll: Can you please remove this unnecessary empty comment?

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -211,7 +211,8 @@ public TextUICommandLine() {
         addOption("-omitVisitors", "v1[,v2...]", "omit named visitors");
         addOption("-chooseVisitors", "+v1,-v2,...", "selectively enable/disable detectors");
         addOption("-choosePlugins", "+p1,-p2,...", "selectively enable/disable plugins");
-        addOption("-adjustPriority", "v1=(raise|lower)[,...]", "raise/lower priority of warnings for given visitor(s)");
+        addOption("-adjustPriority", "v1=(raise|lower|suppress)[,...]",
+                "raise/lower priority of warnings for given detectors (simple or fully qualified class names) or bug patterns, or suppress them completely");
```
JuditKnoll: AFAIK adjustPriority is only for detectors/visitors (e.g. `FindNullDeref`) and not for bug patterns (e.g. `NP_DEREFERENCE_OF_READLINE_VALUE`).

spotbugs/src/main/java/edu/umd/cs/findbugs/config/UserPreferences.java
```diff
@@ -308,7 +329,10 @@ public void write(@WillClose OutputStream out) throws IOException {
         writeProperties(props, KEY_EXCLUDE_FILTER, excludeFilterFiles);
         writeProperties(props, KEY_EXCLUDE_BUGS, excludeBugsFiles);
         writeProperties(props, KEY_PLUGIN, customPlugins);
-
+        props.put(KEY_ADJUST_PRIORITY, adjustPriority.entrySet() //
+                .stream() //
+                .map(entry -> entry.getKey() + "=" + entry.getValue()) //
```
JuditKnoll: Can you please remove these empty comments from the end of the lines?

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
```diff
@@ -211,7 +211,8 @@ public TextUICommandLine() {
         addOption("-omitVisitors", "v1[,v2...]", "omit named visitors");
         addOption("-chooseVisitors", "+v1,-v2,...", "selectively enable/disable detectors");
         addOption("-choosePlugins", "+p1,-p2,...", "selectively enable/disable plugins");
-        addOption("-adjustPriority", "v1=(raise|lower)[,...]", "raise/lower priority of warnings for given visitor(s)");
+        addOption("-adjustPriority", "v1=(raise|lower|suppress)[,...]",
```
JuditKnoll: For suppression of detectors the `omitVisitors` or the `chooseVisitors` parameters can be used. Otherwise these could contradict each other, which needs to be handled and tested carefully.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
ttho: These were added intentionally to prevent the formatter of putting everything on one line, to improve readability. Are there better means to do that?

spotbugs/src/main/java/edu/umd/cs/findbugs/config/UserPreferences.java
ttho: see comment above

CHANGELOG.md
ttho: oh yes, this was done by auto-merging

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
ttho: I did not add support for suppress. The code was already there. I simply updated the documentation.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
ttho: No, support for bug patterns was already there. See TextUICommandLine:574. I simply updated the documentation (and added support for FQCN for detectors, which I found useful).

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
JuditKnoll: Spotless doesn't remove line breaks - or at least not in cases like this. I think it's not a project specific formatter rule, at least on my side this is not reformatted if I delete the comments.
Isn't it a setting in your local system, enforced by your IDE?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+- Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
```
JuditKnoll, line 12: Can you please add the `###Fixed` group as well, since it does not exist yet in Unreleased? Or rather it's not a fix, but an added feature, so it could be under `###Added`.

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
JuditKnoll: Thanks for the explanation and for the fix. (To tell the truth, I'm not so familiar with this part of SpotBugs.)

spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java
JuditKnoll: Thanks for the explanation and for the fix.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
ttho: Yes. Isn't this the default configuration of a fresh Eclipse installation? Anyway, not worth discussing here - I removed them.

JuditKnoll: I use IntelliJ, so this did not come up (and when I used eclipse I had a separate formatter setup). If you want to you can set up [gradle/eclipse-formatter.xml](https://github.com/spotbugs/spotbugs/blob/master/gradle/eclipse-formatter.xml) for the project as a formatter, which is used by spotless.

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -505,6 +505,52 @@ public void setUserPreferences(UserPreferences userPreferences) {
...
+                if (factory != null) {
+                    factory.setPriorityAdjustment(adjustmentAmount);
+                } else {
+                    DetectorFactoryCollection i18n = DetectorFactoryCollection.instance();
+                    BugPattern pattern = i18n.lookupBugPattern(adjustmentTarget);
+                    if (pattern == null) {
+                        throw new IllegalArgumentException("Unknown detector or bug pattern: " + adjustmentTarget);
+                    }
+                    pattern.adjustPriority(adjustmentAmount);
+                }
```
JuditKnoll: This is eerily similar to [TextUICommandLine:L551-L576](https://github.com/spotbugs/spotbugs/blob/ec3ed58fa6b8025652d74df06bf986483bc4c30e/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java#L551-L576). It could be extracted into a util method to avoid code duplication.

ttho: Thanks, now improved. Also refactored parsing.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
ttho: Thanks for the hint. I chose Eclipse as I thought it would be easier for ui testing. How to setup these for Intellij?

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
```diff
@@ -227,6 +240,67 @@ public void widgetSelected(SelectionEvent event) {
...
+
+        updateAdjustPriorityText(getCurrentProps().getAdjustPriority());
+    }
+
+    /**
+     *
+     */
+    protected void updateAdjustPriority() {
+        if (adjustPriorityText.getText() != null) {
+            StringTokenizer tokenizer = new StringTokenizer(adjustPriorityText.getText(), "\n,");
```
JuditKnoll: Are you sure that you wanted `"\n,"`? It seems a bit odd to me if a line starts with a `,`.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/AdjustPrioritiesViaUserPrefsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    public void afterEach() {
+        FindBugs2.resetPriorityAdjustments();
+    }
+
+    @Test
+    void testDefaultPreferencesShouldNotSetAnyAdjustments() {
+        try (FindBugs2 engine = new FindBugs2()) {
+            engine.setUserPreferences(UserPreferences.createDefaultUserPreferences());
+            DetectorFactoryCollection.instance().factoryIterator().forEachRemaining(f -> {
+                assertEquals(0, f.getPriorityAdjustment());
```
JuditKnoll: Can you please add some messages to the asserts describing the problem? Especially to the `assertTrue()` calls.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/AdjustPrioritiesViaUserPrefsTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.BugPattern;
+import edu.umd.cs.findbugs.DetectorFactory;
+import edu.umd.cs.findbugs.DetectorFactoryCollection;
+import edu.umd.cs.findbugs.FindBugs2;
+import edu.umd.cs.findbugs.config.UserPreferences;
+
+class AdjustPrioritiesViaUserPrefsTest extends AbstractIntegrationTest {
```
JuditKnoll: Can you please add a test, where there are several items in the priority adjustment map? Also perhaps one with bad input (e.g. non existent key and/or value)?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
```diff
@@ -26,12 +25,12 @@ private void performAnalysisForRefComp() {
...
         // When there is no priority adjustment, and we enable ref comparison for
         // all, we only get an integer ref comparison bug.
-        setPriorityAdjustment(0); // reset priority
         SystemProperties.setProperty("findbugs.refcomp.reportAll", "true"); // enable ref comparison
 
         performAnalysisForRefComp();
         assertBugTypeCount("RC_REF_COMPARISON", 1);
-        assertBugInMethodAtLineWithConfidence("RC_REF_COMPARISON", "RefComparisons", "integerBadComparison", 5, Confidence.HIGH);
+        assertBugInMethodAtLineWithConfidence("RC_REF_COMPARISON", "RefComparisons", "integerBadComparison", 5,
+                Confidence.HIGH);
```
JuditKnoll: As far as I can see, there is only whitespace change here. Can you please revert this unnecessary unrelated change?

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -505,6 +505,58 @@ public void setUserPreferences(UserPreferences userPreferences) {
...
+        if (userPreferences.getAdjustPriority() != null) {
+            configureAdjustPriority(userPreferences.getAdjustPriority());
+        }
+    }
+
+    /**
+     * Adjusts priorities of detectors and bug patterns.
+     *
+     * @param priorityAdjustments
+     *            Map of detector / bug pattern -> priority change
```
JuditKnoll: The possible priority values could be mentioned or linked.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
```diff
@@ -227,6 +240,67 @@ public void widgetSelected(SelectionEvent event) {
...
+                String token = tokenizer.nextToken();
+                int equalPos = token.indexOf('=');
+                if (equalPos > 0) {
+                    String key = token.substring(0, equalPos).trim();
+                    String value = token.substring(equalPos + 1).trim();
+                    if (!key.isEmpty() && !value.isEmpty()) {
+                        adjustPriority.put(key, value);
+                    }
+                }
+            }
```
JuditKnoll: This is eerily similar to `UserPreferences.parseAdjustPrioirities(String)` function. Maybe it could be reused?

JuditKnoll (changes requested):

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
ttho: Don't know exactly what you meant, but improved the code anyway.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
JuditKnoll: I can see that you modified the code in this class, but it was originally copied from `UserPreferences`, right? Can you please reuse the code instead of copying it?

ttho: I think you looked at an outdated version.

ThrawnCA (approved):

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
JuditKnoll: You are right. Thanks.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
```diff
@@ -227,6 +238,58 @@ public void widgetSelected(SelectionEvent event) {
...
+        });
+
+        updateAdjustPriorityText(getCurrentProps().getAdjustPriority());
+    }
+
+    protected void updateAdjustPriority() {
+        if (adjustPriorityText.getText() != null) {
+            String adjustPriorityStr = adjustPriorityText.getText();
+            // strip spaces
+            adjustPriorityStr = adjustPriorityStr.replaceAll(" ", "");
```
JuditKnoll: `UserPreferences.parseAdjustPriorities` trims each item and token already, so I think replacing the spaces is unnecessary.

ttho: Yes and no. `UserPreferences.parseAdjustPriorities` trims spaces only at ends, whereas here spaces in between are stripped, too. However, this wasn't really intended... ;)
I think we can leave it like this.
Thanks for the thorough review!

JuditKnoll: Since none of the bugpatterns, visitor names and acceptable priority adjustment contain spaces, with a valid value whitespace removal is already handled by the trims in `UserPreferences.parseAdjustPriorities`, and this won't cause any issues in functionality, but isn't necessary either, removing it would slightly improve performance. Additionally, space is not a regex, so `replace` could be used instead of `replaceAll` for efficiency.

If there are some additional spaces inside these values (e.g. `RC_R EF _CO MPARISON` or `r a i s e`), then yes this line removes those and makes the values acceptable. I'm not sure whether we want this.

eclipsePlugin/src/de/tobject/findbugs/properties/ReportConfigurationTab.java
```diff
@@ -227,6 +238,58 @@ public void widgetSelected(SelectionEvent event) {
...
+            public void focusLost(FocusEvent e) {
+                updateAdjustPriority();
+            }
+        });
+
+        updateAdjustPriorityText(getCurrentProps().getAdjustPriority());
+    }
+
+    protected void updateAdjustPriority() {
+        if (adjustPriorityText.getText() != null) {
```
JuditKnoll: `getText()` never returns null, only empty string if it's not set.

iloveeclipse:
I'm going to review this now, the first thing I see are multiple commits fixing / updating change log and few commits for the actual change.

@ttho : it would be great if you could cleanup the PR to only few (one or two) commits that are essential, without extra commit for changelog. Othwerwise everything will be squashed, as the single "fix changelog" commits make no sense.

JuditKnoll (approved):

eclipsePlugin/src/de/tobject/findbugs/builder/FindBugsWorker.java
```diff
@@ -216,7 +216,12 @@ public void work(List<WorkItem> resources) throws CoreException {
             FindbugsPlugin.log("Running SpotBugs");
         }
 
-        runFindBugs(findBugs);
+        try {
+            runFindBugs(findBugs);
+        } finally {
+            FindBugs2.resetPriorityAdjustments();
```
iloveeclipse: Doesn't look right (we have per-project config, why should it work without context), and makes the change not working at all in Eclipse.

iloveeclipse:
The main problem we face here is (at least in Eclipse) that we can run multiple SpotBugs analysis tasks in parallel. Not sure about Maven, Sonar, Gradle and whoever else. I see that the priority adjustment happens for a **static** context, means, all in parallel running SpotBugs instances will be affected if only one of them would apply the prio adjustments.

For Eclipse it could be solved by changing the `MutexSchedulingRule` to make it conflicting if the project preferences contains different `adjust_priority` value.

Not sure yet about other build/QA systems.

iloveeclipse:
> For Eclipse it could be solved by changing the `MutexSchedulingRule`

Nope, the code will run into troubles every time `FindBugs2.setUserPreferences(UserPreferences)` or  `FindBugs2.resetPriorityAdjustments()` are called in parallel, even if adjustments would be same.

iloveeclipse:
The priority adjustments should be made on the level of reporter / current bug collection in the context of a each `FindBugs2` object instance. As they were used before was the "hack" from `TextUICommandLine` that doesn't work for multi-threaded, non-static execution context.

ttho:
I see :(
Ok, I suggest to add it to BugReporter then. It would get an optional adjustPriorities Map as TextUICommandLine does when set up, and clones BugInstances on doReportBug() before further processing, changing their priority accordingly. The priority adjustments on BugPattern and Detector are removed. What do you think?

iloveeclipse:
> The priority adjustments on BugPattern and Detector are removed. What do you think?

If you mean the new code changes you've added: yes, because the BugPattern and Detector. changes are static.
If you mean also removing the old code: only if the behavior stays compatible for TextUICommandLine with the new changes.

ttho:
Yes, of course TextUICommandLine should work as before

ttho:
@iloveeclipse , reworked it. Now adjustments are hosted on the bug reporter, non-statically. While at it, also extended it to support an absolute setting of the priority. ("raise", "lower", "suppress" as before, but also "+n", "-n", "n", the latter being the absolute setting. Found this a nice feature as this gives the client independence over what is preconfigured in the category.
This is a draft and still needs polishing. Please review it from an architectural view only.

iloveeclipse:
@ttho : note, your branch has merge conflicts with master. Could you please rebase?
I will try to review this PR this week.

iloveeclipse:
@ttho : there is errors now with javadoc, however reports doesn't look easy to understand.
Looks like `/home/runner/work/spotbugs/spotbugs/test-harness-core/src/main/java/edu/umd/cs/findbugs/test/AnalysisRunner.java:42: error: reference not found` is the problem. I guess method reference can't be resolved.

ttho:
@iloveeclipse javadoc fixed now. As said, still needs polishing

spotbugs/src/main/java/edu/umd/cs/findbugs/FindBugs2.java
```diff
@@ -505,6 +505,9 @@ public void setUserPreferences(UserPreferences userPreferences) {
         // setAnalysisFeatureSettings(userPreferences.getAnalysisFeatureSettings());
 
         configureFilters(userPreferences);
+        if (userPreferences.getAdjustPriority() != null && !userPreferences.getAdjustPriority().isEmpty()) {
```
iloveeclipse: I would extract `userPreferences.getAdjustPriority()` to a local variable for better readability.

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java
```diff
@@ -0,0 +1,70 @@
+package edu.umd.cs.findbugs;
```
iloveeclipse, line 20: File header is missing.

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java
```diff
@@ -0,0 +1,70 @@
...
+        this.adjustments = new HashMap<>();
+        for (Map.Entry<String, String> entry : adjustments.entrySet()) {
+            String target = checkTarget(entry.getKey());
+            PriorityAdjustment priorityAdjustment = new PriorityAdjustment(entry.getValue());
+            this.adjustments.put(target, priorityAdjustment);
+        }
+    }
+
+    private static String checkTarget(String adjustmentTarget) {
+        DetectorFactory factory = DetectorFactoryCollection.instance().getFactoryByClassName(adjustmentTarget);
```
iloveeclipse: I would extract `DetectorFactoryCollection.instance()` to a local variable for better readability.

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjustment.java
```diff
@@ -0,0 +1,53 @@
+package edu.umd.cs.findbugs;
```
iloveeclipse, line 20: File header is missing

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjustment.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+
+    private IllegalArgumentException wrongFormat(String value) {
+        return new IllegalArgumentException("Invalid priority adjustment value: " + value
+                + ". Must be +[integer], -[integer], [integer] or one of raise|lower|suppress");
+    }
+
+    public int adjust(int priority) {
+        return isDelta ? priority + delta : delta;
```
iloveeclipse: After reading this method I would propose to rename `delta` field to `adjustmentValue`.

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjustment.java
```diff
@@ -0,0 +1,53 @@
...
+            delta = 1;
+        } else if ("suppress".equals(value)) {
+            isDelta = true;
+            delta = +100;
+        } else {
+            throw wrongFormat(value);
+        }
+    }
+
+    private IllegalArgumentException wrongFormat(String value) {
```
iloveeclipse: can be defined `static`

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/AdjustPrioritiesViaUserPrefsTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import edu.umd.cs.findbugs.FindBugs2;
+import edu.umd.cs.findbugs.config.UserPreferences;
+
+class AdjustPrioritiesViaUserPrefsTest extends AbstractIntegrationTest {
+
+    //    @Test
```
iloveeclipse: Is there any reason new commented out code is added? It is not completed yet, or fails, or ... ?

iloveeclipse (commented):
Code wise looks very good, that's even better what I've imagined.
Only few smaller style issue with the code.

Not approving yet as I have to test the behavior in Eclipse, but so far looks very promising.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
```diff
@@ -37,7 +42,7 @@ void testComparisonsNotEnabled() {
     @Test
     void testComparisonsWhenPropertyUnset() {
         // When raising the priority, we get all ref comparison bugs.
-        setPriorityAdjustment(-1); // raise priority
+        setPriorityAdjustment("-1"); // raise priority
```
JuditKnoll, line 45: Are you sure about this change? It looks a bit off.

iloveeclipse: The adjustment can be a string, or what do you mean by "off"?

iloveeclipse (approved):
I've tested in Eclipse now, found nothing to report, everything works as expected.

iloveeclipse:
@ttho : if you would improve the PR a bit as proposed, it would be great. Fell free to force push without extra "fix" commit.

ttho:
Polished it, please review

iloveeclipse:
> Polished it, please review

@ttho : Could you please check that all review comments are addressed and if yes, "resolve" all review comments?

ttho:
> > Polished it, please review
> 
> @ttho : Could you please check that all review comments are addressed and if yes, "resolve" all review comments?

Oh, I see I still have get accustomed to reviews on GitHub

ttho:
Ok, applied your comments now, which I simply did not see before. All resolved, except for Judit's remark.
One more thing: I stumbled upon DetectorFactory.enabledButNonReporting. Isn't this also some static state that is used from a non-static context (ExecutionPlan - FindBugsWorker)?

iloveeclipse:
> One more thing: I stumbled upon DetectorFactory.enabledButNonReporting. Isn't this also some static state that is used from a non-static context (ExecutionPlan - FindBugsWorker)?

Yep, good catch, I haven't noticed that before and I wonder how many "strange" appearing & disappearing bugs we might have seen in Eclipse if two differently configured projects were running analysis in parallel. This is not caused by your PR.

I believe this can be also fixed by removing `DetectorFactory.getPriorityAdjustment()` & `DetectorFactoryChooser` and changing `BugInstance.adjustForDetector(String)` and `BugInstance` creation so that the bugs from "enabled but non reporting" factories are filtered out at  `PriorityAdjuster.adjustPriority(BugInstance)` .

Could you please do that, it would nicely fit to the rest of the changes here? If not, we could merge your PR first and fix afterwards. WDYT?

iloveeclipse:
> Could you please do that, it would nicely fit to the rest of the changes here? If not, we could merge your PR first and fix afterwards. WDYT?

As discussed privately, I've created https://github.com/spotbugs/spotbugs/issues/3753 and will try to fix that.

Beside this, many thanks @ttho for the fix!

ttho:
Thank you also for the quick and thorough reviews. Trying to cause less iterations next time.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindRefComparisonTest.java
JuditKnoll: The acceptable string values were `"raise"`, `"lower"` and `"suppress"` earlier, which were converted into `-1`, `+1` and `+100` respectively.
I can see now that it can be also -[int],+[int] or [int] as well. When I wrote this comment either the docs were not updated yet, or I missed it.

Issue #2339, opened 2023-01-30
`fullDescription` missing from SARIF report


jplewa:
I noticed that the full rule description is not included in the SARIF report. I know that it was briefly included (in https://github.com/spotbugs/spotbugs/pull/1184) and then removed again in https://github.com/spotbugs/spotbugs/pull/1185. I understand that the reasoning was that `fullDescription` is HTML and while it can be included in the `markdown` field, it's the `text` field that's required and the HTML description can't be used to populate that one. However, I'm wondering if there's some approach that that could be used? 

Converting HTML into plaintext isn't that hard of a task, however, as per [SARIF's docs](https://github.com/microsoft/sarif-tutorials/blob/main/docs/Authoring-rule-metadata-and-result-messages.md):
> Plain text messages should be expressed in a single paragraph, without line breaks or formatting elements such as HTML tags or Markdown markup. Markdown messages, on the other hand, may make full use of Markdown's formatting capabilities (for example, line breaks, lists, code fragments, and italic or bold text). Because of this, it might make sense to present the information in the Markdown version of a message in a different order.

While it's easy enough to remove HTML tags from a block of text, I realize it might be harder to adhere to these guidelines. I've played around with some HTML parsing libraries and here's two examples of how they convert HTML into plaintext:

<details>
<summary>HTML</summary>

> <p>
>     Server-Side Request Forgery occur when a web server executes a request to a user supplied destination
>     parameter that is not validated. Such vulnerabilities could allow an attacker to access internal services
>     or to launch attacks from your web server.
> </p>
> <p>
>     URLConnection can be used with file:// protocol or other protocols to access local filesystem and potentially other services.
> <p>
>     <b>Vulnerable Code:</b>
> <pre>
> new URL(String url).openConnection()
> </pre>
> 
> <pre>
> new URL(String url).openStream()
> </pre>
> 
> <pre>
> new URL(String url).getContent()
> </pre>
> </p>
> <p>
>     <b>Solution/Countermeasures:</b><br/>
>     <ul>
>         <li>Don't accept URL destinations from users</li>
>         <li>Accept a destination key, and use it to look up the target destination associate with the key</li>
>         <li>White list URLs (if possible)</li>
>         <li>Validate that the beginning of the URL is part of a white list</li>
>     </ul>
> </p>
> <br/>
> <p>
> <b>References</b><br/>
[...]

PR #3725, opened 2025-09-22
Enhance sarif fullDescription section

adibnaya:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,76 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Added
 - Recognize `jakarta.annotation.Nonnull` and `jakarta.annotation.Nullable` ([#3780](https://github.com/spotbugs/spotbugs/pull/3780))
+- Detect use of `sun.misc.Unsafe` and `jdk.internal.misc.Unsafe` ([#3804](https://github.com/spotbugs/spotbugs/pull/3804))
+- New bug type is introduced: `NCR_NOT_PROPERLY_CHECKED_READ`. Improper validation of the return value from the read() method in InputStream and Reader classes may result in an array not being fully filled. ([#3766](https://github.com/spotbugs/spotbugs/pull/3766))
 
 ### Fixed
 - Fix incorrect bug counts and sizes when unioning reports ([#3721](https://github.com/spotbugs/spotbugs/issues/3721))
 - Enhanced SARIF output with full description sections - adding markdown is still an open issue ([#2339](https://github.com/spotbugs/spotbugs/issues/2339))
+
+### Removed
+- Removed old deprecated methods: 
+  - `assertPresentBugPattern(String, IMarker[])` protected method from `de.tobject.findbugs.test.AbstractQuickfixTest` deprecated since 2014,
+  - `setFontSizeHelper(Component[], float)` protected method from `edu.umd.cs.findbugs.gui2.FBFrame` deprecated since 2010,
+  - `matchedPrefixes(String[], String)` method from `edu.umd.cs.findbugs.gui2.ViewFilter` deprecated since 2010,
+  - `lookupFromUniqueId(String)` method from `edu.umd.cs.findbugs.BugCollection` and `edu.umd.cs.findbugs.SortedBugCollection` deprecated since 2006,
+  - `create(BugReporter)` method from `edu.umd.cs.findbugs.DetectorFactory` deprecated since 2008,
+  - `instantiateDetectorsInPass(BugReporter)` method from `edu.umd.cs.findbugs.plan.AnalysisPass` deprecated since 2008,
+  - `getMessage(String)` method from `edu.umd.cs.findbugs.I18N` deprecated since 2019,
+  - `getElementSignature()` method from `edu.umd.cs.findbugs.OpcodeStack.Item` deprecated since 2008,
+  - `getFieldAnnotation()` method from `edu.umd.cs.findbugs.OpcodeStack.Item` deprecated since 2006,
+  - `PluginLoader(URL)` and `PluginLoader(URL, ClassLoader)` constructors from `edu.umd.cs.findbugs.PluginLoader` deprecated since 2010,
+  - `addSourceDir(String)` method from `edu.umd.cs.findbugs.Project` deprecated since 2017,
+  - `getImplicitClasspathEntryList()` method from `edu.umd.cs.findbugs.Project` deprecated since 2008,
+  - `write(String, boolean, String)` method from `edu.umd.cs.findbugs.Project` deprecated since 2007,
+  - `getInteger(String, int)` method from `edu.umd.cs.findbugs.SystemProperties` deprecated since 2010,
+  - `getId()` method from `edu.umd.cs.findbugs.ba.BasicBlock` deprecated since 2010,
+  - `getArgument(InvokeInstruction, ConstantPoolGen, int, int)` method from `edu.umd.cs.findbugs.ba.Frame` deprecated since 2010,
+  - `findDeclaredExceptions(InvokeInstruction, ConstantPoolGen)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2008,
+  - `findConcreteMethod(JavaClass, String, String)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `findXMethod(JavaClass, String, String, JavaClassAndMethodChooser)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `findXMethod(JavaClass[], String, String)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `findXMethod(JavaClass[], String, String, JavaClassAndMethodChooser)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `findMethod(JavaClass[], String, String)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `isConcrete(XMethod)` method from `edu.umd.cs.findbugs.ba.Hierarchy` deprecated since 2007,
+  - `doesMethodUnconditionallyThrowException(XMethod, JavaClass, Method)` method from `edu.umd.cs.findbugs.ba.PruneUnconditionalExceptionThrowerEdges` deprecated since 2008,
+  - `nameAndSignatureIsCalled(XMethod)` method from `edu.umd.cs.findbugs.ba.XFactory` deprecated since 2020,
+  - `isInterned(XMethod)` method from `edu.umd.cs.findbugs.ba.XFactory` deprecated since 2007,
+  - `canonicalizeString(String)` method from `edu.umd.cs.findbugs.ba.XFactory` deprecated since 2017,
+  - `findXFieldFromValueNumber(Method, Location, ValueNumber, ValueNumberFrame)` method from `edu.umd.cs.findbugs.ba.npe.NullDerefAndRedundantComparisonFinder` deprecated since 2008,
+  - `findFieldAnnotationFromValueNumber(Method, Location, ValueNumber, ValueNumberFrame)` method from `edu.umd.cs.findbugs.ba.npe.NullDerefAndRedundantComparisonFinder` deprecated since 2008,
+  - `findLocalAnnotationFromValueNumber(Method, Location, ValueNumber, ValueNumberFrame)` method from `edu.umd.cs.findbugs.ba.npe.NullDerefAndRedundantComparisonFinder` deprecated since 2008,
+  - `findAnnotationFromValueNumber(Method, Location, ValueNumber, ValueNumberFrame)` method from `edu.umd.cs.findbugs.ba.npe.NullDerefAndRedundantComparisonFinder` deprecated since 2008,
+  - `compact(int[], int)` method from `edu.umd.cs.findbugs.ba.vna.ValueNumberFactory` deprecated since 2008,
+  - `fromResourceName(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `fromFieldSignature(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `isClassResource(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `createClassDescriptorFromSignature(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `createClassDescriptor(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `createClassDescriptor(String[])` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `createClassDescriptorFromDottedClassName(String)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `createClassDescriptor(JavaClass)` method from `edu.umd.cs.findbugs.classfile.ClassDescriptor` deprecated since 2008,
+  - `canonicalizeString(String)` method from `edu.umd.cs.findbugs.classfile.DescriptorFactory` deprecated since 2017,
+  - `isContainerField(XField)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `getReadFields()` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `getWrittenFields()` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `isWrittenOutsideOfInitialization(XField)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `isWrittenDuringInitialization(XField)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `isWrittenInConstructor(XField)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `strongEvidenceForIntendedSerialization(ClassDescriptor)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `existsStrongEvidenceForIntendedSerialization(ClassDescriptor)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `isReflexive(XField)` method from `edu.umd.cs.findbugs.detect.UnreadFields` deprecated since 2011,
+  - `RelationalOp(String)` private constructor from `edu.umd.cs.findbugs.filter.RelationalOp` deprecated since 2008,
+  - `isLibraryFileName(String)` method from `edu.umd.cs.findbugs.util.Archive` deprecated since 2022,
+  - `replace(String, String, String)` method from `edu.umd.cs.findbugs.util.Strings` deprecated since 2010,
+  - `toString(Object[])` method from `edu.umd.cs.findbugs.util.Strings` deprecated since 2010,
+  - `closeSilently(OutputStream)` method from `edu.umd.cs.findbugs.util.Util` deprecated since 2018,
+  - `closeSilently(Closeable)` method from `edu.umd.cs.findbugs.util.Util` deprecated since 2018,
+  - `closeSilently(ZipFile)` method from `edu.umd.cs.findbugs.util.Util` deprecated since 2018,
+  - `getRefConstantOperand()` method from `edu.umd.cs.findbugs.visitclass.DismantleBytecode` deprecated since 2010,
+  - `getDottedFieldSig()` method from `edu.umd.cs.findbugs.visitclass.PreorderVisitor` deprecated since 2006.
+- Removed old deprecated fields:
+  - String `RELEASE` from `edu.umd.cs.findbugs.Version` deprecated since 2018.
 
 ## 4.9.8 - 2025-10-18
 ### Fixed
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,18 +1,17 @@
-Before opening a 'bug'
-
-* Search existing issues and pull requests to see if the issue was already discussed.
-* Check our discussions to see if the issue was already discussed.
-* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
-* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin)
-
 ---
 name: Bug report
 about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
-
 ---
+
+**Before opening a bug**
+
+* Search existing issues and pull requests to see if the issue was already discussed.
+* Check our discussions to see if the issue was already discussed.
+* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
+* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin)
 
 **Describe the bug**
 A clear and concise description of what the bug is.
@@ -24,22 +23,26 @@ Steps to reproduce the behavior:
 3. Scroll down to '....'
 4. See error
 
+Please provide a code example, if applicable. Prefarably [a minimal, verifiable code example](https://stackoverflow.com/help/minimal-reproducible-example).
+
+```java
+public class Bug {
+  public static void foo() {
+    // code example
+  }
+}
+```
+
 **Expected behavior**
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
-
-**Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
-
-**Smartphone (please complete the following information):**
- - Device: [e.g. iPhone6]
- - OS: [e.g. iOS8.1]
- - Browser [e.g. stock browser, safari]
- - Version [e.g. 22]
+**Environment**
+- Java Compiler: [e.g. Eclipse Temurin]
+- Java version: [e.g. JDK 21]
+- SpotBugs Version: [e.g. 4.9.8]
+- How do you use SpotBugs: [e.g. via Gradle Plugin, Maven Plugin, Ant Plugin, SonarQube Plugin, Eclipse IDE Plugin, CLI]
+    - Plugin version: [e.g. SpotBugs Gradle Plugin 6.4.5]
+    - Environment version for Plugin: [e.g. Gradle 9.2.1]
 
 **Additional context**
 Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,10 +1,3 @@
-Before opening a 'pull request'
-
-* Search existing issues and pull requests to see if the issue was already discussed.
-* Check our discussions to see if the issue was already discussed.
-* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
-* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin)
-
 ---
 name: Feature request
 about: Suggest an idea for this project
@@ -13,6 +6,13 @@ labels: ''
 assignees: ''
 
 ---
+
+**Before opening a feature request**
+
+* Search existing issues and pull requests to see if the issue was already discussed.
+* Check our discussions to see if the issue was already discussed.
+* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
+* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin)
 
 **Is your feature request related to a problem? Please describe.**
 A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
     continue-on-error: true
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
         with:
           fetch-depth: 0
       - name: Set up JDK ${{ matrix.java }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,7 +26,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
         with:
           fetch-depth: 0
           fetch-tags: true
--- a/build.gradle
+++ b/build.gradle
@@ -9,9 +9,9 @@ import org.gradle.api.publish.maven.MavenArtifact;
 import org.gradle.api.tasks.util.PatternFilterable;
 
 plugins {
-  id 'org.sonarqube' version '7.0.1.6134'
+  id 'org.sonarqube' version '7.1.0.6387'
   id 'org.gradle.crypto.checksum' version '1.4.0'
-  id 'com.github.spotbugs' version '6.4.4'
+  id 'com.github.spotbugs' version '6.4.5'
   id 'io.github.gradle-nexus.publish-plugin' version '2.0.0'
 }
 
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,6 +1,6 @@
 plugins {
     `kotlin-dsl`
-    id("com.diffplug.spotless") version "8.0.0"
+    id("com.diffplug.spotless") version "8.1.0"
 }
 
 repositories {
@@ -8,5 +8,5 @@ repositories {
 }
 dependencies {
     implementation("com.diffplug.gradle:goomph:4.4.1")
-    implementation("com.diffplug.spotless:spotless-plugin-gradle:8.0.0")
+    implementation("com.diffplug.spotless:spotless-plugin-gradle:8.1.0")
 }
--- a/eclipsePlugin-test/src/de/tobject/findbugs/test/AbstractQuickfixTest.java
+++ b/eclipsePlugin-test/src/de/tobject/findbugs/test/AbstractQuickfixTest.java
@@ -32,7 +32,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
-import javax.annotation.Nonnull;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IResource;
@@ -49,7 +48,6 @@ import org.junit.jupiter.api.BeforeEach;
 import de.tobject.findbugs.FindbugsPlugin;
 import de.tobject.findbugs.FindbugsTestPlugin;
 import de.tobject.findbugs.reporter.MarkerUtil;
-import edu.umd.cs.findbugs.BugPattern;
 import edu.umd.cs.findbugs.plugin.eclipse.quickfix.BugResolution;
 import edu.umd.cs.findbugs.plugin.eclipse.quickfix.BugResolutionGenerator;
 
@@ -196,17 +194,6 @@ public abstract class AbstractQuickfixTest extends AbstractPluginTest {
     protected void assertEqualFiles(URL expectedFile, ICompilationUnit compilationUnit) throws IOException, JavaModelException {
         String expectedSource = readFileContents(expectedFile);
         assertEquals(expectedSource, compilationUnit.getSource());
-    }
-
-    @Deprecated
-    protected void assertPresentBugPattern(@Nonnull String bugPatternType, IMarker[] markers) {
-        for (int i = 0; i < markers.length; i++) {
-            BugPattern pattern = MarkerUtil.findBugPatternForMarker(markers[i]);
-            if (pattern != null && bugPatternType.equals(pattern.getType())) {
-                return;
-            }
-        }
-        fail("Couldn't find pattern " + bugPatternType);
     }
 
     protected void assertPresentBugPatterns(List<QuickFixTestPackage> packages, IMarker[] markers) {
--- a/gradle/checkstyle.gradle
+++ b/gradle/checkstyle.gradle
@@ -2,7 +2,7 @@
 apply plugin: 'checkstyle'
 
 checkstyle {
-  toolVersion = '12.1.1'
+  toolVersion = '12.1.2'
   ignoreFailures = false
   configFile = file("${rootDir}/spotbugs/etc/checkstyle.xml") // TODO : This config file is lame and should be moved out...
 }
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
--- a/spotbugs-tests/build.gradle
+++ b/spotbugs-tests/build.gradle
@@ -20,7 +20,7 @@ dependencies {
   implementation 'org.apache.ant:ant:1.10.15'
   implementation libs.log4j.core
   implementation libs.log4j.slf4j2.impl
-  implementation 'com.google.errorprone:error_prone_annotations:2.43.0'
+  implementation 'com.google.errorprone:error_prone_annotations:2.44.0'
   implementation files(project(':spotbugs').sourceSets.gui.output)
   testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ReadReturnShouldBeCheckedTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ReadReturnShouldBeCheckedTest.java
@@ -1,0 +1,37 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class ReadReturnShouldBeCheckedTest extends AbstractIntegrationTest {
+
+    private static final String NCR_BUG_TYPE = "NCR_NOT_PROPERLY_CHECKED_READ";
+    private static final String RR_BUG_TYPE = "RR_NOT_CHECKED";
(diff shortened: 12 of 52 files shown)


JuditKnoll:
It looks like the build is failing because of Spotless. Please run `.\gradlew spotlessApply` to resolve it.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+### Added
+- Enhanced SARIF output with detailed help sections for each rule, incorporating comprehensive bug pattern descriptions from SpotBugs message database
```
JuditKnoll: Can you please refer this PR for reference?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -124,7 +133,8 @@ static Rule fromBugPattern(BugPattern bugPattern, String formattedMessage) {
             tags = Collections.singletonList(category);
         }
 
-        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailText(), formattedMessage, helpUri,
+        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailText(), formattedMessage, bugPattern
+                .getDetailText(), helpUri,
```
JuditKnoll: Can you please format it in a way to have `bugPattern` and `.getDetailText()` in the same line? (It may happen that this was done by Spotless because of the line is too long.)

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -29,19 +29,23 @@ final class Rule {
...
-    Rule(@NonNull String id, @NonNull String shortDescription, @NonNull String fullDescription, @NonNull String defaultText, @Nullable URI helpUri,
+    Rule(@NonNull String id, @NonNull String shortDescription, @Nullable String fullDescription, @NonNull String defaultText,
+            @Nullable String detailText, @Nullable URI helpUri,
             @NonNull List<String> tags, @NonNull int cweid) {
         this.id = Objects.requireNonNull(id);
         this.shortDescription = Objects.requireNonNull(shortDescription);
-        this.fullDescription = Objects.requireNonNull(fullDescription);
+        this.fullDescription = fullDescription != null ? fullDescription : "";
         this.defaultText = Objects.requireNonNull(defaultText);
+        this.detailText = detailText != null ? detailText : "";
```
JuditKnoll: For reference: the [help property definition in SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791099):

> 3.49.13 help property
> A reportingDescriptor object **MAY** contain a property named help whose value is a localizable multiformatMessageString object (§3.12, §3.12.2) which provides the primary documentation for the reporting item.
> 
> NOTE: This property is useful when help information is not available at a URI, for example, in the case of a custom rule written by a developer, as opposed to one supplied by the tool vendor.

JuditKnoll: Since the object only **may** contain this property - and not should -, please add it as null and not as an empty string.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -29,19 +29,23 @@ final class Rule {
     final String fullDescription;
     @NonNull
     final String defaultText;
+    @NonNull
+    final String detailText;
```
JuditKnoll: As far as I can see, this corresponds to the [help](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791099) property from SARIF. Can you please rename it to not cause confusion?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -29,19 +29,23 @@ final class Rule {
...
     final int cweid;
 
-    Rule(@NonNull String id, @NonNull String shortDescription, @NonNull String fullDescription, @NonNull String defaultText, @Nullable URI helpUri,
+    Rule(@NonNull String id, @NonNull String shortDescription, @Nullable String fullDescription, @NonNull String defaultText,
+            @Nullable String detailText, @Nullable URI helpUri,
             @NonNull List<String> tags, @NonNull int cweid) {
         this.id = Objects.requireNonNull(id);
         this.shortDescription = Objects.requireNonNull(shortDescription);
-        this.fullDescription = Objects.requireNonNull(fullDescription);
+        this.fullDescription = fullDescription != null ? fullDescription : "";
```
JuditKnoll: The [definition of fullDescription property of reportingDescriptor in SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791096):
> 3.49.10 fullDescription property
> A reportingDescriptor object SHOULD contain a property named fullDescription whose value is a localizable multiformatMessageString object (§3.12, §3.12.2) that comprehensively describes the reporting item.
> 
> The fullDescription property SHOULD, as far as possible, provide details sufficient to enable resolution of any problem indicated by the reporting item.
> 
> The beginning of fullDescription (for example, its first sentence) SHOULD provide a concise description of the reporting item, suitable for display in cases where available space is limited. Tools that construct fullDescription in this way do not need to provide a value for shortDescription (§3.49.9). Tools that do not construct fullDescription in this way SHOULD provide a value for shortDescription.
> 
> NOTE:The rationale for this guidance is that in the absence of shortDescription, a viewer with limited display space might display a truncated version of fullDescription, for example, the first sentence (if a sentence is identifiable), the first paragraph, or the first 100 characters. If this guidance is not followed, that truncated version might not be understandable.

JuditKnoll: From this it looks like to me, the annotation of the parameter should stay `@NonNull` and this line should not be changed. Can you please elaborate why did you change this?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +172,119 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+
+        JsonObject ruleJson = rules.get(0).getAsJsonObject();
+        assertThat(ruleJson.get("id").getAsString(), is(type));
+
+        // Verify help section exists but is empty when detail text is null
+        assertThat(ruleJson.has("help"), is(true));
+        JsonObject help = ruleJson.get("help").getAsJsonObject();
+        assertThat(help.has("text"), is(true));
+        String helpText = help.get("text").getAsString();
+        assertThat(helpText, notNullValue());
```
JuditKnoll: According to SARIF's definition the [help property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791099) is an optional [multiformatMessageString object](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790723), where the text property is a not optional non-empty string. So if the help property is null or empty, the rule should have a help property.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +172,119 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        JsonObject ruleJson = rules.get(0).getAsJsonObject();
+        assertThat(ruleJson.get("id").getAsString(), is(type));
+
+        // Verify help section exists and contains empty text
+        assertThat(ruleJson.has("help"), is(true));
+        JsonObject help = ruleJson.get("help").getAsJsonObject();
+        assertThat(help.has("text"), is(true));
+        String helpText = help.get("text").getAsString();
+        assertThat(helpText, notNullValue());
+        assertThat(helpText, is(""));
```
JuditKnoll: According to SARIF's definition the [help property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791099) is an optional [multiformatMessageString object](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790723), where the text property is a not optional non-empty string. So if the help property is null or empty, the rule should have a help property.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +172,119 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        assertThat(ruleJson.get("id").getAsString(), is(type));
+
+        // Verify help section exists and contains the detail text
+        assertThat(ruleJson.has("help"), is(true));
+        JsonObject help = ruleJson.get("help").getAsJsonObject();
+        assertThat(help.has("text"), is(true));
+        String helpText = help.get("text").getAsString();
+        assertThat(helpText, notNullValue());
+        assertThat(helpText, containsString("This is a test detail text"));
+        assertThat(helpText, containsString("<strong>HTML markup</strong>"));
```
JuditKnoll: According to SARIF, this should be plain text. See [the definition](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790726).

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: According to SARIF the text should be plain text. The `BugPattern::getDetailPlainText` gives you back `detailText` as plain text, so you can use that.

JuditKnoll:
Please address the change requests before asking for a new review. Perhaps you forgot to push your changes and/or haven't send your review comments?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: There is a small issue as both fullDescription and also help text uses the same "details" field from messages.xml but one of them (fullDescription) is nonNull and the other is nullable, so the tests creates a null detailText and then the tests fails,
Examining the code it seems as maybe the right field to take the "fullDescription" field of the Rule should acually be getLongDescription() instead of getDetailText()?

CHANGELOG.md
adibnaya: Done

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

adibnaya: Done, not pushed as there's a need to finalize the getDesription issues

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: After further inspection, you are right, the SARIF `fullDescription` really comes from BugPattern's `detailText`.

No, IMO `fullDescription` should stay `detailText`, since `longDescription` is already used at `defaultText`. (See the [usage](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyser.java#L146) of `fromBugPattern` in `BugCollectionAnalyser`'s `processRule()` function.)

So as far as I can see, instead of adding the `help` parameter, it would be better to only fill out the `fullDescription` parameter with the value in `Rule::toJsonObject()`, as the TODO already indicates. https://github.com/spotbugs/spotbugs/issues/2339 is the issue about this.

ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Well, I've just verified that adding fullDescription instead of help would also show in the needed place in the security check screen but it will not change the exact same issue i'm facing with the tests with the help field and the tests, Unless fullDescription **MAY** be there but if it's there it has to have content (Or empty string) and then I will not have the issue with null.

If this is the case I can just modify my code to add full description and will add the TODO in #2339 to the changelog, does this sounds good?

JuditKnoll: Yes, IMO it would be great solution to add full description to the SARIF report.

JuditKnoll: According to SARIF standard, fullDescription should not be an empty string. If there is no value or it's empty, it should not be present at all. This is also true for help text, since both are [multiformatMessageString objects](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790723).

____

**Summary:** `detailText` in `edu.umd.cs.findbugs.BugPattern` is not null, but there may be some cases where it's an empty string (could not find a check for it).

**Reasoning/Details:** 
As far as I can see in SpotBugs `edu.umd.cs.findbugs.sarif.Rule` (which represents SARIF's reportingDescriptor) is only created from `edu.umd.cs.findbugs.BugPattern`, which is created from the findbugs.xml and messages.xml files. Specifically the `detailText` is from the `Details` in messages.xml. The xsd for [messages.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/messages.xml) is [messagecollection.xsd](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/messagecollection.xsd). The definition for BugPattern in the xsd is the following:
https://github.com/spotbugs/spotbugs/blob/660cdf7d2da1d9af1aa4e11a6424151c733c4b4b/spotbugs/etc/messagecollection.xsd#L66-L74
Meaning that if the messages.xml is valid according to the schema, the Details needs to be added for every BugPattern. However it can be an empty string and still be valid.
So if the messages.xml is valid according to the schema, `detailText` in `edu.umd.cs.findbugs.BugPattern` is not null, but it may be an empty string.
From a quick search, I could not find the validating of messages.xml in the code, but it should be validated.

The `edu.umd.cs.findbugs.BugPattern` class has only one constructor and has no setter, except for the deprecated flag. The only constructor sets `detailText` from the parameter. This constructor is called in SpotBugs either in tests - where detailText does have a value -, when [creating the REALLY_UNKNOWN BugPattern](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/BugPattern.java#L98-L103) inside the same class, or [in PluginLoader](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java#L996). PluginLoader does a null check for detailText [inside the getChildText()](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java#L1352-L1358) function.

Currently in SpotBugs in the messages.xmls (all languages) no BugPattern details are empty strings, and they should not be empty. However from a quick search, I could not find a check verifying this. This means that in the future or in a plugin there may be an empty string.

adibnaya: Empty strings are not problematic, I can handle the end cases, the problem was only with using NULL, which was not actually problematic in real usecases but only with the tests as the same value was used for nullable and nonNull values, I'll go ahead and update the code with fullDescription.

adibnaya: Done all the changes, Resolving for now, please let me know of any further changes.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +171,123 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+
+        /* test that rule contains fullDescription with text */
+        JsonArray rules = analyser.getRules();
+        assertThat(rules.size(), is(1));
+
+        JsonObject ruleJson = rules.get(0).getAsJsonObject();
+        assertThat(ruleJson.get("id").getAsString(), is(type));
+
+        /* test that fullDescription exists and contains the detail text */
+        assertThat(ruleJson.has("fullDescription"), is(true));
```
JuditKnoll: Instead of `assertThat(something, is(true))` please use `assertTrue(something)`.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +171,123 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        assertThat(fullDescText, notNullValue());
+        assertThat(fullDescText, is(detailText));
+    }
+
+    @Test
+    void testRuleFullDescriptionHandlesNullDetailText() {
+        String type = "TEST_NULL_DETAIL_TYPE";
+
+        BugPattern bugPattern = new BugPattern(type, "abbrev", "category", false, "shortDescription",
+                "longDescription", null, "https://example.com/help.html", 0);
```
JuditKnoll: The detail text never will be null in the code, see my earlier comment and reasoning at line 43/46 of Rule. So I think it needs to be tested.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +171,123 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        /* test that fullDescription exists and null detail text is handled as empty string */
+        assertThat(ruleJson.has("fullDescription"), is(true));
+        JsonObject fullDescription = ruleJson.get("fullDescription").getAsJsonObject();
+        assertThat(fullDescription.has("text"), is(true));
+
+        /* null values should be converted to empty strings */
+        String fullDescText = fullDescription.get("text").getAsString();
+
+        assertThat(fullDescText, notNullValue());
+        assertThat(fullDescText.isEmpty(), is(true));
```
JuditKnoll: As I mentioned in my earlier comment, according to SARIF, this should never be an empty string.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +171,123 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+
+        /* test that fullDescription exists and empty detail text is preserved */
+        assertThat(ruleJson.has("fullDescription"), is(true));
+        JsonObject fullDescription = ruleJson.get("fullDescription").getAsJsonObject();
+        assertThat(fullDescription.has("text"), is(true));
+
+        String fullDescText = fullDescription.get("text").getAsString();
+
+        assertThat(fullDescText, notNullValue());
+        assertThat(fullDescText.isEmpty(), is(true));
```
JuditKnoll: As I mentioned in my earlier comment, according to SARIF, this should never be an empty string.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -57,11 +57,15 @@ JsonObject toJsonObject() {
         JsonObject messageStrings = new JsonObject();
         messageStrings.add("default", textJson);
 
-        // TODO put 'fullDescription' with both of text and markdown representations
+        JsonObject fullDescJson = new JsonObject();
+        fullDescJson.addProperty("text", fullDescription);
```
JuditKnoll: Please add the markdown representation as well.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: FullDescription is not nullable, it is nonnull. See the details in my previous comment. Please don't change the annotation on it.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -124,7 +128,7 @@ static Rule fromBugPattern(BugPattern bugPattern, String formattedMessage) {
             tags = Collections.singletonList(category);
         }
 
-        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailText(), formattedMessage, helpUri,
-                tags, bugPattern.getCWEid());
+        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailPlainText(), formattedMessage,
```
JuditKnoll: To be able to add both the plaintext and the markdown representation, the original detailtext is needed.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -170,4 +171,123 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+
+        SourceLineAnnotation lineAnnotation = new SourceLineAnnotation("TestFile", "Test.java", 1, 1, 0, 0);
+        bug.addSourceLine(lineAnnotation);
+
+        bugCollection.add(bug);
+        bugCollection.bugsPopulated();
+
+        BugCollectionAnalyser analyser = new BugCollectionAnalyser(bugCollection);
+
+        /* test that rule handles null detail text properly - null values should be converted to empty strings */
```
JuditKnoll: From what I get, it's the other way around: the detail text (given value) is never null, it can be an empty string (unlikely) or a html formatted text. The SARIF fullDescription (the expected value) can be either null, or a json object with text field which is a non-empty plain text. However Rule's - representing the SARIF object - fullDescription field is non null, so if the detail text is an empty string, maybe it could be a template text, just like the one for the REALLY_UNKOWN BugPattern.

So, what should be tested IMO:
- detail text is a non-empty plain text -> the fullDescription.text is the same plain text,
- detail text is a non-empty HTML formatted text -> the fullDescription.text is a plain text with the expected value, (preferably this should have more tests)
- detail text is empty string -> the fullDescription is a template text mentioning that there should be something here, as the fullDescription field of Rule is nonnull?

Please let me know if you disagree with me and share your side.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Yes, but for future possible cases I thought it's importnant to handle the case that it will be, if for some reason detail text is null so it's needed to validate that:
1. No exceptions are thrown
2. fullDescription is an empty string and not null

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: I agree, I also think that all these texts should be held in some external constants class but this is for another PR, I don't want to mix

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Chnaged to template text

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: I thought about it and added but what should be inside? because if it's the same plain text so I see no reason for that and if it should really be markdown text so there's a need of some conversion by some rules, or am I missing something?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: detailText in messages is HTML, not markdown

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -36,11 +36,13 @@ final class Rule {
     @NonNull
     final int cweid;
 
-    Rule(@NonNull String id, @NonNull String shortDescription, @NonNull String fullDescription, @NonNull String defaultText, @Nullable URI helpUri,
-            @NonNull List<String> tags, @NonNull int cweid) {
+    Rule(@NonNull String id, @NonNull String shortDescription, @Nullable String fullDescription, @NonNull String defaultText,
+            @Nullable URI helpUri, @NonNull List<String> tags, @NonNull int cweid) {
```
JuditKnoll: Can you please revert the changes to these lines?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -36,11 +36,13 @@ final class Rule {
...
         this.id = Objects.requireNonNull(id);
         this.shortDescription = Objects.requireNonNull(shortDescription);
-        this.fullDescription = Objects.requireNonNull(fullDescription);
+        this.fullDescription = (fullDescription != null && !fullDescription.equals("null") && !fullDescription.trim().isEmpty()) 
+            ? fullDescription 
+            : "No detailed description available for this bug pattern.";
```
JuditKnoll: The fullDescription parameter is @NonNull, it will never have `null` value, and I don't think we care if it's value is the string `"null"`. Why do you think the `"null"` string should be handled differently?

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: If you take a look at the multiformatMessageString object's definition inside SARIF, you can see that the markdown property should be the following:
> 3.12.4 markdown property
> A multiformatMessageString object **MAY** contain a property named markdown whose value is a non-empty string containing a formatted message (§3.11.4) expressed in GitHub-Flavored Markdown [[GFM](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#GFM)].
> 
> SARIF consumers that cannot (or choose not to) render formatted text **SHALL** ignore the markdown property and use the text property (§3.12.3) instead.

The earlier mentioned https://github.com/spotbugs/spotbugs/issues/2339 issue can also help you in this.

Maybe there is a solution already to this problem in some library, which could be reused.

Or another solution can be, that you leave the part of the TODO mentioning the markdown representation, and make sure that the CHANGELOG entry is in sync with this.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -162,12 +164,173 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        /* test that fullDescription exists and HTML is converted to plain text */
+        assertTrue(ruleJson.has("fullDescription"));
+        JsonObject fullDescription = ruleJson.get("fullDescription").getAsJsonObject();
+        assertTrue(fullDescription.has("text"));
+
+        String fullDescText = fullDescription.get("text").getAsString();
+
+        assertNotNull(fullDescText);
+        /* HTML conversion may add trailing whitespace, so we trim it */
+        assertThat(fullDescText.trim(), is(expectedPlainText));
```
JuditKnoll: The triming should be done in the code, and not in the test.

JuditKnoll (changes requested):

JuditKnoll (changes requested):
The build is failing because of formatting issues. Run `gradlew spotlessApply` to resolve them, just like the build fail is indicating it.
Pls make sure that `gradlew build smokeTest` is successful, before requesting a new review.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: when BugPattern.getDetailPlainText() is called with a null detailText, it's returning the string "null" instead of null.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Decided to wait for next time, addressed it in CHANGELOG

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Done

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: Thank you for highlighting it.
We still need the original detailText to convert the HTML to markdown. When converting to plain text a lot of info is lost.

Can you please keep the original detailText here and do the conversion when setting fullDescription.text in the `toJsonObject()` function?

adibnaya:
I've made the changes, ran `gradlew spotlessApply` and `gradlew build smokeTest`, both finished well

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -162,12 +164,133 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+
+        /* test that fullDescription exists and HTML is converted to plain text */
+        assertTrue(ruleJson.has("fullDescription"));
+        JsonObject fullDescription = ruleJson.get("fullDescription").getAsJsonObject();
+        assertTrue(fullDescription.has("text"));
+
+        String fullDescText = fullDescription.get("text").getAsString();
+
+        assertNotNull(fullDescText);
+        /* HTML conversion may add trailing whitespace, so we trim it */
```
JuditKnoll: This comment is not up to date anymore.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -57,11 +58,27 @@ JsonObject toJsonObject() {
         JsonObject messageStrings = new JsonObject();
         messageStrings.add("default", textJson);
 
-        // TODO put 'fullDescription' with both of text and markdown representations
+        // TODO add markdown representations to 'fullDescription'
```
JuditKnoll:
```suggestion
        // TODO add markdown representation to 'fullDescription'
```

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -57,11 +58,27 @@ JsonObject toJsonObject() {
...
+        JsonObject fullDescJson = new JsonObject();
+        String plainTextDescription;
+        try {
+            plainTextDescription = HTML.convertHtmlSnippetToText(fullDescription);
+            if (plainTextDescription.trim().isEmpty()) {
+                plainTextDescription = "No detailed description available for this bug pattern.";
+            } else {
+                plainTextDescription = plainTextDescription.trim();
+            }
+        } catch (Exception e) {
```
JuditKnoll: Please do not catch all exceptions, just the relevant ones.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -124,7 +141,7 @@ static Rule fromBugPattern(BugPattern bugPattern, String formattedMessage) {
             tags = Collections.singletonList(category);
         }
 
-        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailText(), formattedMessage, helpUri,
-                tags, bugPattern.getCWEid());
+        return new Rule(bugPattern.getType(), bugPattern.getShortDescription(), bugPattern.getDetailText(), formattedMessage,
+                helpUri, tags, bugPattern.getCWEid());
```
JuditKnoll: Can you please revert this whitespace change as there is no meaningful change here to not interfere with the git history?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+### Fixed
+- Enhanced SARIF output with full description sections for each rule, incorporating comprehensive bug pattern descriptions from SpotBugs message database - adding markdown is still am open issue ([#2339](https://github.com/spotbugs/spotbugs/issues/2339))
+
```
JuditKnoll, line 86: Please merge it with the already existing Fixed section under Unreleased. Also it could be slightly simplified.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -57,11 +58,27 @@ JsonObject toJsonObject() {
...
+        try {
+            plainTextDescription = HTML.convertHtmlSnippetToText(fullDescription);
+            if (plainTextDescription.trim().isEmpty()) {
+                plainTextDescription = "No detailed description available for this bug pattern.";
+            } else {
+                plainTextDescription = plainTextDescription.trim();
+            }
+        } catch (Exception e) {
+            plainTextDescription = fullDescription; // Fallback to original if conversion fails
+        }
```
JuditKnoll: Please log at least that there was an issue with the conversion.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -57,11 +58,27 @@ JsonObject toJsonObject() {
...
+        String plainTextDescription;
+        try {
+            plainTextDescription = HTML.convertHtmlSnippetToText(fullDescription);
+            if (plainTextDescription.trim().isEmpty()) {
+                plainTextDescription = "No detailed description available for this bug pattern.";
+            } else {
+                plainTextDescription = plainTextDescription.trim();
+            }
+        } catch (Exception e) {
+            plainTextDescription = fullDescription; // Fallback to original if conversion fails
```
JuditKnoll: IMO it should be trimmed in this case as well, but that could be done when adding the property.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: Okay, thanks :)

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: Thanks for highlighting this :)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
```diff
@@ -162,12 +164,133 @@ void testGetRulesGetResultsGetCweTaxonomyMultipleBugsWithAndWithoutCweId() {
...
+        assertTrue(fullDescription.has("text"));
+
+        String fullDescText = fullDescription.get("text").getAsString();
+
+        assertNotNull(fullDescText);
+        assertThat(fullDescText, is("No detailed description available for this bug pattern."));
+    }
+
+    @Test
+    void testRuleFullDescriptionHandlesHtmlDetailText() {
```
JuditKnoll, line 257: Can you please add one more example with more html tags used in messages.xml? E.g. with CDATA, a table, code, etc.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
JuditKnoll: This could be a simple private util function.

JuditKnoll (changes requested):
Can you please squash your commits as your earlier ones quite different, and not as relevant, and modify the title to represent the change better?

adibnaya:
> Can you please squash your commits as your earlier ones quite different, and not as relevant, and modify the title to represent the change better?

There are a few merges between my commits and I would not want to rewrite history so I'm not sure about how to squash without destroying the branch, I did change the title...

CHANGELOG.md
adibnaya: Done locally, will push soon

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done locally, will push soon

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done locally, will push soon

adibnaya: Done locally, will push soon

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done locally, will push soon

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
adibnaya: Done locally, will push soon

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Done locally, will push soon

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/sarif/BugCollectionAnalyserTest.java
adibnaya: Added a few more, can remove a few of them if you think it's too much

hazendaz:
I see devs on the team ask users to squash pretty frequently.  That is unnecessary, we are going to squash your commits on merge per policy.

ThrawnCA (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Documentation of -adjustPriority parameter
 - Functionality from DetectorFactory setEnabledButNonReporting(), getPriorityAdjustment() methods and BugInstance.adjustForDetector() is deprecated and moved to PriorityAdjuster ([#3753](https://github.com/spotbugs/spotbugs/issues/3753))
 - Improved `FindNakedNotify` to handle the case when the lock is loaded from a field ([#3634](https://github.com/spotbugs/spotbugs/issues/3634))
+- Enhanced SARIF output with full description sections - adding markdown is still an open issue ([#2339](https://github.com/spotbugs/spotbugs/issues/2339))
```
JuditKnoll: Please move this changelog entry to Unreleased section.

spotbugs/src/main/java/edu/umd/cs/findbugs/sarif/Rule.java
```diff
@@ -40,7 +48,7 @@ final class Rule {
             @NonNull List<String> tags, @NonNull int cweid) {
         this.id = Objects.requireNonNull(id);
         this.shortDescription = Objects.requireNonNull(shortDescription);
-        this.fullDescription = Objects.requireNonNull(fullDescription);
+        this.fullDescription = Objects.requireNonNull(fullDescription.trim());
```
JuditKnoll: The trim() should be called on the return value of the Objects.requireNonNull, not on it's parameter like this:
```suggestion
        this.fullDescription = Objects.requireNonNull(fullDescription).trim();
```

adibnaya:
> Two small remarks, but otherwise looks good. Thank you for your work on this!

Done, Thank you for the comments, I've learned a lot.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## 4.9.8 - 2025-10-18
 ### Fixed
 - Maven plugin reporting issue if -adjustPriority is not set ([#3774](https://github.com/spotbugs/spotbugs/issues/3774)) 
+- Enhanced SARIF output with full description sections - adding markdown is still an open issue ([#2339](https://github.com/spotbugs/spotbugs/issues/2339))
```
JuditKnoll: Can you please move it under Unreleased? Feel free to add a `### Fixed` section as necessary.

adibnaya: Done

adibnaya:
All checks are green and the branch is current. Since the PR has approvals, could a maintainer please squash-merge when convenient? Thanks! @hazendaz  @JuditKnoll  @ThrawnCA

hazendaz (approved):

Issue #3740, opened 2025-09-30
Eclipse can use out-of-date settings file


iloveeclipse:
Just found a **very** old bug, found while reviewing https://github.com/spotbugs/spotbugs/pull/3709.

After the `UserPreferences` in Eclipse are read from the disk, they are stored in the project session property and the value is not changed even if the underlined properties file is changed on disk, only if the user changes preferences via UI.
So just opening the `.settings/edu.umd.cs.findbugs.core.prefs` in the text editor and changing any value doesn't make any difference on followed analysis results.

The fix is trivial, I will push a PR in a moment.

PR #3741, opened 2025-09-30
Always read latest preferences state from the disk on build in Eclipse

iloveeclipse:
For the rest of the code it isn't that important (and constant reading can affect performance), but every analysis execution should make sure it is working with up-to-date settings.

Additionally removed unneeded call to getUserPreferences() from readBugCollectionAndProject(), it was not needed anymore after one of the previous code changes.

Fixes https://github.com/spotbugs/spotbugs/issues/3740


iloveeclipse:
I believe I know better place where it could be fixed, will check tomorrow.

PR #3746, opened 2025-10-02
Remove Long Deprecated Priority enum from spotbugs annotations

hazendaz:
This annotation enum attribute was deprecated in 2011 [here](https://github.com/spotbugs/spotbugs/commit/d3b41945e03f854ce8613e1521b16230a62fb9c9).  It had a full replacement and usage in spotbugs only fell back to read that if the replacement 'Confidence' was not found.  Its been a full 14 years, its well past time to get rid of this.  The change log is updated to reflect what users should change to if found to be in that position.  This necessitates bumping up but considering minor to 4.10.x given the wide gap on when this really should have been removed.


ThrawnCA (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix Eclipse not always using latest preferences file state ([#3740](https://github.com/spotbugs/spotbugs/issues/3740)) 
 - Fix exception throw when singleton implementing Cloneable has no clone() method ([#3727](https://github.com/spotbugs/spotbugs/issues/3727)) 
 
+### Refactor
+- Remove deprecated 'Priority' annotation originally deprecated in 2011.  Switch to 'Confidence' for same behaviour. ([#3746](https://github.com/spotbugs/spotbugs/issues/3746))
```
JuditKnoll: There is an additional space between the two sentences.
The link mentions issue, but points to this PR, it finds the right place, but there is an unnecessary redirect, which could be easily avoided. I could not find a relevant issue which could be linked.
```suggestion
- Remove deprecated 'Priority' annotation originally deprecated in 2011. Switch to 'Confidence' for same behaviour. ([#3746](https://github.com/spotbugs/spotbugs/pull/3746))
```

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/DesireNoWarning.java
```diff
@@ -26,10 +26,6 @@
      */
     public String value();
 
-    /** @deprecated - use confidence instead */
-    @Deprecated
-    Priority priority() default Priority.LOW;
-
     /** Want no warning at this priority or higher */
```
JuditKnoll, line 33: Maybe this comment could also be updated? Or does Confidence appear anywhere as Priority (e.g. in some xml)?

CHANGELOG.md
JuditKnoll: As far as I can there are other Priorities in SpotBugs as well (e.g. in [BugInstance](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java#L115) or in [FilterBugsTask](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-ant/src/main/java/edu/umd/cs/findbugs/anttask/FilterBugsTask.java#L74)), which are not removed and should not be removed. I think it would be beneficial to mention that there are still some priorities used in the code.

JuditKnoll (commented):
I searched for "priority" in the project, and there were a lot of hits. There are still other fields, variables, settings, environment variables, and these should stay - at least these are not deprecated. I could not find any priority (field, function or class) which has `@Deprecated` annotation or `@deprecated` in the comments.

I think the documentation should be checked for usages of priority and updated if necessary. @hazendaz Did you check the docs?

spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/DesireNoWarning.java
JuditKnoll: Looked at it a bit more, please ignore this comment.

CHANGELOG.md
JuditKnoll: Looked at it a bit more, feel free to ignore this comment.

hazendaz: the extra space is english sentences so I don't think that should ever be a concern.  The other is fine, I'll fix both for you tough.

as to looking through docs, there are parts to fix but it requires documentation rebuild which is manual and likely not ran since many years back as its listing docker from 2017 timeframe, implies no devs use windows like I do (mac guy did the work..), and it simply doesn't work currently.  I'm going to try to resolve that but if I cannot, I think we simply create an issue to restore working docs and I just stage a branch that contains proposed cleanup in docs.  That way this can move forwards but at any direction, I'll have patch as noted for changelog coming up before it can be merged.

JuditKnoll: Thank you for the fix :) 

Thanks for the insight with the docs as well. To tell the truth I haven't really checked the docs in detail and haven't thought that it's this complex. The readthedocs is generated for every PR regularly and that seems to be quite up-to-date to me.
Fixing the whole docs is definitely out-of-scope of this PR, I only meant the mention of `priority`, but that can be also left to another one.

hazendaz:
I'll get a separate draft pr up on docs part where I've gotten to before I merge this in.

JuditKnoll:
IMO this one could be merged. @hazendaz is there something you still want to add to this or are you waiting for someone to review it?

hazendaz:
Hi @JuditKnoll, only docs which will be separate later really needed addressed as previously noted but will merge shortly as noted this is ready and aligning with cleanup we are doing that is long overdue.

Issue #3753, opened 2025-10-06
DetectorFactory.enabledButNonReporting() should not be used


iloveeclipse:
Follow up on https://github.com/spotbugs/spotbugs/pull/3709#issuecomment-3360757613.
`DetectorFactory.enabledButNonReporting()` is instance related but the instances of DetectorFactory are used in static context.
Because of that, SpotBugs tasks running in parallel and using different configurations might silently & randomly produce unexpected results.

This affects SpotBugs if it executed in parallel in same JVM, like in Eclipse where build can run on multiple unrelated projects.

I brief analysis shows this can be fixed by removing `DetectorFactory.getPriorityAdjustment()` & `DetectorFactoryChooser` and changing `BugInstance.adjustForDetector(String)` and `BugInstance` creation so that the bugs from "enabled but non reporting" factories are filtered out at  `PriorityAdjuster.adjustPriority(BugInstance)`.

This is not a regression but a very old code/behavior since day one.

I will try to provide a PR.

hazendaz:
@iloveeclipse Would I expect to have a number of new bugs added that were not there before?  Both choose visitors and omit visitors tests in the maven plugin integration suite are now failing.  Normally all the integration tests just run so I'm thinking this one noting possible behaviour difference could be what did it.  However, I get the same difference with single module run.

Here is test before (note this is without images)
[choose-visitors-4.9.6.html](https://github.com/user-attachments/files/22915785/choose-visitors-4.9.6.html)

Here is the test after (note this is without images)
[choose-visitors-4.9.7.html](https://github.com/user-attachments/files/22915786/choose-visitors-4.9.7.html)

This change seems most likely to be involved.

Integration test failures seen [here](https://github.com/hazendaz/spotbugs-maven-plugin/actions/runs/18512337068/job/52755789107)

iloveeclipse:
Could it be, maven plugin uses its own BugReporter, which isn't part of SpotBugs core?
The bugs from "disabled" detectors are filtered there via PriorityAdjuster, and by default custom BugReporter would need to provide its own PriorityAdjuster. I will check later today, I've never looked into maven plugin code before.

iloveeclipse:
@hazendaz : what we need is something like
https://github.com/spotbugs/spotbugs/blob/a3667d7b0de3545d236e7cfdc90ce8a669bb6111/spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java#L69-L96

especially 

[...]

PR #3754, opened 2025-10-06
Don't modify DetectorFactory state during analysis

iloveeclipse:
SpotBugs creates new `ExecutionPlan` for analysis, however each `ExecutionPlan` uses same `DetectorFactory` instances created from static `Plugin.getDetectorFactories()` context and the state of the instances could be modified during analysis, affecting analysis results.

`DetectorFactory` state relied on `enabledButNonReporting` flag and could lead to unpredictable analysis results for multiple SpotBugs analysis tasks running in parallel.

If one thread would call `factory.getPriorityAdjustment()` while creating a `BugInstance`, other could change this priority by calling `setEnabledButNonReporting()` from `ExecutionPlan.build()`.

The first problem was that `ExecutionPlan.build()` unconditionally sets `factory.setEnabledButNonReporting(false)` at very beginning, which alone could affect other execution results.

Later on, `ExecutionPlan.build()` changed factory state depending on the results of `DetectorFactoryChooser.choose()`. `DetectorFactoryChooser`is created in `FindBugs2.createExecutionPlan()`, which is based on `FindBugs.isDetectorEnabled()` result which depends on `UserPreferences`. Again, this could lead to unexpected results if the settings used were different.

Removed `getPriorityAdjustment()` and `setEnabledButNonReporting()` from `DetectorFactory`, `adjustForDetector()` from `BugInstance` and moved the related functionality to the `PriorityAdjuster.adjustForDetector()`.

Fixes https://github.com/spotbugs/spotbugs/issues/3753

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix exception throw when singleton implementing Cloneable has no clone() method ([#3727](https://github.com/spotbugs/spotbugs/issues/3727)) 
 - Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
 - Documentation of -adjustPriority parameter
+- Functionality from DetectorFactory setEnabledButNonReporting(), getPriorityAdjustment() methods and BugInstance.adjustForDetector() is deprecated and moved to PriorityAdjuster ([#3753](https://github.com/spotbugs/spotbugs/issues/3753))
 
 ### Changed
 - Support for fully qualified class names for detectors in -adjustPriority parameter
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
@@ -19,18 +19,30 @@
 
 package edu.umd.cs.findbugs;
 
-import java.nio.charset.StandardCharsets;
-import java.util.*;
-import java.util.stream.Collectors;
-
 import static java.util.Objects.requireNonNull;
 
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -51,6 +63,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import edu.umd.cs.findbugs.ba.AccessMethodDatabase;
+import edu.umd.cs.findbugs.ba.AccessMethodDatabase.AccessMethodLocation;
 import edu.umd.cs.findbugs.ba.AnalysisContext;
 import edu.umd.cs.findbugs.ba.CFGBuilderException;
 import edu.umd.cs.findbugs.ba.ClassContext;
@@ -63,7 +76,6 @@ import edu.umd.cs.findbugs.ba.OpcodeStackScanner.UnreachableCodeException;
 import edu.umd.cs.findbugs.ba.XFactory;
 import edu.umd.cs.findbugs.ba.XField;
 import edu.umd.cs.findbugs.ba.XMethod;
-import edu.umd.cs.findbugs.ba.AccessMethodDatabase.AccessMethodLocation;
 import edu.umd.cs.findbugs.ba.bcp.FieldVariable;
 import edu.umd.cs.findbugs.ba.vna.ValueNumberSourceInfo;
 import edu.umd.cs.findbugs.bytecode.MemberUtils;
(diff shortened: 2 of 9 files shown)


iloveeclipse:
If no one has objections, I would like to merge this PR soon.

spotbugs/src/main/java/edu/umd/cs/findbugs/PriorityAdjuster.java
```diff
@@ -79,11 +86,42 @@ BugInstance adjustPriority(BugInstance bugInstance) {
...
+    /**
+     * Adjust priority if the factory was forcibly enabled by {@link ExecutionPlan} although the user did not select it.
+     *
+     * @param bugInstance
+     * @return adjusted bug priority
+     */
+    public int adjustForDetector(BugInstance bugInstance) {
+        DetectorFactory factory = bugInstance.getDetectorFactory();
+        if (factory != null) {
+            if (factoryChooser != null && !factoryChooser.wasFactoryEnabled(factory)) {
```
ThrawnCA: Could be combined with the previous condition.

iloveeclipse: Done

spotbugs/src/main/java/edu/umd/cs/findbugs/DetectorFactory.java
```diff
@@ -293,22 +291,6 @@ public boolean isDefaultEnabled() {
...
-     *
-     * @return the priority adjustment
-     */
-    public int getPriorityAdjustment() {
-        if (enabledButNonReporting) {
-            return 100;
-        }
-        return 0;
-    }
-
```
JuditKnoll, line 311: These two functions could be used outside of SpotBugs, e.g. in a plugin. (I haven't checked it.) Please mark them as deprecated instead of immediately deleting them, and mention it in the changelog as well.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -243,29 +255,17 @@ public Object clone() {
...
-            adjustForDetector(detectorName);
+            rememberDetectorFactory(detectorName);
         }
 
     }
 
     /**
      * @param detectorName
      */
-    public void adjustForDetector(String detectorName) {
```
JuditKnoll, line 256: This functions could be used outside of SpotBugs, e.g. in a plugin. (I haven't checked it.) Please mark it as deprecated instead of immediately deleting it, and mention it in the changelog as well.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix exception throw when singleton implementing Cloneable has no clone() method ([#3727](https://github.com/spotbugs/spotbugs/issues/3727)) 
 - Fix for missing -adjustPriority parameter in Eclipse preferences ([#3687](https://github.com/spotbugs/spotbugs/issues/3687))
 - Documentation of -adjustPriority parameter
+- DetectorFactory.enabledButNonReporting() should not be used ([#3753](https://github.com/spotbugs/spotbugs/issues/3753))
```
JuditKnoll: Can you please elaborate a bit more, mentioning the function removal and maybe the impact?

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/DetectorFactory.java
iloveeclipse: "setEnabledButNonReporting" matches: https://github.com/search?q=.setEnabledButNonReporting&type=code&p=1

"getPriorityAdjustment" matches: https://github.com/search?q=.getPriorityAdjustment+language%3AJava&type=code&l=Java&p=1

All seem not relevant or just copies of same code in spotbugs/findbugs forks.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
iloveeclipse: "adjustForDetector" matches: https://github.com/search?q=adjustForDetector+language%3AJava&type=code&l=Java&p=1

All seem not relevant, just copies of same code in spotbugs/findbugs forks.

CHANGELOG.md
iloveeclipse: Will do.

spotbugs/src/main/java/edu/umd/cs/findbugs/DetectorFactory.java
JuditKnoll: Indeed, these hits don't seem to be relevant.
However there should be much more hits even on the public github as there are more forks of SpotBugs, but even if these searches contain everything from public github, there could be private repos, or GitLab or even private GitHub servers.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
iloveeclipse: I've restored and deprecated methods

CHANGELOG.md
iloveeclipse: Done

spotbugs/src/main/java/edu/umd/cs/findbugs/DetectorFactory.java
iloveeclipse: I've restored and deprecated methods.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -243,29 +255,21 @@ public Object clone() {
...
-            adjustForDetector(detectorName);
+            rememberDetectorFactory(detectorName);
         }
 
     }
 
+    @Deprecated(forRemoval = true)
+    public void adjustForDetector(String detectorName) {
+    }
+
```
JuditKnoll, line 268: If it's used it would break it only runtime, not compile-time, and could cause some mysterious issues. Can you please at least set the `detectorFactory` field here (e.g. call the `rememberDetectorFactory()` function)?

iloveeclipse: Done

JuditKnoll (approved):

PR #3757, opened 2025-10-07
Remove some long deprecated methods and a field

JuditKnoll:
This PR removes some long deprecated methods and a field, these have no other uses or only minimal usage, and deleting them is really straight-forward, so I done it in one PR.

----

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,18 +1,17 @@
-Before opening a 'bug'
-
-* Search existing issues and pull requests to see if the issue was already discussed.
-* Check our discussions to see if the issue was already discussed.
-* Check for specific project we support to raise the issue on, under [spotbugs](https://github.com/spotbugs)
-* Do not open intellij plugin issues here, open them at [intellij-plugin](https://github.com/JetBrains/spotbugs-intellij-plugin)
-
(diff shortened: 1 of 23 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/XFactory.java
```diff
@@ -227,31 +227,6 @@ private boolean isCalledDirectlyOrIndirectly(@CheckForNull ClassDescriptor clazz
...
-     * @deprecated This method does not work as expected. Will be removed in 5.x release.
-     */
-    @Deprecated
-    public boolean nameAndSignatureIsCalled(XMethod m) {
-        return calledMethodSignatures.contains(getDetailedSignature(m));
-    }
-
-    private static String getDetailedSignature(XMethod m2) {
-        return m2.getName() + m2.getSignature() + m2.isStatic();
-    }
```
JuditKnoll, line 240: However this method is not deprecated, it's private and its only usage is in the deleted deprecated method above.

JuditKnoll:
I searched for all the functions (and field) deleted in this PR in both FindSecBugs' and Fb-Contrib's current master. Could not find any usage of any of it.

ThrawnCA (approved):

hazendaz (approved):

PR #3759, opened 2025-10-08
DOC: visitors reference for omitVisitors (issue #1454)

itssaharsh:
This PR adds a small documentation scaffold and a helper script to generate a visitors reference for SpotBugs. It addresses issue #1454 by providing a starting point for a document that lists visitors/detectors, short descriptions and suggested bug-pattern identifiers.

Files added:
- spotbugs-contribution/docs/visitors-reference.rst — includes auto-generated visitor list
- spotbugs-contribution/generated/visitors.inc — auto-generated list-table fragment (Javadoc descriptions and heuristic pattern tokens)
- spotbugs-contribution/scripts/ExtractVisitors.java — helper to scan the source tree and regenerate the include
- spotbugs-contribution/README.md — usage notes and manual steps

Notes for reviewers:
- The generated include contains many placeholder (TBD) descriptions. Manual review is required to:
  - refine descriptions (prefer the first Javadoc paragraph or README snippets)
  - map each visitor to its canonical bug-pattern identifiers
- The helper is heuristic and intended to accelerate documentation work. I can update it to cross-reference `docs/bugDescriptions.rst` to pull exact pattern IDs, if desired.

Follow-ups:
- I can run the helper to produce a more complete mapping and submit a follow-up commit, or we can collaborate on filling the descriptions in this PR.
- Thanks for the feedback everyone! I've updated the PR based on your suggestions:

**What I changed:**
- Created a new extractor that uses `findbugs.xml` instead of guessing from class names (thanks @JuditKnoll!)  
- Cleaned up the redundant string matching logic
- Added explanation for why the generated file is committed (thanks @iloveeclipse!)

The XML-based approach gives us the actual bug pattern IDs and detector metadata directly from SpotBugs' config, so it should be much more accurate.

Let me know what you think of this version!

### DIFF
--- a/spotbugs-contribution/scripts/ExtractVisitors.java
+++ b/spotbugs-contribution/scripts/ExtractVisitors.java
@@ -52,8 +52,8 @@ public class ExtractVisitors {
                         if (parts.length >= 2) {
                             String after = parts[1].trim();
                             String className = after.split("\\s+|\\{|\\(|;")[0];
-                            // Heuristic: class name contains Detector or Visitor or Pattern or starts with Find
-                            if (className.endsWith("Detector") || className.contains("Detector") || className.endsWith("Visitor") || className.contains("Visitor") || className.contains("Pattern") || className.startsWith("Find")) {
+                            // Heuristic: class name contains Detector, Visitor, Pattern or starts with Find
+                            if (className.contains("Detector") || className.contains("Visitor") || className.contains("Pattern") || className.startsWith("Find")) {
                                 String rel = root.relativize(p).toString().replace('\\', '/');
 
                                 // Try to extract a Javadoc comment immediately above the declaration
--- a/spotbugs-contribution/README.md
+++ b/spotbugs-contribution/README.md
@@ -5,6 +5,8 @@ This small scaffold helps extract a list of SpotBugs visitor/detector classes an
 Files added:
 - `docs/visitors-reference.rst` - starter document (generated or edited)
 - `scripts/ExtractVisitors.java` - small scanner that finds likely detector classes in a SpotBugs source tree and emits an RST table to stdout
+- `scripts/ExtractVisitorsImproved.java` - improved version that uses findbugs.xml as authoritative source
+- `generated/visitors.inc` - auto-generated include file (committed for convenience, can be regenerated)
 
 Quick usage
 
@@ -21,4 +23,15 @@ Quick usage
 
 Notes
 
-- The scanner is heuristic-based: it looks for classes whose name or inheritance suggest they are detectors. Manual review is still required to add accurate descriptions and bug pattern mappings.
+- The original scanner (`ExtractVisitors.java`) is heuristic-based: it looks for classes whose name or inheritance suggest they are detectors. 
+- The improved version (`ExtractVisitorsImproved.java`) uses `findbugs.xml` as the authoritative source for all detector classes.
+- Manual review is still required to add accurate descriptions and bug pattern mappings.
+
+## Generated Files Policy
+
+The `generated/visitors.inc` file is committed to git for convenience and to show the current state of the documentation. This follows common practices in documentation projects where generated content is committed when:
+- It provides immediate value to reviewers and users
+- The generation process requires specific build environment setup
+- The generated content is relatively stable and doesn't change frequently
+
+The file can always be regenerated using the scripts, but committing it allows others to see the current documentation state without running the generation tools.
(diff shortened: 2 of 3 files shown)


spotbugs-contribution/generated/visitors.inc
iloveeclipse, line 1: If the file is re-generated, does it need to be added to git? Pardon if the question is stupid, I have no idea abut "rst" "inc" stuff.

JuditKnoll:
Shouldn't [findbugs.xml](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugs.xml#L298) contain all visitors as Detectors? Simply going through the not hidden detectors seems like a reliable option to me. Or do I miss something?

spotbugs-contribution/scripts/ExtractVisitors.java
```diff
@@ -0,0 +1,129 @@
...
+                    String line = lines.get(i).trim();
+                    // look for a class or interface declaration
+                    if (line.matches(".*\\bclass\\b.*") || line.matches(".*\\binterface\\b.*")) {
+                        // extract token after 'class' or 'interface'
+                        String[] parts = line.split("\\bclass\\b|\\binterface\\b");
+                        if (parts.length >= 2) {
+                            String after = parts[1].trim();
+                            String className = after.split("\\s+|\\{|\\(|;")[0];
+                            // Heuristic: class name contains Detector or Visitor or Pattern or starts with Find
+                            if (className.endsWith("Detector") || className.contains("Detector") || className.endsWith("Visitor") || className.contains("Visitor") || className.contains("Pattern") || className.startsWith("Find")) {
```
JuditKnoll: Checking both endsWith and contains with "Detector" and "Visitor" doesn't seem really useful, as the latter is always true when the first one is.

spotbugs-contribution/generated/visitors.inc
itssaharsh: @iloveeclipse Good question! I wasn't sure about this either initially. 

I decided to commit the `.inc` file because it shows what the generated documentation actually looks like without having to run the script yourself. Some doc projects do this when the generation step is a bit involved (needs the full SpotBugs source tree in this case).
I added a note in the README explaining this - basically it's there for convenience so you can see the current state, but it can always be regenerated if needed.
Let me know if you think it should be handled differently!

spotbugs-contribution/scripts/ExtractVisitors.java
itssaharsh: @JuditKnoll You're absolutely right! I was overthinking this.
I took your suggestion and created a new version (`ExtractVisitorsImproved.java`) that parses `findbugs.xml` directly. It's much cleaner and gets all the info we need - the detector classes, their bug patterns, speed settings, everything. No more guessing from class names.
Also fixed that redundant logic you caught - yeah, if `contains("Detector")` is true, then `endsWith("Detector")` is just extra work.
The XML approach is definitely better. Thanks for pointing me in the right direction!

JuditKnoll:
I just looked at https://github.com/spotbugs/spotbugs/issues/1454, the relevant issue, and it looks like it was already solved by https://github.com/spotbugs/spotbugs/pull/1651, just the issue wasn't set to closed/fixed. See [detectors section in the docs](https://spotbugs.readthedocs.io/en/stable/detectors.html).

itssaharsh:
> I just looked at #1454, the relevant issue, and it looks like it was already solved by #1651, just the issue wasn't set to closed/fixed. See [detectors section in the docs](https://spotbugs.readthedocs.io/en/stable/detectors.html).

@JuditKnoll Oops, you're absolutely right! Should have checked existing docs first. 

Looking at that detectors page now - yep, that's exactly what #1454 was asking for. My bad for jumping in without doing proper research first! 😅

Should I just close this PR since the work's already done? Or do you think there's any value in keeping the XML extractor script around for maintaining that documentation? Though honestly, if it's working fine as-is, probably not worth the overhead.

JuditKnoll:
> > I just looked at #1454, the relevant issue, and it looks like it was already solved by #1651, just the issue wasn't set to closed/fixed. See [detectors section in the docs](https://spotbugs.readthedocs.io/en/stable/detectors.html).
> 
> @JuditKnoll Oops, you're absolutely right! Should have checked existing docs first.
> 
> Looking at that detectors page now - yep, that's exactly what #1454 was asking for. My bad for jumping in without doing proper research first! 😅
> 
> Should I just close this PR since the work's already done? Or do you think there's any value in keeping the XML extractor script around for maintaining that documentation? Though honestly, if it's working fine as-is, probably not worth the overhead.

Thank you for your work. Feel free to close it. There are still plenty of issues to choose from, if you have interest. Just please check first that they are still relevant 😄 
If you find an already solved issue in the mean time, please a leave a comment there, so we can close it.

Issue #3774, opened 2025-10-15
Maven plugin reporting issue if "adjustPriority" is not set


iloveeclipse:
> Would I expect to have a number of new bugs added that were not there before?  Both choose visitors and omit visitors tests in the maven plugin integration suite are now failing.  Normally all the integration tests just run so I'm thinking this one noting possible behaviour difference could be what did it.  However, I get the same difference with single module run.
> 
> Here is test before (note this is without images)
> [choose-visitors-4.9.6.html](https://github.com/user-attachments/files/22915785/choose-visitors-4.9.6.html)
> 
> Here is the test after (note this is without images)
> [choose-visitors-4.9.7.html](https://github.com/user-attachments/files/22915786/choose-visitors-4.9.7.html)
> 
> This change seems most likely to be involved.
> 
> Integration test failures seen [here](https://github.com/hazendaz/spotbugs-maven-plugin/actions/runs/18512337068/job/52755789107) 

 _Originally posted by @hazendaz in [#3753](https://github.com/spotbugs/spotbugs/issues/3753#issuecomment-3404127409)_

Regression from https://github.com/spotbugs/spotbugs/issues/3753.

I couldn't find any BugReporter in maven plugin code, so I assume all output is generated somehow by `FindBugs2` created via `AntBuilder` here:

https://github.com/spotbugs/spotbugs-maven-plugin/blob/8bedb8ee38971883cd4fcbd2801bc2a959907bc4/src/main/groovy/org/codehaus/mojo/spotbugs/SpotBugsMojo.groovy#L1157C50-L1158

This should call `FindBugs2.main()` and this should call `TextUICommandLine` and it should actually be configured here

https://github.com/spotbugs/spotbugs/blob/a3667d7b0de3545d236e7cfdc90ce8a669bb6111/spotbugs/src/main/java/edu/umd/cs/findbugs/TextUICommandLine.java#L665-L667
 
I believe `-adjustPriority` is not (always) given by maven. That would result in not created `PriorityAdjuster` instance. As a workaround, one can try to use something like `edu.umd.cs.findbugs.detect.CheckCalls=+0` value for `-adjustPriority` argument (shouldn't change anything except that `PriorityAdjuster` would be created).

I will try to provide a fix for `TextUICommandLine` now.

hazendaz:
To run the tests simply check out master from spotbugs maven plugin and use the first Integration Test noted on the Readme. The tests use source from spotbugs so the first one is remotely pulling that source.

[...]

PR #3775, opened 2025-10-15
Unconditionally initialize PriorityAdjuster for AbstractBugReporter

iloveeclipse:
After commit f03df3d019fba0c42cf3c604231104f2ca164f70 `PriorityAdjuster` is also used to disable/enable factories (by raising their reporting priority) which was implemented as a "static global" feature of `DetectorFactory` before.

However, this requires that we *always* initialize `PriorityAdjuster` in `AbstractBugReporter`, independently if `-adjustPriority` is used or not.

Fixes https://github.com/spotbugs/spotbugs/issues/3774

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+### Fixed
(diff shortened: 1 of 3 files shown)


iloveeclipse:
@hazendaz, could you please check if that fixes the maven plugin? I have no clue how it works.

hazendaz:
thanks, checking now.

hazendaz:
This didn't resolve the issue.  Same results using 4.10.0-SNAPSHOT of spotbugs.

iloveeclipse:
> This didn't resolve the issue. Same results using 4.10.0-SNAPSHOT of spotbugs.

Please add some info on https://github.com/spotbugs/spotbugs/issues/3774 how to run these tests. I have zero experience with maven plugins. Is there any possibility to bisect the test with different commits of SpotBugs to identify which one causes the failure?

Issue #3721, opened 2025-09-19
UnionResults calculates incorrect totals


danielrwhite74:
When UnionResults is used to combine two or more spotbugs reports the package and project totals (size and error counts) are not correct. This is caused by the findings being counted twice during the merge process: first when the bug instance is added to the results bug collection and a second time when the project stats are merged.

From the `merge` method in `UnionResults.java`:
```
for (BugInstance bugInstance : from.getCollection()) {
    if (hashes == null || hashes.add(bugInstance.getInstanceHash())) {
        into.add(bugInstance); //<--- increments the collection total counts
    }
}
ProjectStats stats = into.getProjectStats();
ProjectStats stats2 = from.getProjectStats();
stats.addStats(stats2); //<--- adds the totals a second time
```

iloveeclipse:
Thanks for reporting your findings. Do you want to contribute a fix?

danielrwhite74:
I don't currently have dev environment setup to run gradle builds but I can look into setting up one.

PR #3777, opened 2025-10-16
fixes incorrect stats when unioning results (#3721)

dwhite996:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,12 @@ This is the changelog for SpotBugs. This follows [Keep a Changelog v1.0.0](http:
 Currently the versioning policy of this project follows [Semantic Versioning v2.0.0](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased - 2025-??-??
+### Fixed
+- Fix incorrect bug counts and sizes when unioning reports ([#3721](https://github.com/spotbugs/spotbugs/issues/3721))
 
 ## 4.9.8 - 2025-10-18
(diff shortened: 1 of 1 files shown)


hazendaz:
Can you move the changelog entry up as we release the set you have that in yesterday?  In the unreleased section just add entry for the section to your change.

hazendaz:
@dwhite996 I'm not familiar with the code but fixing there is in the main, wouldn't the 'union' method have the same issue if merge is the cause?

dwhite996:
> @dwhite996 I'm not familiar with the code but fixing there is in the main, wouldn't the 'union' method have the same issue if merge is the cause?

Good point. The fix would be better in the merge method since both main and union call merge.

spotbugs/src/main/java/edu/umd/cs/findbugs/workflow/UnionResults.java
hazendaz, line 112: @dwhite996 For stats here, do we even these lines 112 through 114 since those cause the duplication under the specific condition intended to be fixed?  Since recompute is done instead, I think you would just want to move the lines you have created to this location and delete the old.  That seems more logical to me than trying to update the stats only to recompute them afterwards.

dwhite996: The addStats method adds tracking of the packages in the bug collection while recomputeFromComponents just recalculates the totals (using the packages added by addStats) so both are needed.

PR #3156, opened 2024-10-10
Add support for Jakarta's Nonnull and Nullable

pkini07:
This adds support for `jakarta.annotation.Nonull` and `jakarta.annotation.Nullable`.

PR #3780, opened 2025-10-16
 Add support for Jakarta's Nonnull and Nullable (continued work #3156 by pkini07)

stephanroch:
This PR continues the work done by @pkini07 for the support of Jakarta's Nonnull and Nullable annotations by fixing isues found in #3156. The original developer is not responding to mails and other contact requests and does not solve the issue. @JuditKnoll suggested to open a new PR and to mention the original autor.

- [X] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code (updated PR from #3156 to #3780 already)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## Unreleased - 2025-??-??
 ### Added
 - Recognize `jakarta.annotation.Nonnull` and `jakarta.annotation.Nullable` ([#3780](https://github.com/spotbugs/spotbugs/pull/3780))
+### Fixed
+- Fix incorrect bug counts and sizes when unioning reports ([#3721](https://github.com/spotbugs/spotbugs/issues/3721))
 
 ## 4.9.8 - 2025-10-18
 ### Fixed
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ import org.gradle.api.publish.maven.MavenArtifact;
 import org.gradle.api.tasks.util.PatternFilterable;
 
 plugins {
-  id 'org.sonarqube' version '7.0.0.6105'
+  id 'org.sonarqube' version '7.0.1.6134'
   id 'org.gradle.crypto.checksum' version '1.4.0'
   id 'com.github.spotbugs' version '6.4.4'
   id 'io.github.gradle-nexus.publish-plugin' version '2.0.0'
--- a/gradle/checkstyle.gradle
+++ b/gradle/checkstyle.gradle
@@ -2,7 +2,7 @@
 apply plugin: 'checkstyle'
 
 checkstyle {
-  toolVersion = '12.1.0'
+  toolVersion = '12.1.1'
   ignoreFailures = false
   configFile = file("${rootDir}/spotbugs/etc/checkstyle.xml") // TODO : This config file is lame and should be moved out...
 }
--- a/spotbugs-tests/build.gradle
+++ b/spotbugs-tests/build.gradle
@@ -20,7 +20,7 @@ dependencies {
   implementation 'org.apache.ant:ant:1.10.15'
   implementation libs.log4j.core
   implementation libs.log4j.slf4j2.impl
-  implementation 'com.google.errorprone:error_prone_annotations:2.42.0'
+  implementation 'com.google.errorprone:error_prone_annotations:2.43.0'
   implementation files(project(':spotbugs').sourceSets.gui.output)
   testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/workflow/UnionResults.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/workflow/UnionResults.java
@@ -120,6 +120,9 @@ public class UnionResults {
         for (AnalysisError error : from.getErrors()) {
             into.addError(error);
         }
+
+        into.getProjectStats().getPackageStats().forEach(s -> s.recomputeFromClassStats());
+        into.getProjectStats().recomputeFromComponents();
     }
 
     public static void main(String[] argv) throws IOException {
@@ -140,6 +143,7 @@ public class UnionResults {
             System.exit(1);
             return;
         }
+
         results.setWithMessages(commandLine.withMessages);
         if (commandLine.outputFile == null) {
             results.writeXML(System.out);


volkanto (approved):
🤟

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ### Added
 - Added the unnecessary annotation to the `US_USELESS_SUPPRESSION_ON_*` messages ([#3395](https://github.com/spotbugs/spotbugs/issues/3395))
 - Multi-threaded code checks can be skipped with `@NotThreadSafe` ([#3390](https://github.com/spotbugs/spotbugs/issues/3390))
+- Recognize `jakarta.annotation.Nonnull` and `jakarta.annotation.Nullable` ([#3780](https://github.com/spotbugs/spotbugs/pull/3780))
```
JuditKnoll: Can you please merge master and move this to the unreleased part? (Most likely there will be a merge conflict to resolve.)

hazendaz:
@stephanroch Please move the change log item up to the current location at top under unreleased.  If there is not a section for it yet, please add the section as needed.  Rest looks ok, once that is fixed, I'll execute the build and if all good will merge in for the next release.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/NullnessAnnotation.java
```diff
@@ -39,6 +39,7 @@ boolean match(@DottedClassName String className) {
             return "android.support.annotation.Nullable".equals(className)
```
ThrawnCA, line 39: This really should be a `static final` Collection or array.

hazendaz: While I agree, I think that is out of scope for this PR to change all that especially if we cut a quick release.  If you can, open up an issue related to better handling of such items with link to the code, I can look at taking care of that at a later time or maybe someone else gets it.  For now though scope on this PR I think should just be around the addition of this as long missing.

hazendaz:
actually I'll move change log so I can launch this now.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.nio.file.Paths;
+
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Check if Nonnull annotation from jakarta.annotations is detected.
+ *
+ * @author pkini07
```
hazendaz: I'll accept as is but adding author is extremely legacy and anti pattern.  It should be removed.  I know the repo is full of those so will do at some point separately.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@ExtendWith(SpotBugsExtension.class)
+class JakartaNonnullTest {
+    @Test
+    void checkedJakartaNonnullReturn_isOk(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/CheckedJakartaNonnullReturn.class"));
```
hazendaz: We are java 11, so Paths.get should be Path.of instead.  I can fix that later.

hazendaz:
If build is fine, I'll merge and correct the issues I noted separately through local as I suspect we have a lot of those issues in the code rather than what is here so I'm fine with that as-is.  This is just really long overdue and personally I started using it recently as jspecify has issues with guava and nullaway so that one worked better with less excessive other warnings so having spotbugs too recognize it is a good thing.  For anyone that asks, did javax one have this too, answer is no, it was added in the jakarta one likely to do to attempts to get away from jsr305.  The 'javax' one that did exist is in jsr305.

hazendaz:
For spotbugs maintainers, there was a flag disabled on the project here that was preventing us forcing the branch merge to update on stale branches.  That I think was a typical ask of users to merge in master changes, everyone should see a button now to do that.  I applied that here to get the change log then up to date.

hazendaz:
@stephanroch Didn't compile, please review and correct.  Also look at my notes, go ahead and clean that stuff up to since this is not yet working order. Thanks.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
```diff
@@ -0,0 +1,40 @@
...
+        assertThat(bugCollection, containsExactly(0, bug()));
+    }
+
+    @Test
+    void uncheckedJakartaNonnullReturn_isDetected(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/UncheckedJakartaNonnullReturn.class"));
+        assertThat(bugCollection, containsExactly(1, bug()));
```
JuditKnoll: Can you please add one more assert with the details of the expected bug? (Preferable for every expected bug, so in the other tests as well.)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Check if Nonnull annotation from jakarta.annotations is detected.
+ *
+ * @author pkini07
+ */
+@ExtendWith(SpotBugsExtension.class)
+class JakartaNonnullTest {
```
JuditKnoll: Instead of using `@ExtendWith(SpotBugsExtension.class)`, you can simply extend `AbstractIntegrationTest` ([src](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java)), which has some assert functions for easy and consistent handling of bug checks. It's okay like this as well, but that's a bit more streamlined.
If you decide to migrate to that, there are some things which needs to be changed:
- the tests have no parameter, 
- the `AbstractIntegrationTest` provides the `performAnalysis`, which is currently called on the SpotBugsRunner instance,
- `AbstractIntegrationTest`'s `performAnalysis` function parameters are strings, not Paths and it can be simply the slashed name, so `../spotbugsTestCases/build/classes/java/main/` prefix is not necessary.
- the bugCollection can be accessed via `getBugCollection()`, but it's not even necessary if `AbstractIntegrationTest`'s assert functions are used.

spotbugsTestCases/src/java/nullnessAnnotations/CheckedJakartaNonnullReturn.java
```diff
@@ -0,0 +1,12 @@
+import jakarta.annotation.Nonnull;
+
+public class CheckedJakartaNonnullReturn {
```
JuditKnoll, line 5: Since there are multiple sample test classes, can you please move them into a package? You can use `nonnull` ([src](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java/nonnull)) or  `nullnessAnnotations` ([src](https://github.com/spotbugs/spotbugs/tree/master/spotbugsTestCases/src/java/nullnessAnnotations)), which are quite similar.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author pkini07
+ */
+@ExtendWith(SpotBugsExtension.class)
+class JakartaNullabilityTest {
+
+    @Test
+    void checkedJakartaNullableReturn_isOk(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/CheckedJakartaNullableReturn.class"));
```
JuditKnoll: When running this test, the following is written to the stderr:
```
  jakarta.annotation.Nullable
```

Please add the fakeAnnotation to the analysis or as an aux class path to resolve this issue. Take care, that the error contains the dotted class name, but the slashed class name is need with the `.class` suffix.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
```diff
@@ -0,0 +1,42 @@
...
+        assertThat(bugCollection, containsExactly(0, bug()));
+    }
+
+    @Test
+    void uncheckedJakartaNullableReturn_isDetected(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/UncheckedJakartaNullableReturn.class"));
```
JuditKnoll: When running this test, the following is written to the stderr:
```
  jakarta.annotation.Nullable
```

Please add the fakeAnnotation to the analysis or as an aux class path to resolve this issue. Take care, that the error contains the dotted class name, but the slashed class name is need with the `.class` suffix.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
```diff
@@ -0,0 +1,40 @@
...
+    void checkedJakartaNonnullReturn_isOk(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/CheckedJakartaNonnullReturn.class"));
+        assertThat(bugCollection, containsExactly(0, bug()));
+    }
+
+    @Test
+    void uncheckedJakartaNonnullReturn_isDetected(SpotBugsRunner spotbugs) {
+        BugCollection bugCollection = spotbugs.performAnalysis(
+                Paths.get("../spotbugsTestCases/build/classes/java/main/UncheckedJakartaNonnullReturn.class"));
```
JuditKnoll: When running this test, the following is written to the stderr:
```
  jakarta.annotation.Nonnull
```

Please add the fakeAnnotation to the analysis or as an aux class path to resolve this issue. Take care, that the error contains the dotted class name, but the slashed class name is need with the `.class` suffix.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
JuditKnoll: When running this test, the following is written to the stderr:
```
  jakarta.annotation.Nonnull
```

Please add the fakeAnnotation to the analysis or as an aux class path to resolve this issue. Take care, that the error contains the dotted class name, but the slashed class name is need with the `.class` suffix.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static edu.umd.cs.findbugs.test.CountMatcher.containsExactly;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Check if Nullable annotation from jakarta.annotations is detected.
+ *
+ * @author pkini07
+ */
+@ExtendWith(SpotBugsExtension.class)
+class JakartaNullabilityTest {
```
JuditKnoll: The two tests are similar enough that they could be in one test. It's okay like this as well though.

JuditKnoll (changes requested):
Thank you @stephanroch for taking this up and for your work.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
stephanroch: I'm not sure if this is needed: The method bug() defined in both test classes contains already a matcher for the type of the bug. This should be detailed enough. What do you think?

spotbugsTestCases/src/java/nullnessAnnotations/CheckedJakartaNonnullReturn.java
stephanroch: I will move the two newly introduced classes to nullnessAnnotations.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
stephanroch: I will move two test cases from the other file to this one.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
stephanroch: Tricky, that the fakeAnnotation is not included as part of the classpath if I use AbstractIntegrationTest.performAnalysis.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
stephanroch: Is removed now and replaced by using AbstractIntegrationTest.assert*

stephanroch:
I fixed and modernized the tests and followed your suggestions. Thank you very much. I think you resolve your points after another round of reviews or is this left up to me?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
stephanroch: I left this note because the fame for this should go to the original author pkini07 and for future deletion in the whole repo.

hazendaz: no need, the issue tracking shows all that history and doesn't go away.  From a code perspective, its bad practice to add author tags in recent years for a number of reasons, just remove it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNullabilityTest.java
JuditKnoll: FakeAnnotation doesn't have a separate folder in the compiled code, it's inside the build/classes/java/main together with the files from src/java. So it works with this change 😄

CHANGELOG.md
```diff
@@ -49,7 +51,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ### Cleanup
 - S1481: Unused local variables should be removed ([#3654](https://github.com/spotbugs/spotbugs/pull/3654))
-- Moved test libraries to jakarta namespace including switching off jsr305 where possible for jakarta.annotatoin ([#3695](https://github.com/spotbugs/spotbugs/pull/3695))
+- Moved test libraries to jakarta namespace including switching off jsr305 where possible for jakarta.annotation ([#3695](https://github.com/spotbugs/spotbugs/pull/3695))
```
JuditKnoll, line 56: Good catch! Thanks. Updated this in the description of the [4.9.5](https://github.com/spotbugs/spotbugs/releases/tag/4.9.5) release.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
JuditKnoll: It's a good practice in the project to check both the number of relevant bugtypes (so tests show if a bug is reported multiple times or there is a false positive), and the details of each bug (so if something changes - e.g. a bug is reported at the wrong line - the tests show it immediately). Since these sample codes are quite simple and there is only one bug for each, it can be okay like this as well.

JuditKnoll (approved):
Thank you for your work!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/nullness/JakartaNonnullTest.java
stephanroch: This is fixed allready. Will close this as resolved.

stephanroch:
Is there anything left which should be done for this pull request by my side? There are some open comments, but I think this can go without changes to the scheduled 4.10.0. We are looking forward for this release. Please contact me if I could further assist pulling this in production. Thanks for the review and handling of this task.

volkanto:
@JuditKnoll any plans to merge this and make it available in the upcoming release? thanks in advance.

PR #3804, opened 2025-11-12
Add UnsafeDetector to report uses of Java's Unsafe classes

### LINKED ISSUE #3543: Proposal: Catch usages of sun.misc.Unsafe or jdk.internal.misc.Unsafe
I think that any use of `sun.misc.Unsafe` or `jdk.internal.misc.Unsafe` should be flagged as a [security concern](https://blogs.oracle.com/javamagazine/post/the-unsafe-class-unsafe-at-any-speed). Especially now that the newer, safer [Foreign Function and Memory API](https://docs.oracle.com/en/java/javase/21/core/foreign-function-and-memory-api.html) is available.

So I have started working on a [draft detector](https://github.com/thomasleplus/java-memory-safety/blob/main/spotbugs/src/main/java/org/leplus/spotbugs/UnsafeDetector.java) but before I create a PR, I wanted to check what the community thinks of such a rule?

Cheers,

Tom

thomasleplus:
This PR adds a new detector for Java. It detects uses of the `sun.misc.Unsafe` and `jdk.internal.misc.Unsafe` classes which are a [security concern](https://archive.is/AFw31). The newer, safer [Foreign Function and Memory API](https://docs.oracle.com/en/java/javase/21/core/foreign-function-and-memory-api.html) should be used instead if really needed.

Closes #3543

### DIFF
--- a/spotbugs/etc/findbugs.xml
+++ b/spotbugs/etc/findbugs.xml
@@ -482,7 +482,7 @@
           <Detector class="edu.umd.cs.findbugs.detect.Naming" speed="fast"
                     reports="NM_WRONG_PACKAGE,NM_WRONG_PACKAGE_INTENTIONAL,NM_VERY_CONFUSING,NM_VERY_CONFUSING_INTENTIONAL,NM_CONFUSING,NM_METHOD_CONSTRUCTOR_CONFUSION,NM_LCASE_HASHCODE,NM_LCASE_TOSTRING,NM_BAD_EQUAL,NM_CLASS_NAMING_CONVENTION,NM_FIELD_NAMING_CONVENTION,NM_METHOD_NAMING_CONVENTION,NM_CLASS_NOT_EXCEPTION,NM_SAME_SIMPLE_NAME_AS_SUPERCLASS,NM_SAME_SIMPLE_NAME_AS_INTERFACE"/>
           <Detector class="edu.umd.cs.findbugs.detect.ReadReturnShouldBeChecked" speed="fast"
-                    reports="RR_NOT_CHECKED,SR_NOT_CHECKED"/>
+                    reports="RR_NOT_CHECKED,SR_NOT_CHECKED,NCR_NOT_PROPERLY_CHECKED_READ"/>
           <Detector class="edu.umd.cs.findbugs.detect.SerializableIdiom" speed="fast"
                     reports="SE_NO_SERIALVERSIONID,SE_METHOD_MUST_BE_PRIVATE,WS_WRITEOBJECT_SYNC,RS_READOBJECT_SYNC,SE_INNER_CLASS,SE_BAD_FIELD,SE_BAD_FIELD_STORE,SE_NONLONG_SERIALVERSIONID,SE_NONSTATIC_SERIALVERSIONID,SE_NONFINAL_SERIALVERSIONID,SE_NO_SUITABLE_CONSTRUCTOR,SE_NO_SUITABLE_CONSTRUCTOR_FOR_EXTERNALIZATION,SE_READ_RESOLVE_MUST_RETURN_OBJECT,SE_TRANSIENT_FIELD_NOT_RESTORED,SE_TRANSIENT_FIELD_OF_NONSERIALIZABLE_CLASS,SE_BAD_FIELD_INNER_CLASS,SE_READ_RESOLVE_IS_STATIC,SE_PRIVATE_READ_RESOLVE_NOT_INHERITED,SE_PREVENT_EXT_OBJ_OVERWRITE"/>
           <Detector class="edu.umd.cs.findbugs.detect.StartInConstructor" speed="fast"
@@ -742,6 +742,7 @@
           <BugCode abbrev="VSC"/>
           <BugCode abbrev="DA"/>
           <BugCode abbrev="HSM"/>
+          <BugCode abbrev="NCR"/>
           <!-- Bug patterns -->
           <BugPattern abbrev="JUA" type="JUA_DONT_ASSERT_INSTANCEOF_IN_TESTS" category="BAD_PRACTICE" experimental="true" />
           <BugPattern abbrev="AA" type="AA_ASSERTION_OF_ARGUMENTS" category="BAD_PRACTICE" cweid="617"/>
@@ -1347,4 +1348,5 @@
           <BugPattern abbrev="US" type="US_USELESS_SUPPRESSION_ON_METHOD_PARAMETER" category="STYLE"/>
           <BugPattern abbrev="US" type="US_USELESS_SUPPRESSION_ON_PACKAGE" category="STYLE"/>
           <BugPattern abbrev="UNS" type="UNS_UNSAFE_CALL" category="SECURITY"/>
+          <BugPattern abbrev="NCR" type="NCR_NOT_PROPERLY_CHECKED_READ" category="BAD_PRACTICE" cweid="135"/>
 </FindbugsPlugin>
--- a/spotbugs/etc/messages.xml
+++ b/spotbugs/etc/messages.xml
@@ -974,7 +974,7 @@ malicious code.  </p>
     <Details>
 <![CDATA[
 <p> This detector looks for calls to InputStream.read() or InputStream.skip() where the
-return value is ignored.  </p>
+return value is ignored or improperly checked.</p>
 ]]>
     </Details>
   </Detector>
@@ -4763,6 +4763,11 @@ be confusing to users of this class.</p>
   causing the program to fail only sporadically.</p>
   <p>See <a href="https://cwe.mitre.org/data/definitions/252.html">CWE-252: Unchecked Return Value</a>.
 </p>
+
+  <p>Additionally, it is not sufficient to simply check for the end of the file (EOF)
+  as an indicator that the entire array has been filled. The return value of
+  <code>read()</code> must be verified to ensure that the correct number of bytes
+  have been read.</p>
 ]]>
     </Details>
   </BugPattern>
@@ -9849,6 +9854,26 @@ Using floating-point variables should not be used as loop counters, as they are 
     </Details>
   </BugPattern>
   
+  <BugPattern type="NCR_NOT_PROPERLY_CHECKED_READ">
+    <ShortDescription>Return value of read() is not properly checked, array may be partially filled</ShortDescription>
+    <LongDescription>The return value of method {2.nameAndSignature} is not properly checked in {1.name}, which may result in using a partially filled array.</LongDescription>
+    <Details>
+      <![CDATA[
+        <p>
+          When using the read() method to fill an array, the method may not always fill the entire array. This can occur when the input stream does not have enough bytes available to completely fill the array, leading to a partial read. This issue can result in unexpected behavior if the partially filled array is assumed to be fully populated.
+        </p>
+        <p>
+          It is crucial to verify the return value of the read() method, which indicates the number of bytes actually read, and to ensure that the array is fully filled if necessary. Failing to do so can result in incomplete data processing and potential vulnerabilities.
+        </p>
+        <p>
+          To fix this issue, check the return value of read() and handle cases where fewer bytes are read than expected. Consider looping over the read() call or using methods like readFully() to ensure that the array is filled completely.
+        </p>
+        <p>
+          See SEI CERT rule <a href="https://wiki.sei.cmu.edu/confluence/display/java/FIO10-J.+Ensure+the+array+is+filled+when+using+read%28%29+to+fill+an+array">FIO10-J. Ensure the array is filled when using read() to fill an array</a> for more information.
+        </p>
+      ]]>
+    </Details>
+  </BugPattern>
   <!--
   **********************************************************************
    BugCodes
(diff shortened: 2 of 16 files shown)


spotbugs/etc/findbugs.xml
```diff
@@ -1343,4 +1346,5 @@
           <BugPattern abbrev="US" type="US_USELESS_SUPPRESSION_ON_METHOD" category="STYLE"/>
           <BugPattern abbrev="US" type="US_USELESS_SUPPRESSION_ON_METHOD_PARAMETER" category="STYLE"/>
           <BugPattern abbrev="US" type="US_USELESS_SUPPRESSION_ON_PACKAGE" category="STYLE"/>
+          <BugPattern abbrev="UNS" type="UNS_UNSAFE_CALL" category="SECURITY"/>
```
JuditKnoll, line 1350: As far as I can see, there is no `"UNS"` bug code yet. Can you please add it? The others are around line 715-745 in this file.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/UnsafeDetectorTest.java
```diff
@@ -0,0 +1,97 @@
...
+import edu.umd.cs.findbugs.test.matcher.BugInstanceMatcherBuilder;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/** Tests for the {@link UnsafeDetector}. */
+class UnsafeDetectorTest {
```
JuditKnoll: This test class would be more simple and uniform with the codebase if it extended the `AbstractIntegrationTest` ([src](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java)) class, which provides both `performAnalysis` and assert functions to check the found bugs.

spotbugs/etc/messages.xml
```diff
@@ -9822,6 +9830,18 @@ Using floating-point variables should not be used as loop counters, as they are
...
     </Details>
   </BugPattern>
+
+  <BugPattern type="UNS_UNSAFE_CALL">
+    <ShortDescription>Call to Unsafe.</ShortDescription>
+    <LongDescription>Do not use {0}. As the class name indicates, its methods are not safe to use.
+    </LongDescription>
+    <Details>
+      <![CDATA[
+      <p>Do not use either sun.misc.Unsafe or jdk.internal.misc.Unsafe. As there class name indicates, these methods are not safe to use.</p>
```
JuditKnoll: Can you please provide a solution too? What should the developer do, when they get this bug? Also, can you please provide a bit more explanation and maybe some references?

spotbugs/etc/messages_fr.xml
```diff
@@ -57,6 +57,11 @@
     <!-- Details subelement needed -->
   </BugCategory>
 
+  <BugCategory category="SECURITY">
+    <Description>Sécurité</Description>
+    <Abbreviation>S</Abbreviation>
+    <Details>Une utilisation d’entrées non fiables d’une manière qui pourrait créer une vulnérabilité de sécurité exploitable à distance.</Details>
+  </BugCategory>
```
JuditKnoll, line 64: This part was not added to the english `messages.xml` file.

spotbugsTestCases/src/java/BadUnsafeSun.java
```diff
@@ -0,0 +1,13 @@
+import java.lang.reflect.Field;
+import sun.misc.Unsafe;
+
+class BadUnsafeSun {
```
JuditKnoll: Since this and the other code example, `BadUnsafeSunQualified` are connected - both are example for the same underlying issue - can you please put them in a package together?
If you do, don't forget to update the path in `UnsafeDetectorTest`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
```diff
@@ -0,0 +1,59 @@
...
+     * Inspects bytecode instructions as they are encountered. Specifically, this method checks if the
+     * current opcode is an invocation of a virtual method ({@code INVOKEVIRTUAL}) on the {@code
+     * sun/misc/Unsafe} or {@code jdk/internal/misc/Unsafe} classes. If such a call is detected, it
+     * reports a high-priority bug instance indicating the use of potentially unsafe operations.
+     *
+     * @param seen the opcode of the currently visited instruction
+     */
+    @Override
+    public void sawOpcode(final int seen) {
+        if (seen != Const.INVOKEVIRTUAL) {
```
JuditKnoll, line 64: Is there any function of `sun.misc.Unsafe` or `jdk.internal.misc.Unsafe`, which is invoked with other invoke byte codes? (invoke interface, invoke special, invoke static)

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Inspects bytecode instructions as they are encountered. Specifically, this method checks if the
+     * current opcode is an invocation of a virtual method ({@code INVOKEVIRTUAL}) on the {@code
+     * sun/misc/Unsafe} or {@code jdk/internal/misc/Unsafe} classes. If such a call is detected, it
+     * reports a high-priority bug instance indicating the use of potentially unsafe operations.
+     *
+     * @param seen the opcode of the currently visited instruction
+     */
+    @Override
+    public void sawOpcode(final int seen) {
```
JuditKnoll, line 63: You only want to warning against calling functions of `sun.misc.Unsafe` and `jdk.internal.misc.Unsafe`, but not against other usages of these? E.g. creating a field with this type, casting, inheriting from them.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
```diff
@@ -0,0 +1,59 @@
...
+        if (seen != Const.INVOKEVIRTUAL) {
+            return;
+        }
+        if (getClassConstantOperand().equals("sun/misc/Unsafe")
+                || getClassConstantOperand().equals("jdk/internal/misc/Unsafe")) {
```
JuditKnoll, line 68: Are these classes final or can they be inherited? If they can, don't you want to check for usages of their children as well?
If you want to check usages of child classes, you may find `edu.umd.cs.findbugs.ba.Hierarchy` ([src](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/Hierarchy.java)) useful.

spotbugsTestCases/src/java/BadUnsafeSunQualified.java
```diff
@@ -0,0 +1,12 @@
+import java.lang.reflect.Field;
+
+class BadUnsafeSunQualified {
```
JuditKnoll: As far as I can see, in these code example classes 2 space indentation are used. Can you please use 4 space indentation to have it uniform with the rest of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
```diff
@@ -0,0 +1,59 @@
+package edu.umd.cs.findbugs.detect;
```
JuditKnoll, line 19: Can you please add the license header? You can copy it from another class. The one, starting with "SpotBugs".

thomasleplus:
> Thank you for your contribution! It looks like the [article you linked](https://blogs.oracle.com/javamagazine/post/the-unsafe-class-unsafe-at-any-speed) doesn't work anymore.

Thanks, apparently they took it down so I replaced the link with an archived version.

I'll work on your code review ASAP. I really appreciate it.

spotbugs/etc/findbugs.xml
thomasleplus: I noted that most bug codes are defined in `message.xml` (160) than in `findbugs.xml` (29) so I followed the same approach. But I am happy to have it in `findbugs.xml` (or both) if it is better?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
thomasleplus: They are both final.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
thomasleplus: I think so. They are utility classes. Invoking their methods is the actual security risk. Instantiating, casting, assigning etc is not, although it would be rather pointless to do that if you are not going to use some of the methods at some point. IMHO flagging all the references to these classes might create noise.

spotbugs/etc/messages_fr.xml
thomasleplus: It's already there. I am just taking the opportunity to add the French localization.

JuditKnoll: Thanks for the clarification and the translation both.

spotbugs/etc/findbugs.xml
JuditKnoll: You are right. Thanks for the highlight, I'll look into what might be the reason for this difference. It's good like this for now.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
thomasleplus: All the risky methods are instance methods (i.e. not static). They are not inherited nor are they part of an interface. Some of the methods are native methods but I am not sure if that makes a difference. Just in case I changed one on the test cases so that one uses a native method and the other not, this way we should be covered.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * FindBugs - Find Bugs in Java programs
```
JuditKnoll: This is the old license header. Can you please use the new one, the one starting with SpotBugs? You can copy it e.g. from `edu.umd.cs.findbugs.detect.AbstractAssertDetector` ( [src](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java) ).

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
JuditKnoll: Thanks for the explanation.

spotbugs/etc/messages.xml
thomasleplus: Because its a collection of methods which each have their safer alternative, it's hard to give a short solution. So instead I have added a link to the official OpenJDK spec that led to the deprecation of these methods. It has extensive details on the security issues with these methods and code examples on how to replace them.

spotbugsTestCases/src/java/BadUnsafeSun.java
thomasleplus: Done

spotbugs/etc/messages.xml
JuditKnoll: Thanks 😄

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
JuditKnoll: Thank you

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnsafeDetector.java
thomasleplus: Done. Thanks for catching that.

spotbugs/etc/messages_fr.xml
```diff
@@ -940,6 +945,14 @@ Un accesseur spécial est généré dans ce cas par la compilateur pour écrire
 ]]>
   </Details>
  </Detector>
+  <Detector class="edu.umd.cs.findbugs.detect.UnsafeDetector">
+    <Details>
+<![CDATA[
+<p>Ce détecteur cherches les appels aux classes Unsafe.</p>
```
gtoison: It should be:
"Ce détecteur cherche les appels aux classes Unsafe."

gtoison (changes requested):
There's a small typo in the French translation, otherwise this looks good to me 👍

spotbugs/etc/messages_fr.xml
thomasleplus: Good catch! Merci

hazendaz:
@thomasleplus Can you fix merge conflicts?

thomasleplus:
> @thomasleplus Can you fix merge conflicts?

Sure. Done!

Issue #3812, opened 2025-11-17
Consider splitting spotbugs-annotations to micro-modules


vlsi:
Currently, `spotbugs-annotations` includes multiple loosely-related annotations in a single package.
Clients often need `SuppressFBWarnings`, however, the users get all the other annotations on their autocomplete.

For instance, log4j2 uses `SuppressFBWarnings`: https://github.com/search?q=repo%3Aapache%2Flogging-log4j2+edu.umd.cs.findbugs.annotations&type=code

Log4j2 uses `org.jspecify.annotations.Nullable` annotations, however, there's one case when they accidentally imported `finbugs.annotations.Nullable`: https://github.com/apache/logging-log4j2/blob/4b7065b8b3818cabb88cf1d553da33897b733413/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/SocketAppenderReconnectTest.java#L24

It would be great if spotbugs ships multiple annotation modules like:
* `spotbugs-annotations-suppress` with `@SuppressFBWarnings`
* `spotbugs-annotations-checkreturnvalue` with `@CheckReturnValue`
* `spotbugs-annotations-resourcecleanup` with `@CreatesObligation`, `@DischargesObligation`, and `@CleanupObligation`
* `spotbugs-annotations-defaultannotation` with `@DefaultAnnotation*`
* etc

WDYT?

hazendaz:
Hi @vlsi, this is unlikely to occur because the fact is a lot of the stuff in that is deprecated for decade plus, we are working to delete all the offending content which is going to take a while.  The end result will end up the same.

Issue #3811, opened 2025-11-17
Deprecate and hide @Nullable, @NonNull, and the rest nullability annotations to avoid name clashes with jspecify


vlsi:
Currently `spotbugs-annotation` module brings multiple nullability annotations.

The ecosystem moves towards https://jspecify.dev/, so it would be nice if you could deprecate and hide `@Nullable` and `@NonNull` annotations.

Deprecation is straightforward: `@Deprecated` + release announcement.
Hiding would probably involve adding `ACC_SYNTHETIC` modifier after the compilation. It would still keep the class in the bytecode, so the old libraries would still work fine in the runtime, while the new clients would need to move to a different nullability annotations.

An alternative option could be to rewrite `@Nullable` annotation in Kotlin and use `@Deprecated(level = HIDDEN)`.

WDYT?

hazendaz:
@vlsi Feel free to send PR deprecating the ones that are not currently.

vlsi:
Thanks for the fast response.

Just wondering: is it possible to release spotbugs 4.x with deprecations as well?

AFAIK log4j2 is stuck with spotbugs 4.9 since log4j2 still targets Java 8: https://github.com/apache/logging-log4j2/blob/4b7065b8b3818cabb88cf1d553da33897b733413/.github/dependabot.yaml#L93

iloveeclipse:
> Just wondering: is it possible to release spotbugs 4.x with deprecations as well?

No! As said, we can't deprecate some functionality if there is no replacement available yet (for spotbugs it means, same support for JSpecify as we have for the other Null annotations).

iloveeclipse:
> The ecosystem moves towards https://jspecify.dev/, so it would be nice if you could deprecate and hide @Nullable and @NonNull annotations.

If you want from SpotBugs to deprecate "old" annotations in favour of jspecify, you should make sure first that SpotBugs has full jspecify annotations support, where "full" means at same level as with "old" annotations. One step in this direction would be https://github.com/spotbugs/spotbugs/issues/3143.

So all use cases that were working with old annotations should be supported.

> Hiding would probably involve adding ACC_SYNTHETIC modifier after the compilation. It would still keep the class in the bytecode, so the old libraries would still work fine in the runtime, while the new clients would need to move to a different nullability annotations.

I see you have already solution in mind, but I don't see which problem you are trying to address with this solution? It would be nice if you could elaborate your use case & how the proposed solution should solve that.

[...]

Issue #3143, opened 2024-10-03
Fully Support JSpecify 1.0


blipper:
1.0 is out now!

https://jspecify.dev/blog/

This could be done incrementally but ideally full conformance with https://jspecify.dev/docs/spec/ via https://jspecify.dev/docs/api/org/jspecify/annotations/package-summary.html

jhenkehd:
What is the status of JSpecify support? The linked PR has not been updated for half a year and the plugin repository is archived.

I cannot find any documentation if there are plans to support the JSpecify annotations.

We use spotbugs in our projects to enforce null-correctness (as far as static analysis can archive that) and we would like to introduce JSpecify's `@NullMarked` to reduce the amount of annotations requried in the code base.

hanson76:
Spring Boot 4 will be released in November and they have migrated to Jspecify and are promoting it. The list of open source projects that use Jspecify is growing.

It had been good if users of these projects could keep using SpotBugs.

hazendaz:
Hi @hanson76 honestly that doesn't break usage, we are not even on jakarta fully still.  Many have issues still with jspecify such as guava.  Its not that we won't support it, we have an open PR for quite some time that had issues with it then zero follow up.  If you  want to champion it, then please go pick up work from the open PR where you also already commented.  Otherwise its likely to just sit and age.

PR #3142, opened 2024-10-03
Partial Support of JSpecify Annotations

blipper:
https://jspecify.dev/docs/applying/

PR #3814, opened 2025-11-17
chore: deprecate nullability annotations as the ecosystem moves towards JSpecify

vlsi:
Fixes https://github.com/spotbugs/spotbugs/issues/3812

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## Unreleased - 2025-??-??
 ### Added
 - Recognize `jakarta.annotation.Nonnull` and `jakarta.annotation.Nullable` ([#3780](https://github.com/spotbugs/spotbugs/pull/3780))
+
+### Deprecated
+- Deprecated nullability annotations in favour of JSpecify ([#3811](https://github.com/spotbugs/spotbugs/issues/3811)) 
+
 ### Fixed
 - Fix incorrect bug counts and sizes when unioning reports ([#3721](https://github.com/spotbugs/spotbugs/issues/3721))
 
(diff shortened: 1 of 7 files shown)


JuditKnoll:
IMO first we should have at least one release supporting JSpecify, so with https://github.com/spotbugs/spotbugs/pull/3142 merged, and preferable making that support not only partial.

vlsi:
I might miss the full picture, however, could you please clarify why do you think "jspecify support" should come before deprecating SpotBugs' nullable annotations?

In practice, the support for JSpecify would be non-trivial, especially when it comes to generics, arrays, so the support might take noticeable time.

For instance, https://github.com/uber/NullAway focuses on nullness topic, and they are not fully there yet.
I doubt "full jspecify support" is easily doable.

iloveeclipse:
> I might miss the full picture, however, could you please clarify why do you think "jspecify support" should come before deprecating SpotBugs' nullable annotations?

See https://github.com/spotbugs/spotbugs/issues/3143

We can't deprecate one annotations set in favour of other if the other is not supported by SpotBugs. As the result of the PR here users will see code full of deprecation warnings and if they decide to switch to JSpecify, SpotBugs will not find any NP related issue anymore.

vlsi:
I would not question the usability of SpotBugs `@Nullable` annotations, however what would you think of the following:
release a separate artifact with just `@SuppressFBWarnings` and the corresponding `enum SuppressMatchType`? (they would need to be on a different package to avoid clashes if both jars appear on the classpath).

Then the users who use just `@SuppressFBWarnings` won't suffer from excessive `@Nullable` annotations (SpotBugs brings two of them!)

iloveeclipse:
Closing because of no interest from reporter: https://github.com/spotbugs/spotbugs/issues/3811#issuecomment-3542674850

iloveeclipse:
Solution for specific issue with `SuppressFBWarnings` : https://github.com/spotbugs/spotbugs/issues/3811#issuecomment-3542747802

PR #3819, opened 2025-11-19
Add "Bug Report" issue template

JuditKnoll:
Often times necessary parts (e.g. SpotBugs version) are missed in bug report issues, so I suggest we add issue templates with checklists to provide guidance about the most important info to include.

Try to include everything. LMK if you think I missed anything or it is too long or have any suggestion.


JuditKnoll:
It looks like there is already an issue template, but it doesn't really work, the way it should.

Issue #3822, opened 2025-11-20
False-positive in `CT_CONSTRUCTOR_THROW`


Vampire:
In this class:
```java
public class Foo {
    public Foo(Collection<?> foo) {
        this(new ArrayList<>(foo), bar());
    }

    private Foo(Collection<?> foo, Object bar) {
    }

    private static Object bar() throws RuntimeException {
        return null;
    }
}
```
a `CT_CONSTRUCTOR_THROW` is detected.

If you change
```java
this(new ArrayList<>(foo), bar());
```
to
```java
this(new ArrayList<>(foo), null);
```
or
```java
this(foo, bar());
```
then no false-positive is detected.

PR #3829, opened 2025-11-26
Fix `CT_CONSTRUCTOR_THROW` FP with public and private constructors

JuditKnoll:
Fixes https://github.com/spotbugs/spotbugs/issues/3822.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ConstructorThrowTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ConstructorThrowTest.java
@@ -38,7 +38,8 @@ class ConstructorThrowTest extends AbstractIntegrationTest {
     @Test
     void testConstructorThrowCheck5() {
         performAnalysis("constructorthrow/ConstructorThrowTest5.class");
-        assertBugTypeCount(CT_THROW, 1);
+        assertBugTypeCount(CT_THROW, 2);
+        assertBugAtLine(CT_THROW, 11);
         assertBugAtLine(CT_THROW, 16);
     }
 
@@ -295,6 +296,13 @@ class ConstructorThrowTest extends AbstractIntegrationTest {
     @Test
     void testGoodConstructorThrowCheck21() {
         performAnalysis("constructorthrow/ConstructorThrowNegativeTest21.class");
+        assertNoBugType(CT_THROW);
+    }
+
+    @Test
+    void testGoodConstructorThrowCheck22() {
+        performAnalysis("constructorthrow/ConstructorThrowNegativeTest22.class",
+                "constructorthrow/SuperConstructorThrowNegativeTest22.class");
         assertNoBugType(CT_THROW);
     }
 }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
@@ -23,10 +23,10 @@ import java.util.stream.Collectors;
 import edu.umd.cs.findbugs.OpcodeStack;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.ba.AnalysisContext;
+import edu.umd.cs.findbugs.ba.Hierarchy;
 import edu.umd.cs.findbugs.ba.XMethod;
 import edu.umd.cs.findbugs.internalAnnotations.DottedClassName;
 import edu.umd.cs.findbugs.util.ClassName;
-import edu.umd.cs.findbugs.util.Values;
 import org.apache.bcel.Const;
 
 import edu.umd.cs.findbugs.BugInstance;
@@ -161,8 +161,14 @@ public class ConstructorThrow extends OpcodeStackDetector {
                 }
             }
         } else if (isMethodCall()) {
-            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand()) && getClassConstantOperand().equals(Values.SLASHED_JAVA_LANG_OBJECT)) {
-                hadObjectConstructor = true;
+            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand())) {
+                try {
+                    if (Hierarchy.isSubtype(getDottedClassName(), ClassName.toDottedClassName(getClassConstantOperand()))) {
+                        hadObjectConstructor = true;
+                    }
+                } catch (ClassNotFoundException e) {
+                    AnalysisContext.reportMissingClass(e);
+                }
             }
 
             String calledMethodFQN = getCalledMethodFQN();
--- a/spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest22.java
+++ b/spotbugsTestCases/src/java/constructorthrow/ConstructorThrowNegativeTest22.java
@@ -1,0 +1,36 @@
+package constructorthrow;
+
+import java.io.IOException;
+
+/**
+ * The Exception is thrown before the java.lang.Object constructor exits, so this is compliant, since in this case
+ * the JVM will not execute an object's finalizer.
+ */
+public class ConstructorThrowNegativeTest22 extends SuperConstructorThrowNegativeTest22 {
+    protected static boolean verify() throws IOException {
+        // Returns true if data entered is valid, else throws an Exception
+        // Assume that the attacker just enters invalid data, so this method always throws the exception
+        throw new IOException("Still invalid data!");
+    }
+
+    public ConstructorThrowNegativeTest22() throws IOException {
+        super();
+    }
+}
+
+class SuperConstructorThrowNegativeTest22 {
+    protected static boolean verify() throws IOException {
+        // Returns true if data entered is valid, else throws an Exception
+        // Assume that the attacker just enters invalid data, so this method always throws the exception
+        throw new IOException("Invalid data!");
+    }
+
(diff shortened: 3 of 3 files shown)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ConstructorThrowTest.java
```diff
@@ -38,8 +38,7 @@ void testConstructorThrowCheck4() {
     @Test
     void testConstructorThrowCheck5() {
         performAnalysis("constructorthrow/ConstructorThrowTest5.class");
-        assertBugTypeCount(CT_THROW, 2);
-        assertBugAtLine(CT_THROW, 11);
+        assertBugTypeCount(CT_THROW, 1);
         assertBugAtLine(CT_THROW, 16);
     }
```
JuditKnoll, line 44: The test file for this is here:
https://github.com/spotbugs/spotbugs/blob/fdeb7fe48c9d4edc3ad340a6db207519a5456bdd/spotbugsTestCases/src/java/constructorthrow/ConstructorThrowTest5.java#L7-L23

It's really similar to the edge-case of public and private constructors, and in the no parameter constructor the Object constructor is not called directly. It's called via the other constructor. Fixing the one parameter constructor also fixes the no parameter one, so in this case IMO it's okay to only report the issue at the one parameter constructor.

See the bytecode for the testcase here:
```
constructorthrow.ConstructorThrowTest5.<init> : ()V
  0 aload_0         []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  1 iconst_5        [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  2 invokespecial   [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, 5 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  5 return          []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]

constructorthrow.ConstructorThrowTest5.<init> : (I)V
  0 aload_0         []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
  1 invokespecial   [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
  4 aload_0         []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
  5 iconst_0        [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
  6 putfield        [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, 0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
  9 aload_0         []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
 10 iload_1         [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
 11 putfield        [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
 14 aload_0         []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
 15 invokevirtual   [< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]
 18 return          []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >, < I, IP, r1 >]

constructorthrow.ConstructorThrowTest5.testMethod : ()V
  0 getstatic       []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  3 aload_0         [< Ljava/io/PrintStream;, static java.lang.System.out Ljava/io/PrintStream; >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  4 getfield        [< Ljava/io/PrintStream;, static java.lang.System.out Ljava/io/PrintStream; >, < Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
  7 invokevirtual   [< Ljava/io/PrintStream;, static java.lang.System.out Ljava/io/PrintStream; >, < I, 0:constructorthrow.ConstructorThrowTest5.i I, cbz >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
 10 new             []::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
 13 dup             [< Ljava/lang/RuntimeException;, new >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
 14 invokespecial   [< Ljava/lang/RuntimeException;, new >, < Ljava/lang/RuntimeException;, new >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
 17 athrow          [< Ljava/lang/RuntimeException;, new, return value from java.lang.RuntimeException.<init>()V >]::[< Lconstructorthrow/ConstructorThrowTest5;, IP, r0 >]
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
```diff
@@ -160,7 +160,8 @@ private void reportConstructorThrow(int seen) {
                 }
             }
         } else if (isMethodCall()) {
-            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand())) {
+            String calledClass = getClassConstantOperand();
```
ThrawnCA: Why make a local variable that's only used once?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
```diff
@@ -160,7 +160,8 @@ private void reportConstructorThrow(int seen) {
...
+            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand()) && calledClass.equals("java/lang/Object")) {
```
ThrawnCA: This could use `Values.SLASHED_JAVA_LANG_OBJECT`

JuditKnoll: Good catch! Thanks, fixed it.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
JuditKnoll: Checked it during debug and forgot to remove the variable. Thanks for highlighting it. Fixed it 😄

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
```diff
@@ -160,7 +161,7 @@ private void reportConstructorThrow(int seen) {
                 }
             }
         } else if (isMethodCall()) {
-            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand())) {
+            if (Const.CONSTRUCTOR_NAME.equals(getNameConstantOperand()) && getClassConstantOperand().equals(Values.SLASHED_JAVA_LANG_OBJECT)) {
```
gtoison: I don't completely understand the change but I suspect this should be tested against the superclass (which isn't necessarily java.lang.Object)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/ConstructorThrowTest.java
JuditKnoll: With the latest change, this is no longer true. It works the same way as it did earlier.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/ConstructorThrow.java
JuditKnoll: Thanks, you are right about the superclass. Fixed the code.

So the edge case of public and private constructors is detailed [here](https://wiki.sei.cmu.edu/confluence/spaces/java/pages/88487789/OBJ11-J.+Be+wary+of+letting+constructors+throw+exceptions#OBJ11J.Bewaryoflettingconstructorsthrowexceptions-CompliantSolution(JavaSE6%2CPublicandPrivateConstructors)). Basically if there is an exception raising check before the `java.lang.Object` constructor exits for the relevant object, so the check functions are called in constructor call's argument, then no finalizer attack can be made.

In the code example from https://github.com/spotbugs/spotbugs/issues/3822 the constructor of `ArrayList` is called, then the exception throwing `bar()` method. The detector earlier did not check which constructor is called when looking for the object's constructor, so it incorrectly recognized the `ArrayList`'s constructor as the objects' and flagged the unhandled exception throwing `bar()` call as problematic.

I hope this helps to clarify the change. Please LMK if it's still unclear.

gtoison: Thanks for the explanation and the new test!

PR #3836, opened 2025-11-30
Invalid syntax in findbugsfilter.xsd - 1 line simple fix #3832

### LINKED ISSUE #3832: Invalid syntax in findbugsfilter.xsd - 1 line simple fix
**Describe the bug**
findbugsfilter.xsd refers to itself, which is unintended and not valid syntax.

In IDE editors that strongly validate XML, it will flag both findbugsfilter.xsd and all XMLs written pointing to findbugsfilter.xsd as having a syntax bug. The code still runs, but strictly speaking findbugsfilter.xsd has a bug that strongly validating IDE editors show as an error for all XML file filters users write.

**To Reproduce**
1. Go to any filter file XML (or download findbugsfilter.xsd itself) and bring it up with Eclipse -> Open With -> "Generic Code Editor".
2. See the file is flagged with an error. Note, the XML file doesn't add an error, the original findbugsfilter.xsd itself shows the error when viewed alone.

**Expected behavior**
findbugsfilter.xsd or filter files XML's written should not show syntax errors (unless of course a user writing their filter file XML has some additional bug in their specific file).

**Environment**
- Eclipse version: 2025-12 M3 (4.38.0 M3)  
- Java Compiler: Eclipse Temurin
- Java version: 25
- SpotBugs Version: 4.9.8.2

**Additional context**

### Investigation
I figured out the issue as I encountered this in Eclipse. Eclipse is correct in flagging this XSD file has a bug, which in turn gets flagged as a problem from all downstream XMLs users write which reference this XSD. Possibly many users aren't writing their XMLs in an editor that fully validates the XML and the XSD or may ignore the error. But the XSD does have a bug that pops up if the XML/XSD are being strictly validated.

### Root Cause

- The problem is the findbugsfilter.xsd is referring to itself.

This line should **not** be in the XSD itself.:
`xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0/spotbugs/etc/findbugsfilter.xsd"`

That line goes only in XML files that use this XSD. i.e. findbugsfilter.xsd itself is not being written in the schema defined in findbugsfilter.xsd. Only an XML gets written and points to this XSD.

All XSDs use the schema [XMLSchema.xsd](http://www.w3.org/2001/XMLSchema), but this line is effectively telling XML editors that the findbugsfilter.xsd (the schema file itself) adheres to the schema written inside findbugsfilter.xsd.

### Immediate Simple Fix

1. Delete this 1 line [findbugsfilter.xsd#L3](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/etc/findbugsfilter.xsd#L3).
2. Publish the updated XSD published out to [githubusercontent](https://raw.githubusercontent.com/spotbugs/spotbugs).
3. Users writing XML files pointing to the new XSD will stop seeing the error.

### Slightly Better Longer-Term Fix

1. Update the XSD definition to indicate the XSD is written in the XSD namespace (utilizing **_xmlns:xsd="http://www.w3.org/2001/XMLSchema"_**) like these examples:
    * [Microsoft sample-xsd-file-simple-schema](https://learn.microsoft.com/en-us/visualstudio/xml-tools/sample-xsd-file-simple-schema?view=vs-2022)
    * [IBM files-example-xsd-file](https://www.ibm.com/docs/en/bpm/8.5.6?topic=files-example-xsd-file)
2. Publish the updated XSD published out to [githubusercontent](https://raw.githubusercontent.com/spotbugs/spotbugs).
3. Users writing XML files pointing to the new XSD will stop seeing the error.

patrick-malloy:

### User Documentation Update Note
There is a reference in the user docs to the old XSD:
https://github.com/spotbugs/spotbugs/blob/master/docs/filter.rst

Namely, the two 4.8.4 version numbers within this line:
`xsi:schemaLocation="https://github.com/spotbugs/filter/4.8.4 https://raw.githubusercontent.com/spotbugs/spotbugs/4.8.4/spotbugs/etc/findbugsfilter.xsd">`

After the next spotbugs release goes out, this XSD update should appear at:
https://raw.githubusercontent.com/spotbugs/spotbugs/4.9.10/spotbugs/etc/findbugsfilter.xsd

assuming the next spotbugs release is exactly "4.9.10".

Once that happens, a user could then utilize 4.9.10 version numbers like:
`xsi:schemaLocation="https://github.com/spotbugs/filter/4.9.10 https://raw.githubusercontent.com/spotbugs/spotbugs/4.9.10/spotbugs/etc/findbugsfilter.xsd">`

I am not 100% certain if the next version of spotbugs will be 4.9.10 (what seems likely from build.gradle) or 4.9.9 (since the last release seems to be 4.9.8).

I am happy to add that 1 line user documentation update too. I just didn't want to do that without a confirmation the next version definitely will be 4.9.10. Otherwise, the user doc would have an example pointing to a file that wouldn't exist on https://raw.githubusercontent.com.


spotbugs/etc/findbugsfilter.xsd
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schema targetNamespace="https://github.com/spotbugs/filter/4.8.4" elementFormDefault="unqualified"
-    xsi:schemaLocation="https://github.com/spotbugs/filter/4.8.4 https://raw.githubusercontent.com/spotbugs/spotbugs/4.8.4/spotbugs/etc/findbugsfilter.xsd"
```
ThrawnCA, line 3: Why not just provide the `xsi` namespace?

    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"

patrick-malloy: Absolutely. I made that update and used:

`xmlns:xsd="http://www.w3.org/2001/XMLSchema"
`
to make it as close as possible to findbugsplugin.xsd.

I also updated the docs/filter.rst doc to reference where the schema would show up. I used a version of 4.9.10 based on what I see as the latest version in build.gradle:

https://github.com/spotbugs/spotbugs/commit/3de56f66c4f5fe4f2537aac0fd6c8f1c118fd685

If for any reason the final build version is going to be different, I can change that for docs/filter.rst user documentation.

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -1,427 +1,450 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema targetNamespace="https://github.com/spotbugs/filter/4.8.4" elementFormDefault="unqualified"
-    xsi:schemaLocation="https://github.com/spotbugs/filter/4.8.4 https://raw.githubusercontent.com/spotbugs/spotbugs/4.8.4/spotbugs/etc/findbugsfilter.xsd"
-    xmlns="http://www.w3.org/2001/XMLSchema" xmlns:fb="https://github.com/spotbugs/filter/4.8.4">
+<xsd:schema targetNamespace="https://github.com/spotbugs/filter/4.9.10" elementFormDefault="unqualified"
```
ThrawnCA: Since this whole file is a schema, it makes more sense to keep the schema namespace as the default instead of making it an `xsd` prefix.

patrick-malloy: Absolutely. I made that update.

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema targetNamespace="https://github.com/spotbugs/filter/4.8.4" elementFormDefault="unqualified"
-    xsi:schemaLocation="https://github.com/spotbugs/filter/4.8.4 https://raw.githubusercontent.com/spotbugs/spotbugs/4.8.4/spotbugs/etc/findbugsfilter.xsd"
-    xmlns="http://www.w3.org/2001/XMLSchema" xmlns:fb="https://github.com/spotbugs/filter/4.8.4">
+<schema targetNamespace="https://github.com/spotbugs/filter/4.9.10" elementFormDefault="unqualified"
+    xmlns="http://www.w3.org/2001/XMLSchema"
```
ThrawnCA, line 3: This is now duplicating line 4. And you've still dropped the `schemaLocation` definition instead of fixing it.

    <schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="https://github.com/spotbugs/filter/4.9.10 https://raw.githubusercontent.com/spotbugs/spotbugs/4.9.10/spotbugs/etc/findbugsfilter.xsd

patrick-malloy: For that first item, I removed that duplicate redundant line 4 to clean it up.

The drop of schemaLocation from the XSD was the original fix for a bug that a strict syntax validator picked up. The original findbugsfilter.xsd was referencing itself as its own schemaLocation. XSD's don't place schemaLocation inside themselves as this file did.

The other 3 XSDs in this directory don't do that, so show as fully valid in a syntax validator.

An XML using a particular XSD should point to that XSD schemaLocation to say it will adhere to that XSD.

But the XSD itself wouldn't also point to itself as this original file did. XSDs are written with the elements defined from http://www.w3.org/2001/XMLSchema-instance (element, complexType, annotation, etc.). That's why the schemaLocation is removed from the XSD. schemaLocation would still appear in any XML though.

ThrawnCA (approved):

docs/filter.rst
```diff
@@ -314,9 +314,9 @@ Complete Example
 
     <?xml version="1.0" encoding="UTF-8"?>
     <FindBugsFilter
-		xmlns="https://github.com/spotbugs/filter/4.8.4"
+		xmlns="https://github.com/spotbugs/filter/4.9.10"
```
JuditKnoll: As far as I know there won't be a 4.9.10 version, the next version will be 4.10.0 (see the [milestone](https://github.com/spotbugs/spotbugs/milestone/32)), even though in the `build.gradle` [the 4.9.10 version is mentioned](https://github.com/spotbugs/spotbugs/blob/master/build.gradle#L19).
Even though with the current code it's in sync, this can be easily forgotten to be updated during release and would lead to inconsistencies. I suggest to have 4.10.0 the versions here, and in tandem with this in a separate PR the snapshot version should be updated as well by the SpotBugs team.
@hazendaz WDYT?

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix incorrect bug counts and sizes when unioning reports ([#3721](https://github.com/spotbugs/spotbugs/issues/3721))
 - Enhanced SARIF output with full description sections - adding markdown is still an open issue ([#2339](https://github.com/spotbugs/spotbugs/issues/2339))
 - Added missing null check to `MultipleInstantiationsOfSingletons` detector ([#3823](https://github.com/spotbugs/spotbugs/issues/3823))
+- Fix invalid syntax in findbugsfilter.xsd ([#3823](https://github.com/spotbugs/spotbugs/issues/3832))
```
JuditKnoll: There is a typo here:

```suggestion
- Fix invalid syntax in findbugsfilter.xsd ([#3832](https://github.com/spotbugs/spotbugs/issues/3832))
```

docs/filter.rst
patrick-malloy: I was wondering about something like having the original file be a template with tokens instead of actual versions and then replace those tokens build time. That maybe could be used so any docs always come out saying whatever the latest final version winds up being.

Just playing around locally, it seemed like if:

1. The original filter.rst file was renamed something like "filter.rst.template" or "filter-template.rst" if it is better to keep the file extension
2. Inside that template file, any actual versions get instead replaced by `@VERSION@` in the source control copy
3. build.gradle has some task to replace those `@VERSION@` tokens with the current build version and renames the output to the final real .rst file name.
4. That task gets tied to some TBD normal build task that would run before the separate documentation generation.

```
import org.apache.tools.ant.filters.ReplaceTokens

tasks.register('replaceVersionTokens', Copy) {
  from('docs') {
    include 'filter.rst.template'

    // Apply token replacement
    filter(ReplaceTokens, tokens: [VERSION: project.version])

    // Change the output filename
    rename { fileName ->
      "filter-replaced.rst" // put real final name "filter.rst" in the actual final copy
    }
  }

  // Output goes here
  into("docs")
}
```

Being new to this specific project, not sure if that would be useful, not useful, overkill, etc.

hazendaz: @patrick-malloy I agree with your idea there probably in general here as the version if we are sticking to full version which seems to be the case is going to be missed.  I presume that would cause issues though if we hard code a token then are using IDE to edit it.  Maybe there is a trade off of marking the file such as its left as the current release version then in the gradle file it has something like before value / after value or something similar so during release it increments exactly how we need while staying sound while developing.  For the doc though, sure this seems entirely ok for the doc.

hazendaz: and yes version will be 4.10.0 on next release here.

spotbugs/etc/findbugsfilter.xsd
```diff
@@ -1,7 +1,7 @@
...
+    xmlns:fb="https://github.com/spotbugs/filter/4.9.10">
```
JuditKnoll: @patrick-malloy You forgot to update the versions here. Can you please do it?

patrick-malloy: Absolutely. That should be there as well now.

JuditKnoll (approved):

JuditKnoll:
IMO this is good to go. @hazendaz , @gtoison do you want to take a look at this?

gtoison:
It looks good to me but it should be merged right before the release of 4.9.10 so the URL are up.

hazendaz:
going to release tonight, merging now so not missed.

PR #3842, opened 2025-12-03
Cleanup usages of some deprecated methods

JuditKnoll:
This PR removes usages of deprecated methods. There are still a lot of deprecated method usages, this is only a small cleanup.

----

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
@@ -22,6 +22,7 @@ package edu.umd.cs.findbugs.util;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import edu.umd.cs.findbugs.internalAnnotations.DottedClassName;
 import edu.umd.cs.findbugs.internalAnnotations.SlashedClassName;
+import org.apache.commons.lang3.Strings;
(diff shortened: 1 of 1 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
```diff
@@ -350,7 +349,7 @@ public static boolean matchedPrefixes(String[] classSearchStrings, @DottedClassN
         }
 
         for (String p : pp) {
-            if (!p.isEmpty() && (StringUtils.containsIgnoreCase(className, p) || fuzzyMatch(className, p))) {
+            if (!p.isEmpty() && (org.apache.commons.lang3.Strings.CI.contains(className, p) || fuzzyMatch(className, p))) {
```
ThrawnCA: Why not import this?

hazendaz:
Bet that was AI usage on the import.  It's bad at doing it correctly.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/ClassName.java
JuditKnoll: I love to say that it was AI, but no, it was only me not paying enough attention. The IDE found `edu.umd.cs.findbugs.util.Strings` from the same package, and I thought it's imported, this is why I used FQN. Thanks for the catch, fixed it.

Issue #3838, opened 2025-12-01
Suppression of bad time format on variable doesn't work


jpschewe:
**Describe the bug**
I have a variable that is initialized with a DateTimeFormat with bad options and I want to suppress the warning.
Putting the suppression on the variable doesn't work. However putting the suppression on the class does work.

This doesn't suppress the warning
```java
import java.time.format.DateTimeFormatter;
import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;

public class SuppressTimeFormatBroken {

  @SuppressFBWarnings(value = "FS_BAD_DATE_FORMAT_FLAG_COMBO", justification = "Intentionally don't want to output AM/PM")
  private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("hh:mm");
}
```

This does suppress the warning
```java
import java.time.format.DateTimeFormatter;
import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;

  @SuppressFBWarnings(value = "FS_BAD_DATE_FORMAT_FLAG_COMBO", justification = "Intentionally don't want to output AM/PM")
public class SuppressTimeFormatWorking {

  private static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("hh:mm");
}
```

**Environment**

    * Java Compiler: OpenJDK
    * Java version:
```
openjdk version "21.0.9" 2025-10-21
OpenJDK Runtime Environment (build 21.0.9+10-Ubuntu-124.04)
OpenJDK 64-Bit Server VM (build 21.0.9+10-Ubuntu-124.04, mixed mode, sharing)
```
    * SpotBugs Version: 4.9.4
    * How do you use SpotBugs: [e.g. via Gradle Plugin, Maven Plugin, Ant Plugin, SonarQube Plugin, Eclipse IDE Plugin, CLI]
        Plugin version: SpotBugs Gradle Plugin 6.3.0
        Environment version for Plugin: Gradle 8.13

- Java Compiler: [e.g. Eclipse Temurin]
- Java version: [e.g. JDK 21]
- SpotBugs Version: [e.g. 4.9.8]
- How do you use SpotBugs: [e.g. via Gradle Plugin, Maven Plugin, Ant Plugin, SonarQube Plugin, Eclipse IDE Plugin, CLI]
    - Plugin version: [e.g. SpotBugs Gradle Plugin 6.4.5]
[...]

PR #3868, opened 2026-01-01
Add a field annotation when looking for `FS_BAD_DATE_FORMAT_FLAG_COMBO`

gtoison:
When declaring and initialising a field at the same time the `FS_BAD_DATE_FORMAT_FLAG_COMBO` bug instances found are not associated with the field.
This is because the problematic code is in the `<init>` method, not on the field itself.

This PR adds a class keeping track of the bug instances according to the PC (and next PC) where they were detected.
When the field is assigned we check if a bug instance was found on the previous PC. 
In that case we add a field annotation.

Now that the bug has a field annotation, it is suppressed by field level annotation.


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3838Test.java
```diff
@@ -0,0 +1,24 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3838Test extends AbstractIntegrationTest {
+
+    @Test
+    void testClassSuppression() {
+        performAnalysis("ghIssues/Issue3838.class", "ghIssues/Issue3838$SuppressTimeFormatWorking.class");
```
JuditKnoll: Since both `SuppressTimeFormatWorking` and `SuppressTimeFormatBroken` are nested classes of `Issue3838`, and only one of them is given for each tests, we get the following error messages when running the tests: 
```
  ghIssues.Issue3838$SuppressTimeFormatWorking
```

Can you please either move one of the examples to a separate file or have one tests with all example code analyzed?
The `assertNoBugInClass(String bugtype, String className)` or the `assertBugInClassCount(String bugType, String className, int count)` function can be used to show the intent of testing the `SuppressTimeFormatWorking` class.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugPcMap.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public void accumulateBug(BugInstance sourceLine, BytecodeScanningDetector visitor) {
+        bugInstancesByPc.put(visitor.getPC(), sourceLine);
+        bugInstancesByNextPc.put(visitor.getNextPC(), sourceLine);
+    }
+
+    /**
+     * To be called when the visitor has seen {@link Const#PUTSTATIC} or {@link Const#PUTFIELD}.
+     * If a bug instance was created a the previous PC a field annotation will be added.
```
JuditKnoll:
```suggestion
     * To be called when the visitor has seen {@link Const#PUTSTATIC} or {@link Const#PUTFIELD}.
     * If a bug instance was created at the previous PC a field annotation will be added.
```

spotbugs/src/main/java/edu/umd/cs/findbugs/BugPcMap.java
```diff
@@ -0,0 +1,71 @@
...
+        if (bugInstance != null) {
+            FieldDescriptor fieldDescriptor = visitor.getFieldDescriptorOperand();
+
+            bugInstance.addField(fieldDescriptor);
+        }
+    }
+
+    /**
+     * Report accumulated bug instances to the {@link BugReporter} and clears itself.
+     */
```
JuditKnoll, line 64:
```suggestion
    /**
     * Reports accumulated bug instances to the {@link BugReporter} and clears itself.
     */
```

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
       (See [SEI CERT rule LCK00-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code) and [SEI CERT rule LCK04-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible))
 
 ### Fixed
+- Fix `FS_BAD_DATE_FORMAT_FLAG_COMBO` not suppressed by field-level annotation ([#3868](https://github.com/spotbugs/spotbugs/issues/3868))
```
JuditKnoll: #3868 is this PR, the issue is #3838. However the browser redirects to the PR, this unnecessary redirect can be avoided.

```suggestion
- Fix `FS_BAD_DATE_FORMAT_FLAG_COMBO` not suppressed by field-level annotation ([#3868](https://github.com/spotbugs/spotbugs/pull/3868))
```

JuditKnoll (approved):

PR #3869, opened 2026-01-04
Switch fallthrough fixes

gtoison:
Some (a bit unusual) constructs lead to false positives when looking for a switch fallthrough:
- When the default case is not the last case
- When a switch case ends with a conditional while loop
- When a switch case ends with an unconditional while loop

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
@@ -101,15 +101,15 @@ public class SwitchHandler {
             return false;
         }
 
-        return (pc == getNextSwitchOffset(dbc));
+        return (pc == getNextSwitchCasePC(dbc));
     }
 
-    public int getNextSwitchOffset(DismantleBytecode dbc) {
+    public int getNextSwitchCasePC(DismantleBytecode dbc) {
         int size = switchOffsetStack.size();
         while (size > 0) {
             SwitchDetails details = switchOffsetStack.get(size - 1);
 
-            int nextSwitchOffset = details.getNextSwitchOffset(dbc.getPC());
+            int nextSwitchOffset = details.getNextSwitchCasePC(dbc.getPC());
             if (nextSwitchOffset >= 0) {
                 return nextSwitchOffset;
             }
@@ -130,7 +130,7 @@ public class SwitchHandler {
         while (size > 0) {
             SwitchDetails details = switchOffsetStack.get(size - 1);
 
-            int nextSwitchOffset = details.getNextSwitchOffset(dbc.getPC());
+            int nextSwitchOffset = details.getNextSwitchCasePC(dbc.getPC());
             if (nextSwitchOffset >= 0) {
                 return details;
             }
@@ -308,7 +308,7 @@ public class SwitchHandler {
             this.exhaustive = exhaustive;
         }
 
-        public int getNextSwitchOffset(int currentPC) {
+        public int getNextSwitchCasePC(int currentPC) {
             while ((nextOffset < swOffsets.length) && (currentPC > (switchPC + swOffsets[nextOffset]))) {
                 nextOffset++;
             }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
@@ -339,7 +339,7 @@ public class SwitchFallthrough extends OpcodeStackDetector implements StatelessD
                 // The GOTO brings us back to an earlier PC
                 && branchTarget < getPC()
                 // This is the last instruction of the case
-                && getNextPC() == switchHdlr.getNextSwitchOffset(this)) {
+                && getNextPC() == switchHdlr.getNextSwitchCasePC(this)) {
             // get the entry for <PC of the branch instruction we're getting back to, branch target of that branch instruction>
             Entry<Integer, Integer> nextBranchTargetAfterBranchTarget = branchTargetsByPc.higherEntry(branchTarget);
 
@@ -347,7 +347,7 @@ public class SwitchFallthrough extends OpcodeStackDetector implements StatelessD
                 // the branch target of the GOTO is followed by an IF (it's a while GOTO)
                 if (isIf(getCodeByte(nextBranchTargetAfterBranchTarget.getKey()))
                         // And its target is after the next case so there's no fallthrough
-                        && nextBranchTargetAfterBranchTarget.getValue() > nextSwitchDetails.getNextSwitchOffset(getPC())) {
+                        && nextBranchTargetAfterBranchTarget.getValue() > nextSwitchDetails.getNextSwitchCasePC(getPC())) {
                     return true;
                 }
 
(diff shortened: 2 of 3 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -313,26 +324,46 @@ else if (seen == Const.PUTFIELD && stack.getStackDepth() >= 2) {
...
+        // the branch target is a GOTO instruction as it is the case for a no-op (empty block) in an arrow-syntax switch
+        if (branchTargetCodeByte == Const.GOTO) {
             return true;
         }
 
         SwitchDetails nextSwitchDetails = switchHdlr.getNextSwitchDetails(this);
 
+        if (nextSwitchDetails != null &&
+        // The GOTO brings us back to an earlier PC
+                branchTarget < getPC()
```
JuditKnoll: Can you please move the `&&` to the start of the next line in the condition, so the formatting is consistent?
```suggestion
        if (nextSwitchDetails != null
        // The GOTO brings us back to an earlier PC
                && branchTarget < getPC()
```

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -78,6 +80,8 @@ public class SwitchFallthrough extends OpcodeStackDetector implements StatelessD
 
     private int priority;
 
+    private TreeMap<Integer, Integer> branchTargetsByPc = new TreeMap<>();
```
JuditKnoll: As far as I can see, this could be final.

spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
```diff
@@ -305,15 +320,22 @@ public int getNextSwitchOffset(int currentPC) {
             return switchPC + swOffsets[nextOffset];
         }
 
-        public int getDefaultOffset() {
+        /**
+         * @return The PC of the default case
+         */
+        public int getDefaultOffsetPC() {
```
JuditKnoll: As far as I can see, this function returns the pc of the default case, not the offset of the default case relative to the switch pc. In this case the word `offset` in the name is slightly misleading IMO, instead of that it could contain `case`.
This is also true to the `getLastOffsetPC()` and the corresponding functions in `SwitchHandler` itself.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -313,26 +324,46 @@ else if (seen == Const.PUTFIELD && stack.getStackDepth() >= 2) {
...
-     * <li> or the branch target is before the PC corresponding to the switch instruction</li>
-     * <li> or there's no next switch case (as it is the case for the default case)</li>
-     * </ul>
      */
     public boolean isBranchTargetOutsideOfNextCase() {
         int branchTarget = getBranchTarget();
-        if (getCodeByte(branchTarget) == Const.GOTO) {
+        int branchTargetCodeByte = getCodeByte(branchTarget);
+        // the branch target is a GOTO instruction as it is the case for a no-op (empty block) in an arrow-syntax switch
+        if (branchTargetCodeByte == Const.GOTO) {
```
JuditKnoll: The `branchTargetCodeByte` field is only used once, so no need to have it as a separate variable.

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java
```diff
@@ -429,6 +429,20 @@ public static boolean isSwitch(int opcode) {
...
+     * @param opcode
+     *            the opcode
+     * @return true if instruction is an IF, false if not
+     */
+    public static boolean isIf(int opcode) {
+        return (opcode >= Const.IFEQ && opcode <= Const.IFLE)
+                || (opcode >= Const.IF_ICMPEQ && opcode <= Const.IF_ACMPNE)
+                || opcode == Const.IFNULL
+                || opcode == Const.IFNONNULL;
+    }
```
JuditKnoll, line 443: If we want to optimize this, then the first two part of this condition could be merged, since `IF_ICMPEQ` bytecode is right after `IFLE`. However, this would make the code harder to read, and even checking this implementation requires the code reader to know the opcodes for the relevant bytecode instructions.

See the bytecodes from [Oracle's Java Spec](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.if_acmp_cond):
```
ifeq = 153 (0x99)
ifne = 154 (0x9a)
iflt = 155 (0x9b)
ifge = 156 (0x9c)
ifgt = 157 (0x9d)
ifle = 158 (0x9e)
if_icmpeq = 159 (0x9f)
if_icmpne = 160 (0xa0)
if_icmplt = 161 (0xa1)
if_icmpge = 162 (0xa2)
if_icmpgt = 163 (0xa3)
if_icmple = 164 (0xa4)
if_acmpeq = 165 (0xa5)
if_acmpne = 166 (0xa6)

ifnull = 198 (0xc6)
ifnonnull = 199 (0xc7)
```

JuditKnoll: I looked at other usages of all of these bytecodes in the codebase, and could only find one: in `ifInstructionSet` in `FindUselessControlFlow` ([src](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindUselessControlFlow.java#L55-L74)). Since `FindUselessControlFlow` extends `BytecodeScanningDetector`, which extends `DismantleBytecode`, this implementation could be easily reused there (not necessarily in this PR).
The other usages either include other bytecodes or miss some of these.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
```diff
@@ -313,26 +324,46 @@ else if (seen == Const.PUTFIELD && stack.getStackDepth() >= 2) {
...
      * For loops the branch target will be before the offset of the next case, for breaks we're exiting the switch so the target is actually even after the end of the last case.
      * The branch target might be before the switch when we're inside another structure such as a loop.
-     *
-     * @return <code>true</code> if:
-     * <ul>
-     * <li> the branch target is a GOTO instruction as it is the case for a no-op (empty block) in an arrow-syntax switch
-     * <li> the branch target is after the next switch offset as it is the case for a switch break</li>
-     * <li> or the branch target is before the PC corresponding to the switch instruction</li>
-     * <li> or there's no next switch case (as it is the case for the default case)</li>
-     * </ul>
```
JuditKnoll, line 323: Even though this info is added to the implementation, IMO it's useful to have it in the function javadoc as well - if it's in sync with the implementation.

spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
gtoison: `getDefaultCasePC()` is clearer indeed, thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SwitchFallthrough.java
gtoison: I tried to keep track of the cases in the javadoc but it got messy when introducing nested conditions. I think it is one of those cases where code is the best way to express complexity

gtoison: I tried to keep track of the cases in the javadoc but it got messy when introducing nested conditions. I think it is one of those cases where code is the best way to express complexity

spotbugs/src/main/java/edu/umd/cs/findbugs/SwitchHandler.java
JuditKnoll: The `SwitchHandler::getDefaultOffsetPC` (line 151) and the `SwitchHandler::getLastOffsetPC` (line 158) could also be renamed like this.

gtoison: I have renamed these methods as well, thanks!

PR #3870, opened 2026-01-04
Do not report `EI_EXPOSE_REP` for Android's LiveData

### LINKED ISSUE #3762: False Positive — "May expose internal representation" (EI_EXPOSE_REP) with Android LiveData
When analyzing Android ViewModel classes with the `EI_EXPOSE_REP` rule (either directly from SpotBugs or through SonarQube with the sonar-findbugs plugin), the following pattern is being flagged as a potential exposure of internal representation:

```java
class ExampleViewModel extends ViewModel {
    private final MutableLiveData<String> text = new MutableLiveData<>();

    public LiveData<String> getText() {
        return text; // <-- Rule 'EI_EXPOSE_REP' triggers here
    }
}
```

**Expected Behavior**
This pattern is a standard and recommended Android architecture practice:

- We expose a read-only `LiveData<T>` to the UI layer for observation.
- Internally, the ViewModel maintains the mutable `MutableLiveData<T>` reference.
- The returned `LiveData` does not allow external mutation (`setValue()` and `postValue()` are not accessible).

Therefore, from a design standpoint, this does not expose a mutable internal representation — the external caller cannot modify internal state.

**Actual Behavior**
The rule `EI_EXPOSE_REP` still flags this as a potential issue:

> “May expose internal representation by returning reference to mutable object.”

**Environment**
- **Tool:** SpotBugs (used via SonarQube)
- **Rule ID:** EI_EXPOSE_REP
- **Language:** Java (Android)
- **Relevant library:** androidx.lifecycle.LiveData
- **Plugin version:** 5.x
- **JDK version:** 17
- **SonarQube version:** Enterprise Editoin v10.8

**Rationale**
In this specific case:

- The `LiveData` type itself is not mutable externally.
- The interface prevents external code from changing internal state.
- Returning a `LiveData` reference is the intended pattern for UI observation.

Hence, the warning appears to be a false positive in the context of Android's `LiveData` API.

**Suggested Resolution**
Possible resolutions:

1. Whitelist `androidx.lifecycle.LiveData` and its subclasses as effectively immutable for this rule.
2. Or, provide a mechanism (e.g., annotation or configuration) to mark certain classes as “safe to expose.”

Thanks in advance

gtoison:
Android's LiveData is designed to be exposed, do not report `EI_EXPOSE_REP` when a class exposes it.
This should fix #3762


spotbugsTestCases/src/fakeLibraries/androidx/lifecycle/LiveData.java
```diff
@@ -0,0 +1,13 @@
+package androidx.lifecycle;
+
+public abstract class LiveData<T> {
+    private T value;
+
+    public T getValue() {
+        return value;
+    }
+
+    public void setValue(T value) {
```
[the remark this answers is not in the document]
JuditKnoll: This is indeed protected according to the [Android reference](https://developer.android.com/reference/androidx/lifecycle/LiveData), however it's exposed by [MutableLiveData](https://developer.android.com/reference/androidx/lifecycle/MutableLiveData).

spotbugsTestCases/src/fakeLibraries/androidx/lifecycle/LiveData.java
```diff
@@ -0,0 +1,13 @@
+package androidx.lifecycle;
+
+public abstract class LiveData<T> {
```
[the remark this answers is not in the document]
JuditKnoll: According to [Android reference](https://developer.android.com/reference/androidx/lifecycle/LiveData) this is abstract, so Copilot is making a mistake here.

spotbugsTestCases/src/java/ghIssues/Issue3762.java
```diff
@@ -0,0 +1,12 @@
+package ghIssues;
+
+import  androidx.lifecycle.LiveData;
+import  androidx.lifecycle.MutableLiveData;
+
+public class Issue3762 {
+    private final MutableLiveData<String> text = new MutableLiveData<>();
+
+    public LiveData<String> getText() {
```
JuditKnoll, line 9: The return type also could be `MutableLiveData<String>` here. With this PR the parents are also checked, so that `EI_EXPOSE_REP` is also handled. Maybe that could be also added as an example, but it's not strictly necessary.

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
```diff
@@ -122,6 +125,39 @@ public static boolean mutableSignature(String sig) {
         }
     }
 
+    /**
+     * @return true if the class corresponding to this signature is exposable
+     */
+    public static boolean exposableSignature(String sig) {
```
JuditKnoll, line 131: The `MutableClasses.mutableSignature()` method is called from other detectors than `FindReturnRef` (e.g. `FindAssertionWithSideEffects`, `FindPublicAttributes`, `MutableStaticFields`), and there are other usages in `FindReturnRef` as well.
Did you check the other usages of `MutableClasses.mutableSignature()`?
Would it not worth it to add this edge case to the implementation of `MutableClasses.mutableSignature()` and not have it as a separate method?

gtoison: I looked a bit into it but I'm not sure what the other calls are doing, I'd rather make a conservatively small change for now and maybe we can expand this in another PR?

spotbugsTestCases/src/java/ghIssues/Issue3762.java
gtoison: I've added another method in the test case

spotbugs/src/main/java/edu/umd/cs/findbugs/util/MutableClasses.java
JuditKnoll: Yes, that works too 😄

hazendaz (changes requested):
copilot issues need reviewed

PR #3871, opened 2026-01-04
`BC_IMPOSSIBLE_DOWNCAST_OF_TOARRAY` not detected fix

### LINKED ISSUE #3752: A false negative bug in the BC_IMPOSSIBLE_DOWNCAST_OF_TOARRAY detector
In the following code example, SpotBugs fails to report a BC_IMPOSSIBLE_DOWNCAST_OF_TOARRAY warning at ​​line 5​​. The code incorrectly casts the result of Collection.toArray()to a type more specific than Object[], which is inherently unsafe. This represents a false negative in SpotBugs' detection.

### Minimized Code Example
```java
import java.util.ArrayList;
public class Main {
  public static void main(String args[]) {
    ArrayList<String> a = new ArrayList<String>();
    String s[] = (String[]) a.stream().toArray(); // should report a warning
  }
}
```

### Execution Log of the Code Example
```bash
Exception in thread "main" java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Ljava.lang.String; ([Ljava.lang.Object; and [Ljava.lang.String; are in module java.base of loader 'bootstrap')
        at Main.main(Main.java:6)
```

gtoison:
`BC_IMPOSSIBLE_DOWNCAST_OF_TOARRAY` isn't raised when casting the return of a Stream (or a Collection) toArray().
This should fix #3752


hazendaz (changes requested):
needs changelog, fix per copilot, and moved out of draft.

PR #3872, opened 2026-01-04
Migrate maven badge to maven-badges.sml.io

gtoison:
maven-badges.herokuapp.com was deprecated and must be migrated.

See: https://github.com/softwaremill/maven-badges#a-new-dns-address


hazendaz:
@gtoison That one seems flaky, the community generally switched to shields io instead.  I'm going to update it again as its showing wrong version currently with the change that was performed.

hazendaz:
Fixed directly on master 3e793d64df998bf6f4438f4fdd52aa37b83f6a35

hazendaz:
Now correct version shows.

gtoison:
Thanks !

PR #3897, opened 2026-01-25
Add a field annotation when looking for `DMI_RANDOM_USED_ONLY_ONCE`

### LINKED ISSUE #1613: Spotbugs ignores one particular `@SuppressFBWarnings` annotation.
Spotbugs ignores a suppress warnings annotation. But just one, other are working as expected.

Please take a look at my class `MainFrame.java` line 117:

```java
@SuppressFBWarnings(value = "DMI_RANDOM_USED_ONLY_ONCE", justification = "we need just ONE random number ONCE")
private float hue = new Random().nextFloat();
```

Why is Spotbugs ignoring the suppress warnings annotation? Others in the same project work as expected. I copied the value from the HTML report, so there should not be a typo.

Steps to reproduce:

```
git clone https://github.com/zdila/gpx-animator.git
cd gpx-animator
git checkout spotbugs-issue
./gradlew assemble
```

The build fails on Spotbugs problems. An HTML report can now be found in the build directory:

```
build/reports/checkstyle/main.html
```

I use Java 16.0.1 on macOS 11.4 with Spotbugs Gradle Plugin 4.7.1, Spotbugs 4.3.0, and Spotbugs Annotations 4.3.0.

Is the problem on my side? What have I overlooked?

gtoison:
Similarly to #3868 the bug is in the <init> method but the suppress annotation is on the field.
This should fix #1613


hazendaz (changes requested):
The fix looks good to me and the regression test covers the reported issue. Before this is ready to merge, please add the changelog entry. I'd also consider adding a static-field test since the implementation explicitly handles both PUTFIELD and PUTSTATIC, although I don't consider that a blocker.  And of course take it out of draft.

PR #3932, opened 2026-02-25
Fix(rcn): report RCN when non-null value is on left side of null comparison(#3920)

### LINKED ISSUE #3920: False Negative: RCN missing when non-null value is on the left side of null comparison
**Describe the bug**
SpotBugs fails to report  **RCN: Redundant comparison of non-null value** to null error when the non-null value appears on the left side of the comparison (x == null). The bug is only detected when null appears on the left side (null == x).

**To Reproduce**
```java
public class a {
    // Operand order: non-null == null
    public static void m1() {
        if (System.out == null) {                        // <- should report RCN (FN)
            throw new RuntimeException("System.out is null");
        }
    }

    // Operand order: null == non-null
    public static void m2() {
        if (null == System.out) {                        // <- reported RCN (TP)
            throw new RuntimeException("System.out is null");
        }
    }
}
```

**Expected behavior**
**Both m1() and m2()** should be reported with RCN: Redundant comparison of non-null value to null, since System.out is a known non-null value and both comparisons are semantically equivalent.
**Actual behavior**
Only m2() (where null is on the left side) is reported with RCN. m1() (where System.out is on the left side) is silently ignored.

**Output:**
L D RCN: Redundant comparison of non-null value to null in a.m2()  Redundant null check at a.java:[line 9]
L B THROWS: Method intentionally throws RuntimeException.  At a.java:[line 5]
L B THROWS: Method intentionally throws RuntimeException.  At a.java:[line 10]

**Environment**
- Java version: JDK 21
- SpotBugs Version: 4.9.8
- How do you use SpotBugs:CLI

**Additional context**
At the bytecode level, System.out == null and null == System.out compile to the same IFNULL / IFNONNULL check — the operand order in source code does not affect runtime semantics:
java// These are semantically identical:
```java
System.out == null   // m1() — RCN missed
null == System.out   // m2() — RCN detected
```
System.out is a well-known non-null static field (java.lang.System.out). A null comparison against it is redundant regardless of which side null appears on.

leemeii:
**Description**
When comparing `x == null` where x is a known non-null value, javac compiles this to GETSTATIC + IFNONNULL. The redundant branch was correctly detected, but the priority calculation in FindNullDeref.foundRedundantNullCheck() over-incremented the priority:

  - +1 for infeasibleEdgeSimplyThrowsException
  - +1 for createdDeadCode + !valueIsNull

This pushed the priority above LOW_PRIORITY (3), silently dropping the **RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE** warning.

The semantically equivalent `null == x` form (IF_ACMPNE) was not affected because the infeasibleEdgeSimplyThrowsException increment is only in the `secondValue == null` branch, leaving its priority at LOW_PRIORITY and correctly reporting the bug.

Fix: guard the infeasibleEdgeSimplyThrowsException priority increment with `&& valueIsNull`, so it only applies when checking a null value (not a non-null value), making both forms behave consistently.

Fixes #3920


JuditKnoll:
The build is failing because of failing tests. Please make sure that `.\gradlew build smokeTest` is successful.
See the test fail:
```
DetectorsTest > testAllRegressionFilesJavac() FAILED
    org.opentest4j.AssertionFailedError: Unexpected bugs (1):
    unexpected RCN Redundant null check at Ideas_2011_08_05.java:[line 38] (bugIdeas/Ideas_2011_08_05.java)
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:42)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:143)
        at app//edu.umd.cs.findbugs.DetectorsTest.checkForUnexpectedBugs(DetectorsTest.java:142)
```

leemeii:
> The build is failing because of failing tests. Please make sure that `.\gradlew build smokeTest` is successful. See the test fail:
> 
> ```
> DetectorsTest > testAllRegressionFilesJavac() FAILED
>     org.opentest4j.AssertionFailedError: Unexpected bugs (1):
>     unexpected RCN Redundant null check at Ideas_2011_08_05.java:[line 38] (bugIdeas/Ideas_2011_08_05.java)
>         at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:42)
>         at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:143)
>         at app//edu.umd.cs.findbugs.DetectorsTest.checkForUnexpectedBugs(DetectorsTest.java:142)
> ```

Hi @JuditKnoll,

I have updated the PR .

The previous fix was too broad. I’ve now narrowed the logic to only skip the priority increment when the value comes from a direct field access (using item1.getXField() != null). This ensures that System.out == null is correctly reported (fixing #3920) while maintaining the existing suppression for local variable "throw" patterns.

All CI checks have passed now. Could you please take another look? Thanks!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3920Test.java
```diff
@@ -0,0 +1,25 @@
...
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/3920">GitHub issue</a>
+ */
+class Issue3920Test extends AbstractIntegrationTest {
+
+    @Test
+    void testNonNullOnLeftSideOfNullComparison() {
+        performAnalysis("ghIssues/Issue3920.class");
+
+        // m1(): System.out == null (IFNONNULL) — previously a false negative
+        assertBugInMethod("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE", "Issue3920", "m1");
```
JuditKnoll, line 23: Can you please also check the number of relevant bugtypes?

JuditKnoll (changes requested):
Can you please add an entry to the unreleased part of the `CHANGELOG.md` file?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3920Test.java
leemeii: Hi @JuditKnoll,

I have updated the PR as requested:

Added an entry to the CHANGELOG.md file under the Unreleased section.

Updated Issue3920Test.java to include assertBugTypeCount for both RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE and RCN_REDUNDANT_COMPARISON_OF_NULL_AND_NONNULL_VALUE to ensure the exact number of bug types is verified.

All CI checks are passing now. Please let me know if there's anything else I should address. Thanks!

ThrawnCA (approved):

JuditKnoll (approved):
The build CI wasn't triggered, and I couldn't even approve their run, so I triggered them by updating to latest master.
Thank you for your work!

ThrawnCA (approved):

PR #3933, opened 2026-02-25
Fix: detect NP_LOAD_OF_KNOWN_NULL_VALUE when null check uses instanceof

### LINKED ISSUE #3916: False Negative: NP_LOAD_OF_KNOWN_NULL_VALUE missing when null check is done via instanceof
**Describe the bug**
The rule **NP_LOAD_OF_KNOWN_NULL_VALUE** fails to report null pointer dereference when the null check is performed using instanceof instead of an explicit != null comparison. 

**To Reproduce**
```java
class Foo {
  private void foo1(String s1) {
      if (s1 != null) return;
      String str = s1.toString();  // <- True Positive: NP reported
  }
  
  private void foo2(String s2) {
      if (s2 instanceof String) return;
      String str = s2.toString();  // <- False Negative: NP not reported
  }
}
```

**Expected behavior**
Both foo1() and foo2() should be reported with the NP warning.

**Actual behavior**
Only foo1() is reported with NP . foo2() is silently ignored.

**Environment**
- Java version: JDK21
- SpotBugs Version: 4.9.8
   

**Additional context**
When a parameter with static type T is checked using instanceof T, 
this is semantically equivalent to a null check:
```java
// When x has static type T:
x instanceof T ≡ x != null
```
However, SpotBugs only detects the explicit null check pattern and misses the equivalent instanceof pattern.

leemeii:
When a variable's static type is a subtype of the instanceof check type, the 'not instanceof' branch can only be reached if the value is null. IsNullValueAnalysis.getDecision() now uses Hierarchy.isSubtype() to detect this case and assigns pathSensitiveNullValue() to the 'not instanceof' branch.

Fixes #3916


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/ba/Issue3916Test.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/3916">GitHub issue #3916</a>
+ */
+class Issue3916Test extends AbstractIntegrationTest {
+
+    @Test
+    void testNullCheckViaInstanceof() {
+        performAnalysis("ghIssues/Issue3916.class");
+        assertBugInMethod("NP_LOAD_OF_KNOWN_NULL_VALUE", "Issue3916", "foo1");
+        assertBugInMethod("NP_LOAD_OF_KNOWN_NULL_VALUE", "Issue3916", "foo2");
```
JuditKnoll, line 20: Can you please also check the number of relevant bugtypes?

spotbugsTestCases/src/java/ghIssues/Issue3916.java
```diff
@@ -0,0 +1,26 @@
+package ghIssues;
+
+public class Issue3916 {
```
JuditKnoll, line 3: Can you please add a few more examples with instanceof? E.g. if the condition is negated, and some negative tests with the instanceof checking a subtype.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/npe/IsNullValueAnalysis.java
```diff
@@ -689,9 +696,30 @@ private IsNullConditionDecision getDecision(BasicBlock basicBlock, IsNullValueFr
...
-                ifcmpDecision = isNotInstanceOf ? tos : IsNullValue.pathSensitiveNonNullValue();
-                fallThroughDecision = isNotInstanceOf ? IsNullValue.pathSensitiveNonNullValue() : tos;
+                // Check if the static type of the value is a subtype of the instanceof check type.
+                // If so, "not instanceof" can only happen when the value is null.
+                boolean notInstanceOfImpliesNull = false;
+                try {
+                    TypeFrame typeFrame = typeDataflow.getFactAtLocation(atInstanceOf);
+                    if (typeFrame.isValid()) {
+                        Type tosType = typeFrame.getTopValue();
+                        Type instanceofType = ((TypedInstruction) prev.getInstruction()).getType(methodGen.getConstantPool());
```
JuditKnoll: Are you sure that the previous instruction will always be a `TypedInstruction`?

JuditKnoll (changes requested):
Thank you for your contribution!

Can you please add an entry to the unreleased part of the CHANGELOG.md file?

iloveeclipse:
See also https://github.com/spotbugs/spotbugs/pull/4085 (AI generated)

JuditKnoll (approved):

hazendaz (approved):

PR #3944, opened 2026-03-01
fix(sf): detect SF_SWITCH_NO_DEFAULT when last case contains only a break

### LINKED ISSUE #3905: False Negative: SF_SWITCH_NO_DEFAULT missing when the last case contains only a break statement
**Describe the bug**
The rule **SF_SWITCH_NO_DEFAULT** fails to report a missing default case in a switch statement if the last case block is "empty" (contains only a break statement). However , if the last case block contains any logic , the bug is correctly identified. 
In both scenarios, the default case is missing, and both should be flagged according to the rule definition.

**To Reproduce**
```java
package test;

public class LastCaseContentTest {
    
   
    public void test1() {
        int x = 0;
        switch(x) {
            case 0: x = 0; break;
            case 1: x = 1; break; // <- True Positive
    }

    public void test2() {
        int x = 0;
        switch(x) {
            case 0: x = 0; break;
            case 1: break; // <- should report SF (False Negative)
        }
    }
}
```

**Expected behavior**
SpotBugs should report SF_SWITCH_NO_DEFAULT for both test1() and test2(). The presence or absence of a default case is independent of the content of the preceding case blocks.

**Actual behavior**
Only test1() is reported. test2() is silently ignored .

**Environment**
- Java version: JDK 17
- SpotBugs Version: 4.9.8

leemeii:
## Summary

Fixes #3905 — `SF_SWITCH_NO_DEFAULT` false negative when the last switch case contains only a `break` statement.

### Root Cause

The existing detection relies on natural fall-through from the last case body to the default offset (made possible by the compiler omitting the redundant trailing `goto` for the last case). When the last case is empty (only a `break`), the compiler instead **merges the case offset with the default offset** in the `lookupswitch`/`tableswitch` table — both point to the same bytecode PC:

```
// case 1: x = 1; break  →  compiler drops trailing goto → natural fall-through → detected ✓
lookupswitch { 0: 28, 1: 33, default: 35 }
33: iconst_1; 34: istore_1   // falls through to 35 (reachable=true) → detected

// case 1: break  →  compiler merges case 1 with default → only reachable via GOTO
lookupswitch { 0: 28, 1: 33, default: 33 }  ← same PC!
30: goto 33   // reachable=false at 33 → MISSED (the false negative)
```

When there IS a real `default:` case, the compiler emits a separate `goto` for the empty case so its offset differs from the default offset — the fix condition `getLastCasePC() == getPC()` correctly avoids false positives there:

```
// case 1: break; default: doSomething();
lookupswitch { 0: A, 1: B, default: D }   // B has goto D, D has real code
// getLastCasePC()=B != getPC()=D  → no false positive
```

### Fix

Add a secondary detection path in `SwitchFallthrough.sawOpcode()` that fires when:
- `!reachable` — arrived via GOTO (not natural fall-through)
- `isDefaultOffset` — current PC is the default case PC
- `switchHdlr.getLastCasePC() == getPC()` — last case and default share the same PC (empty last case)
- `getPC() >= biggestJumpTarget` — we are at or past the furthest explicit jump target

### Test plan

- [x] `lastCaseWithLogic` (existing true positive) — still detected
- [x] `lastCaseOnlyBreak` — **now detected** (was false negative, fixes #3905)
- [x] `singleCaseOnlyBreak` — **now detected** (was false negative)
- [x] `lastCaseOnlyBreakWithDefault` — correctly NOT detected (has explicit `default:`)
- [x] `withExplicitEmptyDefault` — correctly NOT detected (has explicit `default:`)
- [x] `Bug2539590` all existing assertions — no regressions
- [x] `Issue3572` enum switch — no regressions


ThrawnCA (approved):

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3905Test.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.Test;
+
+class Issue3905Test extends AbstractIntegrationTest {
+
+    @Test
+    void testSfSwitchNoDefaultWithEmptyLastCase() {
+        performAnalysis("ghIssues/Issue3905.class");
+
+        // True positive: last case has logic - already detected before the fix
+        assertBugInMethod("SF_SWITCH_NO_DEFAULT", "ghIssues.Issue3905", "lastCaseWithLogic");
```
JuditKnoll, line 13: Please assert the number of relevant bugtypes as well.

JuditKnoll (changes requested):
Some tests are failing for Java 21 and 25, making the build fail. Please make sure that the build does not fail with any of the Java versions.
It may be worth to take a look at [JEP 441](https://openjdk.org/jeps/441), which introduced some changes in switches.

Don't forget to add a changelog entry.

Issue #2936, opened 2024-04-10
False Positive: IL_INFINITE_LOOP


raghucssit:
```
public class TestClass {

    public void sampleMethod() throws Exception {

        // some random last index.
        int lastIndex_o1_n1 = 5;
        // start index is always higher than start index.
        int startIndex_o1 = lastIndex_o1_n1 + 1;
        // Loop can't be entered because start is higher.
        for (int i = startIndex_o1; i <= lastIndex_o1_n1; i++) {
            System.out.println("Loop can't be entered. So can never go infinite.");
        }
    }

    public static void main(String[] args) throws Exception {
        new TestClass().sampleMethod();
    }
}

```

![image](https://github.com/spotbugs/spotbugs/assets/10301131/eeb23d25-aec6-4b4f-b5f0-a6ad8584a879)

raghucssit:
This looks like an opposite case of Infinite loop.
i.e. Loop can never be entered making statements inside for loop useless.

iloveeclipse:
I wonder if this is just wrong bug / message reported?

raghucssit:
I think so. This is more of dead code scenario.

JuditKnoll:
I think https://github.com/spotbugs/spotbugs/pull/2073 could help solve this issue.

PR #3945, opened 2026-03-01
fix(il): suppress IL_INFINITE_LOOP false positive for always-false loop conditions

### LINKED ISSUE #3904: False Positive: IL_INFINITE_LOOP incorrectly reported for terminating loops
**Describe the bug**
The rule **IL_INFINITE_LOOP** incorrectly flags terminating inner loops as infinite when they are nested within a true infinite outer loop. While the outer loop is indeed infinite, the inner loop has a clear termination path and should not be flagged.

**To Reproduce**

```java
package test;

public class Bug {

    public void test() {
        int i = 1;
            do {
                int j = 1;
                do {
                } while (j < 1);//<- Should Not Reporte (False Positive)
     
            } while (i < 5); //<- reported correctly
        
    }
}
```

**Expected behavior**
SpotBugs should report IL_INFINITE_LOOP  **only** for the outer while (i < 5) loop. It should **NOT** report it for the inner loop (while (j < 1)), as the inner loop logic is self-terminating regardless of the outer loop's state.

**Actual behavior**
Both of the outer loop and the inner loop are reported as IL_INFINITE_LOOP.

**Environment**
- Java version:JDK 17
- SpotBugs Version:4.9.8

leemeii:
## Summary

- Fixes a false positive where `IL_INFINITE_LOOP` was incorrectly reported for `do-while` loops whose condition evaluates to a constant-false value.
- When `do { } while (j < 1)` with `j == 1`: the condition `1 < 1` is always false, so the backward branch is never taken and the loop exits after its first (and only) iteration — it is **not** an infinite loop.
- The fix evaluates the actual comparison when both operands are known integer constants, and skips reporting when the backward branch would never be taken.

## Root Cause

In `InfiniteLoop.java`, the detection logic for backward conditional branches (`IF_ICMPLT`, `IF_ICMPLE`, etc.) checks only whether the operands are **invariant** (unchanged since the loop start), using `constantSince()`. It does not check whether the comparison itself evaluates to **true** (backward branch taken → loop repeats) or **false** (backward branch not taken → loop exits).

For `do { } while (j < 1)` with `j == 1`:
- `constantSince(j, loop_top) = true` (j not modified in loop body)
- `constantSince(1, loop_top) = true` (literal constant)
- → old code calls `reportPossibleBug` → **FALSE POSITIVE**

## Fix

After confirming both operands are invariant, also evaluate the comparison when both have known `Integer` constant values. If the condition is always false (backward branch never taken), skip reporting.

```java
Object rawConst0 = item0.getConstant();  // top-of-stack (value2)
Object rawConst1 = item1.getConstant();  // below top   (value1)
if (rawConst0 instanceof Integer && rawConst1 instanceof Integer) {
    int v0 = (Integer) rawConst0;
    int v1 = (Integer) rawConst1;
    boolean branchTaken = ...; // evaluates v1 <op> v0 per opcode
    if (!branchTaken) {
        break; // loop exits immediately — not infinite
    }
}
```

## Test plan

- [x] `InfiniteLoop.java` fix added with correct operand ordering (item0=value2, item1=value1)
- [x] `spotbugsTestCases/src/java/ghIssues/Issue3904.java` — four test methods covering:
  - `innerLoopAlwaysFalse`: nested do-while, inner condition always false → only outer loop reported (1 bug)
  - `innerLoopAlwaysTrue`: nested do-while, inner condition always true → inner loop reported
  - `singleLoopAlwaysFalse`: standalone always-false do-while → **not** reported
  - `singleLoopAlwaysTrue`: standalone always-true do-while → reported
- [x] `spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3904Test.java` — integration test
- [x] Verified manually via SpotBugs CLI that all four methods behave as expected
- [x] Existing `infiniteLoop/InfiniteLoop` test cases continue to report correctly

Fixes #3904


ThrawnCA:
Why would there be a legitimate loop whose condition is all constants? Sounds like a bug of some sort _should_ be raised in this case.

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3904Test.java
```diff
@@ -0,0 +1,29 @@
...
+
+class Issue3904Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIlInfiniteLoopFalsePositiveForAlwaysFalseCondition() {
+        performAnalysis("ghIssues/Issue3904.class");
+
+        // innerLoopAlwaysFalse: inner do-while condition (j < 1 with j == 1) is always false.
+        // The inner loop exits immediately — NOT infinite. Only the outer loop should be reported.
+        assertBugInMethodCount("IL_INFINITE_LOOP", "ghIssues.Issue3904", "innerLoopAlwaysFalse", 1);
```
JuditKnoll, line 14: Please also assert the number of expected relevant bugtypes.

JuditKnoll (changes requested):
Please do not forget to add an entry to the changelog's unreleased part.

As far as I can see, this is really similar to https://github.com/spotbugs/spotbugs/issues/2936. Can you please check whether this change resolves that issue as well (e.g. add that as a code example as well)?

hazendaz:
@leemeii You need a changelog.

PR #3946, opened 2026-03-02
fix(st): detect ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD in named inner classes

### LINKED ISSUE #3893: False Negative: ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD missed in non-static inner class with local variable shadowing
**Describe the bug**
The rule **ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD** fails to flag writes to static fields when they occur within a non-static inner class if a local variable shadows the static field name.

While the DLS_DEAD_LOCAL_STORE detector correctly identifies the assignment to the local variable, the ST detector seems to be "masked" or bypassed in this specific nested context.

**To Reproduce**
1. Use the following minimal Java code
 
```java
public class SpotBugsSTInnerLeakMini {
    private static int f;

    // Case 1: Top-level class (Correctly reported)
    void access(SpotBugsSTInnerLeakMini o) {
        f = o.f++; // -> ST triggered
    }

    class Inner {
        public Inner() {}

        // Case 2: Non-static inner class + Shadowing (False Negative)
        void access(SpotBugsSTInnerLeakMini o) {
            int f = o.f++;      // -> Should report ST (FN)
        }

        // Case 3: Overloaded method in inner class (False Negative)
        void access(String s) {
            int f = SpotBugsSTInnerLeakMini.f++;   // -> Should report ST (FN)
        }
    }
}
```
3. Compile the code: javac SpotBugsSTInnerLeakMini.java
4. Run SpotBugs: spotbugs -textui -low -effort:max  .
5. See error

**Expected behavior**
Both the access method in the top-level class AND the methods in the Inner class should trigger the ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD warning.

**Actual behavior**
Top-level class: Correctly reports ST .

Inner class: Completely misses ST. It only reports DLS .

**Environment**
- Java version:  JDK 17
- SpotBugs Version: 4.9.8

leemeii:
## Summary

Fixes #3893 — `ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD` was silently suppressed for all methods in named non-static inner classes.

### Root Cause

In `UnreadFields.java`, the priority-adjustment block for the `ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD` check contained:

```java
if (getClassName().indexOf('$') != -1 || BCELUtil.isSynthetic(getMethod()) || ...
```

Any class whose name contains `'$'` — including legitimate named inner classes like `OuterClass$Inner` — triggered `priority++`, pushing the bug priority from `LOW_PRIORITY` (3) to `EXP_PRIORITY` (4). SpotBugs's default `-low` threshold only reports up to priority 3, so the bug was silently dropped.

### Fix

Replace `getClassName().indexOf('$') != -1` with `getClassDescriptor().isAnonymousClass()`, which matches only anonymous classes (names ending in `$\d+`, e.g. `Outer$1`). Named inner classes (`Outer$Inner`) no longer receive the extra priority bump, so their ST writes are correctly reported.

The existing `BCELUtil.isSynthetic(getMethod())` condition already covers the synthetic accessor methods generated by older Java compilers for private field access across class boundaries.

## Test plan

- Added `spotbugsTestCases/src/java/ghIssues/Issue3893.java` with:
  - Outer class method writing to static field (always worked — control case)
  - Named inner class method writing to static field directly
  - Named inner class method writing to static field via explicit class qualifier
- Added `spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3893Test.java` asserting all three methods produce `ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD`
- Verified `BadStatic.java` still correctly reports ST (not regressed)


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3893Test.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.jupiter.api.Test;
+
+class Issue3893Test extends AbstractIntegrationTest {
+
+    @Test
+    void testStWriteToStaticFromInnerClass() {
+        performAnalysis("ghIssues/Issue3893.class", "ghIssues/Issue3893$Inner.class");
+
+        // True positive (always worked): outer class writes to static field
+        assertBugInMethod("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD", "ghIssues.Issue3893", "setCounter");
```
JuditKnoll, line 13: Can you please also assert the number of relevant bugtypes?

spotbugsTestCases/src/java/ghIssues/Issue3893.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Test cases for GitHub issue #3893:
+ * ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD false negative in non-static inner class.
+ *
+ * When a named non-static inner class writes to a static field of its outer class,
+ * the class name contains '$' (e.g., Issue3893$Inner). The old priority-bump logic
+ * for any class with '$' in its name suppressed reporting, causing a false negative.
+ */
+public class Issue3893 {
```
JuditKnoll, line 11: Can you please add an example where a static field is written from a static inner class?

JuditKnoll (changes requested):
Please don't forget to add a changelog entry in the unreleased part of the `CHANGELOG.md` file.

hazendaz:
@leemeii Please review the requests from you on this to move it forwards.

PR #3947, opened 2026-03-02
fix(ucf): report all UCF instances per method instead of deduplicating (#3894)

### LINKED ISSUE #3894: False Negative: UCF_USELESS_CONTROL_FLOW missed in catch blocks and simple terminal statements
**Describe the bug**
The rule **UCF_USELESS_CONTROL_FLOW**  fails to flag empty if statements located within catch blocks and ignores simple empty if blocks at the end of a method.

**To Reproduce**
```java
public class Bug {
    static int x = 0;
    static int y = 0;
  //case 1
    public static void test() {
        if (x > 0) {} //  <- triggered

 //case 2
        try { trigger(); } 
        catch (Exception e) {
            if (x == 0) {} // <- should report (FN)
        }
  //case 3
        if (y != 0) {} // <- should report (FN)
  //case 4
        if (x == 0 && true) {} // <-triggered 
    }
    static void trigger() throws Exception { throw new Exception(); }
}

```
Command: spotbugs -textui -low -effort:max Bug.class

**Expected behavior**
All four empty if blocks should be flagged as UCF violations regardless of their position or logical redundancy.

**Actual behavior**
Only  Case1 and Case4 are reported.

**Environment**
- Java version: JDK 17
- SpotBugs Version: 4.9.8

leemeii:
## Summary

- **Root cause**: `BugAccumulator` deduplicates bugs with the same hash when `MERGE_SIMILAR_WARNINGS=true` (the default for *all* effort levels). For `UCF_USELESS_CONTROL_FLOW`, the hash is derived from bug type + class name + method name only — no source line — so all UCF occurrences in the same method share the same hash. The `BugAccumulator` keeps only the first and silently drops all subsequent ones.
- **Fix**: Replace `BugAccumulator` with direct `bugReporter.reportBug()` calls and add `.addSourceLine(this)`, so each empty `if` block is reported independently regardless of how many other UCF bugs exist in the same method.
- **Effect**: This resolves two false negatives from issue #3894: empty `if` blocks inside `catch` blocks and empty `if` blocks before terminal statements were both silently dropped when any other UCF bug preceded them in the same method.

## Detailed Analysis

Given this code:

```java
public static void test() {
    if (x > 0) {}           // Case 1 – reported (first occurrence)
    try { trigger(); }
    catch (Exception e) {
        if (x == 0) {}      // Case 2 – FN (same hash as Case 1 → dropped)
    }
    if (y != 0) {}          // Case 3 – FN (same hash as Case 1 → dropped)
    if (x == 0 && true) {}  // Case 4 – reported (different bug type UCF_NEXT_LINE)
}
```

The bytecode for Cases 1, 2, and 3 all have equal `getBranchTarget()` and `getBranchFallThrough()` (the detection condition is satisfied), but Cases 2 and 3 are dropped by `BugAccumulator` deduplication because all three share the hash `"UCF_USELESS_CONTROL_FLOW:ClassName:test"`.

## Test plan

- [ ] `Issue3894Test.testUcfUselessControlFlowNotMissedInCatchOrAfterTryCatch` verifies:
  - All 3 `UCF_USELESS_CONTROL_FLOW` bugs in `multipleCasesInMethod` are reported (regression test for deduplication)
  - Case 2 at line 30 (inside `catch` block) is reported
  - Case 3 at line 33 (after `try-catch`) is reported
  - All 3 `UCF_USELESS_CONTROL_FLOW_NEXT_LINE` bugs in `multipleEmptyIfs` are reported (regression for `_NEXT_LINE` deduplication)
  - No false positives for non-empty `if` blocks

Fixes #3894


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3894Test.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.jupiter.api.Test;
+
+class Issue3894Test extends AbstractIntegrationTest {
+
+    @Test
+    void testUcfUselessControlFlowNotMissedInCatchOrAfterTryCatch() {
+        performAnalysis("ghIssues/Issue3894.class");
+
+        // All three UCF bugs in the same method must each be reported (deduplication regression)
+        assertBugInMethodCount("UCF_USELESS_CONTROL_FLOW", "ghIssues.Issue3894", "multipleCasesInMethod", 3);
```
JuditKnoll, line 13: @leemeii Can you please assert the number of relevant bugtypes as well? This way it's checked that there are no false negatives or duplicated bug reports.

PR #3951, opened 2026-03-03
fix(sa): fix SA_LOCAL_SELF_ASSIGNMENT false positive in try-switch with IINC

### LINKED ISSUE #3929: False Positive: SA_LOCAL_SELF_ASSIGNMENT incorrectly reported in nested try-switch structure
**Description**
The rule **SA_LOCAL_SELF_ASSIGNMENT** incorrectly flags a self-assignment of a local variable in a method containing a switch statement nested within a try-catch block.

While the code performs a legitimate increment (++a), SpotBugs identifies a self-assignment (a = a) at the exit point of the try block (Line 9), which does not exist in the source code. This appears to be a flaw in how the dataflow engine reconciles variable states when multiple control-flow structures (switch + exception handling) are combined.

**To Reproduce**
```java
public class SpotBugsSAMinimal {
    public void reproduce() {
        try {
            int a = 0;
            switch (a) {
                default:
                    ++a; 
            }
        } catch (Exception b) { //  reports a SA here (False Positive)
        }
    }
}
```
**Actual Behavior**
```
L D SA: Self assignment of $L1 in SpotBugsSAMinimal.reproduce()  At SpotBugsSAMinimal.java:[line 9]
M D DLS: Dead store to $L1 in SpotBugsSAMinimal.reproduce()  At SpotBugsSAMinimal.java:[line 7]
M D UC: Method SpotBugsSAMinimal.reproduce() seems to be useless  At SpotBugsSAMinimal.java:[line 11]
```
**Expected behavior**
SA (Line 9): Should NOT be reported. 
There is no a = a or equivalent self-assignment. 

**Environment**
Java version: JDK 21

SpotBugs : 4.8.9 CLI

leemeii:
## Summary

Fixes #3929 — `SA_LOCAL_SELF_ASSIGNMENT` is incorrectly reported in methods that contain a pre/post-increment operator (`++`/`--`) inside a `switch` statement nested within a `try-catch` block.

## Root Cause

`DismantleBytecode` marks the `IINC` instruction as both `isRegisterLoad=true` and `isRegisterStore=true`. `FindLocalSelfAssignment2.sawOpcode()` checks `isRegisterLoad()` first; since `IINC` is a register load, it sets `previousLoadOf = registerOperand` and skips the `else` block that would normally reset state. This leaves a stale "pending load" that survives the forward `GOTO` at the end of the try block, and then falsely matches the catch handler's `astore` (which stores the exception object into the local variable slot reused from the try block).

**Faulty bytecode trace:**
```
iinc 1, 1   →  isRegisterLoad()=true  →  previousLoadOf = 1  (stale!)
goto 19     →  forward, gotoCount = 1,  previousLoadOf stays 1
astore_1    →  slot 1 == previousLoadOf(1), gotoCount(1) < 2  →  FALSE POSITIVE
```

## Fix

Special-case `IINC` at the top of `sawOpcode()` to clear `previousLoadOf` and reset `gotoCount`. `IINC` is an atomic read-modify-write that never pushes a value onto the operand stack — it cannot be part of the `iload X; istore X` pattern that indicates a self-assignment.

```java
if (seen == Const.IINC) {
    // IINC is an in-place read-modify-write; it never pushes a value onto the stack
    // and cannot be part of a self-assignment (load then store) pattern.
    previousLoadOf = -1;
    gotoCount = 0;
    return;
}
```

## Test plan

- [ ] `Issue3929Test` — `assertNoBugInMethod` for `reproduceFalsePositive()` (the exact pattern from the issue) and `incrementBeforeCatch()` (variant with multiple cases)
- [ ] `Issue3929Test` — `assertBugInMethod` for `realSelfAssignment()` verifies true positives are still detected
- [ ] CLI verification: no `SA_LOCAL_SELF_ASSIGNMENT` for the minimal reproducer; `SA` still fires for `a = a`


spotbugsTestCases/src/java/ghIssues/Issue3929.java
```diff
@@ -0,0 +1,42 @@
...
+                    break;
+                default:
+                    break;
+            }
+        } catch (RuntimeException e) {
+        }
+    }
+
+    // True positive: real self-assignment should still be reported
+    @SuppressWarnings("SA_LOCAL_SELF_ASSIGNMENT")
```
JuditKnoll: Why do you suppress this warning here?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3929Test.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.jupiter.api.Test;
+
+class Issue3929Test extends AbstractIntegrationTest {
+
+    @Test
+    void testNoFalsePositiveSaLocalSelfAssignmentInTrySwitchIinc() {
+        performAnalysis("ghIssues/Issue3929.class");
+
+        // False positive (now fixed): ++a inside switch inside try-catch should NOT report SA
+        assertNoBugInMethod("SA_LOCAL_SELF_ASSIGNMENT", "ghIssues.Issue3929", "reproduceFalsePositive");
```
JuditKnoll, line 19: Can you please also assert the number of relevant bugtypes?

JuditKnoll (changes requested):

spotbugsTestCases/src/java/ghIssues/Issue3929.java
```diff
@@ -0,0 +1,56 @@
+package ghIssues;
+
+/**
+ * Test cases for GitHub issue #3929:
+ * SA_LOCAL_SELF_ASSIGNMENT incorrectly reported in nested try-switch structure.
+ */
+public class Issue3929 {
+
+    // False positive (now fixed): ++a inside switch inside try-catch should NOT report SA
+    public void reproduceFalsePositive() {
```
JuditKnoll, line 19: Tried it locally and I could minimize the issue: the switch-case is not necessary to reproduce. Can you please add a few more testcases like these: a minimized testcase with increment inside try-catch, a self assignment inside a try-catch.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindLocalSelfAssignment2.java
```diff
@@ -60,6 +60,13 @@ public void visit(Code obj) {
 
     @Override
     public void sawOpcode(int seen) {
+        if (seen == Const.IINC) {
```
JuditKnoll, line 63: This fixes the issue.
For reference: the loads and stores handled with `isRegisterStore()` and `isRegisterLoad()` functions from `DismantleBytecode`, which gives back the value of fields with the same name. These fields are modified in here:
https://github.com/spotbugs/spotbugs/blob/cf1f36b0448066b4b3cedbb449134544877f4474/spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java#L736-L836

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindLocalSelfAssignment2.java
```diff
@@ -60,6 +60,13 @@ public void visit(Code obj) {
...
+            // IINC is an in-place read-modify-write; it never pushes a value onto the stack
+            // and cannot be part of a self-assignment (load then store) pattern.
+            previousLoadOf = -1;
+            gotoCount = 0;
+            return;
+        }
```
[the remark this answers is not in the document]
JuditKnoll: As `IINC` is already handled, no need to additionally handle `WIDE`+`IINC`.

JuditKnoll (changes requested):

leemeii:
Thanks, I added the requested coverage in 3b24bbe7a. The test case now includes a minimized increment-inside-try-catch false-positive reproducer without the switch, and a real self-assignment inside try-catch to make sure SA_LOCAL_SELF_ASSIGNMENT is still reported there. I also updated the expected bug count accordingly. I verified the updated Issue3929.java source compiles with javac. The focused Gradle test could not complete locally because the wrapper distribution download for Gradle 9.5.1 timed out in this environment; CI should provide the full test result.

JuditKnoll (approved):

PR #3967, opened 2026-03-16
fix(il): fix IL_INFINITE_LOOP false positive for int-float/double/long comparison

leemeii:
## Summary

Fixes #3963

### Root Cause

When a while loop condition uses a widening primitive conversion (e.g., `int > double` via `dcmpl`/`dcmpg`, or `int > long` via `lcmp`), the OpcodeStack correctly constant-folds the comparison instruction to produce an int constant (e.g., `-1` when `1.0 < 2.0`). However, OpcodeStack does **not** propagate this constant through unary `IFLE`/`IFGE`/etc. to determine branch reachability — unlike binary `IF_ICMP*` instructions which do get this treatment.

As a result, the fall-through path after the always-taken exit branch is not marked unreachable. `InfiniteLoop` then sees the backward `goto` as reachable, creates a `BackwardsBranch`, finds the matching `ForwardConditionalBranch` with both items constant (the `dcmpl` result), and incorrectly reports `IL_INFINITE_LOOP`.

### Fix

- Added `opcode` field to `ForwardConditionalBranch` to record which conditional branch opcode created it.
- Added `isExitAlwaysTaken(ForwardConditionalBranch)` helper that evaluates whether the loop-exit branch is statically always taken, using the constant values of the operands and the opcode semantics.
- In `visit()`, skip the bug report when `isExitAlwaysTaken()` returns true — the loop body is never entered, so it is not an infinite loop.

The fix handles:
- Unary comparisons (`IFLE`, `IFGE`, `IFLT`, `IFGT`, `IFEQ`, `IFNE`) from `dcmpl`/`dcmpg`/`lcmp`/`fcmpl`/`fcmpg` results
- Binary integer comparisons (`IF_ICMP*`) as a defense in depth

### Test Cases

`Issue3963.java` / `Issue3963Test.java` cover:
- **False positives (now fixed)**: `foo.length() > 2.0`, `> 2.0f`, `> 2L` with `foo = "a"` (length 1, condition statically false)
- **True positives (still detected)**: same comparisons with `foo = "abc"` (length 3, condition statically true → genuinely infinite)

## Test plan
- [x] `Issue3963Test` integration test passes locally
- [x] `testLoop2()` from the issue report no longer produces `IL_INFINITE_LOOP`
- [x] True positives (genuinely infinite loops via double/float/long comparison) are still detected

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
@@ -6,7 +6,7 @@ import org.junit.jupiter.api.Test;
 class Issue3963Test extends AbstractIntegrationTest {
 
     @Test
-    void testIlInfiniteLoopFalsePositiveForIntDoubleComparison() {
+    void testIlInfiniteLoopFalsePositiveForWideningNumericComparisons() {
         performAnalysis("ghIssues/Issue3963.class");
 
         // False positives (now fixed): loop exits immediately because condition is statically false
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
@@ -1,0 +1,22 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3963Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIlInfiniteLoopFalsePositiveForIntDoubleComparison() {
+        performAnalysis("ghIssues/Issue3963.class");
+
+        // False positives (now fixed): loop exits immediately because condition is statically false
+        assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testDoubleComparisonFP");
+        assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testFloatComparisonFP");
+        assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testLongComparisonFP");
+
+        // True positives: condition is statically true, loop is genuinely infinite
+        assertBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testDoubleComparisonTP");
+        assertBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testFloatComparisonTP");
+        assertBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testLongComparisonTP");
+    }
+}
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
@@ -9,6 +9,9 @@ class Issue3963Test extends AbstractIntegrationTest {
     void testIlInfiniteLoopFalsePositiveForWideningNumericComparisons() {
         performAnalysis("ghIssues/Issue3963.class");
 
+        // Exactly 3 IL_INFINITE_LOOP bugs expected (one per true positive method)
+        assertBugTypeCount("IL_INFINITE_LOOP", 3);
+
         // False positives (now fixed): loop exits immediately because condition is statically false
         assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testDoubleComparisonFP");
         assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testFloatComparisonFP");
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
@@ -139,11 +139,13 @@ public class InfiniteLoop extends OpcodeStackDetector {
     static class ForwardConditionalBranch extends Jump {
         final OpcodeStack.Item item0;
         final OpcodeStack.Item item1;
+        final int opcode;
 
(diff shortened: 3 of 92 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
```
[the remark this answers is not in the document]
leemeii: @copilot apply changes based on [this feedback](https://github.com/spotbugs/spotbugs/pull/3967#discussion_r2939021565)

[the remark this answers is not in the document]
JuditKnoll: It looks like Copilot was not able to apply these changes. Can you please revert these changes yourself?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.jupiter.api.Test;
+
+class Issue3963Test extends AbstractIntegrationTest {
+
+    @Test
+    void testIlInfiniteLoopFalsePositiveForWideningNumericComparisons() {
+        performAnalysis("ghIssues/Issue3963.class");
+
+        // False positives (now fixed): loop exits immediately because condition is statically false
+        assertNoBugInMethod("IL_INFINITE_LOOP", "ghIssues.Issue3963", "testDoubleComparisonFP");
```
JuditKnoll, line 16: Please also assert the number of relevant expected bug hits. This makes sure that none of the bugs are reported multiple times are there are no additional unexpected hits.

spotbugsTestCases/build.gradle
```diff
@@ -55,9 +55,11 @@ tasks.register('exportClasspath') {
...
     dependenciesFile.getParentFile().mkdirs();
 
-    configurations.getByName('runtimeClasspath').resolvedConfiguration.resolvedArtifacts.each { ResolvedArtifact artifact ->
-      dependenciesFile << artifact.file
-      dependenciesFile << '\n'
+    dependenciesFile.withWriter('UTF-8') { writer ->
+      configurations.getByName('runtimeClasspath').resolvedConfiguration.resolvedArtifacts.each { ResolvedArtifact artifact ->
+        writer.write(artifact.file.toString())
+        writer.write('\n')
+      }
```
JuditKnoll: This change here seems unrelated. Is it necessary for the fix?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
```diff
@@ -480,6 +485,66 @@ private int constantSince(Item item1) {
...
+     * statically known to be always true (e.g., from a float/double comparison via dcmpl/dcmpg).
+     */
+    private static boolean isExitAlwaysTaken(ForwardConditionalBranch fcb) {
+        Object c0 = fcb.item0.getConstant();
+        if (!(c0 instanceof Integer)) {
+            return false;
+        }
+        int v0 = (Integer) c0;
+        switch (fcb.opcode) {
+        // Unary comparisons against 0 (item0 == item1, e.g. from dcmpl/dcmpg/lcmp result)
```
JuditKnoll: Can you please elaborate why do you compare against _zero_?

JuditKnoll (changes requested):

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
hazendaz: @copilot apply changes based on [this feedback](https://github.com/spotbugs/spotbugs/pull/3967#discussion_r2939021565)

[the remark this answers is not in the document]
hazendaz: branch was conflicted thus the issue, fixed the conflict now issue is gone.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3963Test.java
leemeii: Done. Added `assertBugTypeCount("IL_INFINITE_LOOP", 3)` to verify exactly 3 bugs are reported (one per true positive method), ensuring no duplicates or unexpected hits.

spotbugsTestCases/build.gradle
leemeii: You're right, it's unrelated. It was a local workaround for Chinese locale (GBK encoding) causing test failures. Reverted in the latest commit.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
leemeii: Good question. Unary `IF*` instructions (`IFEQ`, `IFNE`, `IFLT`, etc.) pop one int operand from the stack and compare it against an **implicit zero** per the JVM spec. These opcodes appear after `dcmpl`/`dcmpg`/`fcmpl`/`fcmpg`/`lcmp` instructions, which produce `-1`, `0`, or `1` as the comparison result. So `v0` holds that result and we check whether the branch (comparing `v0` against 0) is always taken. Updated the comment to explain this more clearly.

JuditKnoll: Thanks for the explanation

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
```diff
@@ -480,6 +485,69 @@ private int constantSince(Item item1) {
...
+        case Const.IF_ICMPLT:
+        case Const.IF_ICMPLE:
+        case Const.IF_ICMPGT:
+        case Const.IF_ICMPGE: {
+            Object c1 = fcb.item1.getConstant();
+            if (!(c1 instanceof Integer)) {
+                return false;
+            }
+            int v1 = (Integer) c1;
+            switch (fcb.opcode) {
```
JuditKnoll, line 529: This switch is a nested one, with the outer and inner switch both checking the same variable's value. Can you please simplify it to use only one switch?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
```diff
@@ -339,11 +344,11 @@ public void sawOpcode(int seen) {
             OpcodeStack.Item item0 = stack.getStackItem(0);
             if (getDefaultSwitchOffset() > 0) {
                 forwardConditionalBranches.add(new ForwardConditionalBranch(item0, item0, getPC(), getPC()
-                        + getDefaultSwitchOffset()));
+                        + getDefaultSwitchOffset(), seen));
```
JuditKnoll, line 347: I know it was like this earlier, but can you please delete the line break here, or - if spotless is complaining about the formatting - have the `getPC() + getDefaultSwitchOffset()` in the same line, so it's a bit easier to read? (It's only a nice to have, and doesn't strictly belongs to this PR.)

spotbugsTestCases/src/java/ghIssues/Issue3963.java
```diff
@@ -0,0 +1,51 @@
...
+ * Test cases for GitHub issue #3963:
+ * IL_INFINITE_LOOP false positive when loop condition uses int-to-double/float/long
+ * comparison that is statically known to be false (loop exits immediately).
+ */
+public class Issue3963 {
+
+    // False positive (now fixed): foo.length()=1 > 2.0 is false, loop exits immediately
+    public void testDoubleComparisonFP() {
+        String foo = "a";
+        while (foo.length() > 2.0) {
```
JuditKnoll, line 13: Does it still work if 2.0 is saved into a variable?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/InfiniteLoop.java
```diff
@@ -480,6 +485,69 @@ private int constantSince(Item item1) {
 
     }
 
+    /**
+     * Returns true if the forward conditional branch (loop exit) would always be taken
+     * based on the constant values of its operands, meaning the loop body is never entered.
+     * This prevents false positive IL_INFINITE_LOOP reports when the exit condition is
+     * statically known to be always true (e.g., from a float/double comparison via dcmpl/dcmpg).
+     */
+    private static boolean isExitAlwaysTaken(ForwardConditionalBranch fcb) {
```
JuditKnoll, line 494: For me this name does not describe well the behavior, however the javadoc explains it well. Can you please rename it?

JuditKnoll (changes requested):

ThrawnCA (approved):

PR #4000, opened 2026-04-12
fix(tq) @Generated behaviour for TQ_ checks (#3999)

### LINKED ISSUE #3999: CheckTypeQualifiers findings do not consider @Generated
So both these

```java
@Generated
public class IssueX {
  @Strict
  Object f;

  public void violation(Object unknown) {
    f = unknown;
  }
}
```

```java
public class IssueX {
  @Strict
  Object f;

  @Generated 
  public void violation(Object unknown) {
    f = unknown;
  }
}
```

will report a `TQ_UNKNOWN_VALUE_USED_WHERE_ALWAYS_STRICTLY_REQUIRED`.

AFAICT, this is because `edu.umd.cs.findbugs.BugInstance#addClassAndMethod(edu.umd.cs.findbugs.ba.XMethod)` does not call `MemberUtils.isUserGenerated`/`foundInAutogeneratedMethod` like the other `addClassAndMethod` methods.

I am already working on a fix, I mainly need this issue to get an issue number for my test.

Tillerino:
Closes #3999

`@Generated` annotations were being ignored by all `TQ_` checks due to a missing call to `isUserGenerated`. Additionally, `edu.umd.cs.findbugs.bytecode.MemberUtils#isUserGenerated(edu.umd.cs.findbugs.ba.ClassMember)` did not check for `@Generated` annotations on the enclosing class.

I had to rename a different `@Generated` annotation in tests because it was relying on this broken behaviour. With the fix, too many bugs were being filtered out for its test.


ThrawnCA (approved):

spotbugsTestCases/src/java/ghIssues/issue3999/Generated.java
```diff
@@ -0,0 +1,4 @@
+package ghIssues.issue3999;
+
+public @interface Generated {
+}
```
JuditKnoll, line 4: Can you please share what's the purpose of this Generated annotation? Why not simply reuse the already existing Generated fake annotation?

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/GeneratedX.java
```diff
@@ -4,7 +4,8 @@
 import java.lang.annotation.Target;
 
 @Target(ElementType.TYPE)
-public @interface Generated {
+// X suffix so that this is not matched by the built-in filter for all "Generated" annotations
+public @interface GeneratedX {
```
JuditKnoll, line 8: The goal of this class is to represent the [original](https://github.com/immutables/immutables/blob/master/value-annotations/src/org/immutables/value/Generated.java) `org.immutables.value.Generated` annotation. Please do not change it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3999Test.java
```diff
@@ -0,0 +1,18 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+public class Issue3999Test extends AbstractIntegrationTest {
```
JuditKnoll, line 6: Please remove the public access modifier as it's unnecessary with Junit Jupiter.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
```diff
@@ -834,8 +834,14 @@ public BugInstance addClassAndMethod(MethodDescriptor methodDescriptor) {
...
 
+    @Nonnull
     public BugInstance addClassAndMethod(XMethod xMethod) {
-        return addClassAndMethod(xMethod.getMethodDescriptor());
+        addClass(xMethod.getClassName());
+        add(MethodAnnotation.fromXMethod(xMethod));
+        if (!MemberUtils.isUserGenerated(xMethod)) {
+          foundInAutogeneratedMethod();
+        }
+        return this;
```
JuditKnoll, line 844: This would affect other bugs as well, not those in TQ category. Please revert it.

JuditKnoll (changes requested):
The bugs from the `TQ` category are reported by `edu.umd.cs.findbugs.detect.CheckTypeQualifiers` detector (see the relevant parts of [findbugs.xml](https://github.com/spotbugs/spotbugs/blob/6b55b3576d0e8467f362270abc9c942cc85588fc/spotbugs/etc/findbugs.xml#L646-L647), [src of the class](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CheckTypeQualifiers.java)). I think the cause of the issue should be there.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
Tillerino: I framed this PR as TQ-related (since that was where I encountered the bug), but the effect on other categories is IMO correct.

AFAICT, Spotbugs _should never_ report bugs in `@Generated` code - independent of the type of bug. TQ is one of the affected categories - it looks like some of the categories are implemented with a different mechanism .

I pretty much yoinked this bit from a different `addClassAndMethod` method: https://github.com/spotbugs/spotbugs/pull/4000/changes/BASE..cc9526f529b6f8203312de6237049621ad8da7f1#diff-1d958cc9dc29ec3d6137b19391626e4d6a26b6ee962bfb162ca1012816f08722R923

I found that mechanism when debugging why other types of bugs were affected by `@Generated`. So yeah, if this is not the way to fix that, then I don't have a fix :)

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/GeneratedX.java
Tillerino: I think that there was a timing issue here. This test ensures that you can _manually_ filter out findings for specified annotations, taking immutables as an example.

Either a general filter for `@Generated` had not been implemented yet when this test was written, or this test just accidentally hit categories that were affected by the bug that I am trying to fix.

Either way, if Spotless were working as what I think is intended (i.e. all `@Generated` code is not reported on), then this test does would not make sense: Spotbugs would _never_ report findings in immutables-generated code. Renaming it a bit seemed like the least invasive way to fix this.

spotbugsTestCases/src/java/ghIssues/issue3999/Generated.java
Tillerino: There were two `Generated` annotations in `issueXXX` packages. Mixing GH issues seemed like bad practice. And the other `Generated` method, I had to rename, see below.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3999Test.java
Tillerino: Sure thing.

I can also see a few spotless findings in the CI. I just ran `gradle test` locally. I think it's a good idea to have spotless run when you go `gradle test`.

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/GeneratedX.java
Tillerino: I can try to reframe this entire annotation filter thing away from immutables, if you want.

Tillerino:
Thanks for the review. I'll wait for answers until I push further changes.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3999Test.java
JuditKnoll: Spotless is run when `gradle build` is run, not for `gradle test`. If you check the [contribution guideline](https://github.com/spotbugs/spotbugs/blob/master/.github/CONTRIBUTING.md#contributing-guideline), you can see that `./gradlew spotlessApply build smoketest` should be run locally to make sure there are no build issues.

spotbugs/src/main/java/edu/umd/cs/findbugs/BugInstance.java
JuditKnoll: > If this is not only related to one bug category, then both the issue and the changelog are quite ambiguous.

There are bugs which still should be reported even in `@Generated` code.

However even if, the goal would be to not report any bugs on `@Generated` code, then the current changes in this PR would not achieve that goal, as if another e.g. an `addClassAndMethod` function with different parameter types is called, then this would not suppress that.

JuditKnoll: > TQ is one of the affected categories - it looks like some of the categories are implemented with a different mechanism .

If this is not only related to one bug category, then both the issue and the changelog are quite ambiguous.

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/GeneratedX.java
JuditKnoll: Not timing, but the intention is different: there are some bugs which still should be reported even in code with `@Generated` annotation.

I'm not convinced that the issue you described is a false positive. Can you please share why do you think the bug should not be reported?

spotbugsTestCases/src/java/ghIssues/issue3999/Generated.java
JuditKnoll: The Generated annotation you renamed is in `fakeAnnotations`, not in `issueXXX` package.
There can be classes with the same names in different packages, so this does not indicate that one of them should be renamed.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3999Test.java
Tillerino: Ok you got me. I didn't pay attention :D

spotbugsTestCases/src/fakeAnnotations/org/immutables/value/GeneratedX.java
Tillerino: > there are some bugs which still should be reported even in code with `@Generated` annotation.

Ok, so then there is a big misunderstanding here. I could not find any documentation about the treatment of generated code at all. To be honest, I just _assumed_ that generated code should not be reported on and just _assumed_ that it was a bug that some findings would slip through. The opaque handling of this (in the addClassAndMethod methods) and the thin tests supported that theory :)

Looks like that was wrong. But is there documentation? Specifically which bug categories should be ignored in generated code?

I have been using findbugs/spotbugs for over a decade now and this has been bugging me for ages. I always chalked it up to spotbugs being buggy and ended up using workarounds. I only now decided to do something about it. If I find the relevant piece of documentation instead of changing the code, that would be an even better outcome :)

JuditKnoll: > But is there documentation? Specifically which bug categories should be ignored in generated code?

I don't think there is documentation specifically about generated code handling. At least there isn't any I know about. It may be mentioned at some bugs at the [bug descriptions](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html) in the docs.

Tillerino: It's mentioned for four bugs, but applies to more. Tbh, I think I'll back off. This needs to be straightened out by somebody who is deeper in the whole framework.

Tillerino:
Closing this, since it seems that there are deeper things at work here that need to be solved by a maintainer.

The way that other generated code is ignored in `addClassAndMethod` is very intransparent and it's only documented for five bug types ([starting here](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#us-useless-suppression-on-a-field-us-useless-suppression-on-field)). I assumed this not working for TQ annotations was just an oversight (one `addClassAndMethod` missing the required call), but it's more complicated.

For me, for now, it's sufficient to just turn off spotless for all my generated code :shrug:

PR #4016, opened 2026-04-24
fix: improve EQ_DOESNT_OVERRIDE_EQUALS warning to mention symmetry risk

### LINKED ISSUE #511: Adhering to EQ_DOESNT_OVERRIDE_EQUALS requires violation of equals method contract
This issue was reported on the FindBugs SourceForge issue tracker but was closed without any discussion (see [#1379](https://sourceforge.net/p/findbugs/bugs/1379/)). I'm reposting the issue here as this has recently surfaced issues within my organisation.

The bug EQ_DOESNT_OVERRIDE_EQUALS is defined as:

> This class extends a class that defines an equals method and adds fields, but doesn't define an equals method itself. Thus, equality on instances of this class will ignore the identity of the subclass and the added fields. Be sure this is what is intended, and that you don't need to override the equals method. Even if you don't need to override the equals method, consider overriding it anyway to document the fact that the equals method for the subclass just return the result of invoking super.equals(o).
> In other words, if class A defines equals, and you declare B extends A, if B adds any fields, it should override equals.

As explained in Effective Java's item 8, doing so is harmful, since you are breaking the invariant of symmetry.
Quoting the official documentation:
> The equals method implements an equivalence relation on non-null object references:
> It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if and only if y.equals(x) returns true.

To be more precise, let's see this example (taken from Effective Java, 2nd edition itself):
```
public class Point {
  private final int x;
  private final int y;
  public Point(int x, int y) {
    this.x = x;
    this.y = y;
  }
  @Override public boolean equals(Object o) {
    if (!(o instanceof Point))
      return false;
    Point p = (Point)o;
    return p.x == x && p.y == y;
  }
  ...
  // Remainder omitted
}
```
This is completely legal, valid, and good looking code.
Now, let's assume we extend it to add a color:
```
public class ColorPoint extends Point {
  private final Color color;
  public ColorPoint(int x, int y, Color color) {
    super(x, y);
    this.color = color;
  }
  ...
  // Remainder omitted
}
```
This class would get reported for violating `EQ_DOESNT_OVERRIDE_EQUALS`
Yet doing:
```
  // Broken - violates symmetry!
  @Override public boolean equals(Object o) {
    if (!(o instanceof ColorPoint))
      return false;
    return super.equals(o) && ((ColorPoint) o).color == color;
  }
```
breaks symmetry, since:
```
new Point(2, 3).equals(new ColorPoint(2, 3, Color.BLACK)); // true!
new ColorPoint(2, 3, Color.BLACK).equals(new Point(2, 3)); // false!

```

vaidehiparashar:
Fixes #511

The previous warning message encouraged users to override equals
in subclasses without mentioning that doing so can break the
symmetry contract of equals. This change improves the message
to warn about the potential symmetry issue when using instanceof
in the superclass equals method.


vaidehiparashar:
I have raised a PR #4016 to address this issue.

vaidehiparashar:
I have raised a PR #4016 to address this issue.

ThrawnCA (approved):

hazendaz (approved):

hazendaz:
No change log as this is a clarity issue with disclaimer in docs only.

PR #4022, opened 2026-04-30
Add NUM01-J: Do not perform bitwise and arithmetic operations on the same data

samugyula:
## Summary

New detector `MixArithmeticAndBitwiseOperationsDetector` targeting [SEI CERT NUM01-J](https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/numeric-types-and-operations-num/num01-j/) violations: using both arithmetic and bitwise operations on the same data.

## Bug types

- `MABO_MIXING_ARITHMETIC_AND_BITWISE_OPERATIONS` -- reported when a method uses both arithmetic and bitwise operations on the same data. 

## Limitations

- Does not take the NUM01-J-EX1 exception into consderation; serialization/deserialization might cause false positives
- The context (numeric or bitwise) is tracked only for primitive local variables and intermediate values on stack 
- Uses intraprocedural analysis, so the noncompliant logical and arithmetic right shift examples from the description are not checked

## Detection approach

- Uses `OpcodeStackDetector` with `@CustomUserValue` annotation
- Tracks the context of values in `Item.userValue`
- The context is determined from the first operation that is used on the data
- The context is passed down by assignment and integer conversion operations

## Changes
- `MixArithmeticAndBitwiseOperationsDetector`: new detector class
- `MixArithmeticAndBitwiseOperationsDetectorTest`: new test class
- `MixArithmeticAndBitwise`: new test cases
- `AnnotationMatcherTest`: change number of expected bugs from 4 to 6 in `testPerformAnalysis`
- `Ideas_2012_01_04` test case: add `@ExpectWarning` to getLongFixed

## Evaluation
A run on Jenkins produces 3 true positives and 1 false positive (due to not handling NUM01-J-EX1)


spotbugsTestCases/src/java/bugIdeas/Ideas_2012_01_04.java
```diff
@@ -22,7 +22,7 @@ static long getLong2(byte [] b) {
         return result;
     }
 
-    @NoWarning("")
+    @ExpectWarning("")
```
JuditKnoll, line 25: This change is misleading. You can use the `@SuppressFBWarnings` annotation to suppress the hit for the freshly introduced bugtype.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsTest.java
```diff
@@ -0,0 +1,30 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+public class MixArithmeticAndBitwiseOperationsTest extends AbstractIntegrationTest {
```
JuditKnoll, line 6: Can you please remove the public access modifier from this class? With Junit Jupiter it's no longer necessary.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private static final String MABO_BUG_TYPE = "MABO_MIXING_ARITHMETIC_AND_BITWISE_OPERATIONS";
```
JuditKnoll, line 8: Since there is only one test, this can be inside the test function, doesn't need to be on the class level.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/filter/AnnotationMatcherTest.java
```diff
@@ -153,7 +153,7 @@ void testPerformAnalysis(SpotBugsRunner spotbugs) {
                 .filter(bugInstanceMatcher::match)
                 .count();
 
-        assertEquals(4, numberOfMatchedBugs);
+        assertEquals(6, numberOfMatchedBugs);
```
JuditKnoll, line 156: This change looks unrelated. Can you please share some insight about it?

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9104,29 @@ object explicitly.</p>
...
+      <![CDATA[
+            <p>
+            Integer variables are frequently intended to represent either a numeric value or a bit collection.
+            Numeric values must be exclusively operated on using arithmetic operations,
+            whereas bit collections must be exclusively operated on using bitwise operations.
+            </p>
+            <p>
+            Performing bitwise and arithmetic operations on the same data can obscure the purpose of
+            the data stored in the variable and often indicates confusion regarding that purpose.
+            Although such operations are valid and will compile, they can reduce code readability.
```
JuditKnoll, line 9123: What should a programmer do when they encounter this bug? Can you please provide a suggestion here?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsDetector.java
```diff
@@ -0,0 +1,212 @@
...
+
+import edu.umd.cs.findbugs.BugAccumulator;
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.OpcodeStack;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import org.apache.bcel.Const;
+import org.apache.bcel.classfile.*;
+
+@OpcodeStack.CustomUserValue
```
JuditKnoll, line 30: Can you please solve this problem without using the `@OpcodeStack.CustomUseValue` annotation and changing the opcode stack? It's not wide-spread in the codebase (I could only find one other detector using it), makes the code hard to read.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsDetector.java
```diff
@@ -0,0 +1,212 @@
...
+        } else if (isArithmeticOpcode(seen)
+                || isBitwiseOpcode(seen)) {
+            if (stack.getStackDepth() < 1) {
+                return;
+            }
+            OpcodeStack.Item item = stack.getStackItem(0);
+            if (isArithmeticOpcode(seen)) {
+                item.setUserValue('A');
+            } else if (isBitwiseOpcode(seen)) {
+                item.setUserValue('B');
```
JuditKnoll, line 114: This values are not really descriptive. Are they really necessary?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsDetector.java
```diff
@@ -0,0 +1,212 @@
...
+        case Const.LSHR:
+        case Const.LUSHR:
+            return true;
+        default:
+            return false;
+        }
+    }
+
+    private boolean isArithmeticOpcode(int seen) {
+        return isUnaryArithmeticOpcode(seen) || isBinaryArithmeticOpcode(seen);
```
JuditKnoll, line 194: These two functions are only used once here. Instead of having two functions with two separate switches on the same field, there could be only one switch here.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MixArithmeticAndBitwiseOperationsDetector.java
```diff
@@ -0,0 +1,212 @@
...
+            OpcodeStack.Item lv = stack.getLVValue(slot);
+            Character t = (Character) lv.getUserValue();
+            if (t != null && t == 'B') {
+                bugAccumulator.accumulateBug(new BugInstance(this, "MABO_MIXING_ARITHMETIC_AND_BITWISE_OPERATIONS", NORMAL_PRIORITY)
+                        .addClassAndMethod(this).addString("arithmetic").addString("bitwise").addSourceLine(this),
+                        this);
+            }
+        } else if (isArithmeticOpcode(seen) || isBitwiseOpcode(seen)) {
+
+            int start = 1, end = 2, depth = 2;
```
JuditKnoll, line 58: These look like magic numbers. Can you please share where these values come from? Are you sure these will be the values of them every time?

spotbugsTestCases/src/java/MixArithmeticAndBitwise.java
```diff
@@ -0,0 +1,166 @@
+public class MixArithmeticAndBitwise {
```
JuditKnoll, line 1: Can you please add a few more test cases? E.g a function where there are two fields: one with pure arithmetic, one with pure bitwise operations.

JuditKnoll (changes requested):

aoto-tech:
@JuditKnoll Since this PR has been open for a while, would you be open to me picking it up?

I'd first like to see whether the existing `CustomUserValue` approach can be made clear and reliable. My understanding is that your concern is mainly readability and maintainability; please correct me if there are broader concerns with this approach.

I would propose:

1. Replacing `'A'` and `'B'` with descriptive names such as `ARITHMETIC` and `BITWISE`.
2. Making the tracking of how values are used explicit across copies, reassignments, and control-flow merges.
3. Addressing the missed detection when the same value is used in separate expressions, such as `x & 1` and `x + 1`. I observed this with the PR's detector running on SpotBugs 4.9.8, but would verify it on the PR branch first.
4. Adding tests for these cases, including cases that should not produce warnings.

Would this be an acceptable direction, or would you prefer an approach without `CustomUserValue`?

JuditKnoll:
> @JuditKnoll Since this PR has been open for a while, would you be open to me picking it up?
> 
> I'd first like to see whether the existing `CustomUserValue` approach can be made clear and reliable. My understanding is that your concern is mainly readability and maintainability; please correct me if there are broader concerns with this approach.
> 
> I would propose:
> 
> 1. Replacing `'A'` and `'B'` with descriptive names such as `ARITHMETIC` and `BITWISE`.
> 2. Making the tracking of how values are used explicit across copies, reassignments, and control-flow merges.
> 3. Addressing the missed detection when the same value is used in separate expressions, such as `x & 1` and `x + 1`. I observed this with the PR's detector running on SpotBugs 4.9.8, but would verify it on the PR branch first.
> 4. Adding tests for these cases, including cases that should not produce warnings.
> 
> Would this be an acceptable direction, or would you prefer an approach without `CustomUserValue`?

@aoto-tech As the author didn't answer for quite some time, feel free to take it up and open a new PR with mentioning him as a co-author. If that gets merged, this PR will be closed with it.
The `@CustomUserValue` approach modifies the stack, this itself makes the code hard to read and may lead to some unexpected consequences, so I would much prefer an approach without using it.

PR #4023, opened 2026-04-30
Add FIO15-J: Do not reset a servlet's output stream after committing it

samugyula:
## Summary

New detector `IllegalServletOutputOperationAfterCommit` targeting [SEI CERT FIO15-J](https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/input-output-fio/fio15-j/) violations: performing operations on a committed `HttpServletResponse` that are either illegal or cause silent failures

## Bug types

- `ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT` -- reported when a method:
-- Calls a method on a committed response that modifies its status code or header
-- Resets the response or the buffer after the response is committed
-- Calls `sendError` or `sendRedirect` after calling `getWriter` or `getOutputStream`
-- Manually flushes an already flushed buffer. This is reported with low priority.

## Limitations

- Assumes one response object per method
- Does not check if a committing method call is in a loop
- Does not check if a violating method call happens on the same execution path as the committing call. To moderate the number of false positives, some cases are treated heuristically, e.g. early return after a committing call inside an `if-else` or `catch`.
- Assumes that once the writer/stream for the response is opened, it is the only such object open

## Detection approach

- Uses `BytecodeScanningDetector`
- Uses a state machine to track if the response is committed and the output builder stream/writer is accessed
- Checks if a response object implements / extends `ServletResponse` from `javax.servlet` or `jakarta.servlet`
- If a `ServletOutputStream` (or subtype) is passed to a method, it assumes it belongs to a response and tracks the behavior. If a `Writer` or `OutputStream` (or subtype) is passed, it is treated as a responses output only if a `ServletResponse` (or subtype) is also passed.
- The inline copies of `finally` handlers are skipped during the bytecode scan

## Changes
- `IllegalServletOutputOperationAfterCommit`: new detector class
- `IllegalServletOutputOperationAfterCommitTest`: new test class
- `IllegalServletOpAfterCommit`: new test cases
- `MyAuditResponse`: new auxiliary class for the test cases

## Evaluation
A run on Jenkins produces 3 false positives in two methods where there is a `sendError` or `sendRedirect` in an `if` branch and a method call modifying the header in the corresponding `else` branch.


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommitTest.java
```diff
@@ -0,0 +1,33 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+public class IllegalServletOutputOperationAfterCommitTest extends AbstractIntegrationTest {
```
JuditKnoll: Can you please remove the `public` access modifier from this class, as with Junit Jupiter it's unnecessary?

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
       </Details>
   </BugPattern>
 
+  <BugPattern type="ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT">
+    <ShortDescription>Method resets or reflushes servlet after commit</ShortDescription>
+    <LongDescription>The call of {2} in method {1} {3} at line {4}</LongDescription>
```
JuditKnoll: This long description doesn't describe the problem. Please make sure that it's easy to understand without the short description and the details.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+
+        boolean hasServletStream = false;
+        boolean hasGenericOut = false;
+        boolean hasResponse = false;
+
+        for (Type argType : obj.getArgumentTypes()) {
+            if (argType instanceof BasicType) {
+                continue;
+            }
+            String className = argType.toString().replace('.', '/');
```
JuditKnoll: There are util functions in `edu.umd.cs.findbugs.util.ClassName` to convert between slashed and dotted classnames. Can you please use those?

JuditKnoll: Can you please use `getClassName()` instead of `toString()` here?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+    public void visitMethod(Method obj) {
+
+        setInitialState(obj);
+
+        stateSnapshots.clear();
+        isDeadEnd = false;
+
+        inlineFinallyRanges.clear();
+
+        org.apache.bcel.classfile.Code code = obj.getCode();
```
JuditKnoll: Is there a reason why the fully qualified name of `Code` is used here instead of importing it?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+
+    @Override
+    public void visitMethod(Method obj) {
+
+        setInitialState(obj);
+
+        stateSnapshots.clear();
+        isDeadEnd = false;
+
+        inlineFinallyRanges.clear();
```
JuditKnoll: These looks like resetting the state to the initial state to me. Why aren't these inside the `setInitialState()` function, which is only used here?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+    private final List<int[]> inlineFinallyRanges = new ArrayList<>();
+
+    private Subtypes2 subtypes2 = null;
+
+    private final int STATE_UNCOMMITTED_UNACCESSED = 1;
+    private final int STATE_UNCOMMITTED_ACCESSED = 2;
+    private final int STATE_COMMITTED_UNACCESSED = 3;
+    private final int STATE_COMMITTED_ACCESSED = 4;
+
+    private int currentState = STATE_UNCOMMITTED_UNACCESSED;
```
JuditKnoll: This could be an enum.

spotbugsTestCases/src/java/servletopaftercommit/IllegalServletOpAfterCommit.java
```diff
@@ -0,0 +1,289 @@
...
+
+    // =========================================================================
+    // NEGATIVE TEST CASES (COMPLIANT - Should not report a bug)
+    // =========================================================================
+
+    /**
+     * 1. The official CERT FIO15-J Compliant example.
+     */
+    public void testCompliant(HttpServletRequest request, HttpServletResponse response)
+            throws IOException, ServletException {
```
JuditKnoll: The `ServletException` can't be thrown from this function. Can you please remove it from the throws declaration?

spotbugsTestCases/src/java/servletopaftercommit/IllegalServletOpAfterCommit.java
```diff
@@ -0,0 +1,289 @@
...
+public class IllegalServletOpAfterCommit {
+
+    // =========================================================================
+    // NEGATIVE TEST CASES (COMPLIANT - Should not report a bug)
+    // =========================================================================
+
+    /**
+     * 1. The official CERT FIO15-J Compliant example.
+     */
+    public void testCompliant(HttpServletRequest request, HttpServletResponse response)
```
JuditKnoll: The first parameter is unused here and in several other examples. Is it necessary? If not, can you please remove them?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+        boolean hasResponse = false;
+
+        for (Type argType : obj.getArgumentTypes()) {
+            if (argType instanceof BasicType) {
+                continue;
+            }
+            String className = argType.toString().replace('.', '/');
+            if (className.contains("[]")) {
+                continue;
+            }
```
JuditKnoll: This and the `if` in line 94 has the same body. Can you please combine them, so there is only one continue in this for cycle?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+            String methodName = getNameConstantOperand();
+
+            boolean isResponse = isResponseSafe(className);
+            boolean isStreamOrWriter = isOutSafe(className);
+
+            if (!isResponse && !isStreamOrWriter) {
+                return;
+            }
+
+            boolean isFlushOrClose = "flush".equals(methodName) || "close".equals(methodName);
```
JuditKnoll: Can you please check the function's signature as well?

JuditKnoll: In the `checkAndReport()` function the `flushBuffer` function is also checked. Is that not necessary to check here?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+            int priority = NORMAL_PRIORITY;
+            String msg = "";
+
+            switch (currentState) {
+            case STATE_UNCOMMITTED_ACCESSED:
+                msg = "sends error or redirects after initializing the output builder object";
+                break;
+            case STATE_COMMITTED_UNACCESSED:
+            case STATE_COMMITTED_ACCESSED:
+                if (methodName.equals("flush") || methodName.equals("flushBuffer")) {
```
JuditKnoll: Can you please check the signature of the function?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+            bugAccumulator.accumulateBug(new BugInstance(this,
+                    "ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT", priority)
+                    .addClassAndMethod(this).addCalledMethod(getXMethodOperand())
+                    .addString(msg)
+                    .addSourceLine(this), this);
+
+        }
+    }
+
+    private boolean isValidStateChange(String methodName) {
```
JuditKnoll: The name of this function is misleading. It indicates that this function only checks and doesn't modify the state, but it changes.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+        ClassDescriptor targetClass = DescriptorFactory.createClassDescriptor(targetClassName);
+        try {
+            for (ClassDescriptor allowed : allowedTypes) {
+                if (subtypes2.isSubtype(targetClass, allowed)) {
+                    return true;
+                }
+            }
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
```
JuditKnoll, line 349: Please log this exception!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+                "setContentLengthLong".equals(methodName) ||
+                "setLocale".equals(methodName);
+    }
+
+    private boolean isStreamOrWriterGetter(String methodName) {
+        return "getOutputStream".equals(methodName) ||
+                "getWriter".equals(methodName);
+    }
+
+    private boolean isErrorOrRedirect(String methodName) {
```
JuditKnoll: This function could be static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+                "setHeader".equals(methodName) ||
+                "setIntHeader".equals(methodName) ||
+                "setStatus".equals(methodName) ||
+                "setDateHeader".equals(methodName) ||
+                "setContentLength".equals(methodName) ||
+                "setContentLengthLong".equals(methodName) ||
+                "setLocale".equals(methodName);
+    }
+
+    private boolean isStreamOrWriterGetter(String methodName) {
```
JuditKnoll: This function could be static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+        return false;
+    }
+
+    private boolean isStrictIllegalPostCommitResponseMethod(String methodName) {
+        return isStreamOrWriterGetter(methodName) ||
+                isRelaxedIllegalPostCommitResponseMethod(methodName) ||
+                isErrorOrRedirect(methodName);
+    }
+
+    private boolean isRelaxedIllegalPostCommitResponseMethod(String methodName) {
```
JuditKnoll: This function could be static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+                currentState = STATE_UNCOMMITTED_ACCESSED;
+            }
+            return true;
+
+        case STATE_COMMITTED_ACCESSED:
+            if (isStrictIllegalPostCommitResponseMethod(methodName)) {
+                return false;
+            } else if ("flushBuffer".equals(methodName) || "flush".equals(methodName)) {
+                return false;
+            }
```
JuditKnoll: Can you please combine these two `if`s, which have the same body?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+    private final ClassDescriptor javaxResponse = DescriptorFactory.createClassDescriptor("javax/servlet/ServletResponse");
+    private final ClassDescriptor jakartaResponse = DescriptorFactory.createClassDescriptor("jakarta/servlet/ServletResponse");
+    private final ClassDescriptor javaOutputStream = DescriptorFactory.createClassDescriptor("java/io/OutputStream");
+    private final ClassDescriptor javaWriter = DescriptorFactory.createClassDescriptor("java/io/Writer");
+    ClassDescriptor servletOutJavax = DescriptorFactory.createClassDescriptor("javax/servlet/ServletOutputStream");
+    ClassDescriptor servletOutJakarta = DescriptorFactory.createClassDescriptor("jakarta/servlet/ServletOutputStream");
+
+    private final Map<String, Boolean> responseTypeCache = new HashMap<>();
+    private final Map<String, Boolean> servletStreamTypeCache = new HashMap<>();
+    private final Map<String, Boolean> genericOutTypeCache = new HashMap<>();
```
JuditKnoll, line 62: As far as I can see these 3 cache maps are only used once when filled with `computeIfAbsent`, but the content is never queried. Are these fields necessary?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+        if ("java/io/OutputStream".equals(className) ||
+                "java/io/Writer".equals(className) ||
+                "java/io/PrintWriter".equals(className)) {
+            return true;
+        }
+        return genericOutTypeCache.computeIfAbsent(className,
+                k -> isSubtypeSafe(k, javaOutputStream, javaWriter));
+    }
+
+    private boolean isOutSafe(String className) {
```
JuditKnoll: Can you please rename this function as it doesn't describe well what it does?
Or since it's only used once, and it's a one-liner, it could be even deleted.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        if (isIf(seen) || isSwitch(seen) || seen == Const.GOTO || seen == Const.GOTO_W) {
+            int target = getBranchTarget();
+            if (target > getPC()) {
+                stateSnapshots.put(target, currentState);
+            }
+        }
+
+        isDeadEnd = isReturn(seen);
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
       </Details>
   </BugPattern>
 
+  <BugPattern type="ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT">
+    <ShortDescription>Method resets or reflushes servlet after commit</ShortDescription>
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,377 @@
...
+import java.util.*;
+
+public class IllegalServletOutputOperationAfterCommit extends BytecodeScanningDetector {
+
+    private final ClassDescriptor javaxResponse = DescriptorFactory.createClassDescriptor("javax/servlet/ServletResponse");
+    private final ClassDescriptor jakartaResponse = DescriptorFactory.createClassDescriptor("jakarta/servlet/ServletResponse");
+    private final ClassDescriptor javaOutputStream = DescriptorFactory.createClassDescriptor("java/io/OutputStream");
+    private final ClassDescriptor javaWriter = DescriptorFactory.createClassDescriptor("java/io/Writer");
+    ClassDescriptor servletOutJavax = DescriptorFactory.createClassDescriptor("javax/servlet/ServletOutputStream");
+    ClassDescriptor servletOutJakarta = DescriptorFactory.createClassDescriptor("jakarta/servlet/ServletOutputStream");
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/etc/messages.xml
```diff
@@ -1742,6 +1742,21 @@ factory pattern to create these objects.</p>
...
+    <Details>
+      <![CDATA[
+            <p>Detector for finding instaces of illegal operations on a HTTP response after its committed
+    </p>
+    <p>
+		    Methods of HttpServletResponse that modify the responses status code or header
+		    should not be called after the response is committed, since these are already sent at this point.
+		    New data should not be added to the response, since it may be ignored (if sendError or
+		    sendRedirect caused the commit), or could mask an Exception thrown by the first commit
+		    by triggering another Exception.
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
...
+	repeatedly reflush the output after the response body has been sent, or to call
+	HttpServletRespone.sendError or HttpServletResponse.sendRedirect after the builder
+	object of the response body has been initialized by ServletResponse.getWriter or
+	ServletResponse.getOutputStream.
+	</p>
+	<p>
+	  See SEI CERT rule <a href="https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/input-output-fio/fio15-j">
+	  FIO15-J. Do not reset a servlet's output stream after committing it</a>.
+	</p>
+            </p>
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
...
+    <ShortDescription>Method resets or reflushes servlet after commit</ShortDescription>
+    <LongDescription>The call of {2} in method {1} {3} at line {4}</LongDescription>
+    <Details>
+      <![CDATA[
+	    <p>
+	  A HTTP response is considered to be committed after the HTTP status code and header
+	has been sent. No method that modifies the status code or the header, or that resets the
+	response or its buffer may be called after this. It is also considered bad practice to
+	repeatedly reflush the output after the response body has been sent, or to call
+	HttpServletRespone.sendError or HttpServletResponse.sendRedirect after the builder
```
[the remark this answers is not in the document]
samugyula: Done

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
...
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
+- New detector `IllegalServletOutputOperationAfterCommit` for new bug type `ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT`. This detector reports a bug if a method modifies a `HttpServletResponse` objects header or status code after the response is committed, resets the response after its committed, calls `sendError` or `sendRedirect` after the output builder has been initialized, or repeatedly flushes the buffer. (See [SEI CERT rule FIO15-J](https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/input-output-fio/fio15-j/))
```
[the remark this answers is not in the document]
samugyula: Done

spotbugs/etc/messages.xml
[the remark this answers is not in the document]
samugyula: Done

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommitTest.java
samugyula: I removed the public access modifier.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I switched to using `ClassName.toSlashedClassName`.

samugyula: I switched to using `getClassName()`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: No, sorry, I switched to importing `Code`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I originally intended `setInitialState` to only handle the setting of the initial state for the state machine (that is, selecting the starting state from the 4 possible states). Now I moved the above mentioned 3 lines there as well. Or should I maybe move the whole body of `setInitialState` into `visitMethod`?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I refactored it into an enum.

spotbugsTestCases/src/java/servletopaftercommit/IllegalServletOpAfterCommit.java
samugyula: I removed it from here and `testNoncompliant1` and `testNoncompliant2`.

spotbugsTestCases/src/java/servletopaftercommit/IllegalServletOpAfterCommit.java
samugyula: Sorry, I removed the request objects from the signatures.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I combined the `if`s.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I added the checking of signatures.

samugyula: This boolean is only used to determine if the called method the detector is investigating is a relevant method of a `Writer` or `OutputStream`. `flushBuffer` is a method of `ServletResponse`, so it is not checked here. I also decided to handle `flushBuffer` differently, because I wanted to allow the response to be committed by sending only the metadata (headers and status) before the output builder object is initialized, and sending the actual response body in a subsequent flush.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I added the checking of sigantures.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I renamed the function to `doStateChangeIfValid`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I added the logging, but I'm not entirely sure if I've done it properly. I followed the example of `edu.umd.cs.findbugs.log.ProfileSummary`.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I've made the function static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I've made the function static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I've made the function static.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: I've combined the `if`s.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: `isResponseSafe`, `isServletStreamSafe`, and `isGenericOutSafe` are called in `sawOpcode`, and these potentially call `computeIfAbsent` on the cache maps. So I don't think they are only used once.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
samugyula: You're right, I deleted the function.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+
+    private boolean isResponseSafe(String className) {
+        if ("javax/servlet/ServletResponse".equals(className) ||
+                "jakarta/servlet/ServletResponse".equals(className) ||
+                "javax/servlet/http/HttpServletResponse".equals(className) ||
+                "jakarta/servlet/http/HttpServletResponse".equals(className) ||
+                "javax/servlet/ServletResponseWrapper".equals(className) ||
+                "jakarta/servlet/ServletResponseWrapper".equals(className) ||
+                "javax/servlet/http/HttpServletResponseWrapper".equals(className) ||
+                "jakarta/servlet/http/HttpServletResponseWrapper".equals(className)) {
```
JuditKnoll, line 310: Can you please reformat so that the `||` is at the beginning of each line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+            return true;
+        }
+
+        return responseTypeCache.computeIfAbsent(className,
+                k -> isSubtypeSafe(k, javaxResponse, jakartaResponse));
+    }
+
+    private boolean isServletStreamSafe(String className) {
+        if ("javax/servlet/ServletOutputStream".equals(className) ||
+                "jakarta/servlet/ServletOutputStream".equals(className)) {
```
JuditKnoll, line 320: Can you please reformat so that the `||` is at the beginning of the line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+            return true;
+        }
+        return servletStreamTypeCache.computeIfAbsent(className,
+                k -> isSubtypeSafe(k, servletOutJavax, servletOutJakarta));
+    }
+
+    private boolean isGenericOutSafe(String className) {
+        if ("java/io/OutputStream".equals(className) ||
+                "java/io/Writer".equals(className) ||
+                "java/io/PrintWriter".equals(className)) {
```
JuditKnoll, line 330: Can you please reformat so that the `||` is at the beginning of each line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+                    targetClassName, e);
+            return false;
+        }
+        return false;
+    }
+
+    private boolean isStrictIllegalPostCommitResponseMethod(String methodName, String methodSignature) {
+        return isStreamOrWriterGetter(methodName) ||
+                isRelaxedIllegalPostCommitResponseMethod(methodName, methodSignature) ||
+                isErrorOrRedirect(methodName);
```
JuditKnoll, line 356: Can you please reformat so that the `II` is at the beginning of each line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+                || ("setDateHeader".equals(methodName)
+                        && "(Ljava/lang/String;J)V".equals(methodSignature))
+
+                || ("setStatus".equals(methodName)
+                        && "(I)V".equals(methodSignature));
+    }
+
+    private static boolean isStreamOrWriterGetter(String methodName) {
+        return "getOutputStream".equals(methodName) ||
+                "getWriter".equals(methodName);
```
JuditKnoll, line 415: Can you please reformat so that the `II` is at the beginning of the line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    private static boolean isErrorOrRedirect(String methodName) {
+        return "sendError".equals(methodName) ||
+                "sendRedirect".equals(methodName);
```
JuditKnoll, line 420: Can you please reformat so that the `II` is at the beginning of the line and not the ending, so the formatting is consistent within the class and with other parts of the codebase?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
```diff
@@ -0,0 +1,423 @@
...
+                    msg = "flushes an already flushed buffer";
+                } else {
+                    msg = "performs illegal servlet response operation after commit";
+                }
+                break;
+            }
+            bugAccumulator.accumulateBug(new BugInstance(this,
+                    "ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT", priority)
+                    .addClassAndMethod(this).addCalledMethod(getXMethodOperand())
+                    .addString(msg)
```
JuditKnoll, line 252: The values of the added `str` string are quite different. It might be worth to consider creating separate bugpatterns for them. It's good like this as well, if you prefer like this.
If you decide to create have separate bugpatterns, don't forget to update the findbugs.xml, messages.xml, the CHANGELOG.md and the tests.

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
...
+    <Details>
+      <![CDATA[
+	    <p>
+	  An HTTP response is considered to be committed after the HTTP status code and headers
+ 	have been sent. No method that modifies the status code or the headers, or that resets the
+	response or its buffer may be called after this. It is also considered bad practice to
+	repeatedly reflush the output after the response body has been sent, or to call
+	HttpServletResponse.sendError or HttpServletResponse.sendRedirect after the builder
+	object of the response body has been initialized by ServletResponse.getWriter or
+	ServletResponse.getOutputStream.
```
JuditKnoll, line 9125: The functions could be formatted as code with the `<code>` tag.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/IllegalServletOutputOperationAfterCommit.java
JuditKnoll: Since this is inside a detector, which has an AnalysisContext, simply `AnalysisContext.reportMissingClass(e)` could be called, which is used allover the code base, so it stays consistent.

spotbugs/etc/messages.xml
```diff
@@ -1742,6 +1742,21 @@ factory pattern to create these objects.</p>
...
+    <Details>
+      <![CDATA[
+            <p>Detector for finding instances of illegal operations on an HTTP response after it is committed
+    </p>
+    <p>
+		    Methods of HttpServletResponse that modify the response's status code or header
+		    should not be called after the response is committed, since these have already been sent at this point.
+		    New data should not be added to the response, since it may be ignored (if sendError or
+		    sendRedirect caused the commit), or could mask an exception thrown by the first commit
+		    by triggering another exception.
```
JuditKnoll, line 1755: IMO it would be better, if this would be at the BugPattern's Details. If it stays here, the user won't see this when they encounter the bug.

spotbugs/etc/messages.xml
```diff
@@ -9092,6 +9107,29 @@ object explicitly.</p>
       </Details>
   </BugPattern>
 
+  <BugPattern type="ISOOAC_ILLEGAL_SERVLET_OUTPUT_OPERATION_AFTER_COMMIT">
+    <ShortDescription>Method performs an illegal servlet response operation after commit or output initialization</ShortDescription>
+    <LongDescription>The call to {2} in method {1} {3} at line {4}</LongDescription>
+    <Details>
```
JuditKnoll, line 9116: What should a programmer do when they encounter this bug? Can you please add a short suggestion what to do?

spotbugsTestCases/src/java/servletopaftercommit/IllegalServletOpAfterCommit.java
```diff
@@ -0,0 +1,287 @@
+package servletopaftercommit;
+
+import jakarta.servlet.ServletOutputStream;
+import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.HttpServletResponseWrapper;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+public class IllegalServletOpAfterCommit {
```
JuditKnoll, line 9: Can you please add a negative test case, where there are two separate responses or outputstreams, and one has an illegal servlet operation after the other was commited?

JuditKnoll (commented):
When opening the PR you mentioned that you found some false positives:
> A run on Jenkins produces 3 false positives in two methods where there is a `sendError` or `sendRedirect` in an `if` branch and a method call modifying the header in the corresponding `else` branch.

Can you please add these or even better simplified version of these to the tests and try to filter them out?

hazendaz:
@samugyula if still interested please address comments.

PR #4027, opened 2026-05-03
Fixes #4003 : flag EI_EXPOSE_REP in package-private classes overriding public super-type methods

### LINKED ISSUE #4003: Public interface and shadow copies lead to false negatives in EI_EXPOSE_REP(2)
**Describe the bug**
I find some false negatives in the `FindReturnRef` detector responsible for the `EI_EXPOSE_REP` and `EI_EXPOSE_REP2` vulnerability rules. 
The rule ignores package-private classes. If a package-private class implements a `public interface` and exposes its internal mutable state through overridden public methods, SpotBugs fails to flag it.

And, I also find that the rule seems to assume that invoking `.clone()` on an array mitigates representation exposure. However, for multi-dimensional arrays (e.g., `byte[][]`) or arrays of mutable objects (e.g, HashTable), `.clone()` may only perform a *shallow copy*. The inner mutable objects are still exposed by reference, but now the rule downgrades or the finding to `LOW_PRIORITY` (Priority 3). I am not sure if this is expected.

**To Reproduce**
```java
// CASE 1: Scope Filtering False Negative
// SpotBugs ignores package-private classes, missing exposures
// that escape through public interfaces to external callers.
public interface Provider { 
    byte[] getData(); 
}
class HiddenProvider implements Provider {
    private byte[] data = {1, 2, 3};
    @Override
    public byte[] getData() {
        // FALSE NEGATIVE: Returns direct reference.
        // Missed entirely because the declaring class is not 'public'.
        return data; 
    }
}
// CASE 2 & 3: Shallow Clone 
public class VulnerableManager {
    private byte[][] matrix;
    private Hashtable<String, byte[]> cache;
    // CASE 2: Complex Array Shallow Clone
    public byte[][] getMatrix() {
        // Downgraded to Priority 3 / Low:
        return matrix.clone(); 
    }
    // CASE 3: Mutable Collection Shallow Clone
    public Hashtable<String, byte[]> getCache() {
        // FALSE NEGATIVE:
        // Both keys in the Hashtable are shallow copies.
        return (Hashtable<String, byte[]>) cache.clone(); 
    }
}
```

**Expected behavior**

**Environment**
- Java Compiler: javac 21.0.10
- Java version: JDK 21
- SpotBugs Version: 4.9.8

mirkoalicastro:
`FindReturnRef` used to skip any non-public class (`getThisClass().isPublic() && obj.isPublic()`), so that a package-private class that implements a public interface and exposes a private mutable field through an overridden public method was never flagged, even if the exposure is reachable from outside the package via the public super-type.

With these changes, the detector still skips non-public methods, but a package-private declaring class is analysed when the public method overrides one declared in an externally-visible super-type. Static methods are excluded from the new path since they aren't subject to virtual dispatch through a super-type.

Added a unit test reproducing the issue #4003 . Without the changes, the unit test would fail and the violation would not be detected.

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
@@ -287,6 +287,15 @@ class FindReturnRefTest extends AbstractIntegrationTest {
     }
 
     @Test
+    void testPackagePrivateImplOfPublicInterface() {
+        performAnalysis("exposemutable/PackagePrivateImplProvider.class",
+                "exposemutable/PackagePrivateHiddenProvider.class");
+
+        assertBugTypeCount("EI_EXPOSE_REP", 1);
+        assertBugInMethodAtField("EI_EXPOSE_REP", "PackagePrivateHiddenProvider", "getData", "data");
+    }
+
+    @Test
     @DisabledOnJre({ JRE.JAVA_8, JRE.JAVA_11 })
     @Disabled
     void testUnmodifiableRecord() {
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
@@ -291,8 +291,32 @@ class FindReturnRefTest extends AbstractIntegrationTest {
         performAnalysis("exposemutable/PackagePrivateImplProvider.class",
                 "exposemutable/PackagePrivateHiddenProvider.class");
 
+        assertBugTypeCount("EI_EXPOSE_BUF", 0);
+        assertBugTypeCount("EI_EXPOSE_BUF2", 0);
         assertBugTypeCount("EI_EXPOSE_REP", 1);
+        assertBugTypeCount("EI_EXPOSE_REP2", 0);
+        assertBugTypeCount("EI_EXPOSE_STATIC_BUF2", 0);
+        assertBugTypeCount("EI_EXPOSE_STATIC_REP2", 0);
+        assertBugTypeCount("MS_EXPOSE_BUF", 0);
+        assertBugTypeCount("MS_EXPOSE_REP", 0);
+
         assertBugInMethodAtField("EI_EXPOSE_REP", "PackagePrivateHiddenProvider", "getData", "data");
+    }
+
+    @Test
+    void testPackagePrivateImplWithDefensiveCopy() {
+        performAnalysis("exposemutable/PackagePrivateImplProvider.class",
+                "exposemutable/PackagePrivateImplNoBug.class");
+
+        assertNoExposeBug();
+    }
+
+    @Test
+    void testPackagePrivateImplOfPackagePrivateInterface() {
+        performAnalysis("exposemutable/PackagePrivateInterface.class",
+                "exposemutable/PackagePrivateImplOfPackagePrivateInterface.class");
+
+        assertNoExposeBug();
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -314,6 +320,23 @@ public void sawOpcode(int seen) {
...
+        XMethod xMethod = XFactory.createXMethod(getThisClass(), method);
+        for (XMethod superMethod : Hierarchy2.findSuperMethods(xMethod)) {
+            try {
+                if (superMethod.getClassDescriptor().getXClass().isPublic()) {
+                    return true;
+                }
+            } catch (CheckedAnalysisException e) {
+                AnalysisContext.logError("Error checking visibility of " + superMethod.getClassDescriptor(), e);
+            }
+        }
```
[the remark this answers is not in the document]
JuditKnoll: This is by design like this.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
```diff
@@ -152,7 +154,11 @@ private void collectData(JavaClass javaClass, Method m) {
 
     @Override
     public void visit(Method obj) {
-        check = getThisClass().isPublic() && obj.isPublic();
+        if (!obj.isPublic()) {
+            check = false;
+            return;
+        }
+        check = getThisClass().isPublic() || overridesMethodFromPublicSupertype(obj);
```
JuditKnoll: If I'm not mistaken, this could be simplified with assigning a value only once to check like this:
```
check = obj.isPublic() || getThisClass().isPublic() || overridesMethodFromPublicSupertype(obj);
```

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
```diff
@@ -286,6 +286,15 @@ void testUnmodifiableClass() {
         assertBugTypeCount("MS_EXPOSE_REP", 0);
     }
 
+    @Test
+    void testPackagePrivateImplOfPublicInterface() {
+        performAnalysis("exposemutable/PackagePrivateImplProvider.class",
+                "exposemutable/PackagePrivateHiddenProvider.class");
+
+        assertBugTypeCount("EI_EXPOSE_REP", 1);
```
JuditKnoll, line 296: Can you please also assert the expected number of other similar bugtypes (which are checked in the `assertNoExposeBug()` function) as well, even if 0 hit is expected?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
```diff
@@ -286,6 +286,15 @@ void testUnmodifiableClass() {
         assertBugTypeCount("MS_EXPOSE_REP", 0);
     }
 
+    @Test
+    void testPackagePrivateImplOfPublicInterface() {
```
JuditKnoll, line 290: Can you please add a few more tests? Both positive (there is a bug expected) and negative tests would be nice.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindReturnRef.java
mirkoalicastro: I think that using OR between all the three is too broad because it will check non-public methods in public classes. 
I've simplified it with:
```
check = obj.isPublic() && (getThisClass().isPublic() || overridesMethodFromPublicSupertype(obj));
```

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
mirkoalicastro: Added, thanks

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindReturnRefTest.java
mirkoalicastro: Done, thanks

JuditKnoll (approved):

hazendaz (approved):

PR #4039, opened 2026-05-11
Micro-optimizations: Replaces common inefficient patterns with their optimized alternative

mirkoalicastro:
While contributing to this project, IntelliJ highlighted several fragments. So I thought to contribute fixing some that lead to micro-optimizations. Here is the full list:

- Replace `"".equals(...)` with `.isEmpty()` to avoid unnecessary string comparison
- Replace string concatenation inside `StringBuilder.append()` with chained `.append()` calls to avoid intermediate allocations
- Replace Boolean.valueOf() with Boolean.parseBoolean() to avoid unnecessary boxing
- Add early break to loops that only set a boolean flag
- Replace manual compareTo implementations with `Integer.compare()`
- Replace manual array copy with `System.arraycopy()`
- Replace `String.indexOf() >= 0` with `String.contains()` for readability

Existing test suite should pass without modification.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/log/Profiler.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/log/Profiler.java
@@ -308,9 +308,7 @@ public class Profiler implements IProfiler, XMLWriteable {
                 return c1.getSimpleName().compareTo(c2.getSimpleName());
             } catch (RuntimeException e) {
                 AnalysisContext.logError("Error comparing " + c1 + " and " + c2, e);
-                int i1 = System.identityHashCode(c1);
-                int i2 = System.identityHashCode(c2);
-                return Integer.compare(i1, i2);
+                return Integer.compare(System.identityHashCode(c1), System.identityHashCode(c2));
             }
         }
     }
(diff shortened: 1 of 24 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/log/Profiler.java
```diff
@@ -310,13 +310,7 @@ public int compare(Class<?> c1, Class<?> c2) {
                 AnalysisContext.logError("Error comparing " + c1 + " and " + c2, e);
                 int i1 = System.identityHashCode(c1);
                 int i2 = System.identityHashCode(c2);
```
JuditKnoll: With the change these two variables are only used once, so they could be in the compare inline, instead of having them as separate variable. However the change itself is already an improvement, so even without this, it's okay.

JuditKnoll (approved):
Thank you for your work! One small remark, but even without it LGTM.

spotbugs/src/main/java/edu/umd/cs/findbugs/log/Profiler.java
mirkoalicastro: Good catch. I inlined the variables. Thanks!

JuditKnoll (approved):

hazendaz (approved):

PR #4056, opened 2026-05-20
Micro-optimizations: cache hot-path patterns, remove dead code, modernize API

mirkoalicastro:
Some more optimizations and cleanups 🙂

Mostly addressing:

- Cache compiled Pattern instances on hot paths in WarningSuppressor and CrossSiteScripting
- Replace Pattern with Set in SAXBugCollectionHandler
- Remove two dead RETURN_OPCODE_SET fields never used in PruneUnconditionalExceptionThrowerEdges and ClassParserUsingASM
- Replace several Collectors.joining with String.join
- Replace keySet().contains() with containsKey()
- Replace Boolean.valueOf with Boolean.parseBoolean
- Replace for loops with Arrays.fill and List.replaceAll
- Replace substring(start, end) with append(str, start, end) to avoid intermediate allocations
- Replace !isPresent() with isEmpty()
- Deduplicate branches from switch

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java
@@ -217,11 +217,6 @@ public class FindSelfComparison extends OpcodeStackDetector {
         case Const.ISUB:
             checkForSelfOperation(seen, "COMPUTATION");
             break;
-        case Const.FCMPG:
-        case Const.DCMPG:
-        case Const.DCMPL:
-        case Const.FCMPL:
-            break;
         case Const.LCMP:
         case Const.IF_ACMPEQ:
         case Const.IF_ACMPNE:
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java
@@ -228,6 +228,10 @@ public class FindSelfComparison extends OpcodeStackDetector {
         case Const.IF_ICMPGE:
             checkForSelfOperation(seen, "COMPARISON");
             break;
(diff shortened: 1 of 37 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison.java
```diff
@@ -217,11 +217,6 @@ public void sawOpcode(int seen) {
         case Const.ISUB:
             checkForSelfOperation(seen, "COMPUTATION");
             break;
-        case Const.FCMPG:
-        case Const.DCMPG:
-        case Const.DCMPL:
-        case Const.FCMPL:
-            break;
```
JuditKnoll, line 224: However this is covered by the default case, it might be worth to keep this to show, that for these no-op should be done and these were intentionally left out from the other comparisons. Or at least a comment could be added to indicate this.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSelfComparison2.java
```diff
@@ -144,11 +140,6 @@ private void analyzeMethod(ClassContext classContext, Method method) throws CFGB
             case ISUB:
                 checkForSelfOperation(classContext, location, valueNumberDataflow, "COMPUTATION", method, methodGen, sourceFile);
                 break;
-            case FCMPG:
-            case DCMPG:
-            case DCMPL:
-            case FCMPL:
-                break;
```
JuditKnoll, line 151: Same as in `FindSelfComparison`: might be worth to keep it to show it was intentionally left out from the other comparisons.

JuditKnoll (commented):
Two small comments, but even without those, this is an improvement. Thank you for your work!

JuditKnoll (approved):

PR #4059, opened 2026-05-22
Fixes #4055 : Record method-reference targets as called methods

### LINKED ISSUE #4055: False positive: Uncallable generic method
Methods are incorrectly reported as uncallable. The following code reports `UMAC` errors when it shouldn't in SpotBugs 4.9.8 (via the Gradle plugin version 6.5.4):

```
	public static <IN extends String> Collector<IN, ?, String> concatenating() {
		class Accumulator {
			final StringBuilder sb = new StringBuilder();
			void accumulate(IN s) { sb.append(s); }
			Accumulator combine(Accumulator other) { sb.append(other.sb); return this; }
			String finish() { return sb.toString(); }
		}
		return Collector.of(
			Accumulator::new,
			Accumulator::accumulate,
			Accumulator::combine,
			Accumulator::finish
		);
	}

	@Test
	void testJoining() {
		assertEquals("abc", Stream.of("a", "b", "c").collect(concatenating()));
	}
```

The errors (edited for brevity):

```
M C UMAC: Uncallable method accumulate defined in anonymous class
M C UMAC: Uncallable method combine defined in anonymous class
M C UMAC: Uncallable method finish defined in anonymous class
```

The errors disappear if I replace `IN` with `String` (ie. no generics).

mirkoalicastro:
UMAC reported methods as uncallable when they were reached only through method references. Those references compile to an invokedynamic instruction with the target stored as a CONSTANT_MethodHandle constant, never as an invoke* opcode. So, CalledMethods never registered them and XFactory.isCalled returned false.

The fix is to actually record their targets.

Fixes #4055


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CalledMethods.java
```diff
@@ -48,6 +56,46 @@ public CalledMethods(BugReporter bugReporter) {
...
+            String slashedClassName = cp.getConstantString(refCP.getClassIndex(), Const.CONSTANT_Class);
+            if (slashedClassName.charAt(0) == '[') {
+                continue;
+            }
+            ClassDescriptor c = DescriptorFactory.instance().getClassDescriptor(slashedClassName);
+            if (!subtypes2.isApplicationClass(c)) {
+                continue;
+            }
+            ConstantNameAndType nameAndType = cp.getConstant(refCP.getNameAndTypeIndex());
+            xFactory.addCalledMethod(new MethodDescriptor(slashedClassName,
```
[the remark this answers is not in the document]
mirkoalicastro: > cp.getConstant(refCP.getNameAndTypeIndex()) returns a Constant, so assigning directly to ConstantNameAndType won’t compile (other code in this repo casts here). Cast the result (or use the getConstant(index, tag) overload) before calling getName/getSignature.

Wrong. It compiles perfectly fine and the cast is redundant (`public <T extends Constant> T getConstant(..)`).

> Also consider using DescriptorFactory.instance().getMethodDescriptor(...) instead of new MethodDescriptor(...) to reuse canonical descriptors and avoid extra allocations.

Done.

ThrawnCA (approved):

iloveeclipse:
See https://github.com/spotbugs/spotbugs/pull/4084 (AI generated)

spotbugsTestCases/src/java/ghIssues/Issue4055.java
```diff
@@ -0,0 +1,57 @@
+import java.util.stream.Collector;
+import java.util.stream.Stream;
+
+/**
+ * Methods of a local class that are only reached through method references
+ * (compiled to {@code invokedynamic}) must not be reported as UMAC.
+ */
+public class Issue4055 {
```
JuditKnoll, line 10: Can you please move this class into the ghIssues package, so it's together with the other sample codes coming from github issues?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CalledMethods.java
```diff
@@ -48,6 +55,46 @@ public CalledMethods(BugReporter bugReporter) {
 
     }
 
+    /**
+     * Records the methods targeted by method-reference handles ({@code Foo::bar}) in the
+     * visited class's constant pool. Such targets appear only as {@code CONSTANT_MethodHandle}
+     * constants, never as {@code invoke*} opcodes, so {@link #sawOpcode(int)} cannot see them;
+     * without this they would look uncalled and cause UMAC false positives.
```
JuditKnoll: These calls do appear with invokedynamic opcodes. Wouldn't it be simpler to handle them there in `sawOpcode(int)`?
There are some detectors and utils (e.g. `BootstrapMethodsUtil` class or `OpcodeStack:provessInvokeDynamic` function) handling invoke dynamic, but unfortunately it isn't generally supported in SpotBugs yet. Maybe parts of these could be reused.

spotbugsTestCases/src/java/ghIssues/Issue4055.java
mirkoalicastro: Good catch, done. Thanks!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/CalledMethods.java
mirkoalicastro: I've tried to extract logic in `BootstrapMethodsUtil`. Let me know your thoughts! 🙏

JuditKnoll (approved):

PR #4061, opened 2026-05-24
Fix: UWF_NULL_FIELD false negative for cast null initializers

### LINKED ISSUE #4034: UWF_NULL_FIELD not reported when null initializer is wrapped in a redundant cast
**Describe the bug**
`UWF_NULL_FIELD` is reported for a field initialized with `null`, but the warning disappears when it is written as `(String) null`, even though both are semantically identical. The cast should not affect the analysis, so this leads to a false negative.

**To Reproduce**

```java
// BEFORE — UWF_NULL_FIELD is reported on strVal
public class DefaultValueFields {
    private String strVal = null;

    public int compute(int input) {
        int result = 0;
        if (strVal != null) {
            result += strVal.length();
        }
        return result + input;
    }
}
```

```java
// AFTER — UWF_NULL_FIELD is NOT reported, although strVal is still only ever null
public class DefaultValueFields {
    private String strVal = (String) null;

    public int compute(int input) {
        int result = 0;
        if (strVal != null) {
            result += strVal.length();
        }
        return result + input;
    }
}
```

Run:

```
spotbugs -textui -effort:max DefaultValueFields.class
```

**Expected behavior**
`UWF_NULL_FIELD` should be reported on `strVal` in both versions, since the field is only ever written as `null` (a cast of `null` to its declared type is still `null`).

**Actual behavior**
`UWF_NULL_FIELD` fires on the BEFORE version but is suppressed on the AFTER version purely because the `null` initializer is wrapped in a `(String)` cast.

**Environment**

- SpotBugs Version: 4.9.8

omer-cengel:
Fixes #4034.

**UWF_NULL_FIELD** was not reported for fields initialized with cast null values _(e.g. (String) null)_, even though they are semantically equivalent to null. This caused a false negative because the existing opcode check only treated direct **ACONST_NULL** assignments as null.

This change treats **ACONST_NULL** followed by **CHECKCAST** as a null assignment, allowing **UWF_NULL_FIELD** to be reported consistently for cast null initializers.


ThrawnCA (approved):

iloveeclipse:
See also https://github.com/spotbugs/spotbugs/pull/4082 (AI generated)

omer-cengel:
Hi @iloveeclipse !

Thanks for the pointer! I noticed #4082 addresses the same issue.

It looks like the two approaches differ slightly: one keeps the existing opcode-based logic and handles cast-null assignments (e.g. (Type) null) as an additional case, while the other relies more broadly on OpcodeStack.Item#isNull() for nullness detection.

Happy to align with whichever approach you prefer.

iloveeclipse:
> Happy to align with whichever approach you prefer.

I honestly don't have time for reviews. The other PR is AI generated and I assume the "author" (the human behind AI agent) will not be able to explain the reasoning behind. So it is fully up to you for now.

omer-cengel:
Hi @iloveeclipse !

Both approaches seem to fix the reported issue. At the moment, this PR includes a test covering the reported case, a changelog entry, and all checks are passing, while #4082 still appears to be missing a changelog entry and has pending checks. From a readiness perspective, this PR should be ready to merge. The final decision is up to you.

Thanks!

JuditKnoll (approved):

Linked the referenced issue, as GitHub did not recognize it properly since there was an additional word ("issue") between the issue link and the "fixes" keyword. (See [GitHub's keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue).)

hazendaz (approved):

PR #4062, opened 2026-05-25
Add regression test for List.copyOf EI_EXPOSE_REP false positive

### LINKED ISSUE #4033: EI_EXPOSE_REP false positive when returning a `List.copyOf` / `Set.copyOf` result
**Describe the bug**
SpotBugs reports `EI_EXPOSE_REP` for getters returning fields initialized with `List.copyOf(...)` or `Set.copyOf(...)`, even though these JDK methods return unmodifiable collections. It correctly suppresses the warning for `List.of(...)` / `Set.of(...)`, but fails to treat `copyOf` as equally immutable, leading to a false positive.

**To Reproduce**

```java
import java.util.List;
import java.util.Set;

public class CollectionHolder {

    // BEFORE — no warning
    private static final List<String> NAMES_OK = List.of("a", "b", "c");
    private static final Set<Integer>  CODES_OK = Set.of(1, 2, 3);

    // AFTER — EI_EXPOSE_REP reported on getNames() / getCodes()
    private static final List<String> NAMES = List.copyOf(List.of("a", "b", "c"));
    private static final Set<Integer>  CODES = Set.copyOf(Set.of(1, 2, 3));
    private final        List<String> tags  = List.copyOf(List.of("x", "y", "z"));

    public List<String> getNames() { return NAMES; }   // flagged
    public Set<Integer> getCodes() { return CODES; }   // flagged
    public List<String> getTags()  { return tags;  }   // flagged
}
```

Both versions return the same immutable JDK collection types (`ImmutableCollections.ListN` / `SetN`), yet only the `copyOf`-initialized version triggers `EI_EXPOSE_REP`.

**Expected behavior**
No `EI_EXPOSE_REP` warning: `List.copyOf` and `Set.copyOf` (and `Map.copyOf`) are contractually guaranteed to return unmodifiable collections, so the getter cannot expose mutable internal state.

**Actual behavior**
`EI_EXPOSE_REP` (priority 2) is raised on each getter whose backing field is initialized via `List.copyOf` / `Set.copyOf`, while the structurally equivalent `List.of` / `Set.of` case is correctly ignored.

**Environment**

- SpotBugs Version: 4.9.8

wyf027:
## Summary

- Add integration test case `Issue4033` matching the reporter's reproduction from #4033
- Assert no `EI_EXPOSE_REP` / `MS_EXPOSE_REP` warnings for fields initialized via `List.copyOf` / `Set.copyOf` (static and instance)

Fixes #4033

## Notes

The underlying behavior appears to already be handled via `IMMUTABLE_RETURNER_MAP` entries for `copyOf` and the `fieldValues` check in `FindReturnRef`. This PR adds an explicit regression test so the reported scenario stays covered.

## Test plan

- [x] `./gradlew :spotbugsTestCases:classesJava17 :spotbugs-tests:test --tests "edu.umd.cs.findbugs.detect.Issue4033Test"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
     - `USBC_UNSAFE_SYNCHRONIZATION_WITH_INHERITABLE_BACKING_COLLECTION` is reported when the backing collection of a lock can be altered by subclasses.
       (See [SEI CERT rule LCK00-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK00-J.+Use+private+final+lock+objects+to+synchronize+classes+that+may+interact+with+untrusted+code) and [SEI CERT rule LCK04-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK04-J.+Do+not+synchronize+on+a+collection+view+if+the+backing+collection+is+accessible))
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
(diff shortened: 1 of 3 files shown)


iloveeclipse:
> Fixes #4033

This PR can't fix the mentioned issue, it only adds test for it.

PR #4082, opened 2026-05-25
Fix UWF_NULL_FIELD false negative for cast null initializers (#4034)

wyf027:
## Summary
- Fix `UnreadFields` to determine null field writes from the stack value at `PUTFIELD`, not the previous opcode alone
- A redundant cast such as `(String) null` no longer incorrectly marks the field as written non-null
- Add regression test `Issue4034` for the reported scenario

## Test plan
- [x] `./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.detect.Issue4034Test"`
- [x] `./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.detect.UnreadFieldsTest"`
- [x] `./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.DetectorsTest.testAllRegressionFilesJavac"`

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4034Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4034Test.java
@@ -1,0 +1,15 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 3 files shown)


iloveeclipse:
See also https://github.com/spotbugs/spotbugs/pull/4061

PR #4089, opened 2026-05-25
fix: report NP_BOOLEAN_RETURN_NULL when null is returned via local (#4037)

### LINKED ISSUE #4037: NP_BOOLEAN_RETURN_NULL not reported when null is returned via a local variable
**Describe the bug**
`NP_BOOLEAN_RETURN_NULL` is reported for `return null;` in a `Boolean` method, but is not triggered when `null` is first assigned to a local variable and then returned. The semantics are identical, leading to a false negative.

**To Reproduce**

```java
public class ReturnNullDemo {

    // BEFORE — NP_BOOLEAN_RETURN_NULL is reported (correct)
    public Boolean isValidDirect(String input) {
        if (input == null) {
            return null;
        }
        return Boolean.TRUE;
    }

    // AFTER — NP_BOOLEAN_RETURN_NULL is NOT reported (false negative)
    public Boolean isValidIndirect(String input) {
        if (input == null) {
            Boolean result = null;
            return result;
        }
        return Boolean.TRUE;
    }

    // AFTER — same issue with a ternary-style method
    public Boolean checkStatus(int code) {
        if (code < 0) {
            Boolean result = null;
            return result;
        }
        return code == 0 ? Boolean.TRUE : Boolean.FALSE;
    }
}
```

Run:

```
spotbugs -textui -effort:max ReturnNullDemo.class
```

**Expected behavior**
`NP_BOOLEAN_RETURN_NULL` should be reported on `isValidIndirect` and `checkStatus` — the methods still return `null` from a `Boolean`-typed method.

**Actual behavior**
No `NP_BOOLEAN_RETURN_NULL` is raised for the indirect variants; only the direct `return null;` form is flagged.

**Environment**

- SpotBugs Version: 4.9.8

wyf027:
## Summary

Fixes #4037.

`TypeReturnNull` only flagged `return null;` (direct `ACONST_NULL` before `ARETURN`). When null is stored in a local variable first (`Boolean result = null; return result;`), the detector missed it.

The fix also checks whether the value on the stack at `ARETURN` is known null via opcode stack analysis (`returnValue.isNull()`). This applies to `NP_BOOLEAN_RETURN_NULL` and `NP_OPTIONAL_RETURN_NULL` (both extend `TypeReturnNull`).

## Test plan

- [x] Added `Issue4037.java` with direct return, indirect return via local, and ternary-style method
- [x] Added `Issue4037Test`
- [x] `RegressionIdeas20111219Test`, `Bug3277814Test`, `Bug3107916Test` pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
 
 ### Fixed
+- Fix `NP_BOOLEAN_RETURN_NULL` false negative when null is returned via a local variable ([#4037](https://github.com/spotbugs/spotbugs/issues/4037))
 - Fix `DM_STRING_TOSTRING` false negative when `toString()` is chained before a method call (e.g., `s.toString().toLowerCase()`); multiple occurrences in the same method are now all reported ([#3966](https://github.com/spotbugs/spotbugs/issues/3966))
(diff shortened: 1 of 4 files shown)


iloveeclipse:
@leno23 : please stop spamming this repo with PR's without even basic human review of them.

See yet another PR from your AI for same issue: https://github.com/spotbugs/spotbugs/pull/4063

wyf027:
Closing as duplicate of #4063 — same fix for #4037. Sorry for the noise.

PR #4090, opened 2026-05-25
fix: report NS_NON_SHORT_CIRCUIT when one operand is a boolean constant (#3976)

wyf027:
## Summary

Fixes #3976.

`FindNonShortCircuit` only flagged `IOR`/`IAND` in boolean context when **neither** operand was a compile-time constant. This missed cases like `true | hasSideEffect()` and `false & hasSideEffect()` where the left operand is a literal but the right still executes unconditionally.

The fix reports unless **both** operands are compile-time constants (e.g. `true | false` remains silent).

## Test plan

- [x] Added `Issue3976.java` with variable/constant left operands and `bothConstants` negative case
- [x] Added `Issue3976Test`
- [x] `FindNonShortCircuitTest` passes


iloveeclipse:
Yet anoter duplicated PR: https://github.com/spotbugs/spotbugs/pull/4069

PR #4091, opened 2026-05-25
fix: treat Objects.nonNull/isNull as null checks in OBL analysis (#4020)

### LINKED ISSUE #4020: [java] OBL_UNSATISFIED_OBLIGATION false positive when null-check uses Objects.nonNull()
**Describe the bug**
SpotBugs reports `OBL_UNSATISFIED_OBLIGATION` when a resource is closed under `Objects.nonNull(x)`, but does not report the same code when using `x != null`. Since `Objects.nonNull(x)` is semantically equivalent to `x != null`, both forms should be handled consistently. This causes a false positive in the `Objects.nonNull(...)` case.

**To Reproduce**

```java
import java.sql.Connection;
import java.sql.PreparedStatement;
import java.sql.SQLException;
import java.util.Objects;

public class CloseResourceDemo {
    public void executeQuery(Connection conn, String sql) {
        PreparedStatement lPreparedStmt = null;
        try {
            lPreparedStmt = conn.prepareStatement(sql);
            lPreparedStmt.execute();
        } catch (SQLException ex) {
            // log
        } finally {
            // BEFORE: no violation
            // if (lPreparedStmt != null) {

            // AFTER: OBL_UNSATISFIED_OBLIGATION reported
            if (Objects.nonNull(lPreparedStmt)) {
                try {
                    lPreparedStmt.close();
                } catch (SQLException pEx) {
                    // log
                }
            }
        }
    }
}
```

**Actual behavior**
Only the `Objects.nonNull(lPreparedStmt)` variant triggers `OBL_UNSATISFIED_OBLIGATION` (priority 2), while the logically identical `lPreparedStmt != null` variant is accepted.

**Environment**

- SpotBugs Version: 4.9.8

wyf027:
## Summary

Fixes #4020.

`ObligationAnalysis` removed obligations on edges where a resource reference was proven null (nothing to close), but only recognized `IFNULL`/`IFNONNULL` and reference equality checks. `Objects.nonNull(x)` compiles to `invokestatic Objects.nonNull` + `IFEQ`/`IFNE` on the boolean result, so the null branch was not recognized and `OBL_UNSATISFIED_OBLIGATION` was reported even when the resource was closed under `Objects.nonNull(stmt)`.

This change treats `Objects.nonNull`/`Objects.isNull` followed by `IFEQ`/`IFNE` equivalently to direct null checks.

## Test plan

- [x] Added `Issue4020.java` with `stmt != null` (baseline) and `Objects.nonNull(stmt)` close patterns
- [x] Added `Issue4020Test` — neither method reports `OBL_UNSATISFIED_OBLIGATION`
- [x] `FindUnsatisfiedObligationTest` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
 
 ### Fixed
+- Fix `OBL_UNSATISFIED_OBLIGATION` false positive when null-check uses `Objects.nonNull()` ([#4020](https://github.com/spotbugs/spotbugs/issues/4020))
 - Fix `DM_STRING_TOSTRING` false negative when `toString()` is chained before a method call (e.g., `s.toString().toLowerCase()`); multiple occurrences in the same method are now all reported ([#3966](https://github.com/spotbugs/spotbugs/issues/3966))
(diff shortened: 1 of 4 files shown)


iloveeclipse:
Duplicates https://github.com/spotbugs/spotbugs/pull/4064

PR #4094, opened 2026-05-26
Fix EI_EXPOSE_REP FN for anonymous class getters (#4032)

### LINKED ISSUE #4032: False negative: EI_EXPOSE_REP not reported when getter is defined in an anonymous class
**Describe the bug**
SpotBugs flags `EI_EXPOSE_REP` for a named nested class returning an internal mutable array, but misses the identical case in an anonymous class implementing the same interface. Since both return the same mutable `String[]` reference, the detector should not behave differently based on whether the class is named or anonymous.

**To Reproduce**

```java
// BEFORE — EI_EXPOSE_REP is reported (correct)
package demo.before;

public class Outer {
    private static final String[] names = new String[] {"a", "b"};

    public static Provider getProvider() {
        return new ProviderImpl();
    }

    public interface Provider {
        String[] getNames();
    }

    public static class ProviderImpl implements Provider {
        @Override
        public String[] getNames() {
            return names; // flagged: EI_EXPOSE_REP
        }
    }
}
```

```java
// AFTER — no warning reported (false negative)
package demo.after;

public class Outer {
    private static final String[] names = new String[] {"a", "b"};

    public static Provider getProvider() {
        return new Provider() {
            @Override
            public String[] getNames() {
                return names; // NOT flagged, but equally exposes internal state
            }
        };
    }

    public interface Provider {
        String[] getNames();
    }
}
```

Run:

```
spotbugs -textui -effort:max Outer.class
```

**Expected behavior**
`EI_EXPOSE_REP` should be reported on `getNames()` in the anonymous class in `AFTER.java`, matching the behavior for the named nested class in `BEFORE.java`.

**Actual behavior**
No violation is reported for the anonymous-class version, although it exposes the internal `names` array identically.

**Environment**

- SpotBugs Version: 4.9.8

wyf027:
## Summary

- When `isFieldOf` walks enclosing classes, fall back to the `NestHost` attribute if `immediateEnclosingClass` is unset (anonymous classes lack `outer_name` in `InnerClasses`).
- Add regression tests comparing named nested vs anonymous `Provider` implementations.

Fixes #4032.

## Test plan

- [x] `./gradlew :spotbugs-tests:test --tests Issue4032Test`
- [x] `./gradlew :spotbugs-tests:test --tests FindReturnRefTest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
 
 ### Fixed
+- Fix `FindReturnRef` false negative for `EI_EXPOSE_REP` when a getter in an anonymous class returns an enclosing nest host field ([#4032](https://github.com/spotbugs/spotbugs/issues/4032))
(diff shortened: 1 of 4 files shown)


iloveeclipse:
Duplicates https://github.com/spotbugs/spotbugs/pull/4065

PR #4119, opened 2026-05-26
Fix SA_FIELD_DOUBLE_ASSIGNMENT false negative for long/double fields (#3975)

wyf027:
## Summary

- Fix false negative for `SA_FIELD_DOUBLE_ASSIGNMENT` when a `long` or `double` field is assigned twice during field initialization (e.g. `boolean b = 1 > (value = 2L);`).
- Root cause: consecutive `PUTFIELD` instructions for wide primitives can be 10 bytecode offsets apart, but the detector only considered pairs within 10 PCs.
- Extend the PC window to 14 for `J`/`D` field signatures; int and other types keep the existing window of 10.

## Test plan

- [x] `Issue3975Test` — int/long/double field initializer regression
- [x] `RegressionIdeas20090616Test` — existing int constructor case still passes
- [x] `DetectorsTest`

Fixes #3975

Made with [Cursor](https://cursor.com)

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3975Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3975Test.java
@@ -1,0 +1,19 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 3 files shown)


JuditKnoll:
Duplicate of https://github.com/spotbugs/spotbugs/pull/4071.

ThrawnCA (approved):

PR #4127, opened 2026-05-26
Fix SF_SWITCH_NO_DEFAULT when default merges with last case (#3905)

### LINKED ISSUE #3905: False Negative: SF_SWITCH_NO_DEFAULT missing when the last case contains only a break statement

wyf027:
## Summary

Fixes #3905.

When the last `switch` case contains only a `break`, javac may compile the implicit default target to the same bytecode offset as that case. The existing fallthrough-based `SF_SWITCH_NO_DEFAULT` logic only fires when control falls through into a separate default offset, so these switches were missed.

This change detects the merged-default pattern at switch entry (`defaultOffset == lastCaseOffset` for non-exhaustive switches) and reports `SF_SWITCH_NO_DEFAULT`.

## Test plan

- [x] Added `Issue3905` / `Issue3905Test` (assignment in last case + break-only last case both report)
- [x] `./gradlew :spotbugs-tests:test --tests edu.umd.cs.findbugs.detect.Issue3905Test`
- [x] `./gradlew :spotbugs-tests:test --tests edu.umd.cs.findbugs.DetectorsTest`

Fixes #3905

Made with [Cursor](https://cursor.com)

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3905Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3905Test.java
@@ -1,0 +1,15 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue3905Test extends AbstractIntegrationTest {
+
+    @Test
+    void testSwitchNoDefaultWithEmptyLastCase() {
+        performAnalysis("ghIssues/Issue3905.class");
+
+        assertBugInMethod("SF_SWITCH_NO_DEFAULT", "ghIssues.Issue3905", "testWithAssignmentInLastCase");
+        assertBugInMethod("SF_SWITCH_NO_DEFAULT", "ghIssues.Issue3905", "testWithBreakOnlyInLastCase");
+    }
+}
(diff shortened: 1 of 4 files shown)


JuditKnoll:
Duplicate of https://github.com/spotbugs/spotbugs/pull/3944.
Build fails on Java 21 and 25. The cause is probably the same as there.

wyf027:
Pushed a follow-up CI fix in 5f89d2f. The merged-last-case no-default compensation is now limited to non-enum switches, so it still covers the Issue3905 `int` switch cases but avoids reintroducing the Issue3572 false positive for enum switches that have an explicit `default` compiled to the same break target as the last case. Waiting on the refreshed CI/docs statuses now.

PR #4144, opened 2026-06-05
Fixes #4138 : Detect yoda-style null guards in lazy-init and related detectors

### LINKED ISSUE #4138: False negative: `LI_LAZY_INIT_STATIC` not reported when the null check is written Yoda-style (`null == x`)
### SpotBugs version

### Check name (bug pattern)

`LI_LAZY_INIT_STATIC` — Incorrect lazy initialization of static field

### Description

The detector recognizes the unsynchronized lazy-initialization idiom only when the guard is written as `field == null`. Rewriting the *same* guard in Yoda style `null == field` — a semantics-preserving, symmetric reference comparison — makes the warning disappear, even though the lazy-init logic, the field, and the runtime behavior are unchanged.

### Reproducer

**Before:**

```java
public static Object getFoo() {
    if (foo == null)                   // reported
        foo = new Object();
    return foo;
}
public static String[] getY() {
    if (y == null) {                   // reported
        y = new String[] { "a", "b", "c", "d", "e" };
    }
    return y;
}
```

**After:**

```java
public static Object getFoo() {
    if (foo == null)                   // not reported
        foo = new Object();
    return foo;
}
public static String[] getY() {
    if (y == null) {                   // not reported
        y = new String[] { "a", "b", "c", "d", "e" };
    }
    return y;
}
```

### Expected behavior

`LI_LAZY_INIT_STATIC` should fire identically on both operand orderings of the null check, since `field == null` and `null == field` are the same condition.

### Actual behavior

- BEFORE: **2** reports (`LazyInit.foo` in `getFoo()`, `LazyInit.y` in `getY()`).
- AFTER:  **0** reports.

mirkoalicastro:
Several detectors are missing the null-check idiom when written in Yoda-style (null == field).

This PR fixes #4138 and extends the fix to other detectors.


spotbugs/src/main/java/edu/umd/cs/findbugs/ba/bcp/IfNull.java
```diff
@@ -52,9 +57,20 @@ public MatchResult match(InstructionHandle handle, ConstantPoolGen cpg, ValueNum
...
 
+    private static boolean isReferenceComparisonWithNull(InstructionHandle handle) {
+        Instruction ins = handle.getInstruction();
+        if (!(ins instanceof IF_ACMPEQ || ins instanceof IF_ACMPNE)) {
+            return false;
+        }
+        InstructionHandle value = handle.getPrev();
+        InstructionHandle nullConstant = value != null ? value.getPrev() : null;
+        return nullConstant != null && nullConstant.getInstruction() instanceof ACONST_NULL;
+    }
```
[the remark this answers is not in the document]
mirkoalicastro: Fixed.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/LazyInitTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class LazyInitTest extends AbstractIntegrationTest {
+
+    @Test
+    void fieldEqualsNull() {
+        performAnalysis("LazyInit.class");
```
JuditKnoll, line 11: Good catch! The class tested here [LazyInit](https://github.com/spotbugs/spotbugs/blob/master/spotbugsTestCases/src/java/LazyInit.java) was created much earlier, but it wasn't tested before this, not even with the annotations.
(Contrary to the comment before `LazyInit::getBar`, that one is correctly not reported since `bar` is volatile.)

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * equivalent Yoda-style guard ({@code null == field}), which javac compiles to
+ * {@code IF_ACMPEQ}/{@code IF_ACMPNE}. These tests pin the expected behavior:
+ * the Yoda form must be reported just like {@code field == null}.
+ *
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/4138">#4138</a>
+ */
+class YodaNullCheckFalseNegativeTest extends AbstractIntegrationTest {
+
+    @Test
+    void doubleCheck() {
```
JuditKnoll, line 19: Can you please assert the number of relevant bugtypes as well in each test?

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - New detector `FindIncreasedAccessibilityOfMethods` for new bug type `IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY`. This detector reports a bug if a class increases the accessibility of overridden or hidden methods. (See [SEI CERT rule MET04-J](https://wiki.sei.cmu.edu/confluence/display/java/MET04-J.+Do+not+increase+the+accessibility+of+overridden+or+hidden+methods))
 
 ### Fixed
+- Fix `LI_LAZY_INIT_STATIC` false negative when the null guard is written in yoda-style (`null == field`) ([#4144](https://github.com/spotbugs/spotbugs/pull/4144))
+- Fix `DC_DOUBLECHECK`, `NP_SYNC_AND_NULL_CHECK_FIELD` and `SP_SPIN_ON_FIELD` false negatives when the null guard is written in yoda-style (`null == field`) ([#4144](https://github.com/spotbugs/spotbugs/pull/4144))
```
JuditKnoll: As there was a version released, can you please move this to unreleased?
These two changelog entries could be combined into one, but it's good like this as well.

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java
```diff
@@ -402,6 +402,62 @@ public int getPrevOpcode(int offset) {
...
 
+    /**
+     * Is the given opcode an {@code IF_ACMPEQ}/{@code IF_ACMPNE} that compares a
+     * reference against the {@code null} constant? This is how javac compiles a
+     * Yoda-style null check such as {@code null == field}: it emits
+     * {@code aconst_null} followed by the load of the value being tested and
+     * then the reference comparison, instead of the {@code ifnull}/
+     * {@code ifnonnull} it would emit for {@code field == null}.
+     */
+    public boolean isNullComparison(int seen) {
```
JuditKnoll: As far as I can see this is only used inside DismantleBytecode, so this function could be private.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
```diff
@@ -0,0 +1,38 @@
...
+        performAnalysis("ghIssues/issue4138/DoubleCheckYoda.class");
+        assertBugInMethod("DC_DOUBLECHECK", "DoubleCheckYoda", "getNormal");
+        assertBugInMethod("DC_DOUBLECHECK", "DoubleCheckYoda", "getYoda");
```
JuditKnoll: As far as I can see, there is a field added to the `DC_DOUBLECHECK` bug as well ([see](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindDoubleCheck.java#L177-L178)). Can you please add that to the assert as well?
You can use the `AbstractIntegrationTest::assertBugInMethodAtField` function for it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Test
+    void syncAndNullCheckField() {
+        performAnalysis("ghIssues/issue4138/SyncNullCheckYoda.class");
+        assertBugInMethod("NP_SYNC_AND_NULL_CHECK_FIELD", "SyncNullCheckYoda", "useNormal");
+        assertBugInMethod("NP_SYNC_AND_NULL_CHECK_FIELD", "SyncNullCheckYoda", "useYoda");
```
JuditKnoll: As far as I can see, there is a field added to the `NP_SYNC_AND_NULL_CHECK_FIELD` bug as well ([see](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SynchronizeAndNullCheckField.java#L101-L103)). Can you please add that to the assert as well?
You can use the `AbstractIntegrationTest::assertBugInMethodAtField` function for it.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Test
+    void spinLoop() {
+        performAnalysis("ghIssues/issue4138/SpinLoopYoda.class");
+        assertBugInMethod("SP_SPIN_ON_FIELD", "SpinLoopYoda", "spinNormal");
+        assertBugInMethod("SP_SPIN_ON_FIELD", "SpinLoopYoda", "spinYoda");
```
JuditKnoll: As far as I can see, there is a field added to the `SP_SPIN_ON_FIELD` bug as well ([see](https://github.com/spotbugs/spotbugs/blob/master/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindSpinLoop.java#L99-L100)). Can you please add that to the assert as well?
You can use the `AbstractIntegrationTest::assertBugInMethodAtField` function for it.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
mirkoalicastro: done, thanks

CHANGELOG.md
mirkoalicastro: Good catch! Moved

spotbugs/src/main/java/edu/umd/cs/findbugs/visitclass/DismantleBytecode.java
mirkoalicastro: Fair enough. Done, thanks!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
mirkoalicastro: I've changed it to use `assertBugInMethodAtField`, thanks

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
mirkoalicastro: Done!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/YodaNullCheckFalseNegativeTest.java
mirkoalicastro: Done!

ThrawnCA (approved):

JuditKnoll (approved):

hazendaz (approved):

PR #4149, opened 2026-06-06
chore(build): Add cleanup action to cleanup old job history

hazendaz:
Currently github will delete the content of build results after a period of time.  What github fails to do is actually delete the jobs which then are useless and do take some space.  This new action is a long well tested action that simply runs daily to delete old jobs from showing up.

I have added no change log as not user facing, this is to just cleanup jobs that are not accessible going back a year then keeping it clean on ongoing basis at 30 days.


hazendaz:
merging now and will force execute first run.

PR #4150, opened 2026-06-06
chore(build): Update comments for commons-compress version details

hazendaz:
comment change as the issue with commons-compress is still present as of gradle 9.5.1.


hazendaz:
didn't wait for build as comment change in build file only for clarity when we see the issue again by renovate.

PR #4195, opened 2026-07-07
Remove full Eclipse IDE from build requirement

hazendaz:
Historically, full eclipse IDE was required by contributors locally to point to as well as downloading on github actions.  That bundle is large, unnecessary, results in user setup that should not exist.  Further, there were 4 repos (only 3 active) that every build would split the package to that were identical other than metadata that p2 doesn't necessarily need.  That has been streamlined to just create a single plugin set.

I have also updated our override to commons-io as still true to gradle 9.6.1 and fixed a bug when copilot raises PR due to spacing in author.

- [X ] Added an entry into `CHANGELOG.md` if you have changed SpotBugs code


hazendaz:
Set as draft for now, the build works, expectation it lays out to our other 3 repos is sound, it looks correct, I have not tested myself yet but will be doing so this evening to make sure it pulls ok.  Added @iloveeclipse to review in case there was a larger eclipse reason to make the larger metadata differences between our various release channels.

hazendaz:
will review the concerns from copilot later tonight.

ThrawnCA (approved):

hazendaz:
this will be the general goal, it doesn't yet fully work, the xml files are not populated enough yet so when trying to install its not fully showing what it should.  The direction is the right direction though overall, I'll keep this in draft until I can get it working fully.

ThrawnCA (approved):

iloveeclipse:
I'm looking into. This PR seem to have few unrelated commits and lot of intermediate commits. Can we rebase on master, remove unrelated commits and squash the rest?

iloveeclipse (changes requested):
* **Blocking issue** so far: generated update site meta-information is not OK, Eclipse doesn't show any items to install if pointing Eclipse to show what can be installed from that site.
* I see few mentions of `local.properties` in different files, this must be removed.
* Probably unrelated to this patch: I see errors while running build which indicate incomplete classpath
```
> Task :eclipsePlugin-test:spotbugsMain
  jakarta.annotation.Nullable
  jakarta.annotation.Nonnull
SpotBugs ended with exit code 3
```
* Another one, probably unrelated too? returns 1 with no obvious explanation:
```
> Task :eclipsePlugin:spotbugsMain
SpotBugs ended with exit code 1
```

ThrawnCA (approved):

hazendaz:
closing for now as not completed, will re-raise when appropriate.

PR #4201, opened 2026-07-14
Fixes #4192 : Stop flagging meaningful signed-byte comparisons

### LINKED ISSUE #4192: False positive for INT_BAD_COMPARISON_WITH_SIGNED_BYTE
**Describe the bug**
Testing that a byte value is strictly less than 127 produces "Low: Bad comparison of signed byte with 127", although that's a valid comparison.

**To Reproduce**
```java
public class Bug
{
    public static boolean isPrintable(byte b)
    {
        return b > 32 && b < 127;
    }
}
```

**Expected behavior**
Spotbugs should not report anything for `b < 127`. Only `b <= 127` or `b < 128` should trigger a diagnostic.

**Environment**
- Java Compiler: Eclipse Temurin
- Java version: JDK 17
- SpotBugs Version: 4.10.2
- How do you use SpotBugs: Maven Plugin
    - Plugin version: 4.10.2.0
    - Environment version for Plugin: Maven 3.8.4

mirkoalicastro:
`INT_BAD_COMPARISON_WITH_SIGNED_BYTE` reported a false positive for b < 127 (and b >= 127), which are valid comparisons for a signed byte, not always-true/never-true.

For the constant 127, this PR only reports the always-true (b <= 127) and never-true (b > 127) cases. Comparisons against 128 / <= -129 are unchanged.

Fixes #4192

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
@@ -1010,21 +1010,19 @@ public class DumbMethods extends OpcodeStackDetector {
                     Object constant1 = item1.getConstant();
                     if (item0.getSpecialKind() == OpcodeStack.Item.SIGNED_BYTE && constant1 instanceof Number) {
                         int v1 = ((Number) constant1).intValue();
-                        if (v1 <= -129 || v1 >= 128 || v1 == 127 && !(seen2 == Const.IF_ICMPEQ || seen2 == Const.IF_ICMPNE)) {
+                        // For a signed byte x (range -128..127), comparing against 127 is only
+                        // suspicious when it is always true (127 >= x, i.e. x <= 127) or never true
+                        // (127 < x, i.e. x > 127). The comparisons x < 127 (127 > x) and x >= 127
+                        // (127 <= x) are meaningful and must not be reported.
+                        if (v1 <= -129 || v1 >= 128 || v1 == 127 && (seen2 == Const.IF_ICMPGE || seen2 == Const.IF_ICMPLT)) {
                             int priority = HIGH_PRIORITY;
                             if (v1 == 127) {
                                 switch (seen2) {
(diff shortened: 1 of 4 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
```diff
@@ -1010,22 +1010,18 @@ else if (special == OpcodeStack.Item.HASHCODE_INT) {
                     Object constant1 = item1.getConstant();
                     if (item0.getSpecialKind() == OpcodeStack.Item.SIGNED_BYTE && constant1 instanceof Number) {
                         int v1 = ((Number) constant1).intValue();
-                        if (v1 <= -129 || v1 >= 128 || v1 == 127 && !(seen2 == Const.IF_ICMPEQ || seen2 == Const.IF_ICMPNE)) {
+                        // For 127, only report x <= 127 (always true) and x > 127 (never true);
+                        // x < 127 and x >= 127 are meaningful comparisons of a signed byte.
+                        if (v1 <= -129 || v1 >= 128 || v1 == 127 && (seen2 == Const.IF_ICMPGE || seen2 == Const.IF_ICMPLT)) {
```
JuditKnoll: Are you sure about this line change? Like this there will be some false negatives with 128 and -129 (see lines 1030-1047).

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/DumbMethods.java
mirkoalicastro: I get your point, but it's actually safe. There was implicit precedence that made it look risky, although correct. To remove the ambiguity, I've added explicit parentheses around the 127 clause. Also, I've added b < 128 and b > -129 to the tests. Thanks!

JuditKnoll (approved):

hazendaz (approved):

PR #4207, opened 2026-07-26
Fix SS_SHOULD_BE_STATIC for final constant fields initialized in initializer blocks

### LINKED ISSUE #4136: `SS_SHOULD_BE_STATIC` missed when a `final` constant field is initialized in an instance-initializer block instead of at the declaration
### SpotBugs version

### Bug pattern

[SS_SHOULD_BE_STATIC](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ss-unread-field-should-this-field-be-static-ss-should-be-static)

### Description

An unread `final` field whose value is a compile-time constant String is flagged when written at the declaration site (`final String staticIsMissingPath = "/search";`) but not when the identical constant is assigned in an instance-initializer block (`final String staticIsMissingPath; { staticIsMissingPath = "/search"; }`).

The constructor bytecode is **identical** for both forms (same `ldc "/search"; putfield staticIsMissingPath`), and the field is unread in both. The only class-file difference is the `ConstantValue` attribute, present only for the declaration-site form:

```
BEFORE  field staticIsMissingPath:  ConstantValue: String /search   <-- present
AFTER   field staticIsMissingPath:  (no ConstantValue)               <-- absent
<init>  (both):  ... ldc "/search"; putfield staticIsMissingPath ...  <-- identical
```

Note that only the `final` field triggers `SS_SHOULD_BE_STATIC`; non-`final` fields in the same class (e.g. `String fileName = "...";`) are not reported in either form, confirming the report keys off the constant/`ConstantValue` status of the `final` field. Because the relocated form drops `ConstantValue` while remaining unread and constant-valued, the detector misses it — a false negative.

### Reproducer

```java
// BEFORE
public class HardcodedURICheckSample {
    final String staticIsMissingPath = "/search";   // reported: should be static
}

// AFTER
public class HardcodedURICheckSample {
    final String staticIsMissingPath;
    { staticIsMissingPath = "/search"; }             // same constant, unread -> NOT reported
}
```

### Expected behavior

An unread `final` field assigned a compile-time constant should be flagged whether the constant is written at the declaration or in an instance-initializer block; the two forms produce identical constructor bytecode and leave the field equally unread and static-eligible.

### Actual behavior

- BEFORE: **1** report — `Unread field: HardcodedURICheckSample.staticIsMissingPath; should this field be static?`
- AFTER:  **0** reports.

omer-cengel:
Fixes https://github.com/spotbugs/spotbugs/issues/4136

Fixes a false negative for **SS_SHOULD_BE_STATIC** when a final field is assigned a compile-time constant in an instance initializer block instead of at the declaration site.

The unread-fields detector previously relied on the field's ConstantValue attribute to identify constant-valued fields. Since this attribute is not emitted for assignments performed in an instance initializer block, those fields were not recognized as constant, causing the detector to miss **SS_SHOULD_BE_STATIC**.

Update the detector to recognize final fields assigned compile-time constant values regardless of whether the assignment occurs at the declaration site or in an instance initializer block, making both initialization forms behave consistently.

Also add a regression sample and test covering the initializer-block scenario.

- [X] Added an entry into `CHANGELOG.md`


ThrawnCA (approved):

JuditKnoll (commented):
The build is failing because of format issues. Please run `./gradlew spotlessApply` to fix it.

ThrawnCA (approved):

PR #4213, opened 2026-07-29
Remove deprecated IO.close(InputStream) usages

mirkoalicastro:
This PR completes the cleanup started in #4194. These were the only spots left where the deprecate method was being used.


ThrawnCA (approved):

JuditKnoll (approved):
Thanks! LGTM!
These usages were only left in the code, because originally `IO.close(InputStream)` was removed in https://github.com/spotbugs/spotbugs/pull/3683. With this removal in these cases `IO.close(AutoCloseable)` was used. However removing `IO.close(InputStream)` was causing some issues, so it was added back as deprecated in https://github.com/spotbugs/spotbugs/pull/3756.

hazendaz (approved):

PR #4214, opened 2026-07-30
Fixes #724: @SuppressFBWarnings annotation on a method or a constructor now suppresses warnings reported in their lambdas too

### LINKED ISSUE #724: @SuppressFBWarnings on lambdas?
I'm trying to migrate from Findbugs 3.0 to Spotbugs 3.1.6. There are quite a few new `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE` warnings that are generated now. I can fix some of them easily, but others have less well-defined behavior, so I'd like to suppress the FB warnings for now.

A lot of these warnings are coming from statements inside lambdas, but there doesn't seem to be a way to apply `@SuppressFBWarnings` to a lambda. Moreover, `@SuppressFBWarnings` on the method containing the lambda also doesn't actually suppress the warning, either; I feel like it used to work in Findbugs, though...

Is there a way to suppress the warning without suppressing warnings for the entire class?

julianstengaard:
Fixes #724

Updated MethodWarningSuppressor to consider lambdas of suppressed methods/constructors too. 

Please note, that overloaded methods + different constructors can be over-suppressed since the lambda names do not specify which signature it belongs to. To keep the change small I avoided implementing a complex workaround for this (but I did add it to the test as a known limitation). Please let me know, if this is not the right decision.

- [x] ./gradlew spotlessApply build smoketest passes.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/MethodWarningSuppressor.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/MethodWarningSuppressor.java
@@ -1,5 +1,7 @@
 package edu.umd.cs.findbugs;
 
+import org.apache.bcel.Const;
+
 import edu.umd.cs.findbugs.annotations.SuppressMatchType;
 import edu.umd.cs.findbugs.bytecode.MemberUtils;
 import edu.umd.cs.findbugs.detect.UselessSuppressionDetector;
@@ -35,7 +37,7 @@ public class MethodWarningSuppressor extends ClassWarningSuppressor {
         }
 
         MethodAnnotation bugMethod = bugInstance.getPrimaryMethod();
-        if (bugMethod != null && !method.equals(bugMethod)) {
+        if (bugMethod != null && !method.equals(bugMethod) && !isLambdaOfSuppressedMethod(bugMethod)) {
             return false;
         }
         if (DEBUG) {
@@ -55,5 +57,17 @@ public class MethodWarningSuppressor extends ClassWarningSuppressor {
     @Override
     public boolean isUselessSuppressionReportable() {
         return userGeneratedMethod && super.isUselessSuppressionReportable();
+    }
+
+    private boolean isLambdaOfSuppressedMethod(MethodAnnotation bugMethod) {
+        if (!method.getClassName().equals(bugMethod.getClassName())) {
+            return false;
+        }
+
+        if (Const.CONSTRUCTOR_NAME.equals(method.getMethodName())) {
+            return bugMethod.getMethodName().startsWith("lambda$new$");
+        }
+
+        return bugMethod.getMethodName().startsWith("lambda$" + method.getMethodName() + "$");
     }
 }
--- a/spotbugsTestCases/src/java/ghIssues/Issue724.java
+++ b/spotbugsTestCases/src/java/ghIssues/Issue724.java
@@ -1,0 +1,66 @@
+package ghIssues;
+
+import java.util.function.Supplier;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/724">GitHub issue #724</a>
+ */
+public class Issue724 {
+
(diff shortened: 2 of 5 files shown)


spotbugs/src/main/java/edu/umd/cs/findbugs/MethodWarningSuppressor.java
```diff
@@ -56,4 +58,16 @@ public BugInstance buildUselessSuppressionBugInstance(UselessSuppressionDetector
...
         return userGeneratedMethod && super.isUselessSuppressionReportable();
     }
+
+    private boolean isLambdaOfSuppressedMethod(MethodAnnotation bugMethod) {
+        if (!method.getClassName().equals(bugMethod.getClassName())) {
+            return false;
+        }
+
+        if (Const.CONSTRUCTOR_NAME.equals(method.getMethodName())) {
+            return bugMethod.getMethodName().startsWith("lambda$new$");
```
JuditKnoll, line 68: Are you sure that for lambdas in constructors this will be the start of the name?

spotbugsTestCases/src/java/ghIssues/Issue724.java
```diff
@@ -0,0 +1,66 @@
+package ghIssues;
+
+import java.util.function.Supplier;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/724">GitHub issue #724</a>
+ */
+public class Issue724 {
```
JuditKnoll, line 10: Can you please also add testcases with static initializers?

spotbugs/src/main/java/edu/umd/cs/findbugs/MethodWarningSuppressor.java
julianstengaard: Hmm, it's the only way I ever remember seeing them, but maybe my knowledge is a bit limited here.

I checked in javac [LambdaToMethod.java:1198](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java#L1198) and it looks like this:

```java
String syntheticMethodNameComponent(Symbol owner) {
    long ownerFlags = owner.flags();
    if ((ownerFlags & BLOCK) != 0) {
        return (ownerFlags & STATIC) != 0 ?
                "static" : "new";
    } else if (owner.isConstructor()) {
        return "new";
    } else {
        return owner.name.toString();
    }
}
```

spotbugsTestCases/src/java/ghIssues/Issue724.java
julianstengaard: Just to be sure I don't misunderstand. These should still be annotated at the class level, right? Because you can't add annotations to them directly.

julianstengaard:
My apologies for the master merge into this one. I am not so familiar with github and clicked the wrong button when trying to update master on my fork. 

@JuditKnoll I managed to get a [bytecode version of this working](https://github.com/julianstengaard/spotbugs/commit/8de80486f40afb4e20d9a55375138390b31b317d) that does not suffer from the suppression issue on overloads. 

Should I close this PR and open another for that?

JuditKnoll:
> My apologies for the master merge into this one. I am not so familiar with github and clicked the wrong button when trying to update master on my fork.
> 
> @JuditKnoll I managed to get a [bytecode version of this working](https://github.com/julianstengaard/spotbugs/commit/8de80486f40afb4e20d9a55375138390b31b317d) that does not suffer from the suppression issue on overloads.
> 
> Should I close this PR and open another for that?

Merging master into the PR branch is not an issue, the PR branch can be updated both by rebase or master merge. In this repo when merging the changes are squashed usually.

If you found a better solution, feel free to close this PR and open a new PR, or you can update this branch with the changes from there.

julianstengaard:
> If you found a better solution, feel free to close this PR and open a new PR, or you can update this branch with the changes from there.

Thank you. I have opened [PR #4232](https://github.com/spotbugs/spotbugs/pull/4232) instead and will close this one.

PR #4228, opened 2026-08-03
Fixes #4148: Track Connection.prepareCall obligations

### LINKED ISSUE #4148: OBL_UNSATISFIED_OBLIGATION not tracked for Connection.prepareCall, only for Connection.prepareStatement
### SpotBugs version

### Check name

`OBL_UNSATISFIED_OBLIGATION`

### Description

The obligation-tracking detector flags a method that obtains a `java.sql.Statement` from `Connection.prepareStatement(...)` and never closes it, but it does not flag the same code when the statement is obtained from `Connection.prepareCall(...)`. Both factory methods return a `java.sql.Statement` subtype (`PreparedStatement` and `CallableStatement` respectively) whose `close()` obligation is exactly what this detector is documented to track — and the warning text itself reads "may fail to clean up java.sql.Statement". The two forms are equivalent with respect to the tracked obligation: in both cases a `Statement` resource is created and leaked. SpotBugs' obligation policy database recognizes `prepareStatement` as creating an obligation but is missing `prepareCall`, so the leak is silently missed for callable statements.

### Reproducer

```java
// BEFORE — prepareStatement, 1 report
import java.sql.Connection;
import java.sql.Statement;
import java.sql.SQLException;

public class DeadStore {
    public void testDatabaseStore(Connection c) throws SQLException {
        Statement s = c.prepareStatement("select something from other");
    }
}
```

```java
// AFTER — prepareCall (also returns a Statement that is leaked), 0 reports
import java.sql.Connection;
import java.sql.Statement;
import java.sql.SQLException;

public class DeadStore {
    public void testDatabaseStore(Connection c) throws SQLException {
        Statement s = c.prepareCall("select something from other");
    }
}
```

### Expected behavior

`OBL_UNSATISFIED_OBLIGATION` should report the leaked `Statement` for both `prepareStatement` and `prepareCall`, since both create a `java.sql.Statement` subtype that carries the same unclosed-resource obligation.

### Actual behavior

- BEFORE (`prepareStatement`): 1 report — `DeadStore.testDatabaseStore(Connection) may fail to clean up java.sql.Statement`.
- AFTER (`prepareCall`): 0 reports.

DragonFSKY:
`Connection.prepareCall` creates a `CallableStatement`, but the obligation policy database only registered `prepareStatement`, so unclosed callable statements were not reported.

This adds the matching policy and regression coverage for all three JDBC overloads while verifying that try-with-resources satisfies the obligation.

Tests: `./gradlew spotlessApply build smoketest`

Fixes #4148


ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4148Test.java
```diff
@@ -0,0 +1,21 @@
...
+class Issue4148Test extends AbstractIntegrationTest {
+
+    private static final String BUG_TYPE = "OBL_UNSATISFIED_OBLIGATION";
+    private static final String CLASS_NAME = "ghIssues.Issue4148";
+
+    @Test
+    void detectsUnclosedCallableStatements() {
+        performAnalysis("ghIssues/Issue4148.class");
+
+        assertBugTypeCount(BUG_TYPE, 3);
```
JuditKnoll, line 16: Since `OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE` is kind of related and reported by the same detector, can you please add an assert for that as well?

spotbugsTestCases/src/java/ghIssues/Issue4148.java
```diff
@@ -0,0 +1,28 @@
...
+    public void prepareCallWithResultSetOptionsNotClosed(Connection connection) throws SQLException {
+        connection.prepareCall("call procedure()", ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
+    }
+
+    public void prepareCallWithHoldabilityNotClosed(Connection connection) throws SQLException {
+        connection.prepareCall("call procedure()", ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY,
+                ResultSet.CLOSE_CURSORS_AT_COMMIT);
+    }
+
+    public void prepareCallClosed(Connection connection) throws SQLException {
```
JuditKnoll, line 23: It would be nice to see one more compliant example like this without try-with-resources, where the statement is closed in another way.

JuditKnoll (commented):
Some small nitpicking, but otherwise looks good! Thank you for your work!

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4148Test.java
DragonFSKY: Thanks, added an exact assertion that `OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE` remains at 0 for this fixture.

spotbugsTestCases/src/java/ghIssues/Issue4148.java
DragonFSKY: Added a `try`/`finally` example that explicitly closes the `CallableStatement`, together with a no-warning assertion.

ThrawnCA (approved):

ThrawnCA (approved):

JuditKnoll (approved):

ThrawnCA (approved):

hazendaz (approved):

PR #4231, opened 2026-08-05
Fix SING_SINGLETON_GETTER_NOT_SYNCHRONIZED for ternary lazy init

### LINKED ISSUE #4154: `SING_SINGLETON_GETTER_NOT_SYNCHRONIZED` not reported when the lazy init uses a ternary instead of an if-statement
### SpotBugs version

### Check name

`SING_SINGLETON_GETTER_NOT_SYNCHRONIZED`

### Description

The detector flags an unsynchronized singleton getter that lazily initializes its static instance field with an `if (instance == null) { instance = new ...(); }` block, but it stops reporting when the same lazy initialization is written as the equivalent ternary assignment `instance = (instance == null) ? new ...() : instance;`. Both forms are unsynchronized lazy initializations of the same static `instance` field inside the same static getter — exactly the anti-pattern the rule targets — and both have the same thread-safety hazard. The detector appears to pattern-match the `if`-statement shape of the lazy init and misses the ternary shape.

### Reproducer

```java
// BEFORE — if-statement lazy init, 1 report
public class Issue3280 {
    private static volatile Issue3280 instance = null;

    public static Issue3280 getInstance() {
        if (null == instance) {
            instance = new Issue3280();
        }
        instance.factorial(5);
        return instance;
    }

    private Issue3280() {
    }

    private int factorial(int n) {
        if (n == 0) {
            return 1;
        }
        return n * factorial(n - 1);
    }
}
```

```java
// AFTER — equivalent ternary lazy init, 0 reports
public class Issue3280 {
    private static volatile Issue3280 instance = null;

    public static Issue3280 getInstance() {
        instance = null == instance ? new Issue3280() : instance;
        instance.factorial(5);
        return instance;
    }

    private Issue3280() {
    }

    private int factorial(int n) {
        if (n == 0) {
            return 1;
        }
        return n * factorial(n - 1);
    }
}
```

### Expected behavior

`SING_SINGLETON_GETTER_NOT_SYNCHRONIZED` should fire on both forms, since both are unsynchronized lazy initializations of the singleton instance field in an unsynchronized getter.

### Actual behavior

- BEFORE (if-statement): 1 report — `Instance-getter method of class using singleton design pattern (Issue3280) is not synchronized.`
- AFTER (ternary): 0 reports.

kalayciburak:
## Summary

- Fixes a false negative for `SING_SINGLETON_GETTER_NOT_SYNCHRONIZED` when lazy initialization uses a ternary assignment (`instance = null == instance ? new C() : instance`) instead of an `if` statement.
- Ternary form merges constructor and field-reload paths before `PUTSTATIC`, so the previous "definitely null at store" check never fired and OpcodeStack lost constructor metadata.
- Detector now tracks `GETSTATIC` of the instance field and self-construction in the method to recognize this multi-path lazy-init shape.

Fixes #4154

## Test plan

- [x] `./gradlew --configure-on-demand :spotbugs-tests:test --tests edu.umd.cs.findbugs.detect.Issue4154Test`
- [x] `./gradlew --configure-on-demand :spotbugs-tests:test --tests edu.umd.cs.findbugs.detect.Issue3280Test`
- [x] `./gradlew --configure-on-demand :spotbugs-tests:test --tests edu.umd.cs.findbugs.detect.MultipleInstantiationsOfSingletonsTest`
- [x] `./gradlew --configure-on-demand :spotbugs:spotlessApply :spotbugs-tests:spotlessApply`


spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
```diff
@@ -143,30 +166,49 @@ public void sawOpcode(int seen) {
...
                     OpcodeStack.Item item = stack.getStackItem(0);
                     XMethod calledMethod = item.getReturnValueOf();
-                    if (calledMethod != null && Const.CONSTRUCTOR_NAME.equals(calledMethod.getName())
-                            && calledMethod.getClassName().equals(getDottedClassName())) {
+                    boolean storesConstructorResult = calledMethod != null && Const.CONSTRUCTOR_NAME.equals(calledMethod.getName())
+                            && calledMethod.getClassName().equals(getDottedClassName());
+                    // Ternary lazy init merges constructor and field-reload paths before PUTSTATIC, so
+                    // OpcodeStack may lose "return value of constructor" metadata. Fall back to: this
+                    // method constructed self, previously read the instance field, and stores a value of
+                    // this class (#4154).
```
JuditKnoll: Please do not leave irrelevant comments in the code. If you want to explain why you modified something, leave comments on the PRs.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4154Test.java
```diff
@@ -0,0 +1,23 @@
...
+        performAnalysis("ghIssues/Issue4154.class");
+
+        assertNoBugInClass("SING_SINGLETON_IMPLEMENTS_CLONEABLE", "Issue4154");
+        assertBugInClassCount("SING_SINGLETON_GETTER_NOT_SYNCHRONIZED", "Issue4154", 1);
+    }
+
+    @Test
+    void testIfFormStillReportsUnsynchronizedGetter() {
+        // Control: existing if-statement fixture must keep working.
+        performAnalysis("ghIssues/Issue3280.class");
```
JuditKnoll: This is already tested in [Issue3280Test](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3280Test.java) class.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
```diff
@@ -143,30 +166,49 @@ public void sawOpcode(int seen) {
...
+                    boolean ternaryStyleLazyStore = !storesConstructorResult
+                            && constructedSelfInMethod
+                            && instanceFieldsReadInMethod.contains(field)
+                            && !Const.STATIC_INITIALIZER_NAME.equals(getMethodName())
+                            && field.getSignature().equals(item.getSignature());
```
JuditKnoll: This is not strict enough IMO and could result in false positives. Checking the incoming CFG branches could help.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
kalayciburak: removed those comments

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4154Test.java
kalayciburak: yeah fair, dropped the Issue3280 control — already covered there

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
kalayciburak: tightened it — only treat as lazy store when the PUTSTATIC join has multiple incoming CFG edges, one path does self-ctor and another reloads the field. pushed

JuditKnoll (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/MultipleInstantiationsOfSingletons.java
```diff
@@ -143,30 +154,42 @@ public void sawOpcode(int seen) {
...
-                    if (calledMethod != null && Const.CONSTRUCTOR_NAME.equals(calledMethod.getName())
-                            && calledMethod.getClassName().equals(getDottedClassName())) {
+                    boolean storesConstructorResult = calledMethod != null && Const.CONSTRUCTOR_NAME.equals(calledMethod.getName())
+                            && calledMethod.getClassName().equals(getDottedClassName());
+                    boolean ternaryStyleLazyStore = !storesConstructorResult
+                            && !Const.STATIC_INITIALIZER_NAME.equals(getMethodName())
+                            && field.getSignature().equals(item.getSignature())
+                            && isMergedConstructorAndFieldStore(field);
+
+                    if (storesConstructorResult || ternaryStyleLazyStore) {
```
ThrawnCA: I don't like this style of conditional when the two flags are mutually exclusive. Why not just use `if...else`?

kalayciburak: fair, using if/else now

JuditKnoll (approved):

ThrawnCA (approved):

hazendaz (approved):

PR #4236, opened 2026-08-06
Fix AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE missed for inner-class accesses to enclosing fields

### LINKED ISSUE #4137: `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` missed when the non-atomic op lives in an inner class
### SpotBugs version

### Bug pattern

[AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#at-operation-on-shared-variable-is-not-atomic-at-nonatomic-operations-on-shared-variable)

### Description

The shared field `num` belongs to a `Thread` subclass and is updated with a non-atomic read-modify-write `num = num + 2`. When that statement is moved verbatim into a (non-static) inner class, `num` still resolves to the *same* outer-instance field and the operation is still a non-atomic RMW on it, but the detector stops reporting.

The inner-class form genuinely keeps the operation on the shared field — it is not optimized away. Disassembly of the inner class shows the read-modify-write going through the synthetic `this$0` outer reference straight onto `AdditionOnSharedVariable.num`:

```
// subAdditionOnSharedVariable.toggle2()
1: getfield this$0
5: getfield this$0
8: getfield AdditionOnSharedVariable.num    <-- non-atomic read of shared field
11: iconst_2
12: iadd
13: putfield AdditionOnSharedVariable.num   <-- non-atomic write of shared field
```

The "sharedness" of `num` is a property of the field/owning class and is unchanged by the rewrite; only the textual scope of the operation moved. The detector's shared-variable analysis is scoped to methods of the `Thread`/`Runnable` class itself and does not follow inner-class accesses to the enclosing instance's fields, so the non-atomic operation is missed — a coverage gap.

### Reproducer

```java
// BEFORE
public class AdditionOnSharedVariable extends Thread {
    private int num = 0;
    public void toggle() { num = num + 2; }            // reported
    public Integer getNum() { return num; }
}

// AFTER
public class AdditionOnSharedVariable extends Thread {
    private int num = 0;
    class subAdditionOnSharedVariable {
        public void toggle2() { num = num + 2; }       // same RMW on outer field -> NOT reported
    }
    public Integer getNum() { return num; }
}
```

### Expected behavior

The non-atomic read-modify-write on the shared field `num` of the `Thread` subclass should be reported regardless of whether it textually sits in a method of the outer class or of an inner class, since `num` is the same shared field in both.

### Actual behavior

- BEFORE: **1** report — non-atomic operation on `num`.
- AFTER:  **0** reports.

Relocating the operation into an inner class silences the detector — a real coverage gap for inner-class accesses to enclosing-instance shared fields.

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

## What

Report `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` when a non-atomic read-modify-write of a shared field lives in a non-static inner class that mutates a field of its enclosing (multi-threaded) class via the synthetic `this$0` reference. Previously the detector silently missed it.

## Why

The "sharedness" of a field is a property of the field and its owning class, not of where the operation textually sits. Moving `num = num + 2` from a `Thread` subclass into a non-static inner class (which still resolves `num` to the same enclosing-instance field, through `this$0`) leaves the operation the exact same non-atomic RMW on the same shared field, yet the detector stopped reporting it — a real coverage gap (#4137).

## How

Two independent gates dropped the inner-class form; both are opened:

1. **Multi-threaded-code gate** (`MultiThreadedCodeIdentifierUtils`): it only inspected the current class, so a non-static inner class of a `Thread`/`Runnable` subtype was never analyzed at all. Now walks `XClass.getImmediateEnclosingClass()` so an inner class inherits the multi-threaded context of its enclosing class. The inheritance flows only through non-static inner-class links (detected via the synthetic `this$0` field): a static nested class has no enclosing instance, so it does not inherit the context and the walk stops at a static-nesting boundary.

2. **`SharedVariableAtomicityDetector`**, which scoped the shared-variable analysis to the field's own class. Four points opened up:
   - `addNonFinalFieldsOfClass` and the PUTFIELD eligibility check required the field's class to equal the method's class → replaced with `isFieldOfMethodClassOrEnclosing`, following the enclosing-class chain;
   - the PUTFIELD target had to be `this` (register 0) → also accept a synthetic `this$0/this$1/...` enclosing-instance reference (`isEnclosingThisReference`);
   - `readFieldsByMethods` and the inner-method-call edges were cleared at every class-visit boundary, so the "is this shared field actually read in another reachable method?" signal could not survive a class boundary → they now accumulate across the detector lifetime;
   - bug emission is therefore deferred to `report()`, where the read map is fully populated. Each PUTFIELD site captures a `PendingBug` (with `SourceLineAnnotation.fromVisitedInstruction` taken while the instruction state is live), and `report()` gates each one on `hasNonSyncedNonPrivateCallToMethod` + `mapContainsFieldWithOtherMethod`.

The PUTFIELD-target relaxation stays conservative: only a synthetic field named `this$0/this$1/...` is recognized as an enclosing-instance reference, so stores whose target is a plain local (e.g. #3363's `p.number = ...` where `p` is a fresh local) are still correctly skipped.

## Root cause

The detector treated shared-variable analysis as class-local in four ways (multi-threaded-code gate, field-ownership check, PUTFIELD target, per-class read-map reset), so any RMW that crossed an inner→outer class boundary via `this$0` was dropped before evaluation.

## Testing

- New `Issue4137Test`: asserts the enclosing `Thread.toggle()` (baseline) and the inner `Issue4137$SubAdditionOnSharedVariable.toggle2()` (the issue's precise case) are reported (count = 2), **and** that a sibling `Issue4137$StaticNested.toggle3()` doing the same RMW on its own field is *not* reported (it has no enclosing instance, so it does not inherit the Thread context). Fails on master (0 reports for `toggle2`); also fails without the static-nesting guard (3 reports, `StaticNested.toggle3` wrongly flagged).
- `SharedVariableAtomicityDetectorTest` (37 tests, including `noBugInnerLocalVarFields` which anchors #3363): all green, no regression.
- FP corpus diff over the multi-threaded test-cases subset (one-shot integration test enumerating every `AT_*` bug): origin/master baseline **38** (24 `AT_NONATOMIC` / 10 `AT_STALE` / 4 `AT_64BIT`) == full fix **38**, byte-identical — no `AT_*` drift in either direction.
- Self-analysis `:spotbugs:spotbugsMain` on SpotBugs' own codebase (which contains many inner classes): no new bugs.
- Gate: `spotlessApply` + `:spotbugs:checkstyleMain` + `:spotbugs:spotbugsMain` + full `:spotbugs-tests:test` all green.

## Behavior change to flag (cross-class read-map accumulation)

The fix changes one semantic beyond the inner-class case itself: `readFieldsByMethods` (and the inner-method-call edges) now accumulate across the **whole detector lifetime** instead of being cleared at each class boundary. So the "is this field read in another reachable method?" gate can now be satisfied by a read in **any** class/method the detector sees, not only the field's own class or its enclosing chain. That is the more correct model of "shared" (a field shared across classes is still shared), and it is needed beyond the immediate #4137 case: a field of a multi-threaded class can be read in a **sibling** inner class (a non-static inner class of the same enclosing class), which is outside any single enclosing chain, so scoping the read map to one chain would re-introduce the gap for that shape. The trigger is wider than strictly #4137, but the multi-threaded-code gate's own blast radius is held back by the static-nesting guard above (static nested classes are not newly analyzed).

Caveat on the FP corpus: the `38 == 38` byte-identical result is over SpotBugs' multi-threaded test-cases subset only — it proves no drift on that set, **not** that no new `AT_NONATOMIC` reports appear on arbitrary real-world code. The self-analysis run on SpotBugs' own codebase adds some confidence (no new bugs), but a large user codebase could still surface new reports.

If a smaller blast radius is preferred, a viable alternative (happy to implement as a second commit or a separate PR) is to scope the read-map accumulation to the enclosing-class chain only — reset the map when leaving the `enclosing` chain of the field under analysis — which closes the #4137 gap without broadening "shared" to cross-arbitrary-class reads.

## Verification of the original issue

Issue #4137's reproducer — `subAdditionOnSharedVariable.toggle2()` doing `num = num + 2` on the enclosing field:

- Before (master): **0** reports — `toggle2` silently dropped.
- After (this PR): `toggle2` reported as `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE`, alongside the baseline `toggle()`.

Fixes #4137


ThrawnCA (approved):

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4137Test.java
```diff
@@ -0,0 +1,28 @@
+package edu.umd.cs.findbugs.detect;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class Issue4137Test extends AbstractIntegrationTest {
+    private static final String OPS_BUG = "AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE";
```
JuditKnoll, line 7: Can you please assert the other bugs reported by the detector as well?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SharedVariableAtomicityDetector.java
```diff
@@ -121,11 +149,10 @@ public void visit(Method method) {
 
     @Override
     public void visitAfter(JavaClass obj) {
-        bugAccumulator.reportAccumulatedBugs();
+        // readFieldsByMethods / nonSyncedMethodCallsByCallingMethods intentionally
+        // keep accumulating across classes; bug emission happens in report().
```
JuditKnoll, line 153: Somewhere these should be cleared! I can understand not clearing them after nested classes, but after outer classes, these definitely should be cleared.

JuditKnoll (changes requested):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4137Test.java
fudianchn: Done in 8dba4d2: the test also asserts `AT_NONATOMIC_64BIT_PRIMITIVE` and `AT_STALE_THREAD_WRITE_OF_PRIMITIVE` now, both at 0 since this corpus only has int-field read-modify-writes.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SharedVariableAtomicityDetector.java
fudianchn: Cleared in 6028f74, right after the evaluation loop, next to `pendingBugs.clear()`. `report()` is the earliest sound point: classes are visited in call-graph order (`FindBugs2#analyzeApplication` sorts the classes by call edges), so the enclosing class of a nested class can be visited before or after it, and every AT_* bug is a pending bug evaluated in `report()` against the full read map. Clearing at an earlier class boundary (including after outer classes) can therefore drop reads a pending bug still needs, reintroducing the #4137 miss under a different visit order. Growth until then is bounded by the multi-threaded-code gate: only classes passing `isPartOfMultiThreadedCode` accumulate.

If holding whole-run state in the detector is a concern regardless, I can move the cross-class facts into a first-pass note detector (the UnreadFields/CalledMethods pattern) as a follow-up, which restores per-class clearing in this detector.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/SharedVariableAtomicityDetector.java
```diff
@@ -48,9 +53,32 @@ public class SharedVariableAtomicityDetector extends OpcodeStackDetector {
     private LockDataflow currentLockDataFlow;
     private boolean isFirstVisit = true;
     private boolean hadOperation = false;
+    // Field reads and inner-method call edges are accumulated across every
+    // analyzed multi-threaded class for the lifetime of the detector: a field
+    // owned by an enclosing class can be read in another method of the
+    // enclosing class (or a sibling inner class), and that "is this shared
+    // field actually read elsewhere?" signal must survive the per-class visit
+    // boundary. Bug emission is deferred to report() so the shared/read
+    // evaluation sees the fully accumulated map regardless of class visit order.
```
JuditKnoll: This comment looks a bit superfluous to me.

PR #4242, opened 2026-08-06
Reset inAssert at method entry in AbstractAssertDetector

### LINKED ISSUE #3483: ASE_ASSERTION_WITH_SIDE_EFFECT_METHOD produce false positives

Eljees:
Fixes the reproducers reported in #3483 by @woess.

## Problem

`AbstractAssertDetector` tracks whether the scanner is inside an `assert` with a single instance field:

```java
// AbstractAssertDetector.java
protected boolean inAssert = false;

    if (inAssert) { detect(seen); }
    if (seen == Const.GETSTATIC && "$assertionsDisabled".equals(getNameConstantOperand())) { inAssert = true; }
    if (seen == Const.NEW && getClassConstantOperand().equals("java/lang/AssertionError")) { inAssert = false; }
}
```

The flag is set by a read of `$assertionsDisabled` and cleared **only** by a `new AssertionError`. A method that reads the flag without ever throwing — a plain `if ($assertionsDisabled) return;` guard, which is what `javac` emits for some shapes and what bytecode generators produce routinely — leaves `inAssert == true` when the method ends.

Nothing resets it. Neither `FindAssertionsWithSideEffects` nor `FindArgumentAssertions` implements `StatelessDetector`, so a single detector instance is reused for the whole analysis run: every method scanned afterwards, **including methods of the following classes**, is analysed as if it were inside an assertion.

## Reproducers

All three reproducers from @woess's comment fail on `master`. The new fixture merges them:

| method | reported before the fix | why it is wrong |
|---|---|---|
| `Issue3483.victim()` | `ASE_ASSERTION_WITH_SIDE_EFFECT` | plain `PUTSTATIC`, no assertion |
| `Issue3483.<init>()` | `ASE_ASSERTION_WITH_SIDE_EFFECT` ×2 | two field assignments in a constructor |
| `Issue3483.<clinit>()` | `ASE_ASSERTION_WITH_SIDE_EFFECT` | static initializer |
| `Issue3483.victimMethod()` | `ASE_ASSERTION_WITH_SIDE_EFFECT_METHOD` | the pattern named in the issue title |
| `Issue3483Victim.victim()` | `ASE_ASSERTION_WITH_SIDE_EFFECT` | different class — leak across a class boundary |

The `..._METHOD` case matters because it is the pattern @dougxc originally reported; the reproducers in the thread only covered `ASE_ASSERTION_WITH_SIDE_EFFECT`.

## Fix

Reset the flag when a method starts, which is what the rest of the codebase does (`visit(Code obj)` is overridden by ~55 detectors in `detect/`, e.g. `FindNakedNotify`):

```java
@Override
public void visit(Code obj) {
    inAssert = false;
    super.visit(obj);
}
```

@gtoison guessed this correctly in the thread and pushed `visitAfter(Code)` on `gtoison/spotbugs:issue-3483`. I verified both variants against the reproducers and **both clear them**, so his diagnosis was right. I went with `visit(Code)` for two reasons:

1. It is the idiomatic hook in this codebase.
2. `visitAfter` is not guaranteed to run. In `PreorderVisitor.visitCode` the call sits *after* the opcode scan and is not in a `finally`:

   ```java
   public void visitCode(Code obj) {
       code = obj;
       super.visitCode(obj);        // opcode scan
       ...
       visitAfter(obj);             // skipped if anything above throws
       code = null;
   }
   ```

   `FindBugs2` swallows analysis failures per detector (`catch (CheckedAnalysisException | RuntimeException e) { logRecoverableException(...); }`) and moves on to the next class, so a single recoverable failure inside the scan would leave `inAssert` set for the remainder of the run. Resetting on entry cannot be skipped.

## Note on the original report

@dougxc said the `visitAfter` patch did not fix the Graal build, and I could not reproduce that setup. This PR fixes the reproducers @woess posted, so I have deliberately **not** written "closes #3483" — @dougxc, could you re-check the Graal case against this branch? If it still reports, the remaining cause is separate from the state leak and worth tracking on its own.

## Verification

- `./gradlew --configure-on-demand :spotbugs-tests:test --tests 'edu.umd.cs.findbugs.detect.Issue3483Test'` fails on `master` (both tests) and passes with the fix.
- `FindAssertionsWithSideEffectsTest` (3 × `ASE_ASSERTION_WITH_SIDE_EFFECT` + 1 × `ASE_ASSERTION_WITH_SIDE_EFFECT_METHOD`) and `FindArgumentAssertionsTest` (23 × `AA_ASSERTION_OF_ARGUMENTS`) are unchanged — real assertions with side effects are still reported.
- `./gradlew spotlessApply build smoketest` per `.github/CONTRIBUTING.md` — green (311 test classes, 1152 tests, 0 failures, 0 errors). `spotlessApply` left every file in this PR unchanged.
- `CHANGELOG.md` updated under `## Unreleased` / `### Fixed`.

No bug type, detector id or metadata was touched, so `findbugs.xml` / `messages*.xml` are untouched as well.

---

*AI-assisted: I used Claude to help explore the detector's state handling and to draft the fixture, the test and this description. I reviewed every line, ran the new test against unpatched `master` first to confirm it fails without the fix, and ran the full `spotlessApply build smoketest` locally; the analysis and the runs are mine.*

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
@@ -19,7 +19,7 @@
 package edu.umd.cs.findbugs.detect;
 
 import org.apache.bcel.Const;
-import org.apache.bcel.classfile.Code;
+import org.apache.bcel.classfile.Method;
 
 import edu.umd.cs.findbugs.BugInstance;
 import edu.umd.cs.findbugs.BugReporter;
@@ -43,18 +43,8 @@ public abstract class AbstractAssertDetector extends OpcodeStackDetector {
      */
     protected abstract void detect(int seen);
 
-    /**
-     * Resets the state before scanning a method.
-     *
-     * <p>An assertion never spans a method boundary, but {@code inAssert} is only cleared when
-     * a {@code new AssertionError} is seen. A method that reads {@code $assertionsDisabled}
-     * without ever throwing (for example a guard that just returns) therefore leaves the flag
-     * set, and every method scanned afterwards is treated as if it were inside an assertion.
-     * Detector instances are reused for the whole analysis run, so the state also leaks into
-     * the following classes.</p>
-     */
     @Override
-    public void visit(Code obj) {
+    public void visit(Method obj) {
         inAssert = false;
         super.visit(obj);
     }
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
@@ -19,6 +19,7 @@
 package edu.umd.cs.findbugs.detect;
 
 import org.apache.bcel.Const;
+import org.apache.bcel.classfile.Code;
 
 import edu.umd.cs.findbugs.BugInstance;
 import edu.umd.cs.findbugs.BugReporter;
@@ -41,6 +42,22 @@ public abstract class AbstractAssertDetector extends OpcodeStackDetector {
      * Implement this method in a concrete detector
      */
     protected abstract void detect(int seen);
+
+    /**
+     * Resets the state before scanning a method.
+     *
+     * <p>An assertion never spans a method boundary, but {@code inAssert} is only cleared when
+     * a {@code new AssertionError} is seen. A method that reads {@code $assertionsDisabled}
+     * without ever throwing (for example a guard that just returns) therefore leaves the flag
+     * set, and every method scanned afterwards is treated as if it were inside an assertion.
+     * Detector instances are reused for the whole analysis run, so the state also leaks into
+     * the following classes.</p>
+     */
+    @Override
+    public void visit(Code obj) {
+        inAssert = false;
+        super.visit(obj);
+    }
 
     /**
      * Searches for assertion opening, and closing points.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 
 ## Unreleased - 2026-??-??
 ### Fixed
+- Fix `ASE_ASSERTION_WITH_SIDE_EFFECT` and `ASE_ASSERTION_WITH_SIDE_EFFECT_METHOD` false positives in every method analysed after a method that reads `$assertionsDisabled` without throwing an `AssertionError` ([#3483](https://github.com/spotbugs/spotbugs/issues/3483))
 - Fix `INT_BAD_COMPARISON_WITH_SIGNED_BYTE` false positive for meaningful comparisons of a signed byte with `127` (`b < 127`, `b >= 127`) ([#4201](https://github.com/spotbugs/spotbugs/pull/4201))
 
 ## 4.10.3 - 2026-07-12
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3483Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3483Test.java
@@ -1,0 +1,36 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/3483">GitHub issue #3483</a>
+ */
+class Issue3483Test extends AbstractIntegrationTest {
+
+    /**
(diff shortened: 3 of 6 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
```diff
@@ -42,6 +43,22 @@ protected AbstractAssertDetector(BugReporter bugReporter) {
...
+     *
+     * <p>An assertion never spans a method boundary, but {@code inAssert} is only cleared when
+     * a {@code new AssertionError} is seen. A method that reads {@code $assertionsDisabled}
+     * without ever throwing (for example a guard that just returns) therefore leaves the flag
+     * set, and every method scanned afterwards is treated as if it were inside an assertion.
+     * Detector instances are reused for the whole analysis run, so the state also leaks into
+     * the following classes.</p>
+     */
+    @Override
+    public void visit(Code obj) {
```
JuditKnoll: The `BetterVisitor#visit(Code)` method is intended for checking the attributes. Can you please use `BetterVisitor#visit(Method)` or `Visitor#visitMethod(Method)` instead, which describe better when they run?

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/AbstractAssertDetector.java
```diff
@@ -42,6 +43,22 @@ protected AbstractAssertDetector(BugReporter bugReporter) {
...
 
+    /**
+     * Resets the state before scanning a method.
+     *
+     * <p>An assertion never spans a method boundary, but {@code inAssert} is only cleared when
+     * a {@code new AssertionError} is seen. A method that reads {@code $assertionsDisabled}
+     * without ever throwing (for example a guard that just returns) therefore leaves the flag
+     * set, and every method scanned afterwards is treated as if it were inside an assertion.
+     * Detector instances are reused for the whole analysis run, so the state also leaks into
+     * the following classes.</p>
```
JuditKnoll: This is a good explanation for the change, but once this is merged, the comment will be outdated. Please remove this comment.

JuditKnoll (changes requested):

Eljees:
Both points addressed in the commit above:

- the reset moved from `visit(Code)` to `visit(Method)` - you are right that it describes the intent better, and the import follows;
- the explanatory comment is gone.

Re-verified on `eclipse-temurin:21-jdk`: `:spotbugs:compileJava` and `:spotbugs-tests:test --tests "*Issue3483*"` both green after the change.

JuditKnoll (approved):

dougxc:
> @dougxc, could you re-check the Graal case against this branch?

It's not worth trying to re-create the Graal case now. I'll report if something similar happens again in future.

Eljees:
Understood, no need to re-create it. The reset is pinned by the new test either way; if something similar turns up later, a reproducer would be welcome.

PR #4245, opened 2026-08-10
Fix naked notify detection after local stores

### LINKED ISSUE #3884: False Negative: NN_NAKED_NOTIFY is incorrectly suppressed by getfield and invokevirtual instructions
**Describe the bug**
The **NN_NAKED_NOTIFY** rule fails to report a naked notify when the synchronized block contains certain non-state-changing instructions (e.g., field reads or method calls).

```java
public class Bug {
    private int count = 0;

//Case 1 
    public synchronized void testWithFieldRead() {
        int temp = this.count; 
        this.notifyAll();  // <- should report 
    }
//Case 2
    public synchronized void testWithPrintln() {
        System.out.println("Log"); 
        this.notifyAll();  // <- should report 
    }
}
```

**Expected behavior**
Reading a variable or printing a log message does not modify the object's state.
Therefore, the notification remains "naked" because any awakened thread will find the condition unchanged and return to the waiting state immediately.

**Actual behavior**
Both of them are ignored.

**Environment**
- Java Compiler: JDK 21
- Java version:  JDK 21
- SpotBugs Version:  4.9.8

**Additional context**
There is a significant logical inconsistency in Case 1. SpotBugs successfully reports DLS_DEAD_LOCAL_STORE for the variable temp in testWithFieldRead. 
This proves the analysis engine knows the getfield instruction has no meaningful impact on the program's logic. However, this same "dead" instruction is allowed to suppress the NN_NAKED_NOTIFY concurrency warning.

Eljees:
A local store after reading a monitored field does not mutate the object, but it reset the detector state and suppressed NN_NAKED_NOTIFY. Keep tracking through register stores and cover both synchronized forms plus mutation controls.

Fixes #3884

----

### DIFF
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3884Test.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3884Test.java
@@ -1,0 +1,20 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue3884Test extends AbstractIntegrationTest {
+
+    @Test
+    void testNakedNotifyAfterNonMutatingRead() {
+        performAnalysis("ghIssues/Issue3884.class");
+
+        assertBugInMethodCount("NN_NAKED_NOTIFY", "ghIssues.Issue3884", "notifyAfterFieldRead", 1);
+        assertBugInMethodCount("NN_NAKED_NOTIFY", "ghIssues.Issue3884", "blockNotifyAfterFieldRead", 1);
+        assertBugInMethodCount("NN_NAKED_NOTIFY", "ghIssues.Issue3884", "nakedNotify", 1);
+
+        assertBugInMethodCount("NN_NAKED_NOTIFY", "ghIssues.Issue3884", "notifyAfterMutation", 0);
+        assertBugInMethodCount("NN_NAKED_NOTIFY", "ghIssues.Issue3884", "notifyAfterFieldAssignment", 0);
+    }
+}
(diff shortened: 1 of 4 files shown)


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNakedNotify.java
```diff
@@ -78,7 +78,11 @@ public void sawOpcode(int seen) {
             }
             break;
         case LOADED:
-            if (isRegisterLoad() || seen == Const.GETSTATIC || seen == Const.GETFIELD) {
+            // Storing a value into a local variable cannot change the state of the monitored
+            // object, so a notify() that follows is still naked. Everything that may mutate
+            // state - PUTFIELD, PUTSTATIC, invocations, arithmetic feeding them - still falls
+            // through to the reset below.
```
JuditKnoll: Please do not leave unnecessary comments in the code.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3884Test.java
```diff
@@ -0,0 +1,20 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue3884Test extends AbstractIntegrationTest {
+
+    @Test
+    void testNakedNotifyAfterNonMutatingRead() {
```
JuditKnoll, line 10: Please also assert the number of relevant bug types! You can use the `assertBugTypeCount(String, int)` function for this.

JuditKnoll (changes requested):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindNakedNotify.java
Eljees: Removed the explanatory block in 071b65d.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue3884Test.java
Eljees: Added assertBugTypeCount("NN_NAKED_NOTIFY", 3) in 071b65d.

JuditKnoll (approved):
This is only a partial fix to the issue, as there are two cases mentioned there. However recognizing which called function actually modify the object's state would need either interprocedural dataflow analysis or really good heuristics to not make the detector noisy.

hazendaz (approved):

PR #4246, opened 2026-08-12
Fixes #3955 : Fix URF_UNREAD_FIELD false negative for unread fields i…

### LINKED ISSUE #3955: False Negative: UrF missing for unread fields in Enum
**Describe the bug**
Tool fails to report UrF: Unread field for a package-private field defined directly within an enum, even though the field is never read.

**To Reproduce**

```java
enum A {
  ENUM_VALUE_3;
  class B {
    int s = 0;       // <- reported UrF (TP)
  }
  boolean i = false; // <- should report UrF (FN)
}
```

**Expected behavior**
Both A$B.s and A.i should be reported with UrF: Unread field. Since both are package-private and neither is ever accessed or read by any other part of the code, they both represent dead code/redundant fields.

**Actual behavior**
Only the field s inside the inner class B is reported. The field i in the enum A is silently ignored.

**Output:**

Plaintext
M P UrF: Unread field: A$B.s  At InputInnerTypeLastClassRootClass2.java:[line 4]

**Environment**
- Java version: JDK 21
- SpotBugs Version: 4.9.8
- How do you use SpotBugs:  CLI

mirkoalicastro:
Fixes #3955 .

Since enums are serializable, UrF skipped the warning for all the enum fields. However, enum serialization consists only of its name. This PR fixes it by excluding enums from that check.


ThrawnCA (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
```diff
@@ -226,7 +226,7 @@ public void visitAfter(JavaClass obj) {
             data.fieldsOfSerializableOrNativeClassed.addAll(data.myFields);
             data.fieldsOfNativeClasses.addAll(data.myFields);
         }
-        if (isSerializable) {
+        if (isSerializable && !obj.isEnum()) {
```
JuditKnoll, line 229: A comment could be added here to explain why enums are an exception even though they are serializable.

JuditKnoll (approved):
Thank you for your work!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/UnreadFields.java
mirkoalicastro: Done, thanks

hazendaz (approved):

PR #4287, opened 2026-09-05
Issue 4276 fix with all use/test cases #4276

### LINKED ISSUE #4276: FALSE NEGATIVE: `LI_LAZY_INIT_STATIC` misses lazy initialization from a method call
Hi, I found a false negative in SpotBugs 4.10.3 when a static field is lazily initialized with the result of a method call.

**Affected tool**

`SpotBugs 4.10.3`

**Affected checker**

`SpotBugs LI_LAZY_INIT_STATIC`

**Minimal reproducer**

```java
class SpotbugsSonarqubeLazyInitMethodCall {
  static String value;

  String test(StringBuilder builder) {
    if (value == null) value = builder.toString();
    return value;
  }
}
```

**Reproduction command**

```bash
spotbugs -version
javac -d classes spotbugs-sonarqube-li-lazy-init-static-fn-method-call.java
spotbugs -textui -effort:max -low -visitors LazyInit classes
```

**Current behavior**

SpotBugs produces no `LI_LAZY_INIT_STATIC` diagnostic.

**Expected behavior**

SpotBugs should report `LI_LAZY_INIT_STATIC` at line 5 because the unsynchronized method lazily initializes the static field `value`.

Aniketsurya:

----

Closes #4276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 ## Unreleased - 2026-??-??
 ### Fixed
 - Fix `MS_EXPOSE_BUF` and `EI_EXPOSE_BUF` false negative when returning `Buffer.array()` ([#4235](https://github.com/spotbugs/spotbugs/pull/4235))
+- Fix `LI_LAZY_INIT_STATIC` false negative where string field initialized with some method lazily ([#4276](https://github.com/spotbugs/spotbugs/issues/4276))
 
 ## 4.10.4 - 2026-08-19
 ### Fixed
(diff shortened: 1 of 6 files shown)


hazendaz:
@Aniketsurya Thanks for taking the time to work on this and for the additional test coverage. This overlaps substantially with #4285, which addresses the original reported issue with a smaller, more targeted change, so I'm going to close this PR as a duplicate for now.

There is one part of this PR that appears to go beyond the original issue: the change to the `String` handling / `sawINVOKE` logic. That may well be a valid issue in its own right, but I'm not yet convinced it is required to fix the reported problem, and I'd prefer not to broaden the detector change without a demonstrated false positive/negative that requires it.

If you believe that additional change is independently necessary, please feel free to open a separate issue (or provide a concrete case here) showing the behavior it fixes and why the existing `String` handling is incorrect. I'd be happy to look at that separately.

Thanks again for the contribution.

Aniketsurya:
Sure @hazendaz!
Will see what we can do, or see if I can pick other issue

PR #4289, opened 2026-09-06
fix: make -conserveSpace equivalent to -effort:min (#4268)

### LINKED ISSUE #4268: `-conserveSpace` is documented as equivalent to `-effort:min`, but reports different findings
Hi, I found a command-line inconsistency in SpotBugs 4.10.3.

The text UI help documents `-conserveSpace` as equivalent to `-effort:min`:

```text
-conserveSpace
    same as -effort:min (for backward compatibility)
```

However, running the two options on the same compiled class files produces different findings.

## Reproducer

```java
public class ConserveSpaceRepro {
    void test(Runnable action) {
        Throwable thrown = null;
        try {
            action.run();
        } catch (RuntimeException exception) {
            thrown = exception;
        } finally {
            if (thrown instanceof Error) {
                throw (Error) thrown;
            }
        }
    }
}
```

Save the program as `ConserveSpaceRepro.java`, compile it once, and run
SpotBugs twice on the same class file:

```sh
SPOTBUGS=/path/to/spotbugs-4.10.3/bin/spotbugs
CLASSES=$(mktemp -d)

javac -g -d "$CLASSES" ConserveSpaceRepro.java

"$SPOTBUGS" -version
"$SPOTBUGS" -textui -conserveSpace "$CLASSES"
"$SPOTBUGS" -textui -effort:min "$CLASSES"
```

## Actual result

With `-conserveSpace`, SpotBugs reports no findings.

With `-effort:min`, SpotBugs reports:

```text
H C BC: Impossible cast from RuntimeException to Error in
ConserveSpaceRepro.test(Runnable)  At ConserveSpaceRepro.java:[line 10]
```

The difference is stable across repeated runs using the same compiled class
file. The reproduced SpotBugs version is 4.10.3.

## Expected result

Since `-conserveSpace` is documented as the same as `-effort:min`, both commands should produce the same finding set, or the documentation/help text should describe the additional behavior of `-conserveSpace`.

ronodhirSoumik:
Fixes #4268

Previously, the launcher scripts (spotbugs, fb, spotbugs2, spotbugs.bat) intercepted `-conserveSpace` at the shell level and injected a JVM system property (`-Dfindbugs.conserveSpace=true`) while never forwarding the flag to the Java process. This caused FindBugsCommandLine to never apply `MIN_EFFORT` settings, resulting in different analysis results compared to `-effort:min`.

**Fix**: pass `-conserveSpace` through to the Java application so it reaches `FindBugsCommandLine.handleOption()`, which already correctly maps it to `FindBugs.MIN_EFFORT` which is identical to `-effort:min`. Remove the now-redundant JVM property injection and associated dead variables (conservespace_arg / jvm_conservespace / conserveSpaceArg).

**Test**
`./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.TextUICommandLineTest.conserveSpaceSetsMinEffort"`

`./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.TextUICommandLineTest.conserveSpaceDoesNotUseDefaultEffort"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix SARIF output writing source-location URI syntax exceptions to stderr when source filenames are unknown ([#1412](https://github.com/spotbugs/spotbugs/issues/1412))
 - Narrow the definition of singletons ([#2985](https://github.com/spotbugs/spotbugs/issues/2985))
 - Fix `NP_LOAD_OF_KNOWN_NULL_VALUE` false negative when null check uses `instanceof` ([#3916](https://github.com/spotbugs/spotbugs/issues/3916))
+- Fix CLI launcher scripts intercepting `-conserveSpace` with a JVM system property instead passing it to Java Application as equivalent to `-effort:min` ([#4268](https://github.com/spotbugs/spotbugs/pull/4289)) 
 
 ## 4.10.4 - 2026-08-19
 ### Fixed
--- a/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
+++ b/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/TextUICommandLineTest.java
@@ -127,4 +127,39 @@ class TextUICommandLineTest {
         String xml = Files.readString(file, StandardCharsets.UTF_8);
         assertThat(xml, containsString("BugCollection"));
     }
+
+    /**
+     * Regression test for Issue 4268:
+     * {@code -conserveSpace} must set the same analysis effort settings as
+     * {@code -effort:min}.
+     */
+    @Test
+    void conserveSpaceSetsMinEffort() {
+        TextUICommandLine conserveSpace = new TextUICommandLine();
+        conserveSpace.handleOption("-conserveSpace", "");
+
+        TextUICommandLine effortMin = new TextUICommandLine();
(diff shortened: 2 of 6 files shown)


hazendaz:
@ronodhirSoumik  Please add a regression test that invokes the SpotBugs launcher with -conserveSpace and verifies that it is handled equivalently to -effort:min. The existing issue reproducer demonstrates the launcher-level failure, so having that covered in the project would prevent this from regressing in one of the platform-specific scripts.

ThrawnCA (approved):

ronodhirSoumik:
> @ronodhirSoumik Please add a regression test that invokes the SpotBugs launcher with -conserveSpace and verifies that it is handled equivalently to -effort:min. The existing issue reproducer demonstrates the launcher-level failure, so having that covered in the project would prevent this from regressing in one of the platform-specific scripts.

Added test file - please review. will squash then

hazendaz:
@ronodhirSoumik Thanks, no need to squash on your end, we force that on all merges.  As soon as builds pass, will merge.

hazendaz:
@ronodhirSoumik Actually add a change log as well as this affects functionality.  Follow other fixes to make the update.  Thanks.

ThrawnCA (approved):

hazendaz (dismissed):
needs changelog

ronodhirSoumik:
> @ronodhirSoumik Actually add a change log as well as this affects functionality. Follow other fixes to make the update. Thanks.

added

PR #4295, opened 2026-09-07
RANGE_ARRAY_INDEX issue fix

### LINKED ISSUE #4275: FALSE POSITIVE: RANGE_ARRAY_INDEX reports inside a contradictory null branch
Hi, I found a false positive in SpotBugs 4.10.3 when value cannot be both null and non-null, so the array access is unreachable.

**Affected tool**

`SpotBugs 4.10.3`

**Affected checker**

`SpotBugs RANGE_ARRAY_INDEX`

**Minimal reproducer**

```java
class SpotBugsUnreachableArrayIndex {
  void test(Object value) {
    if (value == null && value != null) {
      int[] array = new int[1];
      array[2] = 1;
    }
  }
}
```

**Reproduction command**

```bash
spotbugs -version
javac -d classes spotbugs-range-array-index-fp-unreachable.java
spotbugs -textui -effort:max -low -visitors DumbMethods classes
```

**Current behavior**

SpotBugs reports `RANGE_ARRAY_INDEX` at line 5.

**Expected behavior**

SpotBugs should not report `RANGE_ARRAY_INDEX` because value cannot be both null and non-null, so the array access is unreachable.

AparnaPradhan:
fixes #4275

Title
  Fix RANGE_ARRAY_INDEX false positive in unreachable code (impossible null condition)
  
  Body

  ## Summary

  - Fixes a false positive where `RANGE_ARRAY_INDEX` was reported on an array
    access inside a block guarded by an always-false null condition such as
    `if (x == null && x != null)`.
  - Root cause: `OpcodeStack` did not track reference-null semantics through
    `IFNONNULL`/`IFNULL` branches, so the dead fall-through path was treated as
    reachable.

  ## Changes
  
  ### `OpcodeStack.java`
  Split the previously shared `IFNONNULL`/`IFNULL` fall-through into dedicated
  cases with two new behaviors:

  - **`IFNONNULL`** — after recording the jump state, the local-variable slot the
    tested item was loaded from is written with a null item. Code on the
    fall-through ("value *is* null") path now sees the variable as definitively
    null.
  - **`IFNULL`** — if the popped item is already null (`isNull() == true`), the
    jump is unconditional and the fall-through is dead; `setTop(true)` is called
    so `OpcodeStackDetector.beforeOpcode` returns `false` and `sawOpcode` is
    never invoked for instructions in the dead block.

  `addJumpValue` is always called *before* the `lvValues` update, so the branch
  target receives the correct pre-condition state from the jump path.

  ### `Feature323.java`
  Added `unreachableArrayIndex` annotated `@NoWarning("RANGE_ARRAY_INDEX")` to
  lock in the regression fix.

  ### `CHANGELOG.md`
  Added entry under **Unreleased → Fixed** (issue number is a placeholder —
  update once the issue is filed).

  ## Test plan

  - [ ] `./gradlew :spotbugs:compileJava :spotbugsTestCases:compileJava` — clean
    compile
  - [ ] `./gradlew test --tests "*.Feature323*"` — existing range-check tests
    still pass and the new `unreachableArrayIndex` `@NoWarning` case passes
  - [ ] Full `./gradlew test` — no regressions in the broader test suite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - `@SuppressFBWarnings` annotation on a method or a constructor now suppresses warnings reported in their lambdas too ([#724](https://github.com/spotbugs/spotbugs/issues/724))
 
 ### Fixed
+- Fix `RANGE_ARRAY_INDEX` false positive when the array access is inside unreachable code caused by an impossible null condition (e.g. `if (x == null && x != null)`) ([#4295](https://github.com/spotbugs/spotbugs/issues/4295))
 - Fix `NN_NAKED_NOTIFY` false positive when guarded state is updated before the synchronized block that calls `notify`/`notifyAll` ([#3786](https://github.com/spotbugs/spotbugs/issues/3786))
 - Fix `SA_LOCAL_SELF_ASSIGNMENT` false positive in methods with `++`/`--` inside a switch nested in a try-catch block ([#3929](https://github.com/spotbugs/spotbugs/issues/3929))
 - Fix `OS_OPEN_STREAM` false positive when the result of `PrintWriter.append()` is reassigned before closing the writer ([#4274](https://github.com/spotbugs/spotbugs/issues/4274))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - `@SuppressFBWarnings` annotation on a method or a constructor now suppresses warnings reported in their lambdas too ([#724](https://github.com/spotbugs/spotbugs/issues/724))
 
 ### Fixed
-- Fix `RANGE_ARRAY_INDEX` false positive when the array access is inside unreachable code caused by an impossible null condition (e.g. `if (x == null && x != null)`) ([#4295](https://github.com/spotbugs/spotbugs/issues/4295))
+- Fix `RANGE_ARRAY_INDEX` false positive when the array access is inside unreachable code caused by an impossible null condition (e.g. `if (x == null && x != null)`) ([#4275](https://github.com/spotbugs/spotbugs/issues/4275))
 - Fix `NN_NAKED_NOTIFY` false positive when guarded state is updated before the synchronized block that calls `notify`/`notifyAll` ([#3786](https://github.com/spotbugs/spotbugs/issues/3786))
 - Fix `SA_LOCAL_SELF_ASSIGNMENT` false positive in methods with `++`/`--` inside a switch nested in a try-catch block ([#3929](https://github.com/spotbugs/spotbugs/issues/3929))
 - Fix `OS_OPEN_STREAM` false positive when the result of `PrintWriter.append()` is reassigned before closing the writer ([#4274](https://github.com/spotbugs/spotbugs/issues/4274))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - Fix `OBL_UNSATISFIED_OBLIGATION` false negatives for unclosed `CallableStatement`s returned by `Connection.prepareCall` ([#4148](https://github.com/spotbugs/spotbugs/issues/4148))
 - Fix `LI_LAZY_INIT_STATIC` false negative when field is lazily initialized using a method call ([#4276](https://github.com/spotbugs/spotbugs/issues/4276))
 - Fix SARIF output writing source-location URI syntax exceptions to stderr when source filenames are unknown ([#1412](https://github.com/spotbugs/spotbugs/issues/1412))
+- Narrow the definition of singletons ([#2985](https://github.com/spotbugs/spotbugs/issues/2985))
+- Fix `NP_LOAD_OF_KNOWN_NULL_VALUE` false negative when null check uses `instanceof` ([#3916](https://github.com/spotbugs/spotbugs/issues/3916))
 
 ## 4.10.4 - 2026-08-19
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - `@SuppressFBWarnings` annotation on a method or a constructor now suppresses warnings reported in their lambdas too ([#724](https://github.com/spotbugs/spotbugs/issues/724))
 
 ### Fixed
+- Fix `NN_NAKED_NOTIFY` false positive when guarded state is updated before the synchronized block that calls `notify`/`notifyAll` ([#3786](https://github.com/spotbugs/spotbugs/issues/3786))
 - Fix `SA_LOCAL_SELF_ASSIGNMENT` false positive in methods with `++`/`--` inside a switch nested in a try-catch block ([#3929](https://github.com/spotbugs/spotbugs/issues/3929))
 - Fix `OS_OPEN_STREAM` false positive when the result of `PrintWriter.append()` is reassigned before closing the writer ([#4274](https://github.com/spotbugs/spotbugs/issues/4274))
 - Fix `SING_SINGLETON_GETTER_NOT_SYNCHRONIZED` false negative when lazy init uses a ternary assignment instead of an `if` statement ([#4154](https://github.com/spotbugs/spotbugs/issues/4154))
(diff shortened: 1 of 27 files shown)


hazendaz:
@AparnaPradhan Thanks for the fix — the approach looks good to me. Before I merge, could you add a couple of tests covering the null-state behavior directly, in addition to the existing regression test? Since this changes `OpcodeStack` control-flow state, I'd like to have the intended behavior documented for both sides of the conditional.

Also, please add the changelog entry for the fix.

Once those are in, I'm happy to merge. Thanks!

hazendaz (changes requested):
needs additional tests and changelog

AparnaPradhan:
Hi @hazendaz,
Sure will add the additional changes and update you

AparnaPradhan:
Title
  Fix RANGE_ARRAY_INDEX false positive inside unreachable code from impossible null condition
  
  ---
  Body

  ## Summary

  Fixes a false positive where `RANGE_ARRAY_INDEX` was reported on an array
  access inside a block guarded by an always-false null condition such as
  `if (x == null && x != null)`.

  ## Root cause

  `OpcodeStack`'s `IFNONNULL` and `IFNULL` cases fell through to the shared
  handler used by `IFEQ`/`IFGT`/etc., which has no reference-null semantics.
  As a result:

  - After `IFNONNULL`, the fall-through path ("value *is* null") did not record
    that the local variable is null, so a subsequent `IFNULL` on the same
    variable couldn't tell the jump was unconditional.
  - The dead fall-through of `IFNULL` was treated as live code, so
    `RangeCheckSubDetector` was invoked and raised a spurious warning.

  ## Fix

  **`OpcodeStack.java`** — split `IFNONNULL`/`IFNULL` into dedicated cases:

  - **`IFNONNULL`**: after `addJumpValue` (which snapshots `lvValues` *before*
    the update), overwrites the local-variable slot with a null item so that
    code on the fall-through path sees the variable as definitively null.
  - **`IFNULL`**: if the popped item `isNull()`, the jump is unconditional and
    the fall-through is dead — `setTop(true)` is called, causing
    `OpcodeStackDetector.beforeOpcode` to return `false` and skip `sawOpcode`
    for every instruction in the dead block.

  For the reproducer the chain is:

  1. `IFNONNULL` → marks local variable null on fall-through
  2. `ALOAD` → pushes the null item
  3. `IFNULL` → sees null → `setTop(true)`
  4. `IASTORE` (`array[2] = 1`) → `beforeOpcode` returns `false` → `checkRange` never runs

  **`Feature323.java`** — three regression tests:

  | Method | Annotation | What it covers |
  |---|---|---|
  | `unreachableArrayIndex` | `@NoWarning` | Full IFNONNULL→IFNULL chain (the original reproducer) |
  | `reachableArrayIndexInNullCheckBody` | `@ExpectWarning` | IFNONNULL must not set top — body of a live `if (x == null)` block must still be analysed |
  | `unreachableArrayIndexExplicitNull` | `@NoWarning` | IFNULL dead-fall-through in isolation — explicit `null` assignment makes the `if (x != null)` body unreachable |

  **`CHANGELOG.md`** — entry added under Unreleased → Fixed.

  ## Test plan
  
  - [ ] `./gradlew :spotbugs:compileJava :spotbugsTestCases:compileJava`
  - [ ] `./gradlew test --tests "*.Feature323*"`
  - [ ] `./gradlew test` (full suite — no regressions)

hazendaz (approved):

hazendaz:
@AparnaPradhan tests are failing

AparnaPradhan:
## Summary
  - Split `IFNONNULL`/`IFNULL` opcodes in `OpcodeStack` into dedicated cases
  - On `IFNONNULL` fall-through, propagate null state to the local variable slot so a subsequent `IFNULL` on the same variable can detect the impossible condition
  - On `IFNULL`, mark the fall-through as dead code only when the null comes from a local variable (`registerNumber >= 0`), not from a field summary, to avoid false positives
   in detectors like `SE_PREVENT_EXT_OBJ_OVERWRITE` 
  - Added three regression tests in `Feature323.java`
  
  ## Test plan
  - [ ] `@NoWarning(\"RANGE_ARRAY_INDEX\")` on `unreachableArrayIndex` — the impossible-null body is dead code
  - [ ] `@ExpectWarning(\"RANGE_ARRAY_INDEX\")` on `reachableArrayIndexInNullCheckBody` — the body of a live null check is still reachable
  - [ ] `@NoWarning(\"RANGE_ARRAY_INDEX\")` on `unreachableArrayIndexExplicitNull` — `Object value = null; if (value != null)` body is dead
  - [ ] `PreventOverwriteOfExternalizableObjectTest.testGoodReadExternal8` passes (no regression)" 
  
image

spotbugsTestCases/src/java/sfBugsNew/Feature323.java
```diff
@@ -105,5 +105,33 @@ public char charAtIncorrect() {
         return TEST_STRING.charAt(TEST_STRING.length());
     }
 
+    @NoWarning("RANGE_ARRAY_INDEX")
```
JuditKnoll: Can you please create tests with adding testfiles and creating a separate class testing it from a child class of [AbstractIntegrationTest](https://github.com/spotbugs/spotbugs/blob/master/spotbugs-tests/src/test/java/edu/umd/cs/findbugs/AbstractIntegrationTest.java) instead of using these annotations?

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4295Test.java
```diff
@@ -0,0 +1,29 @@
...
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue4295Test extends AbstractIntegrationTest {
+
+    @Test
+    void testImpossibleNullConditionIsNotReported() {
+        performAnalysis("ghIssues/Issue4295.class");
+
+        assertNoBugInMethod("RANGE_ARRAY_INDEX", "Issue4295", "impossibleNullCondition");
```
JuditKnoll: Please also assert the number of relevant bugtypes.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4295Test.java
```diff
@@ -0,0 +1,29 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue4295Test extends AbstractIntegrationTest {
+
+    @Test
+    void testImpossibleNullConditionIsNotReported() {
```
JuditKnoll: Since all tests analyze the same class, one test can be more efficient with several asserts.

JuditKnoll (changes requested):

AparnaPradhan:
⏺ Title: Fix RANGE_ARRAY_INDEX false positive in unreachable null-condition code

  Description:

  RANGE_ARRAY_INDEX was incorrectly reported when an array access appeared inside dead code caused by an impossible null condition such as if (x == null && x != null).

  Root cause: OpcodeStack did not track null state through IFNONNULL/IFNULL fall-throughs for local variables, so the unreachable branch was not detected as dead code.

  Fix: Split IFNONNULL and IFNULL into dedicated opcode cases in OpcodeStack.java:
  - On IFNONNULL fall-through, propagate null state to the local variable slot
  - On IFNULL, mark the fall-through as dead code when the item is a local variable known to be null (registerNumber >= 0), excluding field-summary nulls to avoid false
  positives in other detectors

  Tests:
  - spotbugsTestCases/src/java/ghIssues/Issue4295.java — three test scenarios
  - spotbugs-tests/.../detect/Issue4295Test.java — AbstractIntegrationTest subclass asserting correct behavior for each scenario
 
image

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4295Test.java
AparnaPradhan: resolved

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4295Test.java
```diff
@@ -0,0 +1,25 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue4295Test extends AbstractIntegrationTest {
```
JuditKnoll: #4295 is this PR, it's not an issue. Can you please rename this class together with the example class file to not be confusing? As far as I can see the relevant issue is #4275 .

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Currently the versioning policy of this project follows [Semantic Versioning v2.
 - `@SuppressFBWarnings` annotation on a method or a constructor now suppresses warnings reported in their lambdas too ([#724](https://github.com/spotbugs/spotbugs/issues/724))
 
 ### Fixed
+- Fix `RANGE_ARRAY_INDEX` false positive when the array access is inside unreachable code caused by an impossible null condition (e.g. `if (x == null && x != null)`) ([#4295](https://github.com/spotbugs/spotbugs/issues/4295))
```
JuditKnoll: #4295 is this PR, not an issue. GitHub will redirect from the link to the PR, but it's an unnecessary redirection. Can you please fix the link?

JuditKnoll (dismissed):

CHANGELOG.md
AparnaPradhan: sure checking

AparnaPradhan: Fixed a false positive where RANGE_ARRAY_INDEX was reported inside dead code caused by an impossible null condition (if (x == null && x != null)). Improved
  null-state tracking in OpcodeStack for IFNONNULL/IFNULL opcodes so unreachable branches are correctly identified. Added regression tests in Issue4275.

spotbugsTestCases/src/java/sfBugsNew/Feature323.java
hazendaz: resolved.

hazendaz (approved):

hazendaz (approved):

PR #4299, opened 2026-09-08
code fix for NP_NULL_ON_SOME_PATH #4273

### LINKED ISSUE #4273: FALSE POSITIVE: NP_NULL_ON_SOME_PATH ignores a known-true instanceof check
Hi, I found a false positive in SpotBugs 4.10.3 when the runtime object is exactly Derived, so result is assigned.

**Affected tool**

`SpotBugs 4.10.3`

**Affected checker**

`SpotBugs NP_NULL_ON_SOME_PATH`

**Minimal reproducer**

```java
class SpotBugsKnownInstanceof {
  class Base {}
  final class Derived extends Base {}
  void test() {
    Base value = new Derived();
    Object result = null;
    if (value instanceof Derived) result = new Object();
    result.toString();
  }
}
```

**Reproduction command**

```bash
spotbugs -version
javac -d classes spotbugs-np-null-on-some-path-fp-known-instanceof.java
spotbugs -textui -effort:max -low -visitors FindNullDeref classes
```

**Current behavior**

SpotBugs reports `NP_NULL_ON_SOME_PATH` at line 8.

**Expected behavior**

SpotBugs should not report `NP_NULL_ON_SOME_PATH` because the runtime object is exactly Derived, so result is assigned.

AparnaPradhan:
fixes #4273

Title: Fix NP_NULL_ON_SOME_PATH false positive when instanceof is always true

  Description:

  NP_NULL_ON_SOME_PATH was incorrectly reported when an instanceof check is always true because the value was created with new and its exact type is a subtype of the checked
  type:

  Base value = new Derived();
  Object result = null;
  if (value instanceof Derived) result = new Object();
  result.toString(); // falsely flagged

  Root cause: IsNullValueAnalysis.getDecision() returned null (no decision) whenever the tested value was definitely non-null, bypassing the subtype check entirely. The "not
  instanceof" branch was incorrectly treated as reachable, so the path where result stays null was merged into the post-if state.

  Fix: When tos.isDefinitelyNotNull() and the value's static type is a subtype of the instanceof check type (Hierarchy.isSubtype()), the "not instanceof" branch is marked as
  infeasible — only the instanceof branch is returned as reachable.

  Tests:
  - spotbugsTestCases/src/java/ghIssues/Issue4296.java — two cases: new Derived() (no warning) and unknown parameter (warning expected)
  - spotbugs-tests/.../detect/Issue4296Test.java — asserts exactly 1 NP_NULL_ON_SOM


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4296Test.java
```diff
@@ -0,0 +1,22 @@
+package edu.umd.cs.findbugs.detect;
+
+import org.junit.jupiter.api.Test;
+
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+
+class Issue4296Test extends AbstractIntegrationTest {
```
JuditKnoll: Can you please rename this to reference the relevant issue together with the analyzed class?

spotbugsTestCases/src/java/ghIssues/Issue4296.java
```diff
@@ -0,0 +1,23 @@
+package ghIssues;
+
+public class Issue4296 {
```
JuditKnoll: Can you please add more examples? With negating and not possible instanceof?

JuditKnoll (dismissed):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4296Test.java
AparnaPradhan: - Files renamed to Issue4273 / Issue4273Test with all internal references updated

spotbugsTestCases/src/java/ghIssues/Issue4296.java
AparnaPradhan: - Added knownNegatedInstanceof — !(value instanceof Derived) is always false, body is dead, no NP expected
  - Test asserts 0 bugs in both knownInstanceof and knownNegatedInstanceof, 1 total bug only from unknownInstanceof

PR #4303, opened 2026-09-09
Code fix for NP_ALWAYS_NULL

### LINKED ISSUE #4272: FALSE POSITIVE: `NP_ALWAYS_NULL` ignores known primitive integer values
Hi, I found a false positive in SpotBugs 4.10.3 when two primitive integer values are known to be equal, making a null dereference unreachable.

**Affected tool**

`SpotBugs 4.10.3`

**Affected checker**

`SpotBugs NP_ALWAYS_NULL`

**Minimal reproducer**

```java
class SpotBugsKnownIntegerValues {
  void test() {
    int left = 42;
    int right = 42;
    if (left != right) {
      Object object = null;
      object.toString();
    }
  }
}
```

**Reproduction command**

```bash
spotbugs -version
javac -d classes spotbugs-np-always-null-fp-known-integer-values.java
spotbugs -textui -effort:max -low -visitors FindNullDeref classes
```

**Current behavior**

SpotBugs reports `NP_ALWAYS_NULL` at line 7.

**Expected behavior**

SpotBugs should not report `NP_ALWAYS_NULL` because `left` and `right` are both 42, so the `left != right` branch is unreachable.

Bhaswati-Saha:
fixes #4272 

Title: Fix NP_ALWAYS_NULL false positive when primitive integer values are known equal
Description:
NP_ALWAYS_NULL was incorrectly reported when two primitive integer values were known to be equal, making an integer-comparison branch unreachable:
int left = 42;
int right = 42;

if (left != right) {
    Object object = null;
    object.toString(); // falsely flagged
}
Root cause: IsNullValueAnalysis.getDecision() did not create a path decision for IF_ICMPEQ and IF_ICMPNE instructions. Consequently, both branches were treated as reachable even when value-number analysis determined that both operands represented the same value.
Fix: When both operands of an IF_ICMPEQ or IF_ICMPNE comparison have the same ValueNumber, mark the impossible branch as infeasible. This prevents the unreachable null dereference from contributing an NP_ALWAYS_NULL warning.
Tests:
- spotbugsTestCases/src/java/ghIssues/Issue4272.java — reproduces comparison of two known-equal primitive integers.
- spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/Issue4272Test.java — verifies that no NP_ALWAYS_NULL warning is reported for the unreachable branch.

### DIFF
--- a/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/npe/IsNullValueAnalysis.java
+++ b/spotbugs/src/main/java/edu/umd/cs/findbugs/ba/npe/IsNullValueAnalysis.java
@@ -693,8 +693,6 @@ public class IsNullValueAnalysis extends FrameDataflowAnalysis<IsNullValue, IsNu
                     // ifnonnull
                     fallThroughDecision = tos;
                 }
-            } else if (tos.isDefinitelyNotNull()) {
-                return null;
             } else {
                 // Check if the static type of the value is a subtype of the instanceof check type.
                 // If so, "not instanceof" can only happen when the value is null.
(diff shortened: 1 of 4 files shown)


Bhaswati-Saha:
@hazendaz I have fixed the NP_ALWAYS_NULL issue, could you please review it.

spotbugs/src/main/java/edu/umd/cs/findbugs/ba/npe/IsNullValueAnalysis.java
```diff
@@ -744,6 +744,18 @@ private IsNullConditionDecision getDecision(BasicBlock basicBlock, IsNullValueFr
         }
 
         if (!nullComparisonInstructionSet.get(lastInSourceOpcode)) {
+            if (lastInSourceOpcode == Const.IF_ICMPEQ || lastInSourceOpcode == Const.IF_ICMPNE) {
+                ValueNumberFrame prevVnaFrame = vnaDataflow
+                        .getFactAtLocation(new Location(lastInSourceHandle, basicBlock));
+                ValueNumber tos = prevVnaFrame.getStackValue(0);
```
ThrawnCA, line 750: Why extract these to local variables when they're only used once?

hazendaz: @ThrawnCA I believe they left the tos/nextToTos naming consistent with the existing IF_ACMP* handling in this method.  Since that seems the overall convention of the file and does not add to GC concerns, I feel its reasonable to leave in place for now although it could be more descriptive but will take as-is.

hazendaz (approved):

PR #4319, opened 2026-09-12
Migrate 'CheckForNull' and solidify on 'Nullable' from jspecify inside spotbugs/spotbugs

hazendaz:
Internal change only, no change log.


hazendaz:
found a true concern from copilot and there is one spot that cannot migrate due to support issue so going to fix those first before doing this, will bring out of draft as soon as done (expecting today).

PR #4326, opened 2026-09-17
Fix USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION false positive for new lock in clone()

### LINKED ISSUE #4320: False positive USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION for locks created in `clone()`.
**Describe the bug**

When using a private synchronization lock object in classes, SpotBugs reports an USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION when clones have their own private locks created in the `clone()` implementation.

**To Reproduce**

1. Run Spotbugs on the sample code provided below.
2. It produces the following false positive:

```
[ERROR] Medium: The lock object nom.tam.fits.Foo.lock used in method nom.tam.fits.Foo.bar() is exposed to untrusted code outside of class nom.tam.fits.Foo, through updating or returning its value from nom.tam.fits.Foo.clone()Lnom/tam/fits/Foo;. [nom.tam.fits.Foo] At Foo.java:[lines 52-55] USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION
``` 

```java
public class Foo implements Cloneable {

    private int property; // class property with synchronized access

    private Object lock = new Object(); // private synchronization lock

    @Override
    protected Foo clone() {
        try {
            Foo clone = (Foo) super.clone();
            clone.lock = new Object(); // cloned object to have its own independent synchronization lock.
            return clone;
        } catch (CloneNotSupportedException e) {
            return null;
        }
    }

    public void bar() {
        synchronized (lock) { // Here we get an USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION
            property = 0;
        }
    }
}
```

**Expected behavior**

No USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION error, since the lock is never exposed.

**Environment**
- Java Compiler: OpenJDK
- Java version: JDK 25.0.4.1
- SpotBugs Version: 4.10.4.0
- How do you use SpotBugs: Maven Plugin
    - Plugin version: SpotBugs Maven Plugin 4.10.4.0
    - Environment version for Plugin: 3.9.11

**Additional context**

Note, that clones should always have their own private locks, or else the original class may block operations on the clone and vice-versa. SpotBugs does not complain if `clone()` retains the reference to the original lock object, even though that is a potential synchronization issue, since the clone and the original can block each other.

ethanstoner:

Fixes #4320

`FindImproperSynchronization` treats every write to a lock field in a public/protected method as exposing the lock. A `clone()` that gives the copy its own lock (`clone.lock = new Object()`) was therefore reported as `USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION`, even though nothing outside the class can get at the lock.

The detector now ignores a `PUTFIELD` in `clone()` when the stored value is newly allocated and the target object is not `this`. Other writes still count as exposure, including `this.lock = ...` in `clone()`, writes of values that aren't newly allocated, and writes in other methods.

Tests:
- `SafeSynchronizationWithNewLockInClone` (the code from the issue): no USO bugs. This failed before the fix.
- `UnsafeSynchronizationWithNewLockInCloneAndLockSetter`: same `clone()` plus a public lock setter, still reports one `USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION`.

Ran locally on JDK 21: `FindImproperSynchronizationTest` (17), `FindImproperMixedSynchronizationTest` (2) and `FindImproperSynchronizationWithAccessibleBackingCollectionsTest` (4) all pass; `spotlessApply` and `:spotbugs:spotlessCheck :spotbugs-tests:spotlessCheck` pass. I did not run the full `build smoketest` locally.

Generated with Claude Code on behalf of @ethanstoner.


ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
```diff
@@ -497,6 +497,40 @@ void testSafeSynchronizationWithBuiltinMethodsSeemToExposeLock() {
...
 
+    /**
+     * @see <a href="https://github.com/spotbugs/spotbugs/issues/4320">GitHub issue #4320</a>
+     */
+    @Test
+    void testSafeSynchronizationWithNewLockInClone() {
+        performAnalysis("synchronizationLocks/privateFinalLocks/SafeSynchronizationWithNewLockInClone.class");
+
+        assertNoUnsafeLockBugs();
+        assertBugTypeCount(EXPOSED_LOCK_OBJECT_BUG, 0);
```
JuditKnoll: It looks like this check is missing from `assertNoUnsafeLockBugs()`. Good catch!
As far as I can see, it should be there. Can you please add this check to the function?
(From the name the `testSafeSynchronizationWithBuiltinMethodsSeemToExposeLock` test might have an `USO_UNSAFE_EXPOSED_OBJECT_SYNCHRONIZATION` hit.)

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
```diff
@@ -441,6 +441,24 @@ private boolean isPackagePrivate(XField field) {
...
 
+    /**
+     * Check if the current field update happens in {@code clone()} and stores a newly allocated object into a field of
+     * another instance (typically the copy returned by {@code super.clone()}). Giving the copy its own, fresh lock
+     * object does not expose the lock to untrusted code.
+     *
+     * @param seen the current opcode
+     * @return true if a newly allocated object is assigned to a field of an object other than {@code this} in {@code clone()}
+     */
+    private boolean isNewLockAssignedToCloneCopy(int seen) {
```
JuditKnoll, line 452: I was wondering whether this false positive can only happen in clone, or this should be broaden to other functions where a newly allocated instance of the visited class is not set to `this`? WDYT?
If nothing comes up, it can be broaden later on as well, if an issue or edge case comes up.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
ethanstoner: Done - moved `assertBugTypeCount(EXPOSED_LOCK_OBJECT_BUG, 0)` into `assertNoUnsafeLockBugs()` and dropped the now duplicate line here.

Good news on the suspicion: `testSafeSynchronizationWithBuiltinMethodsSeemToExposeLock` still passes with the assertion enabled - no `USO_UNSAFE_EXPOSED_OBJECT_SYNCHRONIZATION` there. All 23 tests across `FindImproperSynchronizationTest`, `FindImproperMixedSynchronizationTest` and `FindImproperSynchronizationWithAccessibleBackingCollectionsTest` are green, so this was a genuine gap in the helper rather than a hidden failure.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
ethanstoner: I looked into this, and I'd rather keep it scoped to `clone()` for now - the obvious generalizations are either unsafe or don't actually help `clone()`.

The reason `clone()` is safe isn't "the value is newly allocated". `UnsafeSynchronizationWithLockExposedInDeclaringClass.updateLock4()` is exactly `public void updateLock4() { lock4 = new Object(); }` and is an expected true positive - a public method that swaps a live lock breaks mutual exclusion even though nothing is exposed. So the newly-allocated value alone is never sufficient; what makes `clone()` safe is that the *target* instance hasn't escaped yet, so there is no live lock to swap.

That rules out broadening on "target isn't `this`". Counterexample I ran against the current detector:

```java
public class Foo {
    private Object lock = new Object();
    public void doStuff() { synchronized (lock) { ... } }
    public void resetLockOf(Foo other) { other.lock = new Object(); }
}
```

`USO_UNSAFE_ACCESSIBLE_OBJECT_SYNCHRONIZATION` is reported today, and it should be: untrusted code can call `f.resetLockOf(f)` while another thread is inside `doStuff()`, and the next thread in acquires a different object. Dropping the `clone` name check would silence that.

The principled version is "the target object is provably unescaped", which on the opcode stack means `target.isNewlyAllocated()`. I instrumented `isNewLockAssignedToCloneCopy` to print what the stack sees:

```
PROBE method=clone        targetNewlyAllocated=false targetReg=1
PROBE method=create       targetNewlyAllocated=true  targetReg=0   // static factory
PROBE method=resetLockOf  targetNewlyAllocated=false targetReg=1
PROBE method=updateLock4  targetNewlyAllocated=false targetReg=0
```

Two things fall out. `super.clone()` returns from an INVOKESPECIAL, so its result is never `NEWLY_ALLOCATED` - the `isNewlyAllocated` route can't cover `clone()` at all, and the name-based check is needed either way. But a static factory does come back `NEWLY_ALLOCATED`:

```java
public static Foo create() {
    Foo f = new Foo();
    f.lock = new Object();   // false positive today
    return f;
}
```

That's a real, separate false positive that a `target.isNewlyAllocated()` arm would fix cleanly without touching the `resetLockOf` case. It's a different code path from #4320 though, and it'd want its own test cases, so I'd rather not fold it into this PR on top of an existing approval. Happy to open a follow-up for it if you'd like - or leave it until someone reports it, as you suggested.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindImproperSynchronizationTest.java
JuditKnoll: Thanks, sounds great!

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
JuditKnoll: Thank you for looking into this and the detailed explanation.
> Happy to open a follow-up for it if you'd like - or leave it until someone reports it, as you suggested.

It you could open a follow-up PR for it - or at least open an issue about this additional edge case - that would be awesome 😄 

(@ethanstoner Only leaving this unresolved, so you see it, feel free to resolve it.)

JuditKnoll (approved):

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindImproperSynchronization.java
ethanstoner: Opened it as #4331, with the reproducer, the probe output and the two cases the new arm must not silence (`updateLock4` and `resetLockOf`). Happy to follow up with the PR and test cases once this one lands.

Thanks for the review.

PR #4327, opened 2026-09-17
Fix MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR missing methods with arguments

### LINKED ISSUE #4309: FALSE NEGATIVE: `MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR` misses an overridable setter call
Hi, I found a false negative in SpotBugs 4.10.3 when a constructor calls an overridable setter declared in the same class.

**Affected tool**

`SpotBugs 4.10.3`

**Affected checker**

`SpotBugs MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR`

**Minimal reproducer**

```java
class SpotbugsSonarqubeConstructorSetter {
  int value;

  SpotbugsSonarqubeConstructorSetter(int value) {
    setValue(value);
  }

  void setValue(int value) {
    this.value = value;
  }
}
```

**Reproduction command**

```bash
spotbugs -version
javac -d classes spotbugs-sonarqube-mc-overridable-call-fn-setter.java
spotbugs -textui -effort:max -low -visitors FindOverridableMethodCall classes
```

**Current behavior**

SpotBugs produces no `MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR` diagnostic.

**Expected behavior**

SpotBugs should report line 5 because `setValue()` is overridable during construction.

ronodhirSoumik:
Fixes #4309 

In` FindOverridableMethodCall.sawOpcode()`, the receiver object for `INVOKEVIRTUAL/INVOKEINTERFACE `calls was retrieved with` stack.getStackItem(0)`. This only returns the correct receiver (this) for zero-argument methods. But for methods with arguments like `setValue(value)`, the `getStackItem(0)` returns the topmost argument instead, causing `getRegisterNumber() == 0` to fail and the overridable call to go undetected.

Used the `getNumberArguments()` to correctly locate the receiver object on the operand stack for `INVOKEVIRTUAL/INVOKEINTERFACE` calls. 

**Tests**
`./gradlew :spotbugs-tests:test --tests "edu.umd.cs.findbugs.detect.FindOverridableMethodCallTest.testDirectCaseWithArguments"`


ThrawnCA (approved):

spotbugsTestCases/src/java/overridableMethodCall/DirectCaseWithArguments.java
```diff
@@ -0,0 +1,42 @@
+package overridableMethodCall;
+
+/**
+ * Test case for GitHub issue #4309:
+ * MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR should be reported when a constructor
+ * or clone method calls an overridable method that takes arguments (e.g. a setter).
+ * @see <a href="https://github.com/spotbugs/spotbugs/issues/4309">GitHub issue</a>
+ */
+public class DirectCaseWithArguments implements Cloneable {
```
JuditKnoll, line 9: Can you please also add the minimal reproducer from the issue?

JuditKnoll (changes requested):
Please add a changelog entry to the changelog's unreleased part.

spotbugsTestCases/src/java/overridableMethodCall/DirectCaseWithArguments.java
ronodhirSoumik: Added. Should I remove `spotbugsTestCases/src/java/overridableMethodCall/DirectCaseWithArguments.java` and its test in `spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindOverridableMethodCallTest.java`?

JuditKnoll: Thanks. No need to remove, it can be kept.

JuditKnoll (approved):

PR #4330, opened 2026-09-21
Fix TLW_TWO_LOCK_WAIT false negative

sonihitesh:

----


spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindTwoLockWaitTest.java
```diff
@@ -0,0 +1,21 @@
...
+import edu.umd.cs.findbugs.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+class FindTwoLockWaitTest extends AbstractIntegrationTest {
+
+    @Test
+    void reportsWaitWhileHoldingDistinctMonitors() {
+        performAnalysis("TwoLockWaitDistinctMonitors.class");
+
+        assertBugInMethod("TLW_TWO_LOCK_WAIT", "TwoLockWaitDistinctMonitors", "waitWithTwoLocks");
```
utafrali, line 13: The two existing tests cover the false-negative fix (distinct monitors, no notify) and the false-positive guard (same monitor). There is no test for the third path introduced by the restructured code: distinct monitors *with* a `notify()` call present in the same class. After the fix, `bug.addSourceLine(notifyLine).describe("SOURCE_NOTIFICATION_DEADLOCK")` still runs in that branch, but there is no test verifying the bug is still reported *and* the annotation is attached. Adding a fixture like `TwoLockWaitWithNotify.java` and a corresponding `assertBugInMethod` + annotation check would make the test suite complete.

spotbugs/src/main/java/edu/umd/cs/findbugs/detect/FindTwoLockWait.java
```diff
@@ -93,13 +93,13 @@ public void visitClassContext(ClassContext classContext) {
                 bugReporter.logError("Error analyzing " + method.toString(), e);
             }
         }
-        if (!possibleNotifyLocations.isEmpty()) {
-            for (BugInstance bug : possibleWaitBugs) {
+        for (BugInstance bug : possibleWaitBugs) {
```
utafrali, line 96: The restructuring is correct, but worth a short comment to explain *why* `reportBug` is unconditional. Something like `// Notify locations are optional enrichment; always report the wait-under-two-locks bug` would prevent a future reader from reinstating the old guard under the mistaken belief that a corresponding notify is required.

utafrali (approved):
The fix is correct: moving `reportBug` outside the `possibleNotifyLocations` guard directly resolves the false negative, and the tests cover both the new true-positive case and the existing true-negative case. The two comments above are minor gaps rather than blockers.

ThrawnCA (approved):

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindTwoLockWaitTest.java
JuditKnoll, line 13: Please also assert the number of relevant bugs.

spotbugsTestCases/src/java/TwoLockWaitDistinctMonitors.java
```diff
@@ -0,0 +1,12 @@
+public class TwoLockWaitDistinctMonitors {
```
JuditKnoll, line 2: It's a good practice to refer the relevant issue in comment, or even use the gh<issueNumber> naming convention.

spotbugs-tests/src/test/java/edu/umd/cs/findbugs/detect/FindTwoLockWaitTest.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @Test
+    void doesNotReportWaitWhileReenteringTheSameMonitor() {
+        performAnalysis("TwoLocksWhileWaitingFalsePositive.class");
```
JuditKnoll, line 18: Did you forget to upload this sample class?

JuditKnoll (changes requested):

ThrawnCA (approved):

